PR #3, opened 2021-04-10
Gh-386: formerly panics at double-used "shutdown" ; ended with broad redesign of some structures in masq 

bertllll:


USER-INTERFACE-INTERFACE.md
```diff
@@ -458,6 +458,27 @@ Requests the Node descriptor from a Node.
...
+```
+"payload": {
+    "opcode": <string>,
+    }
+}
+```
+##### Description:
+When the Daemon receives a fire-and-forget message (which is a case not being implemented at the time of writing this;
+all such messages go only in the opposite direction now), the normal way to proceed would be to look whether that type 
+of message is known to the Deamon. If not, then it tries to send a redirect message. However, if it turns out, at the
```
dnwiebe: Daemon, not Deamon.

USER-INTERFACE-INTERFACE.md
```diff
@@ -458,6 +458,27 @@ Requests the Node descriptor from a Node.
 ##### Description:
 Contains a Node's Node descriptor.
 
+#### `undelivered`
+##### Direction: Broadcast
```
dnwiebe: Response, not Broadcast--or at least Node/Daemon-to-UI

masq/src/command_context.rs
```diff
@@ -62,9 +61,10 @@ pub trait CommandContext {
 
 pub struct CommandContextReal {
     connection: ConnectionManager,
-    pub stdin: Box<dyn Read>,
-    pub stdout: Box<dyn Write>,
-    pub stderr: Box<dyn Write>,
+    pub stdin: Box<dyn Read + Send>,
```
dnwiebe: Does CommandContext really need to be sent between threads? Maybe only TerminalWrapper does. If so, probably these + Send markers should be removed. If later we need to put something into CommandContext that simply can't be sent between threads, we'll wind up cursing your ancestors for this.

masq/src/command_processor.rs
```diff
@@ -111,17 +127,124 @@ mod tests {
...
+
+    #[derive(Debug)]
+    struct TameCommand {
+        sender: Sender<String>,
+    }
+
+    impl Command for TameCommand {
+        fn execute(&self, _context: &mut dyn CommandContext) -> Result<(), CommandError> {
+            self.sender.send("This is a message".to_string()).unwrap();
+            thread::sleep(Duration::from_millis(10));
```
dnwiebe: Nice. But could be less than 10ms.

masq/src/command_processor.rs
```diff
@@ -111,17 +127,124 @@ mod tests {
...
+        processor
+            .process(Box::new(TameCommand {
+                sender: cloned_stdout_sender,
+            }))
+            .unwrap();
+
+        let tamed_message_as_a_whole = "This is a message which must be delivered as one piece; we'll do all being \
+             possible for that. If only we have enough strength and spirit and determination and support and... snacks. Roger.";
+        let received_output = broadcast_stream_factory_handle.stdout_so_far();
+        assert!(
```
dnwiebe, line 207: Can I see another assert here proving that the broadcast messages are also present here in this string?

masq/src/communications/broadcast_handler.rs
```diff
@@ -16,6 +19,21 @@ pub trait BroadcastHandle: Send {
     fn send(&self, message_body: MessageBody);
 }
 
+pub struct BroadcastHandleInactive {}
+
+impl BroadcastHandle for BroadcastHandleInactive {
+    fn send(&self, message_body: MessageBody) {
+        drop(message_body); //simply dropped (unless we find a better use for such a message)
```
dnwiebe: You could remove this line (except for the comment, that's good) and put an underscore at the beginning of the variable name.

masq/src/communications/broadcast_handler.rs
```diff
@@ -88,9 +119,10 @@ impl BroadcastHandlerReal {
         message_rx: &Receiver<MessageBody>,
         stdout: &mut dyn Write,
         stderr: &mut dyn Write,
+        terminal_interface: TerminalWrapper,
     ) {
         select! {
-            recv(message_rx) -> message_body_result => Self::handle_message_body (message_body_result, stdout, stderr),
+        recv(message_rx) -> message_body_result => Self::handle_message_body (message_body_result, stdout, stderr,terminal_interface)
```
dnwiebe: Maybe put the indent back in?

dnwiebe (commented):
We stopped at line 321 of broadcast_handler.rs.

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+
+        test_generic_for_handle_broadcast(
+            handle_node_not_running_for_fire_and_forget_on_the_way,
+            ffm_undelivered_body,
+            broadcast_output,
+        )
+    }
+
+    fn test_generic_for_handle_broadcast<T, U>(
+        broadcast_handle: T,
```
dnwiebe: Maybe this should be "broadcast_handler"?

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+            asterisks_count
+        );
+    }
+
+    fn background_thread_making_interferences<U, T>(
+        sync: bool,
+        stdout: &mut dyn Write,
+        mut stdout_clone: Box<dyn Write + Send>,
+        synchronizer: TerminalWrapper,
+        broadcast_handle: T,
```
dnwiebe: Again: broadcast_handler, not broadcast_handle

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+        let interference_thread_handle = thread::spawn(move || {
+            sync_tx.send(()).unwrap();
+            let _lock = if sync {
+                Some(synchronizer.lock())
+            } else {
+                None
+            };
+            (0..80).into_iter().for_each(|_| {
+                stdout_clone.write(b"*").unwrap();
+                thread::sleep(Duration::from_millis(1))
```
dnwiebe: Try 0 here and see if it still works.

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+        rx: Receiver<String>,
+    ) -> String
+    where
+        T: FnOnce(U, &mut dyn Write, TerminalWrapper) + Copy,
+        U: Debug + PartialEq + Clone,
+    {
+        let synchronizer_clone = synchronizer.clone();
+        let (sync_tx, sync_rx) = std::sync::mpsc::channel();
+        let interference_thread_handle = thread::spawn(move || {
+            sync_tx.send(()).unwrap();
```
dnwiebe: If you move this down to after the lock is obtained, you might be able to get rid of the 40-millisecond wait below.

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+    }
+
+    fn background_thread_making_interferences<U, T>(
+        sync: bool,
+        stdout: &mut dyn Write,
+        mut stdout_clone: Box<dyn Write + Send>,
+        synchronizer: TerminalWrapper,
+        broadcast_handle: T,
+        broadcast_message_body: U,
+        rx: Receiver<String>,
```
dnwiebe: Maybe in a function parameter list, this should have a little more-descriptive name?

masq/src/communications/broadcast_handler.rs
```diff
@@ -247,4 +317,202 @@ mod tests {
...
+            running: false,
+            values: vec![UiSetupResponseValue {
+                name: "ip".to_string(),
+                value: "4.4.4.4".to_string(),
+                status: UiSetupResponseValueStatus::Set,
+            }],
+            errors: vec![],
+        };
+
+        let broadcast_output = "Daemon setup has changed:
```
dnwiebe, line 378: This puzzles me. Why does this test pass? Doesn't the broadcast handler produce more complex text than this?

masq/src/communications/connection_manager.rs
```diff
@@ -540,15 +541,16 @@ impl RedirectBroadcastHandler {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::communications::broadcast_handler::{BroadcastHandler, StreamFactoryReal};
+    //use crate::communications::node_conversation::ClientError;
```
dnwiebe: Remove commented-out line

masq/src/communications/mod.rs
```diff
@@ -3,3 +3,22 @@ pub mod broadcast_handler;
...
+use masq_lib::messages::UiUndeliveredFireAndForget;
+use std::io::Write;
+
+fn handle_node_not_running_for_fire_and_forget_on_the_way(
+    body: UiUndeliveredFireAndForget,
+    stdout: &mut dyn Write,
+    term_interface: TerminalWrapper,
+) {
+    let _lock = term_interface.lock();
+    write!(
```
dnwiebe: Try your new macro!

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+            current_piece.push(c);
+        }
+    });
+    if !current_piece.is_empty() {
+        pieces.push(current_piece)
+    }
+    pieces
+}
+
+pub fn go_interactive<A, B, F>(
```
dnwiebe: Maybe rename A to CF and B to CP?

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+    A: CommandFactory + ?Sized + 'static,
+    B: CommandProcessor + ?Sized + 'static,
+{
+    loop {
+        let args = match processor.clone_terminal_interface().read_line() {
+            CommandLine(line) => split_quoted_line(line),
+            Break => {
+                short_writeln!(
+                    streams.stdout,
+                    "Terminated on the basis of a user's specific signal"
```
dnwiebe: Hmm. Maybe just "Terminated?"

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+            }
+        };
+        if args.is_empty() {
+            continue;
+        }
+        if args[0] == "exit" {
+            break;
+        }
+        match handle_command(command_factory, processor, args, streams.stderr) {
+            Ok(_) => (),
```
dnwiebe: How are these two options different?

Maybe just `let _ = handle_command(...`?

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+            .make_params(&make_params_arc)
+            .make_result(Ok(Box::new(FakeCommand::new("setup command"))));
+        let terminal_interface_reference_for_inner = TerminalWrapper::new(Box::new(
+            TerminalPassiveMock::new()
+                .read_line_result(TerminalEvent::CommandLine("setup\n".to_string()))
+                .read_line_result(TerminalEvent::CommandLine("exit\n".to_string())),
+        ));
+        let reference_for_counting = Arc::new(Mutex::new(0));
+        let processor = CommandProcessorMock::new()
+            .insert_terminal_interface(terminal_interface_reference_for_inner.clone())
```
dnwiebe: Please rename these to "inject" instead of "insert." "Inject" is a term of art for what you're doing here.

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+            .insert_terminal_wrapper_shared_counter(reference_for_counting.clone())
+            .process_result(Ok(()))
+            .upgrade_terminal_interface_result(Ok(()));
+
+        assert_eq!(*reference_for_counting.lock().unwrap(), 0);
```
dnwiebe: Quibble: the fact that this line is isolated by whitespace makes it look like the Act section of the test, and the fact that it's an assert leaves skidmarks on the inside of my head.

masq/src/line_reader.rs
```diff
@@ -1,242 +1,300 @@
...
+            }
+            Err(e) => TerminalEvent::Error(format!("Reading from the terminal: {}", e)),
+            Ok(ReadResult::Signal(Signal::Resize)) | Ok(ReadResult::Signal(Signal::Continue)) => {
+                TerminalEvent::Continue
+            }
+            _ => TerminalEvent::Break,
         }
     }
+
+    fn add_history_unique(&self, line: String) {
```
dnwiebe: Can this method be moved out of MasqTerminal, since it's not really a public method?

masq/src/line_reader.rs
```diff
@@ -1,242 +1,300 @@
...
+            .stdin
+            .lock()
+            .expect("poisoned mutex")
+            .read(&mut buffer)
+            .expect("reading failed");
+        let _lock = self
+            .lock
+            .lock()
+            .expect("poisoned mutex in IntegrationTestTerminal");
+        short_writeln!(self.stdout.lock().unwrap(), "{}", MASQ_PROMPT);
```
dnwiebe: I see an `unwrap()` in production code! True, it's not _really_ production code, but it looks enough like production code to be distracting.

masq/src/line_reader.rs
```diff
@@ -1,242 +1,300 @@
...
+            Arc::new(Mutex::new(Box::new(MixingStdout::new(tx_cb.clone()))));
+        let terminal = TerminalWrapper::new(Box::new(terminal_interface));
+        let mut terminal_clone = terminal.clone();
+        let (tx, rx) = std::sync::mpsc::channel();
+        let handle = thread::spawn(move || {
+            let mut background_thread_stdout = MixingStdout::new(tx_cb);
+            tx.send(()).unwrap();
+            (0..3).for_each(|_| {
+                write_one_cycle(&mut background_thread_stdout, &mut terminal_clone);
+                thread::sleep(Duration::from_millis(1))
```
dnwiebe, line 181: Maybe 0 would work here?

masq/src/line_reader.rs
```diff
@@ -1,242 +1,300 @@
...
-                .lock()
-                .unwrap()
-                .push(line.to_string());
-            self.add_history_entry_results.borrow_mut().remove(0)
-        }
+    fn write_one_cycle(stdout: &mut MixingStdout, interface: &mut TerminalWrapper) {
+        let _lock = interface.lock();
+        (0..20).for_each(|num: u8| {
+            stdout.write(num.to_string().as_bytes()).unwrap();
+            thread::sleep(Duration::from_millis(1))
```
dnwiebe, line 212: Maybe 0?

dnwiebe (commented):
Resume at non_interactive_mode.rs.

masq/src/command_context.rs
bertllll: You were right. It is not necessary anymore. I remember that the compiler prompted me to go this way so that I fulfilled some requirement and by now that obviously passed away. Good.

masq/src/command_processor.rs
bertllll: Hahaha. A sleep of 0 s may work!! I hope some other machines or platforms won't betray me.

masq/src/command_processor.rs
bertllll: Oh sorry, that I hadn't told you earlier. Actually, that has ever been tested, it is the ending part of this test.

masq/src/communications/broadcast_handler.rs
bertllll: I'm sorry 0 doesn't work. But I reduced the time lapse spent on these tests otherwise.

masq/src/command_processor.rs
bertllll: I'm sorry that is not how it works. I tried to lower the time amount spent on this test but, those seconds don't obey me, heh. Despite pressing it down on those sleeps, the duration didn't move or even got bigger.

masq/src/interactive_mode.rs
```diff
@@ -0,0 +1,437 @@
...
+    processor: &mut B,
+    streams: &mut StdStreams<'_>,
+) -> u8
+where
+    F: Fn(&A, &mut B, Vec<String>, &mut (dyn Write + Send)) -> Result<(), ()>,
+    A: CommandFactory + ?Sized + 'static,
+    B: CommandProcessor + ?Sized + 'static,
+{
+    loop {
+        let args = match processor.clone_terminal_interface().read_line() {
```
bertllll: Try remove clone_terminal_interface() and replace it with a reference to the guts

bertllll: Resolved

masq/src/line_reader.rs
bertllll: I'm sorry; it isn't.

masq/src/line_reader.rs
bertllll: Nope.

masq/src/non_interactive_mode.rs
```diff
@@ -0,0 +1,416 @@
...
+        }
+    }
+
+    fn extract_subcommand(args: &[String]) -> Option<Vec<String>> {
+        let args_vec: Vec<String> = args.to_vec();
+        for idx in 1..args_vec.len() {
+            let one = &args_vec[idx - 1];
+            let two = &args_vec[idx];
+            //tested by an integration test
+            if &args_vec[idx] == "--help" {
```
dnwiebe: I don't think this is necessary here, at least for production code. Investigate why integration tests require it here.

masq/src/non_interactive_mode.rs
```diff
@@ -0,0 +1,416 @@
...
+                return Some(vec!["--help".to_string()]);
+            }
+            if !one.starts_with("--") && !two.starts_with("--") {
+                return Some(args_vec.into_iter().skip(idx).collect());
+            }
+        }
+        None
+    }
+
+    pub fn populate_non_interactive_dependencies() -> (Box<dyn BroadcastHandle>, TerminalWrapper) {
```
dnwiebe: Try making these two functions non-public.

masq/src/non_interactive_mode.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        let output_when_unlocked = test_stream_handle.stdout_so_far();
+        assert_eq!(
+            output_when_unlocked,
+            "\nThe Node\'s database password has changed.\n\n"
+        )
+    }
+
+    #[test]
+    fn populate_interactive_dependencies_produces_a_functional_broadcast_handle() {
```
dnwiebe: This test is probably not necessary in light of the test above.

dnwiebe (commented):
Resume with terminal_interface.rs

masq/src/non_interactive_mode.rs
bertllll: Seems like neither integration tests need it anymore. I'm cutting it out.

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+
+impl Clone for TerminalWrapper {
+    fn clone(&self) -> Self {
+        Self {
+            interface: Arc::clone(&self.interface),
+        }
+    }
+}
+
+fn interface_configurator<F, U, E: ?Sized, D>(
```
dnwiebe: Can this be made simpler?

Please?

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+        };
+
+    if let Err(e) = set_all_settable_or_give_an_error(&mut *interface) {
+        return Err(e);
+    };
+
+    Ok(TerminalReal::new(interface))
+}
+
+fn set_all_settable_or_give_an_error<U>(interface: &mut U) -> Result<(), String>
```
dnwiebe: There might be a better name for this...maybe "configure_interface" or some such?

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+    //here we can add another parameter to be configured,
+    //such as "completer" (see linefeed library)
+
+    Ok(())
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub trait MasqTerminal {
+    fn provide_lock(&self) -> Box<dyn WriterLock + '_> {
```
dnwiebe: Make these two non-optional? I think I mean to say: mandatory.

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+    #[cfg(test)]
+    fn test_interface(&self) -> MemoryTerminal {
+        intentionally_blank!()
+    }
+    #[cfg(test)]
+    fn tell_me_who_you_are(&self) -> String {
+        intentionally_blank!()
+    }
+}
+//declaration of TerminalReal is in line_reader.rs
```
dnwiebe: Instead of specifying a destination, just refer to the struct name

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+#[derive(Default)]
+pub struct TerminalInactive {}
+
+impl MasqTerminal for TerminalInactive {
+    fn provide_lock(&self) -> Box<dyn WriterLock + '_> {
+        Box::new(WriterInactive {})
+    }
+    #[cfg(test)]
+    fn tell_me_who_you_are(&self) -> String {
+        "TerminalIdle".to_string()
```
dnwiebe: Name should be the same as the struct. Suggestion for the name of the struct: TerminalNoninteractive. Then the intention of the struct would be clearer.

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub trait WriterLock {
+    //I failed in attempts to use 'Any' and dynamic casting from Box<dyn WriterGeneric>
```
dnwiebe: Bert has opinions about this comment

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+    //because: Writer doesn't implement Clone and many if not all methods of Any require
+    //'static, that is, it must be an owned object and I cannot get anything else but a referenced
+    //Writer there.
+    //For delivering at least some tests I decided to use this, kind of a hack.
+    #[cfg(test)]
+    fn tell_me_who_you_are(&self) -> String {
+        intentionally_blank!()
```
dnwiebe: Is this the right thing to do, or should this be mandatory?

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+pub trait InterfaceRaw {
+    fn read_line(&self) -> std::io::Result<ReadResult>;
+    fn add_history_unique(&self, line: String);
+    fn lock_writer_append(&self) -> std::io::Result<Box<dyn WriterLock + '_>>;
+    fn set_prompt(&self, prompt: &str) -> std::io::Result<()>;
+}
+
+impl<U: linefeed::Terminal + 'static> InterfaceRaw for Interface<U> {
```
dnwiebe: Is this still necessary?

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+            });
+
+        let given_output = test_terminal_collision(Box::new(closure1), Box::new(closure2));
+
+        //in an extreme case it may be printed like one group is complete and the other is divided
+        let results = [
+            given_output.contains(&"A".repeat(90)),
+            given_output.contains(&"B".repeat(90)),
+        ];
+        let at_least_one = results.iter().find(|bool| **bool == false);
```
dnwiebe: Investigate the any() method: it may do what you want more cleanly. Also: don't name a variable "bool."

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    #[test]
+    fn configure_interface_allows_us_starting_in_memory_terminal() {
+        let term_mock = MemoryTerminal::new();
+        let term_mock_clone = term_mock.clone();
+        let terminal_type = move || -> std::io::Result<MemoryTerminal> { Ok(term_mock_clone) };
+        let subject =
+            interface_configurator(Box::new(Interface::with_term), Box::new(terminal_type));
+        let _ = match subject {
```
dnwiebe: You could probably  use .unwrap() here and get almost the same functionality the match gives you, but simpler.

masq/src/terminal_interface.rs
```diff
@@ -0,0 +1,439 @@
...
+        //Windows: The handle is invalid. (os error 6)
+        //Linux: "Getting terminal parameters: Inappropriate ioctl for device (os error 25)"
+    }
+
+    #[test]
+    fn configure_interface_allows_us_starting_in_memory_terminal() {
+        let term_mock = MemoryTerminal::new();
+        let term_mock_clone = term_mock.clone();
+        let terminal_type = move || -> std::io::Result<MemoryTerminal> { Ok(term_mock_clone) };
+        let subject =
```
dnwiebe: Is this really a subject? It looks like a result. Maybe the subject is actually interface_configurator, which doesn't get a variable.

masq/src/test_utils/mocks.rs
```diff
@@ -162,6 +171,7 @@ pub struct CommandProcessorMock {
     process_params: Arc<Mutex<Vec<Box<dyn Command>>>>,
     process_results: RefCell<Vec<Result<(), CommandError>>>,
     close_params: Arc<Mutex<Vec<()>>>,
+    terminal_interface: Vec<TerminalWrapper>,
```
dnwiebe: Maybe this could be either just a TerminalWrapper or an Option<TerminalWrapper>

masq/src/test_utils/mocks.rs
```diff
@@ -205,7 +227,8 @@ pub struct CommandProcessorFactoryMock {
 impl CommandProcessorFactory for CommandProcessorFactoryMock {
     fn make(
         &self,
-        _broadcast_stream_factory: Box<dyn StreamFactory>,
+        _terminal_interface: TerminalWrapper,
```
dnwiebe: Should these be stored in a make_params field, and asserted on in a test?

masq/src/test_utils/mocks.rs
```diff
@@ -352,9 +393,166 @@ impl TestStreamFactoryHandle {
                     }
                     thread::sleep(Duration::from_millis(100));
                 }
-                Err(_) => break,
+                Err(_) => {
```
dnwiebe: Extra curly braces

masq/src/test_utils/mocks.rs
```diff
@@ -352,9 +393,166 @@ impl TestStreamFactoryHandle {
...
+            .send(std::str::from_utf8(buf).unwrap().to_string())
+            .unwrap();
+        Ok(0)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+    fn write_fmt(&mut self, fmt: Arguments<'_>) -> std::io::Result<()> {
+        self.channel_half.send(fmt.to_string()).unwrap();
```
dnwiebe, line 506: Maybe this should call self.write()?

masq/src/test_utils/mocks.rs
```diff
@@ -352,9 +393,166 @@ impl TestStreamFactoryHandle {
...
+
+    pub fn read_line_result(self, result: TerminalEvent) -> Self {
+        self.read_line_result.lock().unwrap().push(result);
+        self
+    }
+}
+
+//mock incorporating a terminal very similar to the real one, which is run as in-memory though
+
+pub struct TerminalActiveMock {
```
dnwiebe, line 542: Is this still necessary?

dnwiebe (commented):
Resume with communication_tests_integration.rs

masq/src/test_utils/mocks.rs
bertllll: That may be more complicated than now in the end.

masq/src/test_utils/mocks.rs
bertllll: Yes...I had to say to myself that writing my own mock that'd substitute this one would be too demanding. It's 50/50 with benefits from that.

masq/src/terminal_interface.rs
bertllll: I disagree, since one of three takers of this trait doesn't need it anywhere to use. I had to link in a comment explaining why those structures are so complicated. If I knew how to get rid of them I'd do that happily.

masq/src/terminal_interface.rs
bertllll: First I thought it can. Now, I'm afraid, this is how it must stay. I've added an explanation into the code recently and we'll discuss it.

masq/tests/communication_tests_integration.rs
```diff
@@ -9,30 +9,48 @@ use std::time::Duration;
...
+         "Should see 'Daemon setup has changed' at the receiver; instead, saw '{}' at the receiver and '{}' at the setupper.",
+         stdout_receiver,
+         stdout_setupper
     );
+
+    //TODO the following lines are here to cause a failure once GH-438 fixes what it should fix; Please, remove them then
+    let full_output_length = stdout_receiver.len();
+    let wanted_line_length = "Daemon setup has changed".len();
+    let stdout_receiver_without_the_message =
+        stdout_receiver.replace("Daemon setup has changed", "");
```
dnwiebe, line 48: Use replace_all() here. Maybe also modify this comment so that when somebody reads it, he's directed to the //TODO comment above and knows how to modify the test, rather than just removing the assertion that's failing.

masq/tests/startup_shutdown_tests_integration.rs
```diff
@@ -25,51 +26,77 @@ fn masq_without_daemon_integration() {
...
         "error",
         "--neighborhood-mode",
         "zero-hop",
-        "--chain",
     ]);
 
     let (stdout, stderr, exit_code) = masq_handle.stop();
 
-    assert_eq!(&stderr, "", "{}", stderr);
+    assert_eq!(&stderr, "", "ln. 70: {}", stderr);
```
dnwiebe: Change the "ln. 70" message so that it is either gone, or will not drift into incorrectness.

masq/tests/startup_shutdown_tests_integration.rs
```diff
@@ -25,51 +26,77 @@ fn masq_without_daemon_integration() {
...
     assert_eq!(exit_code, 0);
 
-    let masq_handle = MasqProcess::new().start_noninteractive(vec!["start"]);
+    let masq_handle =
+        MasqProcess::new().start_noninteractive(vec!["--ui-port", &port.to_string(), "start"]);
 
     let (stdout, stderr, exit_code) = masq_handle.stop();
 
-    assert_eq!(&stderr, "", "{}", stderr);
+    assert_eq!(&stderr, "", "ln. 83: {}", stderr);
```
dnwiebe: Same here

masq/tests/startup_shutdown_tests_integration.rs
```diff
@@ -25,51 +26,77 @@ fn masq_without_daemon_integration() {
...
     assert_eq!(exit_code, 0);
 
-    let masq_handle = MasqProcess::new().start_noninteractive(vec!["shutdown"]);
+    let masq_handle =
+        MasqProcess::new().start_noninteractive(vec!["--ui-port", &port.to_string(), "shutdown"]);
 
     let (stdout, stderr, exit_code) = masq_handle.stop();
 
-    assert_eq!(&stderr, "", "{}", stderr);
+    assert_eq!(&stderr, "", "ln. 96: {}", stderr);
```
dnwiebe: Same here

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -186,6 +186,24 @@ impl MockWebSocketsServer {
                                     client.send_message(&om).unwrap()
                                 }
                             },
+                            MessagePath::FireAndForget
+                                if message_body.opcode == "broadcastTrigger" =>
+                            {
+                                log(
+                                    do_log,
+                                    index,
+                                    "Responding to a request for FireAndForget message in dirrection to UI",
```
dnwiebe: "direction" is misspelled

masq_lib/src/utils.rs
```diff
@@ -119,6 +119,13 @@ macro_rules! short_writeln {
     };
 }
 
+#[macro_export]
+macro_rules! intentionally_blank {
+    () => {
+        panic!("unrequired method of a trait; left blank here")
```
dnwiebe: Maybe something like "Required method left unimplemented: should never be called."

node/src/daemon/mod.rs
```diff
@@ -258,14 +261,16 @@ impl Daemon {
                         },
                         payload: match body.payload {
                             Ok(json) => json,
-                            Err((_code, _message)) => unimplemented!(),
+                            Err((_code, _message)) => {
+                                panic!("Incoming message from UI with a payload signaling an error")
```
dnwiebe: A bad message from a UI shouldn't crash the Daemon. Put unimplemented!() back here (or maybe the new todo!() if you like that better) and we'll make a card to determine what should be done here.

dnwiebe (commented):
Resume in node/src/daemon/mod.rs at accepts_shutdown_order_discovers_non_running_node_and_returns_redirect_error() (line 1421)

masq/tests/communication_tests_integration.rs
bertllll: It was a mystification, we remember replace_all() although it applies only for Regex, the string-for replace() works for ALL occurrences.

node/src/daemon/mod.rs
```diff
@@ -1408,13 +1416,11 @@ mod tests {
     }
 
     #[test]
-    fn accepts_shutdown_order_discovers_non_running_node_and_returns_redirect_error() {
-        let (ui_gateway, _, _) = make_recorder();
+    fn accepts_unexpected_message_discovers_non_running_node_and_returns_conversational_answer_of_error(
```
dnwiebe, line 1433: This test no longer asserts that `self.node_process_id` and `self.node_ui_port` are set back to `None` when the Daemon discovers that the Node has unexpectedly stopped.

Add another test, perhaps called `remembers_unexpected_node_crash`:

Create a Daemon with `node_process_id` and `node_ui_port` set.
Inject a `VerifierToolsMock` that will reply `false` to `process_is_running`.

Call `subject.port_if_node_is_running()`

Assert that the result is `false`.
Assert that `subject.node_process_id` and `subject.node_ui_port` are now `None`.

dnwiebe, line 1433: This test, since it now tests something different, may no longer need to send and assert two messages: one may be enough now.

node/src/daemon/mod.rs
```diff
@@ -1570,4 +1651,21 @@ mod tests {
         let ui_gateway_recording = ui_gateway_recording_arc.lock().unwrap();
         assert_eq!(ui_gateway_recording.len(), 0);
     }
+
+    #[test]
+    #[should_panic(expected = "Incoming message from UI with a payload signaling an error")]
+    fn handle_unexpected_message_panics_if_receiving_payload_from_ui_being_an_error() {
```
dnwiebe: This will change when we stop panicking because the UI sent us a bad message. Or, if the `panic!()` in the production code has been changed back to `unimplemented!()`, then this test should be removed.

node/src/daemon/mod.rs
```diff
@@ -301,20 +304,24 @@ impl Daemon {
...
             client_id
         );
-        let body = MessageBody {
-            opcode: "redirect".to_string(),
-            path: FireAndForget,
-            payload: Err((
-                NODE_NOT_RUNNING_ERROR,
-                format!("Cannot handle {} request: Node is not running", err_opcode),
-            )),
+        let body = match path {
```
dnwiebe, line 314: This change introduces a new requirement that Conversations must consist of Request/Response pairs that have the same opcode. If the Response opcode is different from the Request opcode, then the error packet sent back (which the UI will understand as a Response) will have the Request opcode, not the expected Response opcode, and there will be problems. Implicitly, this also prohibits the use of multiple Responses (with different opcodes) to a single Request.

This new requirement should be documented somewhere, preferably in `USER-INTERFACE-INTERFACE.md`.

corey7754 (approved):

corey7754 (approved):

corey7754 (approved):

corey7754 (approved):

dnwiebe (approved):

masq/src/notifications/crashed_notification.rs
```diff
@@ -14,10 +14,12 @@ impl CrashNotifier {
         stdout: &mut dyn Write,
         term_interface: &TerminalWrapper,
     ) {
+        let _lock = term_interface.lock();
         if response.crash_reason == CrashReason::DaemonCrashed {
-            exit_process(1, "The Daemon is no longer running; masq is terminating.\n");
+            exit_process_with_sigterm("\nThe Daemon is no longer running; masq is terminating.\n")
+
+          //  exit_process(1, "The Daemon is no longer running; masq is terminating.\n");
```
dnwiebe: Remove comment

masq_lib/src/utils.rs
```diff
@@ -109,12 +111,15 @@ pub fn exit_process(code: i32, message: &str) {
...
-pub fn exit_process_with_sigterm(message:&str) {
+#[cfg(not(target_os = "windows"))]
+pub fn exit_process_with_sigterm(message: &str) {
     if unsafe { RUNNING_TEST } {
         panic!("{}", message);
     } else {
         eprintln!("{}", message);
-      //  nix::sys::signal::raise(nix::sys::signal::SIGTERM).expect("sigterm failure");
+        signal::raise(signal::SIGTERM).expect("sigterm failure");
+        //making sure that this thread won't cause a panic in the shutdown sequence
```
dnwiebe: Maybe something like, "This function must not return; so wait for death."

PR #13, opened 2021-05-04
GH-386: Sixth review pass

dnwiebe:


masq/src/non_interactive_mode.rs
```diff
@@ -31,34 +32,31 @@ impl Default for Main {
...
-        let vec_candidate = args
-            .to_vec()
-            .into_iter()
-            .skip(1 + Self::eliminate_pre_subcommand_items(args[1].as_str()))
-            .collect::<Vec<String>>();
-        if !vec_candidate.is_empty() {
-            Some(vec_candidate)
+        let original = args.iter();
+        let one_item_shifted_forth = args.iter().skip(1);
+        let index = original.zip(one_item_shifted_forth).enumerate().find(
```
dnwiebe, line 44: Don't call this "index," because "index" means a number. Either call it "index_pair" or something similar, or put a .0 at the end of the find()

masq/src/non_interactive_mode.rs
```diff
@@ -31,34 +32,31 @@ impl Default for Main {
...
+            |(_position, (orig, shifted))| Self::both_do_not_start_with_two_dashes(orig, shifted),
```
dnwiebe, line 45: Suggestion: instead of `orig` and `shifted`, how about `first` and `second` or `left` and `right` or something that definitely tells a reader what the order is rather than making him wonder which way the `shifted` one is shifted.

masq/src/non_interactive_mode.rs
```diff
@@ -31,34 +32,31 @@ impl Default for Main {
...
+        );
+        if let Some((index, _)) = index {
```
dnwiebe, line 47: This might be cleaner with index.map(). Maybe not, but it's worth a look.

masq/src/non_interactive_mode.rs
```diff
@@ -221,10 +219,10 @@ mod tests {
             &[
                 "command",
                 "subcommand",
-                "--param3",
-                "value3",
-                "param4",
-                "param5",
+                "--param1",
+                "value1",
+                "param2",
```
dnwiebe, line 224: Why doesn't this param start with two dashes too?

masq/tests/interactive_mode_help_and_version_integration.rs
```diff
@@ -67,7 +67,7 @@ fn interactive_mode_allows_a_version_call_integration() {
     daemon_handle.kill();
     //TODO put this assertion back when GH-446 is played out - paired with the test above
     //assert_eq!(stderr, "");
-    let regex = Regex::new(r"masq> \nmasq \d.\d.\d\nmasq> ").unwrap();
+    let regex = Regex::new(r"masq> \nmasq [1-255]\.[0-255]\.[0-255]\nmasq> ").unwrap();
```
dnwiebe, line 70: \d+\.\d+\.\d+

masq/tests/non_interactive_mode_help_and_version_integration.rs
```diff
@@ -30,7 +30,7 @@ fn masq_non_interactive_version_command_integration() {
     let (stdout, stderr, exit_code) = masq_handle.stop();
 
     assert_eq!(stderr, "");
-    let regex = Regex::new(r"masq \d.\d.\d\n").unwrap();
```
dnwiebe, line 33: Same thing...

dnwiebe (commented):
Resume with mock_websockets_server.rs

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -66,6 +66,7 @@ impl MockWebSocketsServer {
         self
     }
 
+    // I did't want to write a special test for this as it's already used in a test from command_processor() and works good
```
dnwiebe, line 69: Sure: the only reason you need to write tests for test code is if the test code is complex enough that you don't know whether it works or not. This method is obviously not complex; I'd say you probably don't need this explanatory comment.

PR #14, opened 2021-05-06
Gh 386 7pfreeze

dnwiebe:


masq/src/interactive_mode.rs
```diff
@@ -381,13 +381,16 @@ mod tests {
         let background_interface_clone = terminal_interface.clone();
         let mut stdout = ByteArrayWriter::new();
         let (tx, rx) = bounded(1);
+        let (tx_back, rx_back) = bounded(1);
         let handle = thread::spawn(move || {
```
dnwiebe, line 385: I think you can delete one of these tests.

masq/src/non_interactive_mode.rs
```diff
@@ -39,16 +39,13 @@ impl Main {
...
-        );
-        if let Some((index, _)) = index {
-            Some(args.to_vec().into_iter().skip(index + 1).collect())
-        } else {
-            None
-        }
+        original_args
+            .zip(one_item_shifted_forth)
+            .enumerate()
+            .find(|(_position, (left, right))| Self::both_do_not_start_with_two_dashes(left, right))
```
dnwiebe, line 47: Suggestion: the `_position` and `index` variables should have the same name (aside from the `_`) to make it clear that they refer to the same thing.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+    }
+
+    fn handle_conversational_incoming_message(
+        client: &mut Client<TcpStream>,
+        message_body: MessageBody,
+        inner_responses_arc: &Arc<Mutex<Vec<OwnedMessage>>>,
+        context_id: u64,
+        do_log: bool,
+        index: u64,
+    ) -> u16 {
```
dnwiebe, line 255: Maybe a boolean return value? Or if not a boolean value, then perhaps a two-valued enum with values named something like `Stop` and `Continue`?

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -163,119 +141,31 @@ impl MockWebSocketsServer {
...
-                                    }
-                                }
-                                om => {
-                                    log(
-                                        do_log,
-                                        index,
-                                        &format!("Responding with preset OwnedMessage: {:?}", om),
-                                    );
-                                    client.send_message(&om).unwrap()
+                            MessagePath::Conversation(context_id) => {
```
dnwiebe, line 144: Maybe these outer braces could go away

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+                om => {
+                    log(
+                        do_log,
+                        index,
+                        &format!("Responding with preset OwnedMessage: {:?}", om),
+                    );
+                    client.send_message(&om).unwrap()
+                }
+            }
+            //code that can be interpreted as an empty queue
```
dnwiebe, line 304: Please move this comment down into the `else` clause so that it's easy to see what it references.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+        } else {
+            client
```
dnwiebe, line 306: Two suggestions:

1. Instead of having one byte in the Binary message, make it contain an entire message: perhaps `b"EMPTY QUEUE"` or something.
2. Also put a log() call in here so that something appropriate appears in the test output, if logs are turned on.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+        index: u64,
+    ) {
+        log(
+            do_log,
+            index,
+            "Responding to a request for FireAndForget message in direction to UI",
+        );
+        let queued_messages = &mut *inner_responses_arc.lock().unwrap();
+        let (positional_number_of_the_signal_sent_opt,signal_sender_opt, batch_size_of_broadcasts_to_be_released_at_once) =
+            match (UiBroadcastTrigger::fmb(message_body),self.signal_sender.take()) {
```
dnwiebe, line 329: Maybe clone the signal_sender rather than taking it, so it could theoretically be used more than once?

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+            (Ok((trigger_message, _)), Some(sender)) => match trigger_message.position_to_send_the_signal_opt {
```
dnwiebe, line 330: You can combine this match with the next one, since everything in `trigger_message` is available here.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+                Some(position) => match trigger_message.number_of_broadcasts_in_one_batch {
+                    Some(demanded_batch_size) => (Some(position), Some(sender), demanded_batch_size),
```
dnwiebe, line 332: Here you might want to return the lesser of `demanded_batch_size` and `queued_messages.len()` to avoid overruns in cases where the `demanded batch size` limit is accidentally larger than the number of queued messages remaining. On the other hand, you might instead choose to fail the test here to cudgel the tester into writing his test correctly.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+                    None => (Some(position), Some(sender), queued_messages.len())},
+                None => panic!("You provided a Sender<()> but forgot to provide the positional number of the broadcast where it should be sent; settable within the trigger message"),
+            },
+            (Ok((trigger_message, _)), None) => match trigger_message.position_to_send_the_signal_opt {
+                Some(_) => panic!("You require to send a signal but haven't provided Sender<()> by inject_signal_sender()"),
+                None => match trigger_message.number_of_broadcasts_in_one_batch {
+                    Some(demanded_batch_size) => (None, None, demanded_batch_size),
```
dnwiebe, line 339: The same decision should be made here about `min(demanded_batch_size, queued_messages.len())`

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+                        //////////////////////////////////////////////////////////////////////
+                        client.send_message(&queued_messages.remove(0)).unwrap();
+                        already_sent += 1;
+                        if already_sent == batch_size_of_broadcasts_to_be_released_at_once {
+                            break;
+                        }
+                        factor_of_position_reduction += 1;
+                        /////////////////////////////////////////////////////////////////////////////////////////////////
+
+                        //let's end it; we ran into a conversational message in the queue
```
dnwiebe, line 371: Again, please move this comment into the else clause

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+        let mut factor_of_position_reduction = 0_usize; //because I remove each message after I send it
+        let starting_length = queued_messages.len();
+        for i in 0..starting_length {
+            //sending signal if wanted ////////////////////////////
+            if let Some(position) = positional_number_of_the_signal_sent_opt {
+                if position == i {
+                    signal_sender_opt.as_ref().unwrap().send(()).unwrap()
+                }
+            }
+            //filtering broadcasts only from the queue ///////////////////////////////////
```
dnwiebe, line 358: Change this comment to match the code: you're no longer filtering, you're just...well, I can't think of a good word for it now, but not filtering.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -339,6 +215,195 @@ impl MockWebSocketsServer {
...
+                    None => (None, None, queued_messages.len())
+                }
+            },
+            (_,_) => panic!("BroadcastTrigger received but somehow malformed")
+        };
+        let mut already_sent = 0_usize;
+        //////////////////////////////////////////////////////////////////////////////////////////////////////////
+        //here the own algorithm carrying out messaging starts
+
+        let mut factor_of_position_reduction = 0_usize; //because I remove each message after I send it
```
dnwiebe, line 349: Maybe you can combine `already_sent` and `factor_of_position_reduction`: it looks like they always have the same value.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -422,65 +488,41 @@ mod tests {
                 ("param1".to_string(), "reason1".to_string()),
                 ("param2".to_string(), "reason2".to_string()),
             ],
-        }
```
dnwiebe, line 425: After some cogitation, I think three things:
1

1. This test should not test two messages. There should be one test for the happy-path message, and if the sad-path message is to be tested, there should be a separate test for it. Having the two together makes things too confused. If you separate the two tests, make sure to name this test appropriately.
2. The sad-path message probably doesn't require a test, as it should only ever happen when somebody accidentally writes a bad test using the MockWebsocketsServer. Put a `panic!` in the server where it can't unmarshal the message, and either remove the sad-path test entirely or make it look for that `panic!`. (My preference would be to remove it entirely.)
3. (Actually, this is your thought, not mine) The remaining happy-path test is also unnecessary, because its functionality is subsumed by the next test below.

So...wholesale deletion of this test seems to be in order.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -490,120 +532,205 @@ mod tests {
...
-                value: "to UI".to_string(),
-                status: Set,
-            }],
-            errors: vec![],
-        };
-        let expected_ui_setup_response = UiSetupResponse {
-            running: true,
-            values: vec![],
-            errors: vec![],
+    fn conversational_and_broadcast_messages_can_work_together_testing_corner_cases() {
```
dnwiebe, line 535: If this test is going to be depended on to (among all the other things it does) take the place of `two_in_two_out()` above, it should check the guts of at least one request and at least one response. On the other hand, if you're happy with the responsibilities of this test as they stand, then leave the happy-path parts of `two_in_two_out()` above.

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -490,120 +532,205 @@ mod tests {
...
+            (u64, String),
+        > = connection.transact_with_context_id(conversation_hopeless_attempt_in_bad_time, 10000);
 
-        //now let's use a UiBroadcastTrigger again, this time without count boundaries. We should get all the remaining broadcasts from the queue
-        let second_broadcast_trigger = UiBroadcastTrigger {
-            number_of_broadcasts_in_one_batch: None,
-            position_to_send_the_signal_opt: None,
-        };
-        connection.send(second_broadcast_trigger.clone());
+        //sending another broadcast trigger (unlimited) to get the third, fourth and sixth message
```
dnwiebe, line 655: Third, fourth, and fifth?

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -490,120 +532,205 @@ mod tests {
...
+
+        let _ = stop_handle.stop();
+        ////////////////////////////////////////////////////////////////////////////////////////////
+        //assertions for deliberately caused errors
+        let error_message_number_one = naive_attempt_number_one_to_receive_the_third_broadcast
+            .unwrap_err()
+            .1;
+        assert!(
+            error_message_number_one.contains(
+                "Expected a corresponding response pulled out from the queue. \
```
dnwiebe, line 695: Make it clear that it was a broadcast (not a conversation) that we were looking for and couldn't find.

dnwiebe (commented):
Resume with ui_connection.rs

masq_lib/src/test_utils/mock_websockets_server.rs
bertllll: This is a problem. This message actually doesn't arrive from the server - it's generated on the client's end - and technically means that we've waited too long and aborted from that process otherwise it'd have blocked forever.

masq_lib/src/test_utils/mock_websockets_server.rs
bertllll: Maybe this is not intuitive but I'm confident that mine solution works. It always iterates over the actual queue size, counting messages already sent and asking in every iteration if that number equals to the desired batch size, the loop ends always before we could get bigger numbers than the length of the queue.

masq_lib/src/test_utils/mock_websockets_server.rs
bertllll: Same as above, no issue actually...I think.

masq_lib/src/test_utils/ui_connection.rs
```diff
@@ -60,13 +65,29 @@ impl UiConnection {
...
     pub fn receive<T: FromMessageBody>(&mut self) -> Result<T, (u64, String)> {
-        let incoming_msg = self.client.recv_message();
-        let incoming_msg_json = match incoming_msg {
-            Ok(OwnedMessage::Text(json)) => json,
-            x => panic!("Expected text; received {:?}", x),
+        let mut failure_state_holder: Option<WebSocketResult<OwnedMessage>> = None;
+        let start_instant = Instant::now();
+        let incoming_msg_json = loop {
+            if start_instant.elapsed() > Duration::from_millis(150) {
+                //a way to inform that the queque was probably empty, without panicking
```
dnwiebe, line 72: "Queue" is misspelled

masq_lib/src/test_utils/ui_connection.rs
```diff
@@ -60,13 +65,29 @@ impl UiConnection {
...
+                        failure_state_holder
+                    ),
+                ));
+            }
+            match self.client.recv_message() {
+                Ok(OwnedMessage::Binary(numeric_code)) if numeric_code == vec![101] => {
+                    return Err((0, "The queue is empty; all messages are gone.".to_string()))
+                }
+                Ok(OwnedMessage::Text(json)) => break json,
+                x => failure_state_holder = Some(x),
```
dnwiebe, line 86: I'd like to see TimedOut and WouldBlock (meaning, "We're still waiting; a message hasn't arrived yet") handled differently here, so that the loop isn't quite so tight. Maybe if one of those is received, you could delay for ten milliseconds or so before trying again?

bertllll: I really like this suggestion; helped me make clearer responses.

masq_lib/src/test_utils/ui_connection.rs
bertllll: is passé already

PR #15, opened 2021-05-12
Gh 386 8pfreeze

dnwiebe:


masq_lib/src/test_utils/ui_connection.rs
```diff
@@ -64,32 +69,55 @@ impl UiConnection {
...
+                    if std::str::from_utf8(&bytes).unwrap() == "EMPTY QUEUE" =>
+                {
+                    return make_contextually_incorrect_but_polite_error_message_indicating_a_test_s_failure(
+                        "The queue is empty; all messages are gone.".to_string(),
+                    )
+                },
+
                 Ok(OwnedMessage::Text(json)) => break json,
+
                 x => failure_state_holder = Some(x),
```
dnwiebe, line 100: You might want special treatment here for TimedOut and WouldBlock, because they're not failure states. If you put them in `failure_state_holder`, somebody who sees a test failure with them might spend a lot of time trying to track down an error that doesn't actually exist.

masq/src/interactive_mode.rs
```diff
@@ -12,16 +12,16 @@ use masq_lib::command::StdStreams;
...
 where
-    HC: Fn(&CF, &mut CP, Vec<String>, &mut (dyn Write + Send)) -> Result<(), ()>,
-    CF: CommandFactory + ?Sized + 'static,
-    CP: CommandProcessor + ?Sized + 'static,
+    HC: Fn(&CF, &mut CP, Vec<String>, &mut dyn Write) -> Result<(), ()>,
+    CF: CommandFactory,
+    CP: CommandProcessor,
 {
     loop {
         let read_line_result = processor.terminal_wrapper_ref().read_line();
```
dnwiebe, line 27: Make sure all four options of this `match` are tested...

masq/src/interactive_mode.rs
```diff
@@ -113,7 +112,7 @@ mod tests {
     use masq_lib::test_utils::fake_stream_holder::{ByteArrayWriter, FakeStreamHolder};
     use std::sync::{Arc, Mutex};
     use std::thread;
-    use std::time::{Duration, Instant};
+    use std::time::Instant;
```
dnwiebe, line 115: Comment located here because I can't locate it below: Some of these tests--the ones that have `Main` as subject--should be split apart. The functionality that proves that Main calls `go_interactive()` when given the correct parameters should be in `main.rs`, and the functionality that proves the correct operation of `go_interactive()` should be here in this file, but those tests should call `go_interactive()` directly, rather than through Main.

node/tests/tls_through_node_test.rs
```diff
@@ -14,6 +14,7 @@ use std::thread;
 use std::time::Duration;
 
 #[test]
+#[ignore]
```
dnwiebe, line 17: Remove

masq_lib/src/test_utils/mock_websockets_server.rs
```diff
@@ -230,6 +231,10 @@ impl MockWebSocketsServer {
                 log(do_log, index, "No message waiting");
                 None
             }
+            Err(WebSocketError::IoError(e)) if e.kind() == std::io::ErrorKind::ConnectionReset => {
```
dnwiebe, line 234: This is weird. If the peer has reset the connection, there won't be any more data until another connection is set up--or at least there shouldn't be unless the peer is really insane and the operating system is unusually indulgent. Maybe this is no longer required?

bertllll: happily, it can be removed

PR #18, opened 2021-05-20
Gh 386

dnwiebe:


masq/src/interactive_mode.rs
```diff
@@ -89,33 +95,45 @@ fn handle_help_or_version(
         "version" => app()
             .write_version(&mut stdout)
             .expect("masq version set incorrectly"),
-        _ => return false,
+        _ => return,
```
dnwiebe: Prefer to see an unreachable!() here, since it means there's a problem with the code that should be fixed.

masq/src/terminal_interface.rs
```diff
@@ -1,18 +1,22 @@
...
-#[cfg(not(test))]
-use linefeed::DefaultTerminal;
 use linefeed::{Interface, ReadResult, Writer};
 use masq_lib::constants::MASQ_PROMPT;
-#[cfg(test)]
-use masq_lib::intentionally_blank;
 use std::sync::Arc;
 
+#[cfg(not(test))]
+mod not_test_cfg {
```
dnwiebe: Maybe consider `prod_cfg` here instead of `not_test_cfg`?

masq/tests/startup_shutdown_tests_integration.rs
```diff
@@ -60,6 +60,30 @@ fn masq_propagates_errors_related_to_default_terminal_integration() {
     );
 }
 
+#[test]
+fn masq_terminates_in_reaction_to_loss_of_the_connection_with_daemon_or_node_integration() {
```
dnwiebe: This test name is a little misleading. What this test proves is that when `masq` is connected to the Daemon, and the Daemon goes down, `masq` terminates. The Node isn't involved here.

masq_lib/src/utils.rs
```diff
@@ -109,6 +115,18 @@ pub fn exit_process(code: i32, message: &str) {
...
 
+#[cfg(not(target_os = "windows"))]
+pub fn exit_process_with_sigterm(message: &str) {
+    if unsafe { RUNNING_TEST } {
+        panic!("{}", message);
+    } else {
+        eprintln!("{}", message);
+        not_win_cfg::signal::raise(not_win_cfg::signal::SIGTERM).expect("sigterm failure");
+        //This function must not return; so wait for death.
+        std::thread::sleep(not_win_cfg::Duration::from_millis(300))
```
dnwiebe: Maybe this should be longer, in case it takes the SIGTERM handler longer to close everything out. Could be days and months and years, for that matter: it should never complete, ever.

PR #24, opened 2021-06-22
Another GH-244 review PR - don't merge

dnwiebe:


node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
     lazy_static! {
         static ref TEST_DIRECTORY_FOR_DB_MIGRATION: PathBuf =
             PathBuf::from(format!("{}/db_migration", BASE_TEST_DIR));
-    }
 
-    #[test]
-    #[should_panic(expected = "Database is corrupted: schema version syntax")]
-    fn validation_check_panics_if_the_given_schema_has_wrong_syntax() {
-        let _ = DbMigratorReal::schema_initial_validation_check("0.xx.5");
+        static ref CHANGABLE_PANIC_MESSAGE:String = format!("Database claims to be more advanced ({}) than the version {} which is the latest released.",CURRENT_SCHEMA_VERSION + 1,CURRENT_SCHEMA_VERSION);
```
dnwiebe, line 400: CHANGEABLE

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+                target_version,
+                Box::new(migration_utilities),
+                DbMigratorReal::list_of_existing_updates(),
+            )
+        }))
+        .unwrap_err();
 
-        assert_eq!(result,Err(format!("Database claims to be more advanced ({}) than the version {} which is the latest released.",too_advanced,CURRENT_SCHEMA_VERSION)))
+        let panic_message = captured_panic.downcast_ref::<String>().unwrap();
+        assert_eq!(*panic_message,format!("Database claims to be more advanced ({}) than the version {} which is the latest released.",too_advanced,CURRENT_SCHEMA_VERSION))
```
dnwiebe, line 439: Instead of "which is the latest released," maybe something like, "which is the latest version this Node knows about."

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+        let iterator_shifted = list_of_updates.iter().skip(1);
+        iterator.zip(iterator_shifted).for_each(|(first, second)| {
+            if compare_two_numbers(first.old_version(), second.old_version()).not() {
+                panic!("The list of updates for the database is not ordered properly")
+            }
+        });
+    }
 
-        assert_eq!(result, false)
+    fn compare_two_numbers(first: usize, second: usize) -> bool {
```
dnwiebe, line 516: Name is obscure; is function necessary?

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
 
-        assert_eq!(result, false)
-    }
+        let result = last_entry.unwrap().old_version();
 
-    #[derive(Default, Debug)]
-    struct FakeMigrationRecord {
-        version_result: RefCell<&'static str>,
-        sql_statement: RefCell<&'static str>,
+        assert!(compare_two_numbers(result, CURRENT_SCHEMA_VERSION))
```
dnwiebe, line 534: I think this test would be stronger if it asserted that the old_version of the end of the list had to be specifically CURRENT_SCHEMA_VERSION - 1, rather than just less than CURRENT_SCHEMA_VERSION.

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+        let result = subject.execute_upon_transaction(set_of_sql_statements);
 
-        let _ = subject.make_updates("0.0.10", list, Box::new(ConnectionWrapperMock::default()));
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            "no such table: botanic_garden"
+        );
+        let connection = Connection::open(&db_path).unwrap();
+        //when an error occurs the underlying transaction gets rolled back, and we cannot see any changes to the database
+        let assertion: Result<(String, String), Error> = connection.query_row(
```
dnwiebe, line 608: If you query for a list of rows, rather than for a single row, you can assert on the list being empty rather than on an error message.

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+        let config = DBMigratorConfiguration {
+            db_configuration_table: "test".to_string(),
+        };
+        let subject = DbMigratorReal::new();
 
-        assert!(result.is_ok());
-        TestLogHandler::new().exists_log_containing(
-            "INFO: DbMigrator: Database successfully updated from version 0.0.10 to 0.0.11",
+        let result = subject.make_updates(
+            2,
```
dnwiebe, line 756: I don't understand this 2 given the '3' on line 745.

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+        let fourth_record_old_version_param = fourth_record_old_version_p_arc.lock().unwrap();
+        assert_eq!(fourth_record_old_version_param.len(), 2);
+        let fifth_record_old_version_param = fifth_record_old_version_p_arc.lock().unwrap();
+        assert_eq!(fifth_record_old_version_param.len(), 3);
+        let first_record_migration_params = first_record_migration_p_arc.lock().unwrap();
+        assert_eq!(*first_record_migration_params, vec![]);
+        let second_record_migration_params = second_record_migration_p_arc.lock().unwrap();
+        assert_eq!(*second_record_migration_params, vec![]);
+        let third_record_migration_params = third_record_migration_p_arc.lock().unwrap();
+        assert_eq!(*third_record_migration_params, vec![()]);
```
dnwiebe, line 778: This is confusing, given that the database holds schema version 3. If this is just an artifact of testing, then maybe change the schema version back to 2 to make it clear?

node/src/database/db_migrations.rs
```diff
@@ -74,353 +158,684 @@ impl DbMigratorReal {
...
+            .unwrap();
+        assert_eq!(assertion.1, "5")
     }
 
     #[test]
-    fn transacting_migration_sad_path() {
-        init_test_logging();
-        let execution_params_arc = Arc::new(Mutex::new(vec![]));
-        let params_arc_clone = execution_params_arc.clone();
+    fn make_updates_stops_doing_updates_on_the_version_specified_as_a_parameter() {
```
dnwiebe, line 796: make_updates_terminates_at_the_specified_version?

node/src/database/db_migrations.rs
```diff
@@ -431,57 +846,95 @@ mod tests {
...
-        let subject = DbMigratorReal::default();
+        let mut connection_wrapper = ConnectionWrapperReal::new(connection);
+        let config = DBMigratorConfiguration {
+            db_configuration_table: "test".to_string(),
+        };
+        let subject = DbMigratorReal::new();
 
-        let result = subject.make_updates(outdated_schema, list, Box::new(connection_wrapper));
+        let result = subject.make_updates(
+            0,
```
dnwiebe, line 856: Make intention-revealing names for these numbers?

dnwiebe (commented):
Resume at  node/src/db_config/secure_config_layer.rs

node/src/database/db_migrations.rs
bertllll: But I don't query on more than one row here, there was just one valid db operation and I check that that one wasn't persistently written into the db. 

It's just a convenient answer of the rusqlite query in this case, saying "Query returned no rows" even though it concerns just one. 

I was searching for a method allowing to query on multiple rows - despite I don't need to use it in this test - and couldn't find any.

node/src/node_configurator/configurator.rs
```diff
@@ -952,12 +943,12 @@ mod tests {
                 path: MessagePath::Conversation(4321),
                 payload: Err((
                     CONFIGURATOR_WRITE_ERROR,
-                    r#"DatabaseError("Didn\'t work good")"#.to_string()
+                    r#"DatabaseError("Didn't work good")"#.to_string()
```
dnwiebe, line 946: Should be "Didn't work well."

PR #25, opened 2021-06-23
Gh 244 third review - don't merge

dnwiebe:


node/src/database/db_initializer.rs
```diff
@@ -38,13 +38,16 @@ pub trait DbInitializer {
...
+    #[allow(unused_variables)]
     fn initialize_to_version(
         &self,
         path: &Path,
         chain_id: u8,
         target_version: usize,
         create_if_necessary: bool,
-    ) -> Result<Box<dyn ConnectionWrapper>, InitializationError>;
+    ) -> Result<Box<dyn ConnectionWrapper>, InitializationError> {
+        intentionally_blank!()
```
dnwiebe, line 49: I'd really rather see this in the #[cfg(test)] section.

node/src/database/db_migrations.rs
```diff
@@ -57,11 +58,13 @@ trait DatabaseMigration: Debug {
 }
 
 trait DBMigrationUtilities {
```
dnwiebe, line 60: There should be a trait with a collection of functions which the DBMigration writer is supposed to use, and a different  place where functions that he's not supposed to use should live--to prevent him from making assumptions and calling functions he shouldn't be messing with.

node/src/database/db_migrations.rs
```diff
@@ -169,124 +184,96 @@ impl DbMigratorReal {
...
+        let updates_to_process = Self::select_updates_to_process(
+            mismatched_schema,
+            list_of_updates,
+            &*migration_utilities,
+        );
+        for record in updates_to_process {
+            let current_version = record.old_version();
             //necessary for testing
-            if Self::is_target_version_reached(current_state, &target_version) {
+            if Self::is_target_version_reached(&current_version, &target_version) {
```
dnwiebe, line 195: This may no longer be required at all, since the for loop terminates after the last migration in the list has been processed. If you need to execute just some of these updates for testing, perhaps you could extract the guts of this for loop into another function, and then call that function a certain number of times from the test.

node/src/database/db_migrations.rs
```diff
@@ -169,124 +184,96 @@ impl DbMigratorReal {
...
         Err(error_message)
     }
 
-    fn context_between_two_versions(first: usize, second: &str) -> String {
-        format!("from version {} to {}", first, second)
-    }
-
-    fn log_success(&self, versions: &str) {
-        info!(self.logger, "Database successfully updated {}", versions)
+    fn log_success(&self, current_version: &usize) {
```
dnwiebe, line 270: Maybe this variable name could be changed...if the database was successfully updated, then the current version is `current_version + 1`, rather than `current_version.`

Also, `&usize` is no smaller than `usize`.

node/src/database/db_migrations.rs
```diff
@@ -320,16 +307,18 @@ mod tests {
...
     struct DBMigrationUtilitiesMock {
-        update_schema_version_params: Arc<Mutex<Vec<String>>>,
+        update_schema_version_params: Arc<Mutex<Vec<usize>>>,
         update_schema_version_results: RefCell<Vec<rusqlite::Result<()>>>,
         execute_upon_transaction_params: Arc<Mutex<Vec<Vec<String>>>>,
         execute_upon_transaction_result: RefCell<Vec<rusqlite::Result<()>>>,
         commit_params: Arc<Mutex<Vec<()>>>,
         commit_results: RefCell<Vec<Result<(), String>>>,
+        panic_if_the_version_found_is_too_high_params: Arc<Mutex<Vec<usize>>>,
+        //panic_if_the_version_found_is_too_high_results        --an unnecessary filed
```
dnwiebe, line 317: Remove commented-out code

node/src/database/db_migrations.rs
```diff
@@ -320,16 +307,18 @@ mod tests {
 
     #[derive(Default)]
     struct DBMigrationUtilitiesMock {
-        update_schema_version_params: Arc<Mutex<Vec<String>>>,
+        update_schema_version_params: Arc<Mutex<Vec<usize>>>,
         update_schema_version_results: RefCell<Vec<rusqlite::Result<()>>>,
         execute_upon_transaction_params: Arc<Mutex<Vec<Vec<String>>>>,
         execute_upon_transaction_result: RefCell<Vec<rusqlite::Result<()>>>,
```
dnwiebe, line 313: Probably should be `execute_upon_transaction_resultS` to be like the other results fields

node/src/database/db_migrations.rs
```diff
@@ -507,50 +511,54 @@ mod tests {
...
-        let _ = list_validation_check(DbMigratorReal::list_of_existing_updates());
+    fn list_of_approved_updates_is_correctly_ordered() {
+        let _ = list_validation_check(DbMigratorReal::list_of_approved_updates());
         //success if no panicking
     }
 
     #[test]
-    fn list_of_existing_updates_does_not_end_with_version_higher_than_the_current_version() {
-        let last_entry = DbMigratorReal::list_of_existing_updates()
+    fn list_of_approved_updates_does_not_end_with_version_higher_than_the_current_version() {
```
dnwiebe, line 533: The test is stronger than this name implies. How about ...`ends_on_the_current_version`?

node/src/database/db_migrations.rs
```diff
@@ -507,50 +511,54 @@ mod tests {
...
-            5.to_string(),
-            &migration_utilities,
-        );
+        let result =
+            DbMigratorReal::migrate_semi_automated(&mut migration_record, &migration_utilities);
 
         assert_eq!(result, Err(Error::InvalidQuery));
         let update_schema_version_params = update_schema_version_params_arc.lock().unwrap();
-        assert_eq!(*update_schema_version_params, vec!["5".to_string()])
+        assert_eq!(*update_schema_version_params, vec![5])
```
dnwiebe, line 561: If the migration from 4 to 5 failed, should we really be updating the schema version to 5, or should we instead leave it at 4?

dnwiebe (commented):
Continue at db_migration.rs:570

node/src/database/db_migrations.rs
bertllll: This looks different than it is. In the bigger context, this function's failure (update_schema_version()) means that either writing the new version hasn't been completed and thus there is not a waiting change to the db, updating the schema version, or , and what is more important, any failure at that point just means that the db won't modify its state, because we haven't committed any changes we won't because the failure doesn't allow to step up to the final step.
I think it's completely okay. Only this test may confuse somebody, but I don't know how to avoid it without writing unnecessary lines.

node/src/database/db_migrations.rs
```diff
@@ -916,14 +892,101 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            result,
+            Err(
+                r#"Updating database from version 0 to 1 failed: InvalidColumnIndex(2)"#
+                    .to_string()
+            )
+        );
+        let update_schema_version_params = update_schema_version_params_arc.lock().unwrap();
```
dnwiebe, line 938: Could be removed...

node/src/database/db_migrations.rs
```diff
@@ -916,14 +892,101 @@ mod tests {
...
+            .update_schema_version_result(Ok(()))
+            .update_schema_version_result(Ok(()))
+            .commit_params(&commit_params_arc)
+            .commit_result(Err("Committing transaction failed".to_string()));
+        let subject = DbMigratorReal::new();
+
+        let result = subject.make_updates(0, 2, Box::new(migration_utils), list_of_updates);
+
+        assert_eq!(result, Err(String::from("Committing transaction failed")));
+        let first_record_old_version_param = first_record_old_version_p_arc.lock().unwrap();
```
dnwiebe, line 978: Much if not all of this may be unnecessary

PR #30, opened 2021-08-06
GH-379: Third review

bertllll:


masq_lib/src/multi_config.rs
```diff
@@ -131,6 +131,9 @@ impl<'a> MultiConfig<'a> {
 
 //this way I can protect the inner field from having to be public (Default + deref_mut())
 
+//TODO I think this apply for the situation where I'm not allowed to make this cfg(test); this default() is used
+// in a different crate and thus it doesn't see anything with cfg(test), period. What we maybe should do is to go by the way of
+// GH-456
 impl Default for MultiConfig<'_> {
```
dnwiebe: Investigate this; maybe it can be made better.

masq_lib/src/utils.rs
```diff
@@ -124,17 +124,30 @@ pub fn exit_process_with_sigterm(message: &str) {
     } else {
         eprintln!("{}", message);
         not_win_cfg::signal::raise(not_win_cfg::signal::SIGTERM).expect("sigterm failure");
-        //This function must not return; so wait for death.
+        //This function must not return; so may it waits for death.
```
dnwiebe: I think the word "may" doesn't belong here.

Maybe a clearer version: "This function must not return, and the process will be terminated by another thread within micro- or milliseconds, so we wait here for death."

multinode_integration_tests/src/command.rs
```diff
@@ -63,11 +63,10 @@ impl Command {
 
     fn truncate_long_string(mut string: String) -> String {
         if string.len() <= MAX_PRINTED_STRING_SIZE {
```
dnwiebe: This could be further improved by reversing the test (string.len() > MAX_PRINTED_STRING_SIZE) and eliminating the empty then clause, using the else clause instead.

dnwiebe (commented):
Resume at node/src/banned_dao.rs

node/src/daemon/daemon_initializer.rs
```diff
@@ -162,15 +155,33 @@ impl DaemonInitializerReal {
             .collect_vec();
         self.rerunner.rerun(param_vec);
     }
+
+    #[cfg(test)]
```
dnwiebe: I think you can put this inside another `impl DaemonInitializerReal` inside the tests module; then there isn't even non-compilable test code in the production tree.

node/src/dispatcher.rs
```diff
@@ -477,7 +477,7 @@ mod tests {
 
         dispatcher_subs.ui_sub.try_send(msg).unwrap();
 
-        thread::sleep(std::time::Duration::from_millis(15)); //TODO why cannot I get along without this delay; let's look at it with Dan
+        thread::sleep(std::time::Duration::from_millis(15)); //TODO can we look at this again, please? 0 results in RC
```
dnwiebe: I dunno what this is. Maybe change the comment to something like "Required to break unknown race condition, probably inside actix."

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -912,7 +913,7 @@ impl StandardGossipHandler {
                 }
```
dnwiebe, line 911: Maybe put in a comment explaining that we can't satisfy Clippy because the body of flat_map has a side effect in `update_database_record()`.

node/src/node_configurator/mod.rs
```diff
@@ -209,14 +209,21 @@ pub fn update_db_password(
     Ok(())
 }
 
+pub fn real_user_from_m_c_or_populate(
```
dnwiebe: I'd like to see multi_config spelled out here, just to make it easier to read

node/src/run_modes_factories.rs
```diff
@@ -192,37 +191,65 @@ mod tests {
...
             "program",
             "--initialization",
             "--ui-port",
-            &port.to_string(),
-        ]);
+            1234.to_string().as_str(),
+        ]
+        .array_of_borrows_to_vec();
         let mut stream_holder = FakeStreamHolder::default();
         let result = subject.make(&args, &mut stream_holder.streams()).unwrap();
```
dnwiebe, line 213: Please convert this to standard form, with blank lines before and after the call to the method under test.

dnwiebe (commented):
Resume with test_utils/mod.rs

node/src/daemon/daemon_initializer.rs
bertllll: I cannot; it's unusual but this function is needed in a test in a different file where the test can be more complex. Still I have to create the method here because I declared the struct here.

PR #32, opened 2021-08-10
Real third review for GH-379

dnwiebe:


node/src/run_modes.rs
```diff
@@ -139,6 +139,7 @@ impl RunModes {
         privilege_required: bool,
         streams: &mut StdStreams,
     ) {
+        MultiConfig::new_test_only(app_node().get_matches_from(["program"].iter()));
```
dnwiebe: Remove this

dnwiebe (commented):
Approved, but can't actually approve for technical reasons.

PR #35, opened 2021-08-20
GH-401: Make database picky about chain

bertllll:


masq_lib/src/constants.rs
```diff
@@ -3,7 +3,7 @@
 pub const HIGHEST_RANDOM_CLANDESTINE_PORT: u16 = 9999;
 pub const HTTP_PORT: u16 = 80;
 pub const TLS_PORT: u16 = 443;
-pub const DEFAULT_CHAIN_NAME: &str = "mainnet";
+pub const DEFAULT_CHAIN_NAME: &str = "mainnet"; //TODO look, this is different here and in the node's crate; there we call a default 'ropsten' which is quite concerning anyway
```
masqrauder, line 6: I don't understand this comment; DEFAULT_CHAIN_NAME occurs only once -- right here -- in the whole codebase and it only has "mainnet" -- however TEST_DEFAULT_CHAIN_NAME does exist and it is initialized with "ropsten"

node/src/database/config_dumper.rs
```diff
@@ -195,6 +195,7 @@ mod tests {
             other => panic!("Was expecting Value::Object, got {:?} instead", other),
         };
         let expected_value = json!({
+           "chainName": "ropsten".to_string(),
```
masqrauder: this line could be:
` "chainName": TEST_DEFAULT_CHAIN_NAME.to_string(),`
as used above on lines 178 and 183.

node/src/bootstrapper.rs
```diff
@@ -335,6 +335,7 @@ impl BootstrapperConfig {
             blockchain_bridge_config: BlockchainBridgeConfig {
                 blockchain_service_url: None,
                 chain_id: 3u8, /*DEFAULT_CHAIN_ID*/
+                //TODO this seems wrong, why do we want Ropsten to be a default chain?
```
masqrauder, line 338: Yes, it's wrong for go-live on mainnet but correct for the time-being while testing. The comment is there to remind us the go-live on mainnet version line of code will be:
`chain_id: DEFAULT_CHAIN_ID,`

node/src/database/db_initializer.rs
```diff
@@ -731,6 +747,8 @@ mod tests {
             assert_eq!(actual_name, expected_name);
             value
         };
+
+        verify(&mut config_vec, "chain_name", Some("ropsten"));
```
masqrauder: "ropsten" could be replaced with `TEST_DEFAULT_CHAIN_NAME` here.

node/src/database/db_migrations.rs
```diff
@@ -419,18 +482,23 @@ mod tests {
...
+            unimplemented!()
+        }
     }
 
-    lazy_static! {
-        static ref TEST_DIRECTORY_FOR_DB_MIGRATION: PathBuf =
-            PathBuf::new().join(BASE_TEST_DIR).join("db_migration");
+    fn make_external_migration_parameters() -> ExternalMigrationParameters {
+        ExternalMigrationParameters {
+            chain_name: "mainnet".to_string(),
```
masqrauder: would `"mainnet"` be better as `DEFAULT_CHAIN_NAME`?

node/src/database/db_migrations.rs
```diff
@@ -868,6 +939,13 @@ mod tests {
         TestLogHandler::new().exists_log_containing(
             "INFO: DbMigrator: Database successfully updated from version 0 to 1",
         );
+        let make_mig_declaration_utils_params = make_mig_declaration_params_arc.lock().unwrap();
+        assert_eq!(
+            *make_mig_declaration_utils_params,
+            vec![ExternalMigrationParameters {
+                chain_name: "mainnet".to_string()
```
masqrauder, line 947: This assert is best left as literal `"mainnet"` for verification purposes.

masqrauder (commented):
Left a few comments about the use of "ropsten" and "mainnet" literals vs. using already defined constants. If you disagree we can discuss. Otherwise I can approve if you feel strongly they should stand as is.

node/src/bootstrapper.rs
bertllll: This isn't as mild issue as it looks like. First it's quite hard to discover the fact that the BootstrapperConfig always contains this value initially unless it goes through further configuration. I adapted it to he mainnet now just to know and 59 unit tests in the node crate failed because of it, so it quite has an impact. Not everyone can predict this when he's writing tests, especially newcomers.
 
Another argument is that in all kinds of tests except multinode test we don't really connect to the blockchain and if we do (in the multinode tests) then we use a safe solution with a testnet. Which I'd like to stabilize with my changes from GH-473, where I want to make it clear that there is an only option for the multinode_test for the chain and I try to name constants there with clear names saying that alone. 

So my decision is, finally, I don't want to make a storm, I just want to replace this number 3 with a constant saying TEST_DEFAULT_CHAIN or something, so you know exactly and instantly. It's enough for me. But that will be settled in the other card.

masq_lib/src/constants.rs
bertllll: The problem is not with DEFAULT_CHAIN_NAME, this works correct, it points to 'mainnet'. Troubles start with DEFAULT_CHAIN_ID defined in utils.rs, even though it's placed in the test_utils directory, every time you see this in the code it can confuse you. Because it seems to be symmetrical to DEFAULT_CHAIN_NAME which it isn't, both point differently, the second to 'ropsten', its index respectively.

So I think I'll settle this with proper names which I'm going to set up with GH-473.

node/src/actor_system_factory.rs
```diff
@@ -1174,4 +1196,25 @@ mod tests {
         let result = candidate.decode(&crypt_data).unwrap();
         assert_eq!(result, plain_data);
     }
+
+    #[test]
+    fn database_chain_validity_happy_path() {
+        let chain_id = 1;
```
dnwiebe: Can you put a manifest constant here? It'll make it easier to find when you change the constant names.

node/src/actor_system_factory.rs
```diff
@@ -1174,4 +1196,25 @@ mod tests {
...
+
+        let _ = validate_database_chain_correctness(chain_id, &persistent_config);
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Database with the wrong chain name detected; expected: ropsten, was: mainnet"
+    )]
+    fn database_chain_validity_sad_path() {
+        let chain_id = 3; //Ropsten
```
dnwiebe: Another manifest constant

node/src/database/db_initializer.rs
```diff
@@ -818,41 +836,73 @@ mod tests {
...
+    fn existing_database_with_the_wrong_version_comes_to_migrator_that_makes_it_gradually_migrate_to_upper_versions(
     ) {
         init_test_logging();
         let home_dir = ensure_node_home_directory_exists(
             "db_initializer",
-            "existing_database_with_the_wrong_version_comes_to_migrator_and_is_happily_gradually_migrated_to_upper_versions",
+            "existing_database_with_the_wrong_version_comes_to_migrator_that_makes_it_gradually_migrate_to_upper_versions",
         );
-        let db_path = &home_dir.join(DATABASE_FILE);
+        let template_db_path_dir = &home_dir.join("template");
```
dnwiebe: Better name than "template" here. "from_scratch" might work.

node/src/database/db_initializer.rs
```diff
@@ -818,41 +836,73 @@ mod tests {
...
+        )
+        .unwrap();
+        let conn_template = Connection::open_with_flags(
+            &template_db_path_dir.join(DATABASE_FILE),
+            OpenFlags::SQLITE_OPEN_READ_ONLY,
+        )
+        .unwrap();
+        let extract_from_updated = subject.extract_configurations(&conn_updated);
+        let extract_from_template = subject.extract_configurations(&conn_template);
+        //please, write all rows with irregular values here
```
dnwiebe: Maybe "unpredictable" instead of "irregular?"

node/src/database/db_initializer.rs
```diff
@@ -818,41 +836,73 @@ mod tests {
...
+        let sieve = |updated: &(&String, &Option<String>),
```
dnwiebe: The only thing you use here is the name of one of the rows; it looks like you don't need the other row name or either of the values.

node/src/database/db_migrations.rs
```diff
@@ -302,28 +354,28 @@ mod tests {
...
     use std::fs::create_dir_all;
     use std::panic::{catch_unwind, AssertUnwindSafe};
-    use std::path::PathBuf;
     use std::sync::{Arc, Mutex};
 
     #[derive(Default)]
     struct DBMigrationUtilitiesMock {
         too_high_found_schema_will_panic_params: Arc<Mutex<Vec<usize>>>,
         make_mig_declaration_utils_results: RefCell<Vec<Box<dyn MigDeclarationUtilities>>>,
+        make_mig_declaration_utils_params: Arc<Mutex<Vec<ExternalMigrationParameters>>>,
```
dnwiebe, line 377: Prefer params-before-results order, but not a big deal

node/src/database/db_migrations.rs
```diff
@@ -940,4 +1019,42 @@ mod tests {
...
+        {
+            subject
+                .initialize_to_version(&dir_path, DEFAULT_CHAIN_ID, 1, true)
+                .unwrap();
+        }
+
+        let result = subject.initialize_to_version(&dir_path, DEFAULT_CHAIN_ID, 2, true);
+
+        let (chn_name, chn_value, chn_encrypted): (String, Option<String>, u16) =
+            assurance_query_for_config_table(
```
dnwiebe, line 1034: If you modify this function to take `result` instead of `connection`, you can use it to assert on `result` much more strongly than below.

node/src/run_modes.rs
```diff
@@ -739,7 +739,7 @@ parm2 - msg2\n"
     fn daemon_and_node_modes_version_call() {
         use chrono::offset::Utc;
         use chrono::NaiveDate;
-        if Utc::today().and_hms(0, 0, 0).naive_utc().date() >= NaiveDate::from_ymd(2021, 9, 30) {
+        if Utc::today().and_hms(0, 0, 0).naive_utc().date() >= NaiveDate::from_ymd(2021, 10, 30) {
```
dnwiebe, line 743: It occurs to me: we probably need a comment here to explain this to new developers, because it will definitely get a WTH?! reaction if we don't.

node/tests/http_through_node_test.rs
```diff
@@ -11,10 +11,8 @@ use std::str::FromStr;
 use std::time::Duration;
 
 #[test]
-#[allow(unused_variables)] // 'node' below must not become '_' or disappear, or the
```
dnwiebe, line 14: Can you preserve a modified version of this comment below in the code?

node/tests/utils.rs
```diff
@@ -109,6 +109,15 @@ impl MASQNode {
         format!("stdout:\n{}\nstderr:\n{}", stdout, stderr)
     }
 
+    #[allow(dead_code)]
+    pub fn start_standard_without_preparation(data_dir: &PathBuf) -> MASQNode {
```
dnwiebe: Maybe a name like...`start_standard_in_prepared_environment`? Or something like that? Or maybe using the word "unsterilized" somehow?

dnwiebe (approved):

PR #36, opened 2021-08-28
GH-446: Replaced mpsc with crossbeam_channel

FinsaasGH:


masq/src/commands/shutdown_command.rs
```diff
@@ -405,10 +405,11 @@ mod tests {
 
     #[test]
     fn shutdown_awaiter_sad_path() {
+        use crossbeam_channel::unbounded;
```
bertllll: It's a bit unusual to write imports into the function's body and if you do so you should have a good reason for that. Normally, we simply leave it among those imports at the header of the test module.

mod tests{
#[test]
fn one_of_your_tests(){
}
}

masq_lib/src/multi_config.rs
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) 2017-2019, Substratum LLC (https://substratum.net) and/or its affiliates. All rights reserved.
-
+#![allow(unused_imports)]
```
bertllll: This is even worse than what you did last time. The exclamation mark means apply this rule for the entire file. You suppressed the checks from the compiler. 

Instead you should try to find a way not to use this special attribute at all, in other words, this means most likely that you need to identify the imports that are unused and just remove them. Only if there is a serious reason - which I cannot think of right now - you might use this attribute but it sounds to me very very exotic. I don't remember using it myself.  

Hint: if there are potential unused imports just let the compiler tell you where and what it is. Probably the easiest way is to remove your attribute  #[allow(unused_imports)] and do something (means removing) with what will be shown to you as unused.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -530,7 +530,7 @@ mod tests {
         let to = "0x3f69f9efd4f2592fd70be8c32ecd9dce71c472fc";
         let port = find_free_port();
 
-        let (tx, rx) = mpsc::sync_channel(1337);
+        let (tx, rx) = bounded(1337);
```
bertllll: This is just a guess, I might be wrong here.
This weird number meaning the number of messages to be together maximally inside the internal queue makes me wonder that it isn't ideal. (Somebody probably tried to pick some random number madly large so it doesn't reach its limits. Or I'm wrong and it works with bytes separately and this is number of bytes that we are willing to transport in this test maximally but I doubt it).
Please just try to change it to unbounded() crossbeam channel and try this single test immediately after. If it passes use it if not leave it as you have it now.

node/src/daemon/dns_inspector/dynamic_store_dns_inspector.rs
```diff
@@ -248,13 +248,21 @@ impl StoreWrapperReal {
...
     use core_foundation::boolean::CFBoolean;
     use core_foundation::propertylist::CFPropertyListSubClass;
     use core_foundation::string::CFString;
     use std::cell::RefCell;
-    use std::sync::Arc;
-    use std::sync::Mutex;
+    use std::collections::HashMap;
+    use std::net::IpAddr;
+    use std::str::FromStr;
+    use std::sync::{Arc, Mutex};
```
bertllll: This is really problematic when I don't know how this happened; why these imports started to be needed although there is no change elsewhere in this file which would have answered that. 

Do you see the green lines here? Also look at the header of this smaller window, it belong only to one file, right? The fact that we can see new imports that haven't been obviously brought in because of some seeable need from elsewhere in the file means logically that these imports are very likely redundant. 

It doesn't necessarily mean that these things they are about are not used but it may mean - and I do believe so - that they were named unnecessarily given that there is the line saying "use super::* "
Remember, this line says that everything imported in a higher level (in the superior module) than the module you're in is automatically imported into this module inside the higher. 
Here it means that module "tests" is inside module "dynamic_store_dns_inspector" which is the file itself. If we state this there it very likely means that all the green lines or most of them would be covered by the "super*" but by some unknown reason they got in named explicitly.

I recommend to delete all the green lines and try to compile. The compiler will tell you what imports are missing. Easy.

node/src/daemon/launch_verifier.rs
```diff
@@ -334,9 +334,11 @@ mod tests {
     }
 
     #[test]
+
```
bertllll: Remove this blank line please.

node/src/proxy_server/mod.rs
```diff
@@ -532,7 +532,7 @@ impl ProxyServer {
                                 Err(e) => {
                                     error!(
                                         logger,
-                                        "Neighborhood refused to answer route request: {}", e
+                                        "Neighborhood refused to answer route request: {:?}", e
```
bertllll: Please check if it really needs to be changed to be able to accept Debug instead of Display. Try to google the difference of them for Rust. :) 

Again if it can stay the former way make it look back like that. Otherwise we'll accept.

node/src/stream_handler_pool.rs
```diff
@@ -785,8 +784,9 @@ mod tests {
 
     #[test]
     fn terminal_packet_is_transmitted_and_then_stream_is_shut_down() {
+        use crossbeam_channel::unbounded;
```
bertllll: As I said already. There is no need to use this use clause inside a test; add it to imports above.

node/src/stream_handler_pool.rs
```diff
@@ -1137,6 +1137,7 @@ mod tests {
 
     #[test]
     fn stream_handler_pool_creates_nonexistent_stream_for_reading_and_writing() {
+        use crossbeam_channel::unbounded;
```
bertllll, line 1140: Again...but off course, the previous new import will cover this one. Just remove this.

node/src/stream_handler_pool.rs
```diff
@@ -1244,6 +1245,7 @@ mod tests {
 
     #[test]
     fn transmit_data_msg_handler_finds_ip_from_neighborhood_and_transmits_message() {
+        use crossbeam_channel::unbounded;
```
bertllll: Again.

node/src/stream_handler_pool.rs
```diff
@@ -1435,6 +1437,7 @@ mod tests {
 
     #[test]
     fn node_query_response_handler_resends_transmit_data_msg_when_connection_is_in_progress() {
+        use crossbeam_channel::unbounded;
```
bertllll: Again.

node/src/stream_handler_pool.rs
```diff
@@ -1534,6 +1537,7 @@ mod tests {
     #[test]
     fn when_a_new_connection_fails_the_stream_writer_flag_is_removed_and_another_connection_is_attempted_for_the_next_message_with_the_same_stream_key(
     ) {
+        use crossbeam_channel::unbounded;
```
bertllll: Here we go.

node/src/stream_handler_pool.rs
```diff
@@ -1876,10 +1880,11 @@ mod tests {
 
     #[test]
     fn stream_handler_pool_logs_error_and_returns_when_local_connection_is_gone() {
+        use crossbeam_channel::unbounded;
```
bertllll: And again.

port_exposer/src/main.rs
```diff
@@ -1,8 +1,8 @@
+use crate::port_exposer::PortExposer;
```
bertllll: I don't understand this move at all. 

Please move it back to the place where it was. At least somewhere under the copyright clause.

masq_lib/src/multi_config.rs
FinsaasGH: I had to clear the project and clone it from repo again, and this came with it. I have removed it and this will be in the next commit.

masq/src/commands/shutdown_command.rs
FinsaasGH: I didn't see the mod tests, I have moved this to the header.

port_exposer/src/main.rs
FinsaasGH: This didn't move from anywhere, there was an error failed to resolve: use of undeclared type `PortExposer`. I imported and it put it there. I have moved it below the copyright clause.

node/src/proxy_server/mod.rs
FinsaasGH: image
Mailbox error doesn't work with {}, but it does display with {:?}

node/src/daemon/dns_inspector/dynamic_store_dns_inspector.rs
FinsaasGH: These imports are required for the tests.

node/src/blockchain/blockchain_interface.rs
FinsaasGH: No issue changing to unbounded.

node/src/daemon/dns_inspector/dynamic_store_dns_inspector.rs
bertllll: They aren't, believe me. Try without them. Don't let yourself get disorient by seeing weird things. This code counts in for Mac. 

Remove the green lines and all should be fine.

masq_lib/src/multi_config.rs
bertllll: This couldn't have happened as you describe it - I think.

I hate dealing with this so long-hand. 

Anyway, the bang at the attribute is terrible. At least we need it to be without it.

node/src/daemon/dns_inspector/dynamic_store_dns_inspector.rs
bertllll: Seems like I wasn't exact. Removing all the green lines...true. Except that Mutex and Arc. (Look at it again - you can click on this reference though as "outdated". Arc and Mutex were included into a formatting operation. You can deduce because you see how it looked before - in red - and how it looks afterwards - green and together. The change is obvious, still it implies to be a work of the formater. )

So put back Mutex and Arc...and think of the context. Why we need to do this and how we progressed up to this moment.

node/src/proxy_server/mod.rs
bertllll: You are too pro-active. Here it reveal on you that you abide by the hints by the Rust plug-in. The red line in the code seemingly highlight a real error but not always it's true. You should not listen to this blindly. 

Always follow the hints of the compiler, he's a smarter guy. 

You're told here to add {:?} because they say the trait Display isn't implemented. Well, if you go deep into the structure you'll find that MailboxError is equipped with a derive trait. That's the trait which is easiest to impl, you just name it after the keyword "derive".

Next, if you click on the trait in the derive impl, you'll see that the trait has supertraits - alias traits whose methods it inherits too. You can see Display among those supertraits. Here we are, the Display trait is actually in the place. 

Please remove the {:?} formatting mark, there is no need for it. We could call it a bug of the Rust plugin.

port_exposer/src/main.rs
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) 2017-2019, Substratum LLC (https://substratum.net) and/or its affiliates. All rights reserved.
-pub mod port_exposer;
 
 use crate::port_exposer::PortExposer;
 
+pub mod port_exposer;
```
dnwiebe: Convention is to put `mod` statements before `use` statements. Please move the `pub mod port_exposer` line back above the `use crate::port_exposer::PortExposer` line.

dnwiebe (approved):

PR #38, opened 2021-09-09
GH-457: Remove dead parameter processing

FinsaasGH:


node/src/node_configurator/mod.rs
```diff
@@ -302,37 +154,6 @@ pub fn check_for_past_initialization(
...
-                streams.stdout,
-                &format!("Could not elicit wallet encryption password: {:?}\n", e),
-            );
-            None
-        }
-        Err(PasswordError::InternalError(_)) => panic!("Can't happen: no code path"),
-    }
-}
-
 //TODO this code (and the allied -- see thorough this file) is very likely to go away when GH-457 is played
```
bertllll: Once you resolve all more factual issues please don't forget to delete all comments (or here TODO) which refer to this card. But again, first we must be sure that everything what should go away really did and everything what should stay is still here.

node/src/node_configurator/mod.rs
```diff
@@ -1173,578 +854,6 @@ mod tests {
         assert_eq!(true, user_specified);
     }
 
-    #[test]
-    fn request_database_password_happy_path() {
```
bertllll: I've got a problem with this test having been removed since the function which it's supposed to test is still left there intact.

You either need to remove the function request_existing_db_password too or you shouldn't take a test away because we may still need the function. I don't know. Why did you decide that this function is obsolete? 

After some thinking, my opinion is that we may want to keep this function for the purposes of another card, GH-458. 
If I worked on this I might create a link between this code and the card so that when we start to work on that card we also reconsider this code to be utilized within the functionality which may become true with GH-458. But that's all a lot unclear. Anyway, I have knowledge that --db--password in Deamon's setup_reporter isn't treated as some may have been thinking it has. It will need some re-architecture and this code here could be handy... but I can see an argument that instead of drafting old code to something new it may be easier and safer to write it up anew from scratch. Ideally ask Dan, my experience isn't large enough for this kind of decision.

node/src/node_configurator/mod.rs
```diff
@@ -1173,578 +854,6 @@ mod tests {
...
-        assert_eq!(
-            stdout_writer.get_string(),
-            "Decrypt wallet\n\
-             Enter password: "
-                .to_string()
-        );
-    }
-
-    #[test]
-    fn request_database_password_rejects_blank_password() {
```
bertllll: The same issue.

This is the worst case...leaving untested code alone. Resolve the uncertainty about if it should stay or go.

node/src/node_configurator/mod.rs
```diff
@@ -264,31 +141,6 @@ pub fn data_directory_from_context(
...
-        &DirsWrapperReal {},
-        &real_user,
-        &data_directory_opt,
-        &chain_name,
-    );
-    let persistent_config_box = initialize_database(&directory, chain_id_from_name(&chain_name));
-    Ok((multi_config, persistent_config_box))
-}
-
 pub fn check_for_past_initialization(
```
bertllll: Though newer code than the most of the code we analyze here; I noticed that this function isn't used anywhere except tests in this file. It makes me suspicious that this is another example of some clinically dead code (but we've been overlooking it because it's been used - only in tests. Do you see? I suggest to remove it and run all tests we have, if its okay it can be removed)

node/src/node_configurator/mod.rs
```diff
@@ -419,34 +232,6 @@ where
...
-            } else {
-                flushed_write(streams.stdout, mismatch_msg);
-                Err(PasswordError::Mismatch)
-            }
-        }
-        Err(msg) => Err(PasswordError::VerifyError(msg)),
-    }
-}
-
 pub fn request_password_with_retry<R>(
```
bertllll, line 221: I'm afraid that if you decide that request_existing_password should go away then this function gets the same destiny.

node/src/node_configurator/mod.rs
```diff
@@ -382,14 +203,6 @@ pub fn request_existing_db_password(
...
-    if password.is_empty() {
-        Err("Password cannot be blank.".to_string())
-    } else {
-        Ok(())
-    }
-}
-
 #[derive(Debug, PartialEq, Clone)]
 pub enum PasswordError {
     Mismatch,
```
bertllll, line 194: The life of these all (probably) also depends on your decision about keeping or deleting request_existing_db_password

node/src/node_configurator/mod.rs
```diff
@@ -302,37 +154,6 @@ pub fn check_for_past_initialization(
...
-                &format!("Could not elicit wallet encryption password: {:?}\n", e),
-            );
-            None
-        }
-        Err(PasswordError::InternalError(_)) => panic!("Can't happen: no code path"),
-    }
-}
-
 //TODO this code (and the allied -- see thorough this file) is very likely to go away when GH-457 is played
 pub fn request_existing_db_password(
```
dnwiebe, line 144: This function should go away. It won't be useful for GH-458; its only use is to accept interactive input at the command line when starting the Node, and we don't do that anymore because we have both a CLI and a GUI for that now.

node/src/node_configurator/mod.rs
dnwiebe: My opinion is that `request_existing_db_password` should be removed along with all its tests. (Remove `request_existing_db_password` first, and then let the compiler find its tests.)

bertllll: And I agree with Dan; now it makes more sense to me. :)

PR #40, opened 2021-09-15
GH-414: Changed default mnemonic phrase length from 24 to 12. 

FinsaasGH:
Possible phrase length remains 12-24.


bertllll (approved):

masq/src/commands/generate_wallets_command.rs
```diff
@@ -107,7 +107,7 @@ pub fn generate_wallets_subcommand() -> App<'static, 'static> {
             .long("word-count")
             .value_name("WORD-COUNT")
             .required(false)
-            .default_value("24")
+            .default_value("12")
```
bertllll, line 110: I've accidentally come across the fact that there is also help text that refers to the default word count. You will need to change it to be correct.

masqrauder: I'd like to go on record that I don't really agree with this change. I think we should default to the most secure options when people's funds are involved. For the people that don't know what they're doing, choosing the longer mnemonic phrase sets them up for being more secure when they learn the importance in the future. Another consideration is that some hardware wallets default to 24 word mnemonic phrases. Originally we had hoped people would use them. We considered the possibility of incorporating hardware wallet compatibility/connectivity with Node. As long as we don't lose the ability to choose the longer mnemonics I can see how this would be perceived to be more user friendly -- probably much of the blame should be placed on MetaMask and their default of 12 word phrases.

PR #41, opened 2021-09-17
Gh-457: Removed dead parameter processing

FinsaasGH:


node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -635,26 +635,14 @@ pub mod standard {
 
     pub fn get_db_password(
         multi_config: &MultiConfig,
-        streams: &mut StdStreams,
+        _streams: &mut StdStreams,
```
bertllll: I'd consider removing this parameter fully since we don't need it and this function is not a member of a trait so you can afford it easily.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -635,26 +635,14 @@ pub mod standard {
...
         config: &mut BootstrapperConfig,
         persistent_config: &mut dyn PersistentConfiguration,
     ) -> Result<Option<String>, ConfiguratorError> {
         if let Some(db_password) = &config.db_password_opt {
             return Ok(Some(db_password.clone()));
         }
-        let db_password_opt = match value_user_specified_m!(multi_config, "db-password", String) {
```
bertllll, line 645: Also, I'd check out whether value_user_specified_m! is still being used somewhere else after you removed it here. It's quite an unusual macro so it's possible that we don't need it existing at all.

masqrauder (approved):

node/src/node_configurator/mod.rs
```diff
@@ -264,250 +131,6 @@ pub fn data_directory_from_context(
...
-}
-
-pub fn flushed_write(target: &mut dyn io::Write, string: &str) {
-    write!(target, "{}", string).expect("Failed console write.");
-    target.flush().expect("Failed flush.");
-}
-
 pub fn port_is_busy(port: u16) -> bool {
     TcpListener::bind(SocketAddr::new(localhost(), port)).is_err()
 }
```
dnwiebe, line 136: Some or all of these validators are probably no longer used. You can get rid of the dead ones.

PR #44, opened 2021-09-24
GH-473:  New system of employing chains, internally (for developers) and externally (for users), not only for the expansion to L2 networks 

bertllll:


masq_lib/src/test_utils/utils.rs
```diff
@@ -3,8 +3,11 @@
 use std::fs;
 use std::path::PathBuf;
 
-pub const DEFAULT_CHAIN_ID: u8 = 3u8; //For testing only
-pub const TEST_DEFAULT_CHAIN_NAME: &str = "ropsten"; //For testing only
+pub const TEST_DEFAULT_CHAIN_ID: u8 = 2;
```
masqrauder: This is inconsistent with the Ethereum chain network IDs. 
mainnet = 1
undefined = 2
ropsten = 3
rinkeby = 4
etc.

I strongly encourage that we stick with and not deviate from the standard network IDs. The network IDs are used to create valid transactions on the appropriate network.

EthereumNetworkNumbers

The reason I chose network 2 for dev (multi-node tests) was because it was not already defined and used by an actual network. The multi-node tests automatically choose network Id 2 "dev" using the Ganache docker container. The default as `ropsten` (network ID 3) are so that the tests -- by default -- run as the production code.

multinode_integration_tests/src/masq_node.rs
```diff
@@ -83,15 +104,18 @@ impl NodeReference {
         public_key: PublicKey,
         ip_addr_opt: Option<IpAddr>,
         ports: Vec<u16>,
+        chain_id: u8,
     ) -> NodeReference {
         match ip_addr_opt {
             Some(ip_addr) => NodeReference {
                 public_key,
                 node_addr_opt: Some(NodeAddr::new(&ip_addr, &ports)),
+                chain_id: chain_id,
```
masqrauder: This could be collapsed to `chain_id`

multinode_integration_tests/src/masq_node.rs
```diff
@@ -83,15 +104,18 @@ impl NodeReference {
...
             },
             None => NodeReference {
                 public_key,
                 node_addr_opt: None,
+                chain_id: chain_id,
```
masqrauder: This could be collapsed to `chain_id,`

multinode_integration_tests/src/masq_node.rs
```diff
@@ -26,20 +30,30 @@ use std::time::Instant;
 pub struct NodeReference {
     pub public_key: PublicKey,
     pub node_addr_opt: Option<NodeAddr>,
+    pub chain_id: u8,
 }
 
 impl FromStr for NodeReference {
     type Err = String;
 
+    //TODO inaccurate - this works only as long as we don't parse the real Node descriptor including its URL prefix which we don't now, see extract_node_reference()
```
masqrauder: With some slight changes this could almost be implemented using standard URL/URI classes where:
The standard URI is:
\<protocol\>://\<user\>:\<password\>@\<host or IP\>:\<port\>
So instead of
`masq://eth.QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU:1.2.3.4:1234`
we have
`masq://eth:QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU@1.2.3.4:1234`
protocol = masq
user = eth
password = QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU
host/ip = 1.2.3.4
port = 1234

`user` and `password` aren't actually user and password but the values we want to use in their place.

The one challenge is the comma delimited port numbers. Although those could be handled as standard query parameters or a path parameter such as:

`masq://eth:QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU@1.2.3.4:1234/5678`

I've confirmed that this code works:
```
    use uri_rs::{scheme::Scheme, Path, UserInfo};
    use uri_rs::host_name::HostName;

    let str_descriptor = "masq://eth:as45cs5c5@1.2.3.4:1234/5678";
    let uri = uri_rs::Uri::parse(str_descriptor).unwrap();
    let authority = uri.authority().unwrap();
    assert_eq!(uri.to_string(), str_descriptor);
    assert_eq!(uri.schema(), &Scheme::new("masq".to_string()));
    assert_eq!(uri.path(), &Path::of_abempty_from_strs(&vec!["5678"]));
    assert_eq!(authority.host_name(), &HostName::new("1.2.3.4".to_string()));
    assert_eq!(authority.user_info(), Some(&UserInfo::new("eth".to_string(), Some("as45cs5c5".to_string()))));
    assert_eq!(authority.port(), Some(1234u16));
```

```
    use uri_rs::{scheme::Scheme, Path, UserInfo};
    use uri_rs::host_name::HostName;

    let str_descriptor = "masq://eth:QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU@1.2.3.4:1234/5678";
    let uri = uri_rs::Uri::parse(str_descriptor).unwrap();
    let authority = uri.authority().unwrap();
    assert_eq!(uri.to_string(), str_descriptor);
    assert_eq!(uri.schema(), &Scheme::new("masq".to_string()));
    assert_eq!(uri.path(), &Path::of_abempty_from_strs(&vec!["5678"]));
    assert_eq!(authority.host_name(), &HostName::new("1.2.3.4".to_string()));
    assert_eq!(authority.port(), Some(1234u16));
    assert_eq!(authority.user_info(), Some(&UserInfo::new("eth".to_string(), Some("QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU".to_string()))));
```

node/src/blockchain/blockchain_interface.rs
```diff
@@ -900,7 +877,8 @@ mod tests {
             2u64,
         );
 
-        transport.assert_request("eth_sendRawTransaction", &[String::from(r#""0xf8a801847735940082dbe894384dec25e03f94931767ce4c3556168468ba24c380b844a9059cbb00000000000000000000000000000000000000000000000000626c61683132330000000000000000000000000000000000000000000000000000082f79cd900029a0b8e83e714af8bf1685b496912ee4aeff7007ba0f4c29ae50f513bc71ce6a18f4a06a923088306b4ee9cbfcdc62c9b396385f9b1c380134bf046d6c9ae47dea6578""#)]);
+        //TODO remove this comment: to my defence in the review, I replaced these strings because I believe that chain_id takes part in the signing process and because of changing the number the sign must look different now
```
masqrauder: Yes, the network ID is required to be included in the transaction and because the transaction is signed the payload changes if the network changes. The transaction is not valid and will be rejected if the network ID is not correct.

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,414 @@
+// Copyright (c) 2019-2021, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+use crate::blockchain::blockchain_interface::{
+    MAINNET_CONTRACT_ADDRESS, MULTINODE_TESTNET_CONTRACT_ADDRESS, RINKEBY_TESTNET_CONTRACT_ADDRESS,
+    ROPSTEN_TESTNET_CONTRACT_ADDRESS,
+};
+use crate::blockchain::blockchains_specific_constants::{
```
masqrauder: `blockchains` should not be plural here. It's like deers vs deer.

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,414 @@
...
+    pub non_num_chain_id: Blockchain,
+    pub in_command_name: &'static str,
+    pub directory_by_platform: &'static str,
+    pub chain_label: &'static str,
+    pub contract: Address,
+    pub contract_creation_block: u64,
+}
+
+#[track_caller]
+fn return_right_record<F>(closure: Box<F>) -> &'static BlockchainRecord
```
masqrauder: This method name is awkward. I think this could be implemented using a Map<String, BlockchainRecord> where a simple Map::get() -> Option<BlockchainRecord> (I know this is not valid Rust but hopefully it conveys the idea)

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,414 @@
...
+}
+
+fn return_right_record_opt<F>(closure: Box<F>) -> Option<&'static BlockchainRecord>
+where
+    F: FnMut(&&BlockchainRecord) -> bool,
+{
+    CHAINS.iter().find(closure)
+}
+
+fn return_right_record_by_chain_id(chain_id: u8) -> &'static BlockchainRecord {
```
masqrauder: This could be implemented with a map where the key is the `chain_id`.

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,414 @@
...
+    return_right_record(Box::new(|b: &&BlockchainRecord| b.num_chain_id == chain_id))
+}
+
+fn return_right_record_by_chain_name(chain_name: &str) -> &'static BlockchainRecord {
```
masqrauder: This could be implemented with a map where the key is the `chain_name`

node/src/blockchain/signature.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2017-2019, Substratum LLC (https://substratum.net) and/or its affiliates. All rights reserved.
 use serde_derive::{Deserialize, Serialize};
 
+//TODO what is this?; it seems unused.
```
masqrauder: I'll have to look into this to refresh my memory.

masq_lib/src/test_utils/utils.rs
```diff
@@ -3,8 +3,11 @@
...
+pub const TEST_DEFAULT_CHAIN_NAME: &str = "ropsten";
```
masqrauder: I guess I'm unclear why `DEFAULT_CHAIN_NAME` and `DEFAULT_CHAIN_ID` within the `test_utils` namespace needs to be declared with `TEST_` as a prefix. Is it too confusing? It does make one check to make sure the correct module was imported?

node/src/node_configurator/mod.rs
```diff
@@ -219,7 +219,8 @@ pub fn real_user_from_multi_config_or_populate(
     }
 }
 
-pub fn real_user_data_directory_opt_and_chain_name(
+#[allow(non_snake_case)]
+pub fn real_user__data_directory_opt__chain_name(
```
masqrauder: Why the extra underscores? It forces you to use the `#[allow(non_snake_case)]` directive. It doesn't really communicate extra information on it's own. Is following the naming standard while including a comment better?

node/src/blockchain/blockchains_specific_constants.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2019-2021, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+pub const MAINNET_CONTRACT_CREATION_BLOCK: u64 = 9_415_932;
+pub const ROPSTEN_TESTNET_CONTRACT_CREATION_BLOCK: u64 = 8_688_171;
+pub const RINKEBY_TESTNET_CONTRACT_CREATION_BLOCK: u64 = 5_893_771;
+pub const MULTINODE_TESTNET_CONTRACT_CREATION_BLOCK: u64 = 0;
```
masqrauder: Should this constant be in production code? Should it be limited to multinode tests?

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,414 @@
+// Copyright (c) 2019-2021, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+use crate::blockchain::blockchain_interface::{
```
masqrauder: Is there a specific need to introduce another file beyond `blockchain_interface.rs` and `blockchain_bridge.rs` to resolve a circular dependency or another similar reason?

node/src/sub_lib/neighborhood.rs
```diff
@@ -115,105 +119,123 @@ impl NeighborhoodMode {
...
 
         Ok(NodeDescriptor {
+            blockchain,
             encryption_public_key,
-            mainnet,
             node_addr_opt,
         })
     }
 
+    fn try_dismantle_str(str_descriptor: &str) -> Result<(Blockchain, &str, &str), String> {
```
masqrauder: Would a more common verb used in a scenario like this be `parse`? `try_parse() -> Option(Blockchain)` or `parse() -> Option(Uri)`

node/src/test_utils/neighborhood_test_utils.rs
```diff
@@ -92,7 +93,7 @@ pub fn neighborhood_from_nodes(
                 root.node_addr_opt().unwrap(),
                 vec![NodeDescriptor::from((
                     neighbor,
-                    DEFAULT_CHAIN_ID == chain_id_from_name(DEFAULT_CHAIN_NAME),
+                    Blockchain::EthRopsten, //TODO this is weird - it always was like wrong TEST_DEFAULT_CHAIN_ID and default_chain_name didn't point to the same thing (maybe it was intended but bad names used)
```
masqrauder: I don't believe this was always wrong. I can't recall exactly but I suspect these tests were to prove that the production values `"mainnet"`, network ID 1 along with the mainnet Contract Address would work properly when used with CryptDE mechanisms for signing and verifying signatures for network gossip.

masqrauder (changes requested):
There are a few things I'd like to discuss around the potential use of Uri (uniform resource identifier) format of node descriptors -- it would require a couple of changes to the format but would allow the use of standardized parsers. I also think keeping `BlockchainRecord`s grouped by various keys stored as `Map<>` as needed for easy lookup is good vs writing many of functions to repeatedly iterate `BlockchainRecord` vectors searching for matches.

masq_lib/src/constants.rs
```diff
@@ -12,6 +14,11 @@ pub const CURRENT_LOGFILE_NAME: &str = "MASQNode_rCURRENT.log";
 
 pub const MASQ_PROMPT: &str = "masq> ";
 
+pub const ETH_MAINNET_LABEL: &str = "eth";
+pub const ETH_ROPSTEN_LABEL: &str = "eth_t1";
```
masqrauder: Technically this should probably be `"eth_t3"`... I'm all for avoiding underscores at every opportunity since Rust requires them in so many places. For consistency with `"eth-mainent"` I recommend this be `"eth-t3"` or `"eth-ropsten"`.

masq_lib/src/constants.rs
```diff
@@ -12,6 +14,11 @@ pub const CURRENT_LOGFILE_NAME: &str = "MASQNode_rCURRENT.log";
...
+pub const ETH_RINKEBY_LABEL: &str = "eth_t2";
```
masqrauder: Technically this should probably be `"eth_t4"`... For consistency with `"eth-mainent"` I recommend this be `"eth-t4"` or `"eth-rinkeby"`

masq_lib/src/test_utils/utils.rs
bertllll: Thank you for this. I wouldn't notice without your help.

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,447 @@
...
+        num_chain_id: 5,
+        literal_chain_id: Blockchain::Dev,
+        in_command_name: "dev",
+        directory_by_platform: "dev",
+        chain_label: DEV_LABEL,
+        contract: MULTINODE_TESTNET_CONTRACT_ADDRESS,
+        contract_creation_block: MULTINODE_TESTNET_CONTRACT_CREATION_BLOCK,
+    },
+];
+
```
dnwiebe: I wonder if it would be helpful to have a series of constants here:
```
pub const ETH_MAINNET: &BlockchainRecord = &CHAINS[0];
```
etc. Maybe, maybe not...

When I look at this, I also feel a yearning for a chain enum. Like so:
```
pub enum Chain {Mainnet, Ropsten, Rinkeby, Dev}

impl Chain {
    pub fn id(&self) -> u8 {
        match (self) {
            Mainnet => 0,
            [etc]
```
Then you could do stuff like
```
pub fn get_chain(chain: Chain) -> &BlockchainRecord {
    &CHAINS[chain.id()]
}
[...]
let contract = get_chain(Chain::Rinkeby).contract;
```
But this could very well be silly wishing.

node/src/blockchain/blockchains.rs
```diff
@@ -0,0 +1,447 @@
...
+pub struct BlockchainRecord {
+    pub num_chain_id: u8,
+    pub literal_chain_id: Blockchain,
+    pub in_command_name: &'static str,
+    pub directory_by_platform: &'static str,
+    pub chain_label: &'static str,
+    pub contract: Address,
+    pub contract_creation_block: u64,
+}
+
```
dnwiebe: Instead of all these functions below that are tightly coupled to the structure of `BlockchainRecord`, could you do something like `pub fn chain_from_id (id: u8) -> &BlockchainRecord`?

Then, rather than having a special `contract_address()` method, you could say `chain_from_id(3).contract`, and if somebody added a field to `BlockchainRecord` he wouldn't have to add another special method to match it.

multinode_integration_tests/src/masq_node.rs
bertllll: I appreciate this as well. This is a great idea. I will work towards building it in the code.

multinode_integration_tests/src/masq_node.rs
```diff
@@ -476,7 +474,7 @@ mod tests {
 
     #[test]
     fn node_reference_can_display_itself() {
-        let chain_id = TEST_DEFAULT_MULTINODE_CHAIN_ID;
+        let chain_id = TEST_DEFAULT_MULTINODE_CHAIN;
```
kauri-hero: this still references `chain_id` - is this object supposed to be changed to `chain`??
@bertllll

masq_lib/Cargo.toml
```diff
@@ -20,6 +21,7 @@ serde_derive = "1.0.99"
 serde_json = "1.0.40"
 tiny-hderive = "0.2.1"
 toml = "0.5.3"
+web3 = {version = "0.8.0", default-features = false}
```
masqrauder, line 24: will this stay `0.8.0` or update to `0.11.0`?

masq_lib/src/blockchains/chains.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn default() -> Self {
+        DEFAULT_CHAIN
+    }
+}
+
+impl From<&str> for Chain {
+    fn from(str: &str) -> Self {
+        match str {
+            "polygon-mainnet" => Chain::PolyMainnet,
+            "eth-mainnet" => Chain::EthMainnet,
```
masqrauder: `polygon` is not abbreviated yet `eth` is. Does this feel too asymmetrical? Should it be `poly` and `eth` or `polygon` and `ethereum` for symmetry? I also wonder if having `eth-ropsten` and `polygon-mumbai` in the descriptor will mislead people to try to use those as command line parameters. The `from` function can just as easily accept "poly-mainnet", "poly-mumbai", "polygon-mumbai", "eth-ropsten", "ropsten", "eth-mainnet", "mainnet".

multinode_integration_tests/src/masq_node.rs
```diff
@@ -122,13 +149,13 @@ impl NodeReference {
...
             String::from(slice)
-                .split(',')
+                .split(NodeAddr::PORTS_SEPARATOR)
                 .map(|x| x.parse::<i64>().unwrap_or(-1))
                 .collect()
         };
         if port_list_numbers.contains(&-1) {
             return Err(format!(
-                "The port list must be a comma-separated sequence of valid numbers, not '{}'",
+                "The port list must be a semicolon-separated sequence of valid numbers, not '{}'",
```
masqrauder: Should `semicolon` be `/` instead?

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -331,7 +331,7 @@ fn assert_balances(
     }
 }
 
-fn deploy_smart_contract(wallet: &Wallet, web3: &Web3<Http>, chain_id: u8) -> Address {
+fn deploy_smart_contract(wallet: &Wallet, web3: &Web3<Http>, chain: Chain) -> Address {
```
masqrauder, line 338: It occurs to me that this may still be the SUB smart contract. We may want to consider updating this to the actual MASQ smart contract in case someone checks. Not to be done in this card but I may create a new card after some discussion.

kauri-hero:
Yes this was also brought up in QA testing this week.
I have addressed this wirth @bertllll  and commentd on the other PR version where you tagged me

multinode_integration_tests/src/masq_node.rs
bertllll: A good catch.

multinode_integration_tests/tests/verify_bill_payment.rs
bertllll: Please, do it, I'd appreciate if it were you who made the card.

masqrauder: I created an issue using the "Reference in new issue" function. I'll try to convert to a card next.

masq_lib/src/blockchains/chains.rs
bertllll: Could be... but I'd have to take care of clap's App configuration to accept these including the alternative.

Please, I'd really like to hear advice regarding what the look should be like...I can program it but there different preferences of everybody from us. 
Actually I don't care so much...

Now I quite like the idea of short and long version. We could do that.

But as for the abbreviated polygon to just "pol"y; I don't think it's right. If I look at their sites etc I always see them using the full name and so I'd consider it disrespectful to them to mispronounce their name - that's my insight. 
Also...Polygon is not as broadly known as Ethereum is and so I think it's in place to use the full name to help first-contact people understand. On the other hand if I use "eth" I suppose everybody who touches crypto must know. 
That's my reasoning - and even though I'm aware of the asymmetry I would hold it.
Plus another argument - first I claim I didn't take their idea and use it - I noticed later with Alchemy that they resolved the issue the same way as me. They also use "polygon-mumbai" and "eth-ropsten" iand etc. in their URLs.

bertllll: @kauri-hero Please add your opinion.

masq_lib/Cargo.toml
bertllll: I will leave it. Because in this card there is no need to touch it. We know that I will but it will be in another card, hopefully after we complete this one.

But yes the change to 0.11.0 is planned.

node/src/blockchain/raw_transaction.rs
```diff
@@ -90,55 +92,192 @@ mod tests {
...
+                    Bip32ECKeyPair::from_raw_secret(&signed.private_key.0.as_ref()).unwrap(),
+                ),
+                chain,
             );
+            assert_eq!(sign, signed.signed);
+            assert_eq!(sign[..length], **constant_part)
         }
     }
 
+    //TODO I think the tests below just duplicate what the other tests already tested
```
masqrauder, line 281: The distinction with these additional tests help to ensure the code creates valid signed transactions even with various chain IDs especially since Polygon Mumbai's ID is such a large number. It would be good to have a test including the latter.

masq_lib/src/blockchains/chains.rs
kauri-hero: Thanks @bertllll 
thinking more about it all, I think I agree more with your thoughts of ETH being an understand term as abbreviation fo ETH.

I think we should keep eth-mainnet, eth-ropsten, polygon-mainnet & polygon-mumbai for the chain names and descriptor formats.

do you think we should make a card to possibly change the `--chain` acceptable values to be in line with this?
Currently the `--chain` values accepted are:
`mainnet`
`ropsten`
`polygon`
`mumbai`

I think we can live without this alignment, but we need to be clear in our documentation about `--chain` and the fact that descriptors will appear different from the commandline visible `--chain` value that appears in CONFIG table and `setup`/`configuration` values

node/src/blockchain/signature.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
 use serde_derive::{Deserialize, Serialize};
 
+//TODO what is this for? It seems unused.
```
masqrauder: You could try deleting this and see if anything breaks.

bertllll: Yeah, I tried and it really seems to be used in serialization process, probably by things we do need to keep.

node/src/blockchain/raw_transaction.rs
bertllll: I can understand the general purpose of the tests in this file but I struggle to understand why it speaks of seemingly two situations with different kind of transactions (well it was more obvious before I changed the first sections of the tests - so please look at the master branch to see it unbiased) there have been tests like:

"test_signs_transaction_ropsten()" 
and later on another one called as 
"test_transfer_transaction_signing_ropsten()"

In my eyes they test the pretty much same thing. But I'm asking for confirming it. 
I'd like to get rid of the last two (maybe extra) tests if you agree.

masqrauder (approved):

masqrauder (approved):

PR #46, opened 2021-10-03
GH-372: Automap

dnwiebe:


automap/src/logger.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2019-2021, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
dnwiebe, line 1: Can this be swapped out in favor of the logger in `masq_lib`?

PR #57, opened 2021-10-22
GH-418: Update copyright tag in all files

FinsaasGH:


node/src/accountant/README.md
```diff
@@ -12,4 +12,4 @@ is a better place to start.
 
 Copyright (c) 2019-2021, MASQ Network
```
dnwiebe: Extraneous copyright message

dnwiebe (commented):
Two general comments.

First of two, the LICENSE file in the root directory is a little out of date and needs to be reviewed. One issue is that its copyright ends in 2020, not 2021, and another is that it gives a dead URL for the Substratum source code. (I don't think Substratum's source is publicly available anywhere anymore.)

Second of two, the reason we've been able to take over the copyright from Substratum on unchanged files is that Substratum--for some reason--terminated its copyright in 2019. That was unnecessary. Do we want to terminate our copyright in 2021, then, even though we don't have to? Maybe all those "Copyright 2019-2021" messages should just say "Copyright 2019." Then we wouldn't have to update all of them every year or be in danger of somebody stealing our copyright. (What if we say 2019-2021, and on January 1 2022 somebody publishes a forked repo with all "Copyright 2022 Usurper, LLC." messages before we can get our own 2022 updates in?)

And third of two, we noticed that several XXXXX.md files had duplicate copyright messages in them. Perhaps all of them do.

node/src/accountant/README.md
FinsaasGH: These have been removed

kauri-hero:
I agree with all the proposals and changes made @FinsaasGH @dnwiebe 

thanks for this - I will update the license.md file in the repo Root directory as well, as part of the rev2 branch I created for documentation updates

node/src/run_modes_factories.rs
```diff
@@ -61,11 +61,7 @@ pub trait ServerInitializerFactory {
     fn make(&self) -> Box<dyn ServerInitializer<Item = (), Error = ()>>;
 }
 pub trait DaemonInitializerFactory {
-    fn make(
-        &self,
-        args: &[String],
-        streams: &mut StdStreams,
-    ) -> Result<Box<dyn DaemonInitializer>, ConfiguratorError>;
+    fn make(&self, args: &[String]) -> Result<Box<dyn DaemonInitializer>, ConfiguratorError>;
```
kauri-hero: are these code changes just cleaning up formatting etc?

kauri-hero (approved):
I agree with all the proposals and changes made @FinsaasGH @dnwiebe 

thanks for this - I will update the license.md file in the repo Root directory as well, as part of the rev2 branch I created for documentation updates

node/src/run_modes_factories.rs
FinsaasGH: These formatting changes we recommended by IntelliJ.

bertllll:
Oh, @FinsaasGH, an issue has appeared. We made some effort with Dan to update your branch so that it's ready to merge. Since recently I didn't wake up to that this open pull request has been triggered remotely with a different underlying branch from the one which I open if I fetch GH-418 from the server. So we wrote the changes to a factually different branch and so this PR stayed idle.

Suggestion: please merge this branch from your remote repository into the branch called just GH-418 and then make a pull request out of the branch because both branches now lack a piece of code and you should make it one piece.

Thanks

By the way, in times like these, which are hectic, keep your eye on your branches and continuously merge in stuff if master has changed.

PR #62, opened 2021-10-25
Gh-478: Addition of config values written to database + updating db initialization and migration + better tooling in ntegration tests for Node

bertllll:


multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -64,51 +66,14 @@ fn verify_bill_payment() {
...
-        .build();
+    let (serving_node_1_config, serving_node_1_wallet) =
+        build_config(&blockchain_server, &seed, derivation_path(0, 1));
+    let (serving_node_2_config, serving_node_2_wallet) =
+        build_config(&blockchain_server, &seed, derivation_path(0, 2));
+    let (serving_node_3_config, serving_node_3_wallet) =
+        build_config(&blockchain_server, &seed, derivation_path(0, 3));
 
     let amount = 10u64
         * u64::try_from(node_lib::accountant::PAYMENT_CURVES.permanent_debt_allowed_gwub).unwrap();
```
steveswing, line 79: `gwub` is actually short for `$SUB in Gwei` so if you felt so inclined to change it to `gwei` or `qwaq` or something else that would be fine. Or you could leave it too... that's fine. It's one of those 'you had to be there to know' kind of things.

node/src/accountant/mod.rs
```diff
@@ -697,6 +700,10 @@ impl Accountant {
             })
             .expect("UiGateway is dead");
     }
+
+    pub fn dao_factory(data_directory: &Path) -> DaoFactoryReal {
+        DaoFactoryReal::new(data_directory, false, MigratorConfig::panic_on_migration())
+    }
 }
 
 // At the time of this writing, Rust 1.44.0 was unpredictably producing
```
steveswing, line 709: Is there an issue we can watch to determine when the function below is no longer required?

multinode_integration_tests/tests/verify_bill_payment.rs
masqrauder, line 83: `gwub` is actually short for `$SUB in Gwei` so if you felt so inclined to change it to `gwei` or `qwaq` or something else that would be fine. Or you could leave it too... that's fine. It's one of those 'you had to be there to know' kind of things.

node/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 #[macro_use]
 pub mod sub_lib;
 
-#[cfg_attr(test, macro_use)]
+//#[cfg_attr(test, macro_use)]
```
masqrauder: Should this be deleted?

masqrauder (approved):

multinode_integration_tests/tests/verify_bill_payment.rs
kauri-hero: I think we will likely change the naming of this sometime soon - we are looking to change the naming of the payment curve values anyways, since they are very confusing!

kauri-hero: `gwasq` ?

bertllll: I'm going to take care of the nomenclature in a different card because I've started that in there and I want to do it just at one place.

PR #72, opened 2021-11-15
GH-305: Partial panic and actors hold together until one crashes and then all of them. 

bertllll:


masq/src/communications/connection_manager.rs
```diff
@@ -393,6 +395,7 @@ impl ConnectionManagerThread {
         inner.active_port = Some(redirect_order.port);
         inner.listener_to_manager_rx = listener_to_manager_rx;
         inner.talker_half = talker_half;
+        //TODO this is for conversations but redirected fire and forget isn't handled if absolutely first message: GH-487");
```
dnwiebe: Nit-pick: "):

masq_lib/src/shared_schema.rs
```diff
@@ -65,7 +65,7 @@ pub const NEIGHBORS_HELP: &str = "One or more Node descriptors for running Nodes
           masq://eth-ropsten:OHsC2CAm4rmfCkaFfiynwxflUgVTJRb2oY5mWxNCQkY@150.60.42.72:6642\n\n\
      Notice each of the different chain identifiers in the masq protocol prefix - they determine a family of chains \
      and also the network the descriptor belongs to (mainnet or a testnet).\n\n\
-     If you have more than one descriptor, separate them with commas (but no spaces). There is no default value; \
+     If you have more than one descriptor, separate them with commas. There is no default value; \
```
dnwiebe: Mention here that Node descriptors need quotes, because they contain characters meaningful to the shell. Also, maybe in one of the examples you should supply more than one port number, so that the syntax for that is clear. (Not that it'll be used soon, but just so that we don't lose it.)

node/src/bootstrapper.rs
```diff
@@ -300,8 +300,8 @@ pub struct BootstrapperConfig {
     pub port_configurations: HashMap<u16, PortConfiguration>,
     pub data_directory: PathBuf,
     pub node_descriptor_opt: Option<String>,
-    pub main_cryptde_null_opt: Option<CryptDENull>,
-    pub alias_cryptde_null_opt: Option<CryptDENull>,
+    pub main_cryptde_null_opt: Option<Box<CryptDENull>>,
```
dnwiebe: UnBox?

node/src/bootstrapper.rs
```diff
@@ -1540,15 +1562,18 @@ mod tests {
                 NodeAddr::new(&IpAddr::from_str("1.2.3.4").unwrap(), &[4321]),
                 vec![NodeDescriptor::from((
                     cryptde.public_key(),
-                    &NodeAddr::new(&IpAddr::from_str("1.2.3.4").unwrap(), &[1234]),
+                    Some(&NodeAddr::new(
+                        &IpAddr::from_str("1.2.3.4").unwrap(),
+                        &[1234],
```
dnwiebe: Why does the test pass when this is 1234 rather than port?

node/src/bootstrapper.rs
```diff
@@ -1865,6 +1899,7 @@ mod tests {
             &self,
             config: BootstrapperConfig,
             _actor_factory: Box<dyn ActorFactory>,
+            _persist_config: &dyn PersistentConfiguration,
```
dnwiebe, line 1913: I'm a little concerned that ignoring this parameter may mean that its value isn't tested. Maybe there's an integration test for this somewhere.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -361,71 +362,98 @@ pub mod standard {
...
+                            }
+                        Err(e) => ParamError::new("neighbors", &e).wrap_to_err()
+                    }
+                }
+            )
+            .collect_vec()
+    }
+
+    fn validate_mandatory_node_addr(
+        supplied_version: &str,
```
dnwiebe: Don't say "version;" it makes the reader think of "1.2.3" or something. Maybe user_supplied_descriptor? Or just supplied_descriptor, since it might come from the environment or the config file as well as the command line?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -361,71 +362,98 @@ pub mod standard {
...
+        descriptor: NodeDescriptor,
+    ) -> Result<NodeDescriptor, ParamError> {
+        match descriptor.node_addr_opt.is_some(){
+            true => Ok(descriptor),
+            false => Err(ParamError::new(
+                "neighbors",&format!("Neighbors supplied without ip addresses and ports are not valid: '{}--NA--:--NA--",
```
dnwiebe: "--NA--" would be more idiomatic as "<N/A>"

dnwiebe (commented):
Resume at node_configurator_standard.rs, line 457.

node/src/sub_lib/cryptde_real.rs
```diff
@@ -180,6 +181,10 @@ impl CryptDE for CryptDEReal {
     fn digest(&self) -> [u8; 32] {
         self.digest
     }
+
+    fn as_any(&self) -> &dyn Any {
+        intentionally_blank!()
```
dnwiebe: Can you put a comment here explaining why this should never be called? A future developer might want to know...

node/src/sub_lib/hopper.rs
```diff
@@ -126,6 +127,7 @@ pub struct HopperConfig {
     pub per_routing_service: u64,
     pub per_routing_byte: u64,
     pub is_decentralized: bool,
+    pub crashable: bool,
```
dnwiebe, line 130: Is the Hopper crashability tested?

node/src/sub_lib/neighborhood.rs
```diff
@@ -126,32 +126,17 @@ pub struct NodeDescriptor {
     pub node_addr_opt: Option<NodeAddr>,
 }
 
-impl From<(&PublicKey, &NodeAddr, Chain, &dyn CryptDE)> for NodeDescriptor {
-    fn from(tuple: (&PublicKey, &NodeAddr, Chain, &dyn CryptDE)) -> Self {
-        let (public_key, node_addr, blockchain, cryptde) = tuple;
+impl From<(&PublicKey, Option<&NodeAddr>, Chain, &dyn CryptDE)> for NodeDescriptor {
```
dnwiebe: Is the no-IP-address version of this necessary? Can that Option<> be removed? All the tests seem to test Some(node_addr), none seem to test None.

node/src/sub_lib/neighborhood.rs
```diff
@@ -818,50 +809,13 @@ mod tests {
         );
     }
 
-    #[test]
-    fn node_descriptor_from_key_and_mainnet_flag_works_with_cryptde_null() {
```
dnwiebe, line 822: Maybe these tests should come back in and test the new 4-tuple version of From.

node/src/sub_lib/proxy_client.rs
```diff
@@ -24,6 +25,7 @@ pub struct ProxyClientConfig {
     pub dns_servers: Vec<SocketAddr>,
     pub exit_service_rate: u64,
     pub exit_byte_rate: u64,
+    pub crashable: bool,
```
dnwiebe, line 28: Looks like there's a crashable flag but no actual crashability functionality here.

node/src/sub_lib/proxy_server.rs
```diff
@@ -79,6 +80,7 @@ pub struct ProxyServerSubs {
     pub add_route: Recipient<AddRouteMessage>,
     pub stream_shutdown_sub: Recipient<StreamShutdownMsg>,
     pub set_consuming_wallet_sub: Recipient<SetConsumingWalletMessage>,
+    pub node_from_ui: Recipient<NodeFromUiMessage>,
```
dnwiebe, line 83: Looks like there's a crashable flag but no actual crashability functionality here.

node/src/sub_lib/utils.rs
```diff
@@ -97,19 +99,39 @@ pub fn make_new_multi_config<'a>(
...
-        info!(logger, "Rejected crash attempt: '{}'", msg.panic_message);
+}
+
+fn crash_request_analyzer(
+    msg: NodeFromUiMessage,
+    logger: &Logger,
+    crashable: bool,
+    crash_key: &str,
+) -> Option<UiCrashRequest> {
+    match (crashable, UiCrashRequest::fmb(msg.body)) {
```
dnwiebe: I'd like to see a separate if (!crashable) {return} here, because most of the time this won't be crashable, and UiCrashRequest::fmb() is expensive enough that saving a call to it, if we can, is worthwhile.

node/src/sub_lib/utils.rs
```diff
@@ -97,19 +99,39 @@ pub fn make_new_multi_config<'a>(
...
+        (false, _) => None,
+        (true, Err(_)) => None,
+        (true, Ok((msg, _))) if msg.actor == crash_key => Some(msg),
+        (true, Ok((msg, _))) => {
+            debug!(
```
dnwiebe: I'd rather see this message when a crash request comes for this particular actor, but it's not accepting crash requests.  However, in order to do that, you'd have to parse the UiCrashRequest message even when you're not accepting crash requests, to see whether it was aimed at this actor, and I don't want to spend that much time when !crashable. Therefore, I guess the right thing to do is never to log rejection of a crash request.

Another possibility would be to find a way to put the parsing of the crash request inside the call to logger.debug(), so that it only got executed if the log level was debug or higher. That might require a bit of finagling, though.

node/src/test_utils/mod.rs
```diff
@@ -551,6 +556,23 @@ pub mod pure_test_utils {
         }
     }
 
+    pub fn prove_that_crash_request_handler_is_hooked_up<
```
dnwiebe, line 562: It'd be better if this didn't hang forever on failure.

node/src/ui_gateway/mod.rs
```diff
@@ -3,38 +3,47 @@
...
     logger: Logger,
 }
 
 impl UiGateway {
-    pub fn new(config: &UiGatewayConfig) -> UiGateway {
+    pub fn new(config: &UiGatewayConfig, crashable: bool) -> UiGateway {
         UiGateway {
             port: config.ui_port,
-            websocket_supervisor: None,
+            //websocket_supervisor_factory: Some(Box::new(WebsocketSupervisorFactoryReal)),
```
dnwiebe: Um...

node/src/ui_gateway/mod.rs
```diff
@@ -217,7 +295,102 @@ mod tests {
...
+        };
+
+        subject_addr.try_send(msg.clone()).unwrap();
+
+        System::current().stop();
+        system.run();
+        let random_actor_recording = accountant_recording_arc.lock().unwrap();
+        assert_eq!(random_actor_recording.len(), 0);
+        TestLogHandler::new().exists_log_containing(
+            "WARN: UiGateway: Deserialization error: expected value at line 1 column 1",
```
dnwiebe: It would help forensic investigation if this log contained, say, the first 100 characters of the JSON. Not all of it, because it could be megabytes and megabytes, but the first few bytes could be really helpful to see what was going on.

node/src/ui_gateway/mod.rs
```diff
@@ -217,7 +295,102 @@ mod tests {
...
+        .tmb(0);
+        let crashable = false;
+        let subject = UiGateway::new(&UiGatewayConfig { ui_port: 123 }, crashable);
+
+        let result = subject.deserialization_validator_with_crash_request_handler(crash_request);
+
+        assert_eq!(result, None)
+    }
+
+    #[test]
```
dnwiebe, line 396: Is there a test for the actual crash?

node/src/sub_lib/hopper.rs
bertllll: Yes, it is. This snippet zooms only in its config. The actual actor is elsewhere.

node/src/sub_lib/proxy_client.rs
bertllll: Just seemingly...it has a test in a different file.

node/src/sub_lib/proxy_server.rs
bertllll: Same thing

node/src/sub_lib/neighborhood.rs
bertllll: I think the times we needed this variation of the constructor (if ever) has passed away.

masq_lib/src/shared_schema.rs
```diff
@@ -62,10 +62,12 @@ pub const NEIGHBORS_HELP: &str = "One or more Node descriptors for running Nodes
...
           masq://eth-mainnet:gBviQbjOS3e5ReFQCvIhUM3i02d1zPleo1iXg_EN6zQ@86.75.30.9:5542\n\
           masq://polygon-mumbai:A6PGHT3rRjaeFpD_rFi3qGEXAVPq7bJDfEUZpZaIyq8@14.10.50.6:10504\n\
-          masq://eth-ropsten:OHsC2CAm4rmfCkaFfiynwxflUgVTJRb2oY5mWxNCQkY@150.60.42.72:6642\n\n\
+          masq://eth-ropsten:OHsC2CAm4rmfCkaFfiynwxflUgVTJRb2oY5mWxNCQkY@150.60.42.72:6642/4789/5254\n\n\
      Notice each of the different chain identifiers in the masq protocol prefix - they determine a family of chains \
-     and also the network the descriptor belongs to (mainnet or a testnet).\n\n\
-     If you have more than one descriptor, separate them with commas (but no spaces). There is no default value; \
+     and also the network the descriptor belongs to (mainnet or a testnet). See also the last descriptor which shows \
+     a configuration with multiple clandestine ports.\n\n\
+     If you have more than one descriptor, separate them with commas. Either single or multiple descriptors, they \
```
dnwiebe: "Either" should be "Whether"

dnwiebe (approved):

PR #75, opened 2021-11-19
GH-305: second review

bertllll:


masq_lib/src/shared_schema.rs
```diff
@@ -62,10 +62,12 @@ pub const NEIGHBORS_HELP: &str = "One or more Node descriptors for running Nodes
...
           masq://eth-mainnet:gBviQbjOS3e5ReFQCvIhUM3i02d1zPleo1iXg_EN6zQ@86.75.30.9:5542\n\
           masq://polygon-mumbai:A6PGHT3rRjaeFpD_rFi3qGEXAVPq7bJDfEUZpZaIyq8@14.10.50.6:10504\n\
-          masq://eth-ropsten:OHsC2CAm4rmfCkaFfiynwxflUgVTJRb2oY5mWxNCQkY@150.60.42.72:6642\n\n\
+          masq://eth-ropsten:OHsC2CAm4rmfCkaFfiynwxflUgVTJRb2oY5mWxNCQkY@150.60.42.72:6642/4789/5254\n\n\
      Notice each of the different chain identifiers in the masq protocol prefix - they determine a family of chains \
-     and also the network the descriptor belongs to (mainnet or a testnet).\n\n\
-     If you have more than one descriptor, separate them with commas (but no spaces). There is no default value; \
+     and also the network the descriptor belongs to (mainnet or a testnet). See also the last descriptor which shows \
+     a configuration with multiple clandestine ports.\n\n\
+     If you have more than one descriptor, separate them with commas. Either single or multiple descriptors, they \
```
dnwiebe: "Either" should be "Whether"

node/src/actor_system_factory.rs
```diff
@@ -471,11 +476,123 @@ mod tests {
...
+                    BootstrapperConfig,
+                    Box<dyn ActorFactory>,
+                )>,
+            >,
+        >,
+        prepare_initial_messages_results: RefCell<Vec<StreamHandlerPoolSubs>>,
+        main_cryptde_ref_results: RefCell<Vec<&'static dyn CryptDE>>,
+        alias_cryptde_ref_results: RefCell<Vec<&'static dyn CryptDE>>,
+        database_chain_assertion_params: Arc<Mutex<Vec<Chain>>>,
+        database_chain_assertion_comapre_to_pointer:
```
dnwiebe: "compare," not "comapre"

node/src/actor_system_factory.rs
```diff
@@ -990,7 +1098,6 @@ mod tests {
             blockchain_bridge_param.consuming_wallet,
             Some(make_wallet("consuming"))
         );
-        let _stream_handler_pool_subs = rx.recv().unwrap();
         // more...more...what? How to check contents of _stream_handler_pool_subs?
```
dnwiebe: This comment should go along with the line above it that's already gone

node/src/actor_system_factory.rs
```diff
@@ -1089,15 +1195,13 @@ mod tests {
                 ),
             },
         };
-        let (tx, _) = unbounded();
         let system = System::new("MASQNode");
 
-        ActorSystemFactoryReal::prepare_initial_messages(
+        let _ = ActorSystemFactoryToolsReal.prepare_initial_messages(
```
dnwiebe, line 1202: A little concerned that the return value here is always ignored and never asserted.

node/src/actor_system_factory.rs
```diff
@@ -1323,6 +1402,99 @@ mod tests {
...
             &persistent_config,
+            &ActorSystemFactoryToolsReal,
+        );
+    }
+
+    #[test]
+    fn make_and_start_actors_happy_path() {
+        let database_chain_assertion_params_arc = Arc::new(Mutex::new(vec![]));
+        let prepare_initial_messages_params_arc = Arc::new(Mutex::new(vec![]));
+        let recorder = Recorder::new();
```
dnwiebe: You might be able to use make_recorder() here more simply

node/src/actor_system_factory.rs
```diff
@@ -1323,6 +1402,99 @@ mod tests {
...
+        bootstrapper_config.blockchain_bridge_config.chain = Chain::PolyMainnet;
+        let irrelevant_data_dir = PathBuf::new().join("big_directory/small_directory");
+        bootstrapper_config.data_directory = irrelevant_data_dir.clone();
+        bootstrapper_config.db_password_opt = Some("chameleon".to_string());
+        let main_cryptde = main_cryptde();
+        let main_cryptde_public_key_before = public_key_for_dyn_cryptde_being_null(main_cryptde);
+        let alias_cryptde = alias_cryptde();
+        let alias_cryptde_public_key_before = public_key_for_dyn_cryptde_being_null(alias_cryptde);
+        let actor_factory = Box::new(ActorFactoryReal {}) as Box<dyn ActorFactory>;
+        let actor_factory_raw_address = addr_of!(*actor_factory);
```
dnwiebe, line 1427: Either this should be called _to_compare, or the one above should be called persistent_config_raw_address for consistency

node/src/actor_system_factory.rs
```diff
@@ -1323,6 +1402,99 @@ mod tests {
...
+            bootstrapper_config_after.data_directory,
+            irrelevant_data_dir
+        );
+        assert_eq!(
+            bootstrapper_config_after.db_password_opt,
+            Some("chameleon".to_string())
+        );
+        assert_eq!(addr_of!(*actor_factory_after), actor_factory_raw_address);
+        let system =
+            System::new("make_and_start_actors_and_prepare_initial_messages_are_connected");
```
dnwiebe: This isn't the name of the test

node/src/bootstrapper.rs
```diff
@@ -1559,13 +1579,13 @@ mod tests {
         let mut config = BootstrapperConfig::new();
         config.neighborhood_config = NeighborhoodConfig {
             mode: NeighborhoodMode::Standard(
-                NodeAddr::new(&IpAddr::from_str("1.2.3.4").unwrap(), &[4321]),
```
dnwiebe, line 1562: I'll bet this value was originally put in to prove that the result wasn't coming from here--that this value was being overridden. I think it'd be clearer if you put it back in.

node/src/sub_lib/utils.rs
```diff
@@ -119,19 +119,21 @@ fn crash_request_analyzer(
...
-        (true, Ok((msg, _))) => {
-            debug!(
-                logger,
-                "Rejected crash instruction for '{}' with message '{}'",
-                msg.actor,
-                msg.panic_message
-            );
-            None
+    if !crashable {
+        if logger.debug_enabled() {
```
dnwiebe, line 123: This functionality is already in Logger. Try something like

logger.debug (|| {
   match ...
});

node/src/ui_gateway/mod.rs
```diff
@@ -328,9 +336,15 @@ mod tests {
         system.run();
         let random_actor_recording = accountant_recording_arc.lock().unwrap();
         assert_eq!(random_actor_recording.len(), 0);
-        TestLogHandler::new().exists_log_containing(
-            "WARN: UiGateway: Deserialization error: expected value at line 1 column 1",
-        );
+        let expected_msg_example = "some bad bite for a jason processor; filling up to 120 characters: abcdefghijklmnopqrstuvwxyzabcdefg";
```
dnwiebe: Some of this following code might be simpler if you used

let mut expected_msg_example = payload();
expected_msg_example.truncate(100);

node/tests/initialization_test.rs
```diff
@@ -193,7 +193,7 @@ fn started_without_explicit_chain_parameter_runs_finely() {
     );
 
     node.wait_for_log("UIGateway bound", Some(5000));
-    assert_eq!(node.kill().unwrap().code().unwrap(), 0);
```
dnwiebe, line 196: "started_without_explicit_chain_parameter_runs_smoothly" and "started_without_explicit_chain_parameter_runs_finely" should be the same above.

node/src/sub_lib/utils.rs
bertllll: I guess I'll leave it unchanged. The reason is at the additional condition that we want to print a log only if the actor is the right one otherwise we don't care but even in such a case we would have to give out a String because logger.debug() requires it. And we wouldn't have where to find any good solution...

node/src/node_test_utils.rs
```diff
@@ -274,19 +275,28 @@ pub fn start_recorder_ref_opt(recorder: &RefCell<Option<Recorder>>) -> Addr<Reco
 }
 
 pub fn make_stream_handler_pool_subs_from(
-    stream_handler_pool_opt: &RefCell<Option<Recorder>>, //the input type is complex in order to cover a wider range of implementations
+    stream_handler_pool_either_opt: Either<&RefCell<Option<Recorder>>, Option<Recorder>>,
```
dnwiebe: I don't understand the reason for the Either, or the necessity for the RefCell. Can you explain that?

node/src/node_test_utils.rs
```diff
@@ -274,19 +275,28 @@ pub fn start_recorder_ref_opt(recorder: &RefCell<Option<Recorder>>) -> Addr<Reco
...
 ) -> StreamHandlerPoolSubs {
-    #[allow(unused_assignments)]
-    let mut sentinel_owner = RefCell::new(None);
-    let stream_handler_pool = if stream_handler_pool_opt.borrow().is_some() {
-        stream_handler_pool_opt
+    let addr = if stream_handler_pool_either_opt.is_left() {
```
dnwiebe: This should really be a
```
match stream_handler_pool_either_opt {
    Left(inner) => [stuff],
    Right(recorder) => [stuff],
}
```

node/src/node_test_utils.rs
```diff
@@ -274,19 +275,28 @@ pub fn start_recorder_ref_opt(recorder: &RefCell<Option<Recorder>>) -> Addr<Reco
...
+        #[allow(unused_assignments)]
+        let mut sentinel_owner = RefCell::new(None);
+        let inner = stream_handler_pool_either_opt.left().unwrap();
+        let stream_handler_pool = if inner.borrow().is_some() {
```
dnwiebe: This probably should be a `match` too

dnwiebe (approved):

PR #91, opened 2021-12-24
GH-492: Update accountant to not use futures in scanning for payables/receivables

FinsaasGH:


node/src/accountant/mod.rs
```diff
@@ -277,12 +289,13 @@ impl Accountant {
             report_new_payments: addr.clone().recipient::<ReceivedPayments>(),
             report_sent_payments: addr.clone().recipient::<SentPayments>(),
             ui_message_sub: addr.clone().recipient::<NodeFromUiMessage>(),
+            scan_for_payables: addr.clone().recipient::<ScanForPayables>(),
```
bertllll: I think you can remove these two recipients because while using notify() for the messages you always know your own address. Other actors don't need to know too because they don't send these messages.

node/src/accountant/mod.rs
```diff
@@ -940,7 +899,12 @@ pub mod tests {
             self
         }
 
-        fn non_pending_payables_result(self, result: Vec<PayableAccount>) -> Self {
+        fn non_pending_payables_params(mut self, params: &Arc<Mutex<Vec<()>>>) -> Self {
+            self.non_pending_payables_params = params.clone();
+            self
+        }
+
+        pub(crate) fn non_pending_payables_result(self, result: Vec<PayableAccount>) -> Self {
```
bertllll: Please remove this 'crate' word because it is unnecessary. Also at other places down here.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
-            .non_pending_payables_result(vec![PayableAccount {
-                wallet: expected_wallet.clone(),
-                balance: PAYMENT_CURVES.permanent_debt_allowed_gwub + 1000,
-                last_paid_timestamp: from_time_t(
-                    now - PAYMENT_CURVES.balance_decreases_for_sec - 10,
-                ),
+    fn accountant_sends_report_accounts_payable_to_blockchain_bridge_when_qualified_payments_found()
+    {
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let blockchain_bridge = blockchain_bridge.retrieve_transactions_response(Ok(vec![]));
```
bertllll: This is not needed anymore. You can try to delete it and run it without it. It will pass I think.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
-        let payable_dao = PayableDaoMock::new()
-            .non_pending_payables_result(vec![PayableAccount {
-                wallet: expected_wallet.clone(),
-                balance: PAYMENT_CURVES.permanent_debt_allowed_gwub + 1000,
-                last_paid_timestamp: from_time_t(
-                    now - PAYMENT_CURVES.balance_decreases_for_sec - 10,
-                ),
+    fn accountant_scans_for_received_payments_and_sends_message_to_blockchain_bridge() {
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let blockchain_bridge = blockchain_bridge.retrieve_transactions_response(Ok(vec![]));
```
bertllll: It should never be used because now we don't send Actix messages waiting for a response, we send just a one-way message in this new design.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
-        });
+            },
+            PayableAccount {
+                wallet: make_wallet("foo"),
+                balance: PAYMENT_CURVES.balance_to_decrease_from_gwub + 100,
+                last_paid_timestamp: from_time_t(PAYMENT_CURVES.payment_suggested_after_sec + 5),
+                pending_payment_transaction: None,
+            },
+        ];
+        let payable_dao = PayableDaoMock::new().non_pending_payables_result(accounts.clone());
```
bertllll: I think this mock belonging to payable_dao is not important for this test and so it doesn't have to produce any results, an empty vector instead of these accounts should be just fine.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
+                AccountantConfig {
+                    payables_scan_interval: Duration::from_secs(100_000), //scan intervals deliberately big to demonstrate that we don't do the intervals yet
+                    receivables_scan_interval: Duration::from_secs(100_000),
+                },
+                earning_wallet.clone(),
+            )),
+            Some(payable_dao),
+            Some(receivable_dao),
+            None,
+            None,
```
bertllll, line 1571: Here, there is a problem. Writing this test, the right thing to do would have been to look at what function calls are there under scan_for_receivables(). You can find a call to persistent_configuration, namely start_block(). If you look at it closely you can see persistent_config is a trait object, that is something unspecific but you know it implements the trait which it is called after.

Then, if this happens, you should think of that it might be so because it is a mockable thing (a complicated kind of object which we usually want to replace with a pretender just acting like the real one but almost empty and unable to do more complicated computations).

So here we have one. You mostly prefer to work with mocks in your tests. To write a strong enough test when there is a chance to assert on values which are supposed to come from a mocked object - speaking of start_block() in this case - you want to give the mock a value which will make sure for observers that the values was handed over and got finally to you thorough the code. Here it could be whatever, say 555666, meaning the start block as if saved in the database and you query for it. 

You would set up the mock like this: 
let persistent_configuration = PersistentConfiguratiionMock::default().start_block_result(Ok(555666));

Once you have it you inject that into the subject that holds it.  Here it is the subject, Accountant.
You therefore probably want to supply it in the process of creating it - where you say make_subject(....).
Instead of just None you will paste in Some(persistent_configuration) in the right place, with this parameter.

You will also have to change your assertion slightly because the previous will bleed through to the final message you're asserting on.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
+        send_bind_message!(accountant_subs, peer_actors);
+        send_start_message!(accountant_subs);
 
-        let accountant_recording = accountant_recording_arc.lock().unwrap();
-        assert_eq!(0, accountant_recording.len());
+        System::current().stop();
+        system.run();
+        let blockchain_bridge_recorder = blockchain_bridge_recording_arc.lock().unwrap();
+        assert_eq!(blockchain_bridge_recorder.len(), 2); //the second message is from the scan for receivables
+        let scans_for_received_payments_msgs: Vec<&RetrieveTransactions> = (0
```
bertllll: This variable (and the test result in one) should rather be named by retrieve_transactions_msgs;

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
+            ..blockchain_bridge_recorder.len())
+            .flat_map(|index| {
+                blockchain_bridge_recorder.get_record_opt::<RetrieveTransactions>(index)
+            })
+            .collect();
+        assert_eq!(
+            *scans_for_received_payments_msgs,
+            vec![&RetrieveTransactions {
+                start_block: 0u64,
```
bertllll: This will change as I said above - and it will make this test stronger.

node/src/accountant/mod.rs
```diff
@@ -1532,151 +1502,127 @@ pub mod tests {
...
+                recipient: earning_wallet.clone()
+            }]
+        );
     }
 
     #[test]
```
bertllll, line 1598: I will have to find a way how to sort out  the following test, for now, just don't resolve this comment and work on the other things.

node/src/accountant/mod.rs
```diff
@@ -1742,129 +1688,105 @@ pub mod tests {
...
     fn accountant_logs_if_no_transactions_were_detected() {
         init_test_logging();
-        let earning_wallet = make_wallet("earner3000");
-        let blockchain_bridge = Recorder::new().retrieve_transactions_response(Ok(vec![]));
-        let blockchain_bridge_awaiter = blockchain_bridge.get_awaiter();
-        let blockchain_bridge_recording = blockchain_bridge.get_recording();
-        let (accountant_mock, _, accountant_recording_arc) = make_recorder();
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let blockchain_bridge = blockchain_bridge.retrieve_transactions_response(Ok(vec![]));
```
bertllll: Could be also redundant now - basically the whole mechanism behind this could disappear - it is placed at Recorder in recorder.rs, I believe.

node/src/accountant/mod.rs
```diff
@@ -1742,129 +1688,105 @@ pub mod tests {
...
-            send_start_message!(subject_subs);
-
-            system.run();
-        });
+        System::current().stop();
+        system.run();
 
-        blockchain_bridge_awaiter.await_message_count(1);
+        let blockchain_bridge_recorder = blockchain_bridge_recording_arc.lock().unwrap();
+        assert_eq!(blockchain_bridge_recorder.len(), 1);
```
bertllll: If you record out just a single message you don't need to use the big apparatus below to get the message. You simply write something like  
let received_msg = blockchain_bridge_recorder.get_record::<RetrieveTransactions>(0);

node/src/accountant/mod.rs
```diff
@@ -1742,129 +1688,105 @@ pub mod tests {
...
-        blockchain_bridge_awaiter.await_message_count(1);
-        let retrieve_transactions_recording = blockchain_bridge_recording.lock().unwrap();
-        let retrieve_transactions_message =
-            retrieve_transactions_recording.get_record::<RetrieveTransactions>(0);
-        assert_eq!(earning_wallet, retrieve_transactions_message.recipient);
+        addr.try_send(RetrieveTransactions {
+            start_block: 0u64,
+            recipient: earning_wallet.clone(),
+        })
+        .expect("query failed, aborting message");
```
bertllll: Sorry, this is really bad.

It's a misunderstanding of the concept of testing...
You cannot end the test prematurely and act as if it was no problem. 
See that the test aims at asserting around the logging and the assertion lies exactly at the very end of the test, but you panics earlier, before you get as far as where the assertion lies. 

I don't know what is worse:
a) you really don't assert at anything relevant in this test (because you don't start making the assertion)
b) you have a weird panic in the middle of the test and you don't care (and you instead rewrite the test so that it accepts the panic and looks green and fine)

You will have to try harder here. 

Hint: you cannot send messages after you shut down the system, which means after 

```
System::current().stop();
system.run();
```

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -175,25 +174,53 @@ impl BlockchainBridge {
...
+                .expect("Accountant is unbound")
+                .try_send(ReceivedPayments {
+                    payments: transactions,
+                })
+                .expect("Accountant is dead."),
+            Err(e) => warning!(self.logger, "{}", e),
+        }
+    }
+
+    fn handle_retrieve_transactions_inner(&self, msg: RetrieveTransactions) -> Transactions {
```
bertllll: You're following the pattern given by me, I get it, but not always can you. It doesn't fit here because the 'inner' is very simple and it doesn't deserve its own function, at the end of the day, this is just one function call wrapped into another one and that always smells fishy unless you really want to construct a wrapper (which is a class on its own).

I did that just so I could spare some space in tests, writing less complicated tests, tailored just for the inner's code.

In fact, I think I evaluated it badly I could just do without this finesse. The tests would have probably looked pretty the same, with small differences, maybe. 

So my wish here is that you should get rid of my "something_inner" and your "somethingelse_inner" and reset the tests to address this situation.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -653,13 +455,171 @@ mod tests {
...
 
-        let result = subject.handle_report_accounts_payable(request);
+        let result = subject.handle_report_accounts_payable_inner(request);
 
-        assert_eq!(result.0, Err("No consuming wallet specified".to_string()));
+        assert_eq!(result, Err("No consuming wallet specified".to_string()));
+    }
+
+    #[test]
+    fn report_accounts_payable_sends_transactions_to_blockchain_interface() {
```
bertllll: My bad. This should rather tell:  handle_report_accounts_payable_transacts_and_sends_finished_payments_back_to_accountant()

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -653,13 +455,171 @@ mod tests {
...
+    }
+
+    #[test]
+    fn retrieve_transactions_sends_transactions_to_blockchain_interface() {
+        let system =
+            System::new("retrieve_transactions_sends_transactions_to_blockchain_interface");
+        let (accountant, _, accountant_recording_arc) = make_recorder();
+        let earning_wallet = make_wallet("somewallet");
+        let amount = 42u64;
+        let amount2 = 55u64;
```
bertllll: You don't have to say what numeric data type it is, Rust can elide it alone.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -653,13 +455,171 @@ mod tests {
...
+        let subject_subs = BlockchainBridge::make_subs_from(&addr);
+        let peer_actors = peer_actors_builder().accountant(accountant).build();
+        send_bind_message!(subject_subs, peer_actors);
+
+        let retrieve_transactions = RetrieveTransactions {
+            start_block: 0u64,
+            recipient: earning_wallet.clone(),
+        };
+        addr.try_send(retrieve_transactions)
+            .expect("Blockchain bridge is dead");
```
bertllll: You don't need to use this expect in a test we do that in the production code but unwrap() is just enough here.

You should also keep some alignment which means that above retrieve_transactions wouldn't be a blank line but would be after the block of code so that addr.try_send() stands alone.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -653,13 +455,171 @@ mod tests {
...
+            System::new("retrieve_transactions_sends_transactions_to_blockchain_interface");
+        let (accountant, _, accountant_recording_arc) = make_recorder();
+        let earning_wallet = make_wallet("somewallet");
+        let amount = 42u64;
+        let amount2 = 55u64;
+        let expected_gas_price = 5u64;
+        let expected_transactions = vec![
+            Transaction {
+                block_number: 7u64,
+                from: earning_wallet.clone(),
```
bertllll, line 610: These earning wallets probably shouldn't be the same as our earning wallet which you can see at 595. 
But I'm talking about the real world - this test may not be affected by that, still...

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -653,13 +455,171 @@ mod tests {
...
+                    amount: 210,
+                    timestamp: from_time_t(0), //cannot be asserted directly, this field is just a sentinel
+                    transaction: H256::from("someothertransactionhash".keccak256())
+                })
+            ]
+        )
+    }
+
+    #[test]
+    fn retrieve_transactions_sends_transactions_to_blockchain_interface() {
```
bertllll: A better name, mentioning that we're sending a message at the end of the test (and whom), would be becoming, this one says just a half of the story.

node/src/accountant/mod.rs
FinsaasGH: This is the test you made to send a ReportAccountsPayable msg to blockchain bridge. Are you sure it should be deleted?

node/src/accountant/mod.rs
FinsaasGH: This is the test I made using your template of the above test, to send a RetrieveTransactions msg to blockchain bridge. Are you sure this should be deleted?

node/src/accountant/mod.rs
FinsaasGH: Without a valid payable account only 1 msg will be sent, not 2

node/src/accountant/mod.rs
FinsaasGH: Some tests don't have persistantconfigmock and some do have it. I flipped a coin whether it should have it or not.

node/src/accountant/mod.rs
FinsaasGH: I have changed this test a bit and I am pushing changes to repo now

node/src/accountant/mod.rs
```diff
@@ -48,12 +47,12 @@ use payable_dao::PayableDao;
...
 use std::path::Path;
-use std::thread;
 use std::time::{Duration, SystemTime};
 
 pub const CRASH_KEY: &str = "ACCOUNTANT";
-pub const DEFAULT_PAYABLE_SCAN_INTERVAL: u64 = 3600; // one hour
-pub const DEFAULT_PAYMENT_RECEIVED_SCAN_INTERVAL: u64 = 3600; // one hour
+//TODO evaluate if these should be configurable by user in UI and/or CLI
+pub const DEFAULT_PAYABLES_SCAN_INTERVAL: u64 = 3600; // one hour
+pub const DEFAULT_RECEIVABLES_SCAN_INTERVAL: u64 = 300; // one hour
```
dnwiebe: This is five minutes, not one hour

node/src/accountant/mod.rs
```diff
@@ -48,12 +47,12 @@ use payable_dao::PayableDao;
 use receivable_dao::ReceivableDao;
 use std::ops::Add;
 use std::path::Path;
-use std::thread;
 use std::time::{Duration, SystemTime};
 
 pub const CRASH_KEY: &str = "ACCOUNTANT";
-pub const DEFAULT_PAYABLE_SCAN_INTERVAL: u64 = 3600; // one hour
-pub const DEFAULT_PAYMENT_RECEIVED_SCAN_INTERVAL: u64 = 3600; // one hour
+//TODO evaluate if these should be configurable by user in UI and/or CLI
```
dnwiebe: These should, absolutely, and there's a card for it; maybe this comment is no longer necessary.

node/src/accountant/mod.rs
```diff
@@ -1510,369 +1467,273 @@ pub mod tests {
...
-            expected_pending_payment_transaction.clone();
-
-        let payable_dao = PayableDaoMock::new()
-            .non_pending_payables_result(vec![PayableAccount {
-                wallet: expected_wallet.clone(),
-                balance: PAYMENT_CURVES.permanent_debt_allowed_gwub + 1000,
+    fn accountant_sends_report_accounts_payable_to_blockchain_bridge_when_qualified_payments_found()
+    {
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let accounts = vec![
```
dnwiebe, line 1492: These accounts would be a little more distinctive--and the test a little stronger--if they differed by more than just the wallet address.

node/src/accountant/mod.rs
```diff
@@ -1510,369 +1467,273 @@ pub mod tests {
...
-                payments: vec![Ok(expected_payment)]
-            }
+            report_accounts_payables_msgs,
+            vec![&ReportAccountsPayable { accounts }]
         );
     }
 
     #[test]
-    fn accountant_logs_warn_when_blockchain_bridge_report_accounts_payable_errors() {
+    fn accountant_scans_for_received_payments_and_sends_message_to_blockchain_bridge() {
```
dnwiebe: The Accountant isn't really doing the scan here; it's instructing the BlockchainBridge to do the scan.

node/src/accountant/mod.rs
```diff
@@ -1510,369 +1467,273 @@ pub mod tests {
...
             },
             earning_wallet.clone(),
         );
+        let new_delinquent_account = ReceivableAccount {
+            wallet: wallet_to_be_banned.clone(),
+            balance: 4567,
+            last_received_timestamp: from_time_t(200_000_000),
+        };
+        let system =
+            System::new("accountant_payment_received_scan_timer_triggers_scanning_for_payments");
```
dnwiebe: This is the old name for the test

node/src/accountant/mod.rs
```diff
@@ -1510,369 +1467,273 @@ pub mod tests {
...
-        blockchain_bridge_awaiter.await_message_count(1);
-        let retrieve_transactions_recording = blockchain_bridge_recording.lock().unwrap();
-        let retrieve_transactions_message =
-            retrieve_transactions_recording.get_record::<RetrieveTransactions>(0);
-        assert_eq!(earning_wallet, retrieve_transactions_message.recipient);
-
-        TestLogHandler::new().exists_log_containing(
-            r#"WARN: Accountant: Unable to retrieve transactions from Blockchain Bridge: QueryFailed("really bad")"#,
-        );
+        TestLogHandler::new().exists_log_containing("DEBUG: Accountant: Scanning for payments");
```
dnwiebe: The original test was looking for a log saying "DEBUG: Accountant: No payments detected." The value of the new assertion alone is dubious. This entire test may have been overtaken by events.

node/src/accountant/mod.rs
```diff
@@ -1510,369 +1467,273 @@ pub mod tests {
...
     }
 
     #[test]
     fn accountant_receives_new_payments_to_the_receivables_dao() {
+        init_test_logging();
```
dnwiebe: Doesn't look necessary; I don't think we're asserting on any logs here.

node/src/accountant/mod.rs
```diff
@@ -1925,59 +1784,74 @@ pub mod tests {
...
-            let subject_subs = Accountant::make_subs_from(&subject_addr);
-
-            send_bind_message!(subject_subs, peer_actors);
-            send_start_message!(subject_subs);
+        let non_pending_payables_params_arc = Arc::new(Mutex::new(vec![]));
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let receivable_dao = ReceivableDaoMock::new()
+            .new_delinquencies_result(vec![])
+            .paid_delinquencies_result(vec![]);
+        let system = System::new("accountant_payable_scan_timer_triggers_scanning_for_payables");
```
dnwiebe: Test name has changed

node/src/accountant/mod.rs
```diff
@@ -1925,59 +1784,74 @@ pub mod tests {
...
+        let config = bc_from_ac_plus_earning_wallet(
+            AccountantConfig {
+                payables_scan_interval: Duration::from_millis(100),
+                receivables_scan_interval: Duration::from_secs(50_000), //deliberately big to refrain from starting off this scanning
+            },
+            make_wallet("hi"),
+        );
+        let now = to_time_t(SystemTime::now());
+        // slightly above minimum balance, to the right of the curve (time intersection)
```
dnwiebe: Unnecessary

node/src/accountant/mod.rs
```diff
@@ -1925,59 +1784,74 @@ pub mod tests {
...
+                blockchain_bridge_recorder.get_record_opt::<ReportAccountsPayable>(index)
+            })
+            .collect();
+        assert_eq!(
+            report_accounts_payables_msgs,
+            vec![&ReportAccountsPayable {
+                accounts: vec![account]
+            }]
+        );
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
```
dnwiebe: Not really helpful

node/src/accountant/mod.rs
```diff
@@ -1988,25 +1862,25 @@ pub mod tests {
...
-
         send_bind_message!(subject_subs, peer_actors);
+
         send_start_message!(subject_subs);
 
         System::current().stop();
         system.run();
-
         let tlh = TestLogHandler::new();
         tlh.await_log_containing("DEBUG: Accountant: Scanning for payables", 1000u64);
```
dnwiebe: Shouldn't need an "await" here

node/src/accountant/mod.rs
```diff
@@ -2109,91 +1982,31 @@ pub mod tests {
...
+        let subject_addr = subject.start();
+        let accountant_subs = Accountant::make_subs_from(&subject_addr);
+        send_bind_message!(accountant_subs, peer_actors);
 
-            system.run();
-        });
+        send_start_message!(accountant_subs);
 
-        blockchain_bridge_awaiter.await_message_count(1);
+        system.run();
```
dnwiebe, line 1944: Try System.current().stop() here...

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -426,223 +358,242 @@ mod tests {
...
         let blockchain_interface_mock = BlockchainInterfaceMock::default()
-            .retrieve_transactions_result(result)
-            .contract_address_result(TEST_DEFAULT_CHAIN.rec().contract);
-        let retrieve_transactions_parameters = blockchain_interface_mock
-            .retrieve_transactions_parameters
+            .get_transaction_count_result(Ok(web3::types::U256::from(1)))
+            .send_transaction_result(Err(BlockchainError::TransactionFailed(String::from(
+                "mock payment failure",
+            ))));
+        let transaction_count_parameters = blockchain_interface_mock
```
dnwiebe: Wha...?

Maybe define a local parameters list and use it the same way we're accustomed to doing?

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -426,223 +358,242 @@ mod tests {
...
                 PayableAccount {
                     wallet: make_wallet("foo"),
-                    balance: 21,
-                    last_paid_timestamp: SystemTime::now(),
+                    balance: 210,
+                    last_paid_timestamp: from_time_t(120_000_000),
                     pending_payment_transaction: None,
                 },
             ],
         });
```
dnwiebe: There should be an .unwrap() here

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -426,223 +358,242 @@ mod tests {
...
-            transaction_count_parameters.lock().unwrap()[0],
-            consuming_wallet.clone(),
+            *get_transaction_count_params,
+            vec![consuming_wallet.clone(), consuming_wallet.clone()]
         );
+        let accountant_received_payment = accountant_recording_arc.lock().unwrap();
+        assert_eq!(accountant_received_payment.len(), 1);
+        let sent_payment_msg = accountant_received_payment.get_record::<SentPayments>(0);
+        let mut sent_payments_cloned = sent_payment_msg.payments.clone();
+        sent_payments_cloned
```
dnwiebe, line 508: Pull the actual timestamps out into a separate list here, before replacing them with 0, and then assert separately on the list of timestamps that all of them are between when the Act started and when it finished.

Also, replace the `if let` with an `unwrap()` so that if there is an error in the list, we get an immediate failure. (Unless there are meant to be errors in the list, of course: then assert that they're what they should be.)

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -426,223 +358,242 @@ mod tests {
...
+        let peer_actors = peer_actors_builder().accountant(accountant).build();
+        send_bind_message!(subject_subs, peer_actors);
+        let retrieve_transactions = RetrieveTransactions {
+            start_block: 6,
+            recipient: earning_wallet.clone(),
         };
 
-        let result = subject.handle_report_accounts_payable(request);
+        addr.try_send(retrieve_transactions)
+            .expect("Blockchain bridge is dead");
```
dnwiebe: This could be an .unwrap()

dnwiebe (approved):

PR #92, opened 2021-12-28
GH-496: Updated dependencies to latest available

FinsaasGH:


dnwiebe (approved):
I created a card (GH-537) to address the `serde::export` -> `serde::__private` issue; therefore, I have no further objections here.

PR #95, opened 2021-12-31
Gh 477 - Store consuming wallet private key in database

dnwiebe:


node/src/database/db_migrations.rs
```diff
@@ -832,6 +962,54 @@ mod tests {
         );
     }
 
+    #[test]
+    fn db_password_works() {
```
bertllll, line 963: I don't think you really need a true database for this test. ConnectionWrapperMock could be enough here.

node/src/database/db_migrations.rs
```diff
@@ -1244,4 +1424,144 @@ mod tests {
...
+            let schema4_conn = subject
+                .initialize_to_version(&data_path, 4, false, migrator_config)
+                .unwrap();
+            let mut stmt = schema4_conn.prepare ("select count(*) from config where name in ('consuming_wallet_derivation_path', 'consuming_wallet_public_key', 'seed')").unwrap();
+            let cruft = stmt
+                .query_row([], |row| Ok(row.get::<usize, u32>(0)))
+                .unwrap()
+                .unwrap();
+            assert_eq!(cruft, 0);
+            let mut stmt = schema4_conn.prepare ("select value, encrypted from config where name = 'consuming_wallet_private_key'").unwrap();
```
bertllll: You could theoretically use the utility widely used in other tests: assurance_query_for_config_table() it allows you to get easily information about all three values from a single row. 

...now when I'm thinking of it...that function I mentioned might deserve a better name (so may a native speaker neaten it if willing to)

node/src/database/db_migrations.rs
```diff
@@ -1244,4 +1424,144 @@ mod tests {
...
+            .initialize_to_version(&data_path, 4, false, migrator_config)
+            .unwrap();
+
+        let mut stmt = schema4_conn.prepare ("select count(*) from config where name in ('consuming_wallet_derivation_path', 'consuming_wallet_public_key', 'seed')").unwrap();
+        let cruft = stmt
+            .query_row([], |row| Ok(row.get::<usize, u32>(0)))
+            .unwrap()
+            .unwrap();
+        assert_eq!(cruft, 0);
+        let mut stmt = schema4_conn
```
bertllll: Here you could also use the utility for assertions made over the config table.

node/src/sub_lib/accountant.rs
```diff
@@ -12,6 +12,8 @@ use std::str::FromStr;
 use std::time::Duration;
 
 lazy_static! {
+    // TODO: How is this used? Does it ever get stored in the database, or is it a default assigned
+    // if the database is blank?
```
bertllll: This should be probably investigated. Maybe a question to @kauri-hero too.

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -118,25 +115,27 @@ impl PersistentConfiguration for PersistentConfigurationMock {
...
-        Self::result_from(&self.consuming_wallet_derivation_path_results)
+    fn consuming_wallet_private_key(
+        &self,
+        db_password: &str,
+    ) -> Result<Option<String>, PersistentConfigError> {
+        self.consuming_wallet_private_key_params
+            .lock()
+            .unwrap()
+            .push(db_password.to_string());
+        Self::result_from(&self.consuming_wallet_private_key_results)
```
bertllll, line 146: I don't like these ever returning mock methods; one day I think I will restrict them somehow to work so only if you ask for that explicitly. I think we should contain the trait of mocks getting you notified when an unprepared result is queried and so it blows up - it's a very useful attribute and it can commit troubles if you break this by using mocks with refiling of their queues by automation.   

But it's not required to make it right now, just consider not enlarging the line of tests being dependent on this trick.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -306,54 +308,86 @@ pub fn get_wallets(
...
+    let earning_wallet_opt = earning_opt.map(|earning_address| {
+        Wallet::from_str(&earning_address).unwrap_or_else(|_| {
+            panic!(
+                "Wallet corruption: bad value for earning wallet address: {}",
+                earning_address
+            )
+        })
+    });
+    config.consuming_wallet = consuming_wallet_opt;
+    config.earning_wallet = earning_wallet_opt.unwrap_or_else(|| DEFAULT_EARNING_WALLET.clone());
```
bertllll, line 408: Here it shows how we use the default for the earning wallet.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -928,7 +786,7 @@ mod tests {
     }
 
     #[test]
-    fn get_earning_wallet_from_address_handles_attempted_wallet_change() {
+    fn get_earning_wallet_handles_attempted_wallet_change() {
```
bertllll: I don't understand. I found the function "get_earning_wallet{}" at the end of this file, looking like a testing aid.
Do we use it in production somehow?
Does this test need to exist?
...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -306,54 +308,86 @@ pub fn get_wallets(
...
+    };
+    let consuming_opt = match (&mc_consuming_opt, &pc_consuming_opt) {
+        (None, _) => pc_consuming_opt,
+        (Some(_), None) => mc_consuming_opt,
+        (Some(m), Some(c)) if wallet_parms_are_equal(m, c) => pc_consuming_opt,
+        _ => {
+            return Err(ConfiguratorError::required(
+                "consuming-private-key",
+                "Cannot change to a private key different from that previously set",
+            ))
```
bertllll, line 364: I don't know what should I think about this newly written body of this method. It looks legit, indeed, but I have a feeling like the complexity (with the nested conditions) of the code doesn't correspond to the amount of changes in the test section. It's hard to say. I guess you might have covered all the variations, when I imagine without the panics which you obviously didn't test and I think it's all right.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -873,39 +754,16 @@ mod tests {
             vec![Box::new(CommandLineVcl::new(args.into()))];
         let multi_config = make_new_test_multi_config(&app_node(), vcls).unwrap();
         let persistent_config = PersistentConfigurationMock::new()
-            .earning_wallet_from_address_result(Err(PersistentConfigError::NotPresent));
+            .earning_wallet_result(Err(PersistentConfigError::NotPresent));
 
-        let result = get_earning_wallet_from_address(&multi_config, &persistent_config);
+        let result = get_earning_wallet(&multi_config, &persistent_config);
```
bertllll: Ups, please, read the comment further on regarding this function.

node/src/node_configurator/configurator.rs
```diff
@@ -199,79 +198,24 @@ impl Configurator {
...
             )),
-            Ok(kp) => Ok(Wallet::from(kp)),
+            (Ok(None), Ok(Some(_))) => {
+                panic!("Database corrupted: earning wallet exists but consuming wallet does not")
+            }
+            (Ok(Some(_)), Ok(None)) => {
+                panic!("Database corrupted: consuming wallet exists but earning wallet does not")
+            }
+            (Err(ce), _) => Err((CONFIGURATOR_READ_ERROR, format!("{:?}", ce))),
+            (_, Err(ee)) => Err((CONFIGURATOR_READ_ERROR, format!("{:?}", ee))),
```
bertllll: Don't you want to add a guidance of what wallet the error is attached to within your message?

node/src/node_configurator/configurator.rs
```diff
@@ -1695,6 +1672,114 @@ mod tests {
...
+            db_password: db_password.clone(),
+            seed_spec_opt: Some(UiRecoverSeedSpec {
+                mnemonic_phrase: make_meaningless_phrase_words(),
+                mnemonic_phrase_language: "English".to_string(),
+                mnemonic_passphrase_opt: None,
+            }),
+            consuming_derivation_path_opt: None,
+            consuming_private_key_opt: None,
+            earning_derivation_path_opt: None,
+            earning_address_opt: Some(earning_address.clone()),
```
bertllll: I've noticed that you're sometimes cloning the strings representing the wallet parameters, all despite you may not use the string again if I see well. I can see it in more than one tests around here. But I don't know why Clippy wouldn't tell you...maybe it cannot do that though.

node/src/node_configurator/configurator.rs
```diff
@@ -1695,6 +1672,114 @@ mod tests {
...
+        assert_eq! (result, Err((MISSING_DATA, "If you supply seed information, you must supply either the consuming wallet derivation path or the consuming wallet private key".to_string())));
+    }
+
+    #[test]
+    fn unfriendly_handle_recover_wallets_handles_seed_but_nothing_about_earning_wallet() {
+        let db_password = "password".to_string();
+        let consuming_private_key =
+            "0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF".to_string();
+        let msg = UiRecoverWalletsRequest {
+            db_password: db_password.clone(),
```
bertllll: The same with the password.

node/src/node_configurator/configurator.rs
```diff
@@ -2132,29 +2301,36 @@ mod tests {
...
-            .collect::<Vec<String>>(),
-            mnemonic_passphrase_opt: Some("ebullient".to_string()),
-            mnemonic_phrase_language: "English".to_string(),
-            consuming_derivation_path: derivation_path(0, 4),
-            earning_wallet: "0x005e288d713a5fb3d7c9cf1b43810a98688c7223".to_string(),
+            seed_spec_opt: Some(UiRecoverSeedSpec {
+                mnemonic_phrase: vec![
+                    "parent", "prevent", "vehicle", "tooth", "crazy", "cruel", "update", "mango",
+                    "female", "mad", "spread", "plunge", "tiny", "inch", "under", "engine",
+                    "enforce", "film", "awesome", "plunge", "cloud", "spell", "empower", "pipe",
```
bertllll: If you want you might use the "make_.." instant method to set up the phrase as well as you've done at other places.

node/src/node_configurator/configurator.rs
```diff
@@ -2132,29 +2301,36 @@ mod tests {
...
+                .into_iter()
+                .map(|x| x.to_string())
+                .collect::<Vec<String>>(),
+                mnemonic_passphrase_opt: Some("ebullient".to_string()),
+                mnemonic_phrase_language: "English".to_string(),
+            }),
+            consuming_derivation_path_opt: Some(derivation_path(0, 4)),
+            consuming_private_key_opt: None,
+            earning_derivation_path_opt: Some(derivation_path(0, 5)),
+            // earning_wallet: "0x005e288d713a5fb3d7c9cf1b43810a98688c7223".to_string(),
```
bertllll: Some remains that can be deleted.

node/src/database/db_migrations.rs
dnwiebe: The code needs to make a transaction in order to run. I'm not sure how to do that without a real database.

node/src/database/db_migrations.rs
dnwiebe: So this utility function executes any statement that requests one row with a String, an Option<String>, and a u16?

That's kind of weird.

Maybe I can play with it a little.

node/src/node_configurator/node_configurator_standard.rs
dnwiebe: Good point.

node/src/node_configurator/node_configurator_standard.rs
dnwiebe: I'll check.

USER-INTERFACE-INTERFACE.md
```diff
@@ -640,41 +670,73 @@ If the UI is remembering the database password, it should forget it when this me
...
-by the Node.
+`earningDerivationPathOpt` is the derivation path from the generated seed number to be used to generate the earning
+wallet. By convention, it is "m/44'/60'/0'/0/1", but you can supply whatever path you want. Note that if
+you change any of the numbers ending in ', you may have trouble getting other software and hardware to work with your
+wallet. If you don't supply `earningDerivationPathOpt`, you must supply `earningAddressOpt`.
 
-`earningWallet` is either the derivation path from the mnemonic phrase that was used to generate the earning
-wallet, or--if the derivation path is unknown or the earning wallet is not related to the mnemonic phrase--the
-address of the earning wallet.
+`earningAddressOpt`, if specified, is the address of the consuming wallet, represented as "0x" followed by a string
```
kauri-hero: @dnwiebe I notice there needs to be a correction to the wording here, where it says:
"if specified, is the address of the consuming wallet..."

should be:

if specified, is the address of the earning wallet..."

USER-INTERFACE-INTERFACE.md
```diff
@@ -359,8 +359,9 @@ during creation of the database, it never changes. It's basically a read-only va
 
 * `clandestinePort`: The port on which the Node is currently listening for connections from other Nodes.
 
-* `consumingWalletDerivationPathOpt`: This is the derivation path (from the mnemonic seed) of the consuming wallet.
-  More than likely, it's m/44'/60'/0'/0/0.
+* `consumingWalletPrivateKey`: This is the private key of the consuming wallet, as a 64-digit hexadecimal number.
```
kauri-hero, line 362: It may pay to add text to explain this is a secret value, so won't be displayed without the db-password supplied

@dnwiebe

USER-INTERFACE-INTERFACE.md
```diff
@@ -560,18 +561,31 @@ of payments that have not yet been confirmed. Only cumulative account balances a
...
+    "earningDerivationPathOpt": <optional string>
 }
 ```
 ##### Description:
-This message directs the Node to generate a pair of wallets and report their mnemonic phrase and their addresses
-back to the UI. If the database already contains a wallet pair, the wallet generation will fail.
+This message directs the Node to generate a pair of wallets and report their vital statistic back to the UI. 
+If the database already contains a wallet pair, the wallet generation will fail.
+
+Wallets can be generated in two ways:
```
kauri-hero: @dnwiebe reading through this a few times, it's unclear to me if the user will see the mnemonic seed words output if they run the generateWallets request.

In either of the two generation ways (using derivation paths, or at random), its clear that the consuming and earning wallets and private keys are output to the user.
But does the Node retain the original functionality that it will display (one time only) the seed words generated by the BIP39 algorithm?
Because even if the user writes down the private keys, they should be retaining the seed words as well for security overall, since they were used to derive one/both wallets, depending on the user choices (many users will opt for our default user experience which is simply one wallet for earning and consuming)

USER-INTERFACE-INTERFACE.md
```diff
@@ -560,18 +561,31 @@ of payments that have not yet been confirmed. Only cumulative account balances a
...
+* Entirely at random: in this case no direction should be given in the request, and the response will contain
+  only the address and private key for each wallet generated.
 
-`dbPassword` is the current database password. If this is incorrect, the wallet generation will fail.
+`dbPassword` is the current database password. If this is incorrect or absent, the wallet generation will fail.
+
+`seedSpecOpt` gives the parameters for generating the seed. This only makes sense if one or more of the 
+derivation paths is supplied. If no derivation paths are supplied, this parameter is ignored.
 
 `mnemonicPhraseSize` is the number of words that should be generated in the mnemonic phrase. The acceptable values
```
kauri-hero: I believe we have changed the default phrase size to 12 words, so may need to adjust the wording here.

@FinsaasGH @bertllll I'm pretty confident this was changed - can you confirm?

USER-INTERFACE-INTERFACE.md
```diff
@@ -640,41 +670,73 @@ If the UI is remembering the database password, it should forget it when this me
...
 `dbPassword` is the current database password. If this is incorrect, the wallet recovery will fail.
 
-`mnemonicPhrase` is the mnemonic phrase that was used to generate the consuming wallet and possibly the earning
-wallet as well. It must have 12, 15, 18, 21, or 24 words.
+`seedSpecOpt` gives the parameters for generating the seed. This only makes sense if one or more of the
+derivation paths is supplied. If no derivation paths are supplied, this parameter is ignored.
+
+`mnemonicPhrase` is the mnemonic phrase that represents the seed. It must have 12, 15, 18, 21, or 24 words.
+
+`mnemonicPhraseLanguage` is the language in which the mnemonic phrase is supplied. Acceptable values are
```
kauri-hero: this should not be a required value under seedSpecOpt parameters, (as should be defaulted to English), but its worth making sure this is the case @dnwiebe ?

kauri-hero (commented):
Main question is regarding the seed words being output for user to save them as part of security

General comments on the help/documentation output

USER-INTERFACE-INTERFACE.md
FinsaasGH: This was changed to default to 12 words

dnwiebe: No, the default here is 24 words. The GUI project decided that if their user didn't specify a phrase length, they would specify 12 words here, but the default for Node if you don't specify this value is still 24 words, as it should be. Node defaults to more security; the GUI folks default to less, for user experience.

kauri-hero: that makes sense so thanks for clarifying @dnwiebe

bertllll (approved):

PR #96, opened 2022-01-02
GH-523: pending payment confirmation or cancellation

bertllll:


masq_lib/src/messages.rs
```diff
@@ -518,8 +518,8 @@ pub struct UiPayableAccount {
     pub wallet: String,
     pub age: u64,
     pub amount: u64,
-    #[serde(rename = "pendingTransaction")]
-    pub pending_transaction: Option<String>,
+    #[serde(rename = "pendingTransactionRowid")]
```
dnwiebe: This should have stayed a transaction hash rather than a database row ID. The user can't look up a row in the database, but he can look up a transaction on the blockchain.

masq_lib/src/messages.rs
```diff
@@ -715,7 +715,7 @@ mod tests {
                 wallet: "wallet".to_string(),
                 age: 3456,
                 amount: 4567,
-                pending_transaction: Some("5678".to_string()),
```
dnwiebe, line 740: Same here

masq_lib/src/messages.rs
```diff
@@ -828,7 +828,7 @@ mod tests {
                     "wallet": "wallet",
                     "age": 3456,
                     "amount": 4567,
-                    "pendingTransaction": "transaction"
```
dnwiebe, line 853: Same here

masq_lib/src/messages.rs
```diff
@@ -857,7 +857,7 @@ mod tests {
                         wallet: "wallet".to_string(),
                         age: 3456,
                         amount: 4567,
-                        pending_transaction: Some("transaction".to_string())
```
dnwiebe, line 882: Same here

node/src/accountant/mod.rs
```diff
@@ -97,8 +117,11 @@ pub struct Accountant {
     earning_wallet: Wallet,
     payable_dao: Box<dyn PayableDao>,
     receivable_dao: Box<dyn ReceivableDao>,
+    pending_payments_dao: Box<dyn PendingPaymentsDao>,
```
dnwiebe: I'd like to see the word Payable used in the name of this DAO.

node/src/accountant/mod.rs
```diff
@@ -66,9 +73,22 @@ lazy_static! {
     };
 }
 
-#[derive(PartialEq, Debug, Clone, Copy)]
-pub enum PaymentError {
+pub const DEFAULT_PENDING_TOO_LONG_SEC: u64 = 21_600; //6 hours
+
+#[derive(Debug, PartialEq)]
+pub enum DebtRecordingError {
```
dnwiebe: This nest of error types is confusing. Could you either name them more descriptively or rearrange them so they're clearer?

node/src/accountant/mod.rs
```diff
@@ -141,42 +167,52 @@ impl Handler<StartMessage> for Accountant {
...
 
-impl Handler<SentPayments> for Accountant {
+impl Handler<ScanForPendingPayments> for Accountant {
     type Result = ();
 
-    fn handle(&mut self, msg: SentPayments, _ctx: &mut Self::Context) -> Self::Result {
-        self.handle_sent_payments(msg);
+    fn handle(&mut self, _msg: ScanForPendingPayments, ctx: &mut Self::Context) -> Self::Result {
+        self.scanners.pending_payments.scan(self);
+        self.pending_payments_notify_later_assertable(ctx);
```
dnwiebe: If possible I'd like to see consistency here...either this method everywhere or nowhere.

node/src/accountant/mod.rs
```diff
@@ -255,8 +361,11 @@ impl Accountant {
             earning_wallet: config.earning_wallet.clone(),
             payable_dao: payable_dao_factory.make(),
             receivable_dao: receivable_dao_factory.make(),
+            pending_payments_dao: pending_payments_dao_factory.make(),
             banned_dao: banned_dao_factory.make(),
             crashable: config.crash_point == CrashPoint::Message,
+            scanners: Scanners::default(),
+            transaction_confirmation: TransactionConfirmationTools::default(),
```
dnwiebe: I think I'd like to see this called "transaction_confirmation_tools," or even just "tools."

node/src/accountant/mod.rs
```diff
@@ -387,6 +494,42 @@ impl Accountant {
             .expect("BlockchainBridge is dead");
     }
 
+    fn scan_for_pending_payments(&self) {
+        debug!(self.logger, "Scanning for pending payments");
+        let filtered_pending_payments = self.pending_payments_dao.return_all_payment_backups();
+        if !filtered_pending_payments.is_empty() {
```
dnwiebe: This would look a little better if the clauses were reversed, so that we could do xxx.is_empty() rather than !xxx.is_empty(), but that's a nitpick.

node/src/accountant/mod.rs
```diff
@@ -584,46 +729,54 @@ impl Accountant {
...
-                    ),
-                },
-                Err(e) => warning!(
-                    self.logger,
-                    "{} Please check your blockchain service URL configuration.",
-                    e
-                ),
-            })
+    fn handle_sent_payments(&self, sent_payments: SentPayments) {
+        debug!(
```
dnwiebe: Think about how these logs would look to somebody doing forensic analysis of an error somewhere here, and be gentle with him. He might be you.

node/src/accountant/mod.rs
```diff
@@ -584,46 +729,54 @@ impl Accountant {
...
+    }
+
+    fn discard_incomplete_transaction_with_a_failure(&self, hash: H256) {
+        if let Some(rowid) = self.pending_payments_dao.payment_backup_exists(hash) {
+            debug!(
+                self.logger,
+                "Deleting an existing backup for a failed transaction {}", hash
+            );
+            if let Err(e) = self.pending_payments_dao.delete_payment_backup(rowid) {
+                panic!("Database unmaintainable; payment backup deletion has stayed undone due to {:?}",e)
```
dnwiebe: Can you put something in here about which payment is wrong? If we know, we might be able to manually fix the database.

node/src/accountant/mod.rs
```diff
@@ -710,9 +863,7 @@ impl Accountant {
                     .expect("Bad interval")
                     .as_secs(),
                 amount: account.balance as u64,
-                pending_transaction: account
```
dnwiebe, line 713: This will change back, I hope....

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+    fn handle_confirm_pending_transaction(&self, msg: ConfirmPendingTransaction) {
+        if let Err(e) = self.payable_dao.transaction_confirmed(&msg.payment_backup) {
+            panic!(
+                "Was unable to uncheck pending payment '{}' after confirmation due to '{:?}'",
+                msg.payment_backup.hash, e.0
+            )
+        } else {
+            debug!(
+                self.logger,
+                "Confirmation of transaction {}; record for payable table took change",
```
dnwiebe: "was modified?"

dnwiebe (commented):
Continue with separate_early_errors() in accountant/mod.rs

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            .payments
+            .into_iter()
+            .partition(|payment| payment.is_ok());
+        (
+            ok.into_iter().flatten().collect(),
+            err.into_iter().map(|item|{
+                let error = item.expect_err("partition failed");
+                logger.warning(||
+                        match &error {
+                            BlockchainError::TransactionFailed { msg: _, hash_opt: _ } => format!("Encountered transaction error that occurred close to the actual sending due to '{:?}'", error),
```
dnwiebe: Instead of "close to the actual sending," which a user isn't really going to understand, something like "after successful connection to the blockchain client" or something similar?

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            }
+        }
+    }
+
+    fn separate_early_errors(
+        sent_payments: SentPayments,
+        logger: &Logger,
+    ) -> (Vec<Payment>, Vec<BlockchainError>) {
+        type SentPaymentType = Vec<Result<Payment, BlockchainError>>;
+        let (ok, err): (SentPaymentType, SentPaymentType) = sent_payments
```
dnwiebe: It might work better to use fold() here. I'm pretty sure partition() is implemented using fold() anyway, but you could put a match{} inside the fold() and end up with one collection of naked successes and another collection of naked errors.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            logger: &Logger,
+        ) -> PendingTransactionStatus {
+            info!(logger,"Pending transaction '{}' couldn't be confirmed at attempt {} at {}ms after its sending",payment.hash, payment.attempt, elapsed_in_ms(payment.timestamp));
+            let elapsed = payment.timestamp.elapsed().expect("we should be older now");
+            let transaction_id = TransactionId {
+                hash: payment.hash,
+                rowid: payment.rowid,
+            };
+            if pending_interval <= elapsed.as_secs() {
+                warning!(logger,"Pending transaction '{}' has exceeded the maximum time allowed ({}sec) \
```
dnwiebe: Should be an error, rather than a warning, because this will require user intervention.

Also: "has exceeded the maximum time allowed for being pending" would sound better as "has exceeded the maximum pending time."

And, "manual resolving" should be "manual resolution."

Instead of "to make the transaction paid," try "to complete the transaction."

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            pending_interval: u64,
+            logger: &Logger,
+        ) -> PendingTransactionStatus {
+            info!(logger,"Pending transaction '{}' couldn't be confirmed at attempt {} at {}ms after its sending",payment.hash, payment.attempt, elapsed_in_ms(payment.timestamp));
+            let elapsed = payment.timestamp.elapsed().expect("we should be older now");
+            let transaction_id = TransactionId {
+                hash: payment.hash,
+                rowid: payment.rowid,
+            };
+            if pending_interval <= elapsed.as_secs() {
```
dnwiebe, line 1025: This works, but it would be easier to read if it was comparing the real-world value to the expected standard, rather than the other way around: `if elapsed.as_secs() > pending_interval`

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+                payment.attempt,
+                elapsed_in_ms(payment.timestamp)
+            );
+            PendingTransactionStatus::Confirmed(payment)
+        }
+        fn handle_status_with_failure(
+            payment: &PaymentBackupRecord,
+            logger: &Logger,
+        ) -> PendingTransactionStatus {
+            warning!(logger,"Pending transaction '{}' announced as a failure, checking out attempt {} after {}ms from its sending",payment.hash,payment.attempt,elapsed_in_ms(payment.timestamp));
```
dnwiebe: Also should be an error, because user intervention is required.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+                    None => {
+                        debug!(self.logger,"DEBUG: Accountant: Interpreting a receipt for transaction '{}' but none was given; attempt {}, {}ms since sending", payment.hash, payment.attempt,elapsed_in_ms(payment.timestamp));
+                        PendingTransactionStatus::StillPending(TransactionId{ hash: payment.hash, rowid: payment.rowid })
+                    }
+                }
+            )
+            .collect()
+    }
+
+    fn check_out_transaction_receipt(
```
dnwiebe: "Interpret" would be better than "check out" in this name. Or maybe "process," or "handle," or...something.

node/src/accountant/mod.rs
dnwiebe: Change this "checking out" to whatever you use for the method name.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+        cancellations: Vec<PendingTransactionStatus>,
+        ctx: &mut Context<Self>,
+    ) {
+        cancellations.into_iter().for_each(|status| {
+            if let PendingTransactionStatus::Failure(transaction_id) = status {
+                self.cancel_failed_transaction(transaction_id, ctx)
+            } else if let PendingTransactionStatus::Confirmed(payment) = status {
+                self.confirm_transaction(payment, ctx)
+            }
+        });
```
dnwiebe, line 1094: Do we need a final else{} here to explicitly blow up if we get a PendingTransactionStatus we don't expect?

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+                    .update_backup_after_scan_cycle(id.rowid)
+                {
+                    Ok(_) => trace!(self.logger, "Updated backup for rowid: {} ", id.rowid),
+                    Err(e) => panic!("Failure on updating payment backup due to {:?}", e),
+                },
+                _ => unreachable!("we are operating behind a filter; should not happen"),
+            })
+    }
+
+    fn handle_transaction_cancellation(
```
dnwiebe: Maybe "finalize_transaction?" Or  "handle_transaction_termination?" Just not "cancellation," because that's only one of the two or three possibilities you're handling here.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            );
+    }
+
+    fn handle_payment_backup(&self, msg: PaymentBackupRecord) {
+        match self
+            .pending_payments_dao
+            .insert_payment_backup(msg.hash, msg.amount, msg.timestamp)
+        {
+            Ok(_) => debug!(self.logger, "Processed a backup for payment '{}'", msg.hash),
+            Err(e) => warning!(
```
dnwiebe: Does this require user intervention? If so, it should be an error. Does it make further operation impossible? If so, it should be a panic.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+            ),
+        }
     }
 }
 
 // At the time of this writing, Rust 1.44.0 was unpredictably producing
 // segfaults on the Mac when using u64::try_from (i64). This is an attempt to
 // work around that.
-pub fn jackass_unsigned_to_signed(unsigned: u64) -> Result<i64, PaymentError> {
+pub fn jackass_unsigned_to_signed(unsigned: u64) -> Result<i64, u64> {
```
dnwiebe: Sounds like this can be removed now.

node/src/accountant/mod.rs
```diff
@@ -750,45 +901,323 @@ impl Accountant {
...
+    fn from(payment_backup: &PaymentBackupRecord) -> Self {
+        Self {
+            hash: payment_backup.hash,
+            rowid: payment_backup.rowid,
+        }
     }
 }
 
 #[cfg(test)]
 pub mod tests {
```
dnwiebe: This "pub" is a little ominous. Are tests from other files calling stuff in here? If they are, that stuff should be moved out to somewhere common. (Not your fault, but still weird.)

node/src/accountant/mod.rs
```diff
@@ -1126,144 +1577,21 @@ pub mod tests {
...
         let payable_dao = PayableDaoMock::new();
         let payable_dao_factory =
-            PayableDaoFactoryMock::new(payable_dao).called(&payable_dao_factory_called);
+            PayableDaoFactoryMock::new(Box::new(payable_dao)).called(&payable_dao_factory_called);
         let receivable_dao_factory_called = Rc::new(RefCell::new(false));
         let receivable_dao = ReceivableDaoMock::new();
         let receivable_dao_factory =
             ReceivableDaoFactoryMock::new(receivable_dao).called(&receivable_dao_factory_called);
+        let payment_recover_dao_factory_called = Rc::new(RefCell::new(false));
+        let payment_recover_dao = PendingPaymentsDaoMock::default();
```
dnwiebe: The names here don't match.

node/src/accountant/mod.rs
```diff
@@ -1406,34 +1739,42 @@ pub mod tests {
...
-        let expected_amount = 1;
-        let expected_hash = H256::from("transaction_hash".keccak256());
-        let mut expected_payment = Payment::new(
+        let expected_payment = Payment::new(
             expected_wallet.clone(),
             expected_amount,
             expected_hash.clone(),
+            expected_timestamp,
         );
         let send_payments = SentPayments {
```
dnwiebe: Should be "sent_payments"

node/src/accountant/mod.rs
```diff
@@ -1483,9 +1829,87 @@ pub mod tests {
 
         System::current().stop();
         system.run();
-        TestLogHandler::new().await_log_containing(
-            r#"WARN: Accountant: Blockchain TransactionFailed("Payment attempt failed"). Please check your blockchain service URL configuration."#,
-            1000,
+        let log_handler = TestLogHandler::new();
+        log_handler.exists_no_log_containing(
+            "DEBUG: Accountant: Deleting an existing backup for a failed transaction",
```
dnwiebe, line 1465: If you modify this log to contain some identifier for the failed transaction, then A) somebody reading the log will be able to figure out what it's talking about, and B) you can design this test so that it won't fail because some other test logs this message.

node/src/accountant/mod.rs
```diff
@@ -1483,9 +1829,87 @@ pub mod tests {
...
+        );
+        log_handler.exists_log_containing("WARN: Accountant: Encountered transaction error that occurred close to \
+         the actual sending due to 'TransactionFailed { msg: \"SQLite migraine\", hash_opt: Some(0x0000000000000000000000000000000000000000000000000000000000003039) }'");
+        log_handler.exists_log_containing(
+            r#"WARN: Accountant: Failed transaction with a hash '0x0000…3039' but without the backup - thrown out"#,
+        );
+    }
+
+    #[test]
+    fn handle_sent_payments_discover_failed_transaction_and_payment_backup_was_really_created() {
```
dnwiebe: should be "discovers"

node/src/accountant/mod.rs
```diff
@@ -1483,9 +1829,87 @@ pub mod tests {
...
+        let subject = AccountantBuilder::default()
+            .payable_dao_factory(Box::new(PayableDaoFactoryMock::new(Box::new(payable_dao))))
+            .pending_payments_dao_factory(Box::new(PendingPaymentsDaoFactoryMock::new(
+                pending_payments_dao,
+            )))
+            .build();
+        let wallet = make_wallet("blah");
+        let hash_tx_1 = H256::from_uint(&U256::from(5555));
+        let hash_tx_2 = H256::from_uint(&U256::from(12345));
+
```
dnwiebe: Blank line could be removed

node/src/accountant/mod.rs
```diff
@@ -1599,6 +2036,148 @@ pub mod tests {
...
+                earning_wallet.clone(),
+            )),
+            Some(PayableDaoMock::new().non_pending_payables_result(vec![])),
+            Some(receivable_dao),
+            None,
+            None,
+            None,
+        );
+
+        let system = System::new("accountant_receives_new_payments_to_the_receivables_dao");
```
dnwiebe, line 1693: These two lines look like the Act portion of the test, but aren't.

node/src/accountant/mod.rs
```diff
@@ -1599,6 +2036,148 @@ pub mod tests {
...
+            .try_send(ReceivedPayments {
+                payments: vec![expected_payment.clone(), expected_payment.clone()],
+            })
+            .expect("unexpected actix error");
+        System::current().stop();
+        system.run();
+        let more_money_received_params = more_money_received_params_arc.lock().unwrap();
+        assert_eq!(1, more_money_received_params.len());
+
+        let more_money_received_params = more_money_received_params.get(0).unwrap();
```
dnwiebe: These two as well...

dnwiebe: This more_money_received_params has a different type than the more_money_received_params it's replacing, which makes reading the code difficult.

node/src/accountant/mod.rs
```diff
@@ -1599,6 +2036,148 @@ pub mod tests {
...
+        system.run();
+        let tlh = TestLogHandler::new();
+        tlh.await_log_containing("DEBUG: Accountant: Scanning for payables", 1000u64);
+        tlh.exists_log_containing(&format!(
+            "DEBUG: Accountant: Scanning for payments to {}",
+            make_wallet("hi")
+        ));
+        tlh.exists_log_containing("DEBUG: Accountant: Scanning for delinquencies");
+        tlh.exists_log_containing("DEBUG: Accountant: Scanning for pending payments");
+        //some more week proofs but still good enough
```
dnwiebe: "Weak," not "week"

node/src/accountant/mod.rs
```diff
@@ -1684,80 +2269,96 @@ pub mod tests {
...
+            &RequestTransactionReceipts {
+                pending_payments: vec![payment_backup_record],
+            }
+        );
     }
 
     #[test]
     fn accountant_payable_scan_timer_triggers_periodical_scanning_for_payables() {
-        //in the very first round we scan without waiting but we cannot find any payables
+        //in the very first round we scan without waiting but we cannot find any payable records
```
dnwiebe, line 1983: You said you didn't like this comment...so here's a pointer to it.

node/src/accountant/mod.rs
```diff
@@ -2025,13 +2599,102 @@ pub mod tests {
...
+                when_pending_too_long_sec: DEFAULT_PENDING_TOO_LONG_SEC,
+            },
+            make_wallet("mine"),
+        );
+        let system = System::new("pending payments scan");
+        let mut subject = AccountantBuilder::default()
+            .pending_payments_dao_factory(Box::new(pending_payments_dao_factory))
+            .bootstrapper_config(config)
+            .build();
+        let blockachain_bridge_addr = blockchain_bridge.start();
```
dnwiebe: Misspelled

node/src/accountant/mod.rs
```diff
@@ -2608,25 +3299,50 @@ pub mod tests {
         ));
     }
 
+    #[test]
+    #[should_panic(
+        expected = "Recording services provided for 0x000000000000000000000000000000626f6f6761 \
+    but has hit fatal database error: RusqliteError(\"we cannot help ourself; this is baaad\")"
```
dnwiebe: "Ourself" isn't a word, because "our" is plural and "self" is singular. (At least, it _shouldn't_ be a word.)

dnwiebe (commented):
Continue at record_service_consumed_panics_on_fatal_errors().

node/src/accountant/mod.rs
```diff
@@ -2636,55 +3352,324 @@ pub mod tests {
...
+                    "we cannot help ourself; this is baaad".to_string(),
+                ),
+            ))),
+            None,
+            None,
+            None,
+            None,
+        );
+
+        let _ = subject.record_service_consumed(i64::MAX as u64, 1, 2, &wallet);
```
dnwiebe, line 3043: You said you didn't like this constant.

node/src/accountant/mod.rs
```diff
@@ -2636,55 +3352,324 @@ pub mod tests {
...
+            None,
+            None,
+        );
+
+        let _ = subject.mark_pending_payments(vec![payment]);
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Database unmaintainable; payment backup deletion has stayed undone due to RecordDeletion(\"we slept over, sorry\")"
```
dnwiebe: Should there be parameters in here so that we know which payment we're talking about?

node/src/accountant/mod.rs
```diff
@@ -2636,55 +3352,324 @@ pub mod tests {
...
+        let payment_2 = Payment::new(make_wallet("booga"), 6789, payment_hash_2, now_system);
+        let payment_3 = Err(BlockchainError::TransactionFailed {
+            msg: "closing hours, sorry".to_string(),
+            hash_opt: None,
+        });
+        let payments = SentPayments {
+            payments: vec![payment_1, Ok(payment_2.clone()), payment_3],
+        };
+        let pending_payments_dao = PendingPaymentsDaoMock::default()
+            //one seems to be missing, not really, the execution short-circuited at the outer error
```
dnwiebe: You wanted to delete this comment?

node/src/accountant/mod.rs
```diff
@@ -2769,6 +3754,724 @@ pub mod tests {
...
+        };
+        let payment_2_backup_first_round = PaymentBackupRecord {
+            rowid: rowid_for_account_2,
+            timestamp: payment_timestamp_2,
+            hash: pending_tx_hash_2,
+            attempt: 1,
+            amount: payable_account_balance_2 as u64,
+            process_error: None,
+        };
+        let mut payment_1_backup_second_round = payment_1_backup_first_round.clone();
```
dnwiebe: Memory is hazy, but I may remember a Rust way to do this more concisely. Maybe involving two or three dots in a row?

dnwiebe (commented):
Continue at accountant_handles_payment_backup()

node/src/accountant/mod.rs
```diff
@@ -2769,6 +3754,724 @@ pub mod tests {
...
+        let system = System::new("ordering payment backup test");
+        System::current().stop();
+        assert_eq!(system.run(), 0);
+        let insert_payment_backup_params = insert_payment_backup_params_arc.lock().unwrap();
+        assert_eq!(
+            *insert_payment_backup_params,
+            vec![(tx_hash, amount, timestamp)]
+        );
+        TestLogHandler::new().exists_log_containing(
+            "DEBUG: Accountant: Processed a backup for payment '0x0000…0037'",
```
dnwiebe: If we're going to be searching the blockchain for this wallet, it'd be nice to have the whole address in the log, rather than the abbreviated version.

node/src/accountant/mod.rs
```diff
@@ -2769,6 +3754,724 @@ pub mod tests {
...
+
+        let _ = subject.update_backup_of_pending_transaction(pending_payments);
+
+        let update_after_cycle_params = update_after_cycle_params_arc.lock().unwrap();
+        assert_eq!(*update_after_cycle_params, vec![rowid_1, rowid_2])
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Failure on updating payment backup due to UpdateFailed(\"yeah, bad\")"
```
dnwiebe: Again, it would be nice to see some numbers here for forensic investigation.

node/src/accountant/payable_dao.rs
```diff
@@ -28,29 +32,30 @@ pub struct Payment {
...
         wallet: &Wallet,
-        amount: u64,
-        confirmation_noticed_timestamp: SystemTime,
-        transaction_hash: H256,
+        transaction_id: TransactionId,
     ) -> Result<(), PaymentError>;
 
+    fn transaction_confirmed(&self, payment: &PaymentBackupRecord) -> Result<(), PaymentError>;
+
+    //TODO this method doesn't have a use now; what is its future?
```
dnwiebe: Maybe remove this comment? (Or at least investigate it?)

node/src/accountant/payable_dao.rs
```diff
@@ -76,66 +81,93 @@ pub struct PayableDaoReal {
...
+            .expect("Internal Error");
+        let params: &[&dyn ToSql] = &[
+            &i64::try_from(transaction_id.rowid)
+                .expect("SQLite counts up to i64::MAX; should never happen"),
+            wallet,
+        ];
+        match stm.execute(params) {
+            Ok(1) => Ok(()),
+            Ok(num) => Err(PaymentError(
+                PaymentErrorKind::RusqliteError(format!("Returned {} rows but expected 1", num)),
```
dnwiebe: Should be "Affected," not "Returned."

node/src/accountant/payable_dao.rs
```diff
@@ -76,66 +81,93 @@ pub struct PayableDaoReal {
...
+                PaymentErrorKind::RusqliteError(e.to_string()),
+                transaction_id,
+            )),
+        }
     }
 
     fn account_status(&self, wallet: &Wallet) -> Option<PayableAccount> {
         let mut stmt = self.conn
-            .prepare("select balance, last_paid_timestamp, pending_payment_transaction from payable where wallet_address = ?")
+            .prepare("select rowid, balance, last_paid_timestamp, pending_payment_rowid from payable where wallet_address = ?")
```
dnwiebe: Looks like row_id is unnecessary here.

dnwiebe: select
    payable.balance,
    payable.last_paid_timestamp,
    pending_payable.pending_payment_transaction
from 
    payable left outer join pending_payable 
        on 
            payable.pending_payment_rowid = pending_payable.rowid;

node/src/accountant/payable_dao.rs
```diff
@@ -182,7 +214,7 @@ impl PayableDao for PayableDaoReal {
                     balance,
                     last_paid_timestamp,
                     wallet_address,
-                    pending_payment_transaction
+                    pending_payment_rowid
                 from
                     payable
```
dnwiebe, line 177: You'll probably need another left outer join here

dnwiebe (commented):
Continue at payable_dao.rs: try_decrease_balance()

node/src/accountant/payable_dao.rs
```diff
@@ -407,111 +464,349 @@ mod tests {
...
+            Err(PaymentError(
+                PaymentErrorKind::RusqliteError("attempt to write a readonly database".to_string()),
+                TransactionId { hash, rowid }
+            ))
         )
     }
 
     #[test]
-    fn payment_sent_works_for_overflow() {
+    fn transaction_canceled_returned_different_row_count_than_expected() {
```
dnwiebe: Modify this test to demand a panic if the database is corrupt in this fashion.

node/src/accountant/payable_dao.rs
```diff
@@ -407,111 +464,349 @@ mod tests {
...
-        assert!(before_account_status.is_none());
+        let after_account_status = subject.account_status(&wallet).unwrap();
+        let mut after_expected_status = before_expected_status;
+        after_expected_status.pending_payment_rowid_opt = Some(pending_payments_rowid);
+        assert_eq!(after_account_status, after_expected_status)
+    }
 
-        subject.payment_sent(&payment).unwrap();
+    #[test]
+    fn mark_pending_payment_returned_different_row_count_than_expected() {
```
dnwiebe, line 480: Modify this test to demand a panic if the database is corrupt in this fashion.

node/src/accountant/payable_dao.rs
```diff
@@ -307,17 +341,40 @@ impl PayableDaoReal {
...
     use ethereum_types::BigEndianHash;
     use masq_lib::test_utils::utils::ensure_node_home_directory_exists;
+    use rusqlite::Connection as RusqliteConnection;
     use rusqlite::{Connection, OpenFlags};
-    use std::str::FromStr;
+    use std::path::Path;
     use web3::types::U256;
 
+    #[test]
+    fn try_decrease_balance_changed_no_rows() {
```
dnwiebe, line 329: Modify this test to demand a panic if the database is corrupt in this fashion.

node/src/accountant/payable_dao.rs
```diff
@@ -407,111 +464,349 @@ mod tests {
...
+            result,
+            Err(PaymentError(
+                PaymentErrorKind::RusqliteError("attempt to write a readonly database".to_string()),
+                TransactionId { hash, rowid }
+            ))
+        )
+    }
+
+    #[test]
+    fn transaction_confirmed_works_for_overflow() {
```
dnwiebe: I don't see the creation of a PAYABLE account here. Maybe before you create it, figure out why this tests passes like this: it doesn't seem as though it should.

node/src/accountant/payable_dao.rs
```diff
@@ -598,13 +885,13 @@ mod tests {
                     wallet: make_wallet("foobar"),
```
dnwiebe, line 728: This wallet value and the one in the insert() call above should look much more alike, so that a reader can easily connect them. Either use the long form both places, the make_wallet() call both places, or assign the wallet address to a local variable which is used in both places.

node/src/accountant/payable_dao.rs
```diff
@@ -624,28 +911,40 @@ mod tests {
...
-        assert_eq!(result, Err(PaymentError::SignConversion(u64::MAX)))
+        assert_eq!(result, Err(DebtRecordingError::SignConversion(u64::MAX)))
     }
 
     #[test]
-    fn payable_amount_errors_on_update_balance_when_out_of_range() {
+    fn payable_amount_error_on_transaction_confirmed_updating_balance_out_of_range() {
         let home_dir = ensure_node_home_directory_exists(
             "payable_dao",
             "payable_amount_precision_loss_panics_on_update_balance",
```
dnwiebe: Change this to be the name of the test so that we can find it if we need to

node/src/accountant/payable_dao.rs
```diff
@@ -624,28 +911,40 @@ mod tests {
 
         let result = subject.more_money_payable(&make_wallet("foobar"), u64::MAX);
 
-        assert_eq!(result, Err(PaymentError::SignConversion(u64::MAX)))
+        assert_eq!(result, Err(DebtRecordingError::SignConversion(u64::MAX)))
     }
 
     #[test]
-    fn payable_amount_errors_on_update_balance_when_out_of_range() {
+    fn payable_amount_error_on_transaction_confirmed_updating_balance_out_of_range() {
```
dnwiebe: It looks like this test duplicates transaction_confirmed_works_for_overflow().

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+    fn payment_backup_exists(&self, transaction_hash: H256) -> Option<u64>;
+    fn return_all_payment_backups(&self) -> Vec<PaymentBackupRecord>;
+    fn insert_payment_backup(
+        &self,
+        transaction_hash: H256,
+        amount: u64,
+        timestamp: SystemTime,
+    ) -> Result<(), PendingPaymentDaoError>;
+    fn delete_payment_backup(&self, id: u64) -> Result<(), PendingPaymentDaoError>;
+    fn update_backup_after_scan_cycle(&self, id: u64) -> Result<(), PendingPaymentDaoError>;
```
dnwiebe: Naming: Scan cycles are things that occur at a higher level than DAOs; I think DAOs should not know about or pretend to know about scan cycles. This is just a naming thing, not a functionality thing, but I'd like to see no reference to scan cycles anywhere in the DAOs.

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+    fn mark_failure(&self, id: u64) -> Result<(), PendingPaymentDaoError>;
+}
+
+impl PendingPaymentsDao for PendingPaymentsDaoReal<'_> {
+    fn payment_backup_exists(&self, transaction_hash: H256) -> Option<u64> {
```
dnwiebe: This name implies that the return value will be a boolean. If you're returning the rowid, the name should be different.

If you need a method that returns a boolean (which could justifiably be called payment_backup_exists()), you could write it to simply return \[this method\]().is_some().

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+            Ok(PaymentBackupRecord {
+                rowid: u64::try_from(rowid).expectv("positive value"),
+                timestamp: from_time_t(timestamp),
+                hash: H256::from_str(transaction_hash.as_str()).expectv("string hash"),
+                attempt: u16::try_from(attempt).expectv("positive and low value"),
+                amount: u64::try_from(amount).expectv("positive value"),
+                process_error: None,
+            })
+        })
+        .expect("behaves quite infallible")
```
dnwiebe: Maybe a different message?

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+    fn return_all_payment_backups(&self) -> Vec<PaymentBackupRecord> {
+        let mut stm = self.conn.prepare("select rowid, transaction_hash, amount, payment_timestamp, attempt from pending_payments where process_error is null").expect("Internal error");
+        stm.query_map([], |row| {
+            let rowid: i64 = Self::get_with_expect(row, 0);
+            let transaction_hash: String = Self::get_with_expect(row, 1);
+            let amount: i64 = Self::get_with_expect(row, 2);
+            let timestamp: i64 = Self::get_with_expect(row, 3);
+            let attempt: i64 = Self::get_with_expect(row, 4);
+            Ok(PaymentBackupRecord {
+                rowid: u64::try_from(rowid).expectv("positive value"),
```
dnwiebe: Check to see if rusqlite can natively convert to u64 (etc.)

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+                timestamp: from_time_t(timestamp),
+                hash: H256::from_str(transaction_hash.as_str()).expectv("string hash"),
+                attempt: u16::try_from(attempt).expectv("positive and low value"),
+                amount: u64::try_from(amount).expectv("positive value"),
+                process_error: None,
+            })
+        })
+        .expect("behaves quite infallible")
+        .flatten()
```
dnwiebe: Make sure this flatten() isn't hiding something.

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+        transaction_hash: H256,
+        amount: u64,
+        timestamp: SystemTime,
+    ) -> Result<(), PendingPaymentDaoError> {
+        let signed_amount = jackass_unsigned_to_signed(amount)
+            .map_err(PendingPaymentDaoError::SignConversionError)?;
+        let mut stm = self.conn.prepare("insert into pending_payments (rowid, transaction_hash, amount, payment_timestamp, attempt, process_error) values (?,?,?,?,?,?)").expect("Internal error");
+        let params: &[&dyn ToSql] = &[
+            &Null, //to let it increment automatically by SQLite
+            &format!("{:x}", transaction_hash),
```
dnwiebe: :? would work as well as :x

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+        &self,
+        transaction_hash: H256,
+        amount: u64,
+        timestamp: SystemTime,
+    ) -> Result<(), PendingPaymentDaoError> {
+        let signed_amount = jackass_unsigned_to_signed(amount)
+            .map_err(PendingPaymentDaoError::SignConversionError)?;
+        let mut stm = self.conn.prepare("insert into pending_payments (rowid, transaction_hash, amount, payment_timestamp, attempt, process_error) values (?,?,?,?,?,?)").expect("Internal error");
+        let params: &[&dyn ToSql] = &[
+            &Null, //to let it increment automatically by SQLite
```
dnwiebe: I'm pretty sure you don't need to mention rowid in this insert.

node/src/accountant/pending_payments_dao.rs
```diff
@@ -0,0 +1,630 @@
...
+        let hash = H256::from_uint(&U256::from(45466));
+        let amount = 55556;
+        let timestamp = from_time_t(200_000_000);
+        let subject = PendingPaymentsDaoReal::new(wrapped_conn);
+
+        let _ = subject
+            .insert_payment_backup(hash, amount, timestamp)
+            .unwrap();
+
+        let assertion_conn = Connection::open(home_dir.join(DATABASE_FILE)).unwrap();
```
dnwiebe: You might be able to use another production method here to retrieve the result, rather than doing it by hand...

node/src/accountant/test_utils.rs
```diff
@@ -21,10 +41,491 @@ pub fn make_receivable_account(n: u64, expected_delinquent: bool) -> ReceivableA
...
+            pending_payments_dao_factory: Box::new(PendingPaymentsDaoFactoryMock::new(
+                PendingPaymentsDaoMock::default(),
+            )),
+            banned_dao_factory: Box::new(BannedDaoFactoryMock::new(BannedDaoMock::new())),
+            config_dao_factory: Box::new(ConfigDaoFactoryMock::new(ConfigDaoMock::new())),
+        }
+    }
+}
+
+impl AccountantBuilder {
```
dnwiebe, line 95: Suggestion: Add methods that accept naked DAOs, rather than factories. Have those methods create factory mocks that return the supplied DAOs when instructed to make().

If these new methods make the old methods that accept factories obsolete, remove the old methods.

node/src/accountant/test_utils.rs
```diff
@@ -21,10 +41,491 @@ pub fn make_receivable_account(n: u64, expected_delinquent: bool) -> ReceivableA
...
-        last_paid_timestamp: from_time_t(now - (n as i64)),
-        pending_payment_transaction: None,
+        wallet: recipient,
+        balance,
+        last_paid_timestamp: timestamp_opt.unwrap_or(SystemTime::now()),
+        pending_payment_rowid_opt: None,
+    }
+}
+
+//TODO should I replace this with the builder?
```
dnwiebe: If you make the modifications suggested below, probably.

node/src/actor_system_factory.rs
```diff
@@ -415,6 +415,7 @@ impl ActorFactory for ActorFactoryReal {
         let cloned_config = config.clone();
         let payable_dao_factory = Accountant::dao_factory(data_directory);
         let receivable_dao_factory = Accountant::dao_factory(data_directory);
+        let payment_recover_dao_factory = Accountant::dao_factory(data_directory);
```
dnwiebe: Rename variable

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -51,13 +63,16 @@ impl Handler<BindMessage> for BlockchainBridge {
...
 
     fn handle(&mut self, msg: BindMessage, _ctx: &mut Self::Context) -> Self::Result {
-        self.set_consuming_wallet_subs_opt = Some(vec![
-            msg.peer_actors
-                .neighborhood
-                .set_consuming_wallet_sub
-                .clone(),
-            msg.peer_actors.proxy_server.set_consuming_wallet_sub,
-        ]);
+        //TODO is this dead code?
```
dnwiebe: I don't think it's precisely _dead_ code, but it's currently-unused code. I'd say comment it back in; we made a card to use it.

dnwiebe (commented):
Resume at blockchain_bridge.rs:process_payments()

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -383,20 +472,24 @@ mod tests {
                 wallet: make_wallet("blah"),
                 balance: 42,
                 last_paid_timestamp: SystemTime::now(),
-                pending_payment_transaction: None,
+                pending_payment_rowid_opt: Some(789),
             }],
         };
+        let (accountat, _, _) = make_recorder();
```
dnwiebe: accountant

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -570,6 +651,60 @@ mod tests {
...
+        assert_eq!(
+            received_message,
+            &ReportTransactionReceipts {
+                payment_backups_with_receipts: vec![
+                    (Some(TransactionReceipt::default()), payment_backup_1),
+                    (None, payment_backup_2),
+                ]
+            }
+        );
+        let get_transaction_params = get_transaction_receipt_params_arc.lock().unwrap();
```
dnwiebe: Naming

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -595,6 +730,69 @@ mod tests {
...
+                payment_backup_2.clone(),
+                payment_backup_3.clone(),
+            ],
+        };
+
+        let _ = subject.handle_request_transaction_receipts(msg);
+
+        let get_transaction_receipts_params = get_transaction_receipt_params_arc.lock().unwrap();
+        assert_eq!(*get_transaction_receipts_params, vec![hash_1, hash_2]);
+        let accountant_recording = accountant_recording_arc.lock().unwrap();
```
dnwiebe, line 835: Shouldn't we be sending the good receipts to the Accountant? (Maybe not: I dunno what I'm talking about.)

node/src/blockchain/blockchain_interface.rs
```diff
@@ -100,7 +107,12 @@ pub trait BlockchainInterface {
...
     fn get_transaction_count(&self, address: &Wallet) -> Nonce;
 
-    fn send_transaction_tools<'a>(&'a self) -> Box<dyn SendTransactionToolsWrapper + 'a> {
+    fn get_transaction_receipt(&self, hash: H256) -> Receipt;
+
+    fn send_transaction_tools<'a>(
+        &'a self,
+        _backup_recipient: &'a dyn PaymentBackupRecipientWrapper,
+    ) -> Box<dyn SendTransactionToolWrapper + 'a> {
         intentionally_blank!()
```
dnwiebe: Should this really be default-implemented, or should the implementation be left out here, and the intentionally_blank!() go in the implementations that don't use this method?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -403,25 +441,125 @@ where
...
+    }
+}
+
+impl Display for BlockchainTransactionError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::UnusableWallet(msg) => write!(f, "UnusableWallet: {}", msg),
+            Self::Signing(msg) => write!(f, "Signing: {}", msg),
+            Self::Sending(msg, _) => write!(f, "Sending: {}", msg),
+            Self::SentinelVariant => write!(f, "This is sentinel for BlockchainTransactionError"),
```
dnwiebe: Investigate SentinelVariant and make sure it's still necessary.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -904,31 +986,83 @@ mod tests {
...
         transport.assert_no_more_requests();
-        assert_eq!(result, Ok(H256::from_uint(&U256::from(1))));
+        let (hash, timestamp) = result;
+        assert_eq!(
+            hash,
+            H256::from_str("e26f2f487f5dd06c38860d410cdcede0d6e860dab2c971c7d518928c17034c8f")
+                .unwrap()
+        );
+        let comparable_now = to_time_t(SystemTime::now());
+        let result_as_better_comparable_in_secs = to_time_t(timestamp);
```
dnwiebe: The standard way of asserting this kind of thing is to take a timestamp before the Act and another one after the Act, and then make sure the timestamp produced within the Act is not less than the before timestamp, and not greater than the after timestamp.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -904,31 +986,83 @@ mod tests {
...
-    //test with real web3
     #[test]
     fn blockchain_interface_non_clandestine_can_transfer_tokens() {
+        init_test_logging();
         let mut transport = TestTransport::default();
         transport.add_response(json!(
-            "0x0000000000000000000000000000000000000000000000000000000000000001"
+            "0xe26f2f487f5dd06c38860d410cdcede0d6e860dab2c971c7d518928c17034c8f"
         ));
+        let (recorder, _, recording_arc) = make_recorder();
```
dnwiebe: Could you name `recorder` and `recording_arc` so that it's clear what they're recording? For example, you could call them `accountant` and `accountant_recording_arc`, unless you're philosophically opposed to the BlockchainInterface knowing about the Accountant, which is reasonable; but still, something more intention-revealing than "recorder" would help the readability.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1059,27 +1204,24 @@ mod tests {
...
-    fn assert_gas_limit_is_not_under<T: Transport + Debug>(
+    fn assert_gas_limit_is_not_under<T: Transport + Debug + 'static>(
         subject: BlockchainInterfaceNonClandestine<T>,
         not_under_this_value: u64,
         not_above_this_value: u64,
     ) {
         let sign_transaction_params_arc = Arc::new(Mutex::new(vec![]));
         let consuming_wallet_secret_raw_bytes = b"my-wallet";
-        let send_transaction_tools = &SendTransactionToolsWrapperMock::default()
+        let send_transaction_tools = &SendTransactionToolWrapperMock::default()
```
dnwiebe: I think this name shouldn't have changed.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1059,27 +1204,24 @@ mod tests {
         assert_gas_limit_is_not_under(subject, 55000, 65000)
     }
 
-    fn assert_gas_limit_is_not_under<T: Transport + Debug>(
+    fn assert_gas_limit_is_not_under<T: Transport + Debug + 'static>(
```
dnwiebe: You're passing in a minimum and a maximum here; should the name of this function be something like `assert_gas_limit_is_between`?

dnwiebe (commented):
Resume at blockchain_interface.rs: constructor_for_send_transaction_inputs_handles_value_out_of_range

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1534,4 +1753,117 @@ mod tests {
...
+
+        let result =
+            SendTransactionInputs::new(&payable_account, &consuming_wallet, U256::from(4545), 130);
+
+        assert_eq!(result, Err(BlockchainError::SignedValueConversion(-100)))
+    }
+
+    #[test]
+    fn conversion_between_errors_work() {
+        assert_eq!(
```
dnwiebe: Is there a way you can hack a `match` in here so that the compiler will complain if somebody adds another BlockchainTransactionError without mentioning it in this test?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1534,4 +1753,117 @@ mod tests {
...
+
+    #[test]
+    #[should_panic(expected = "this operation should not happen")]
+    fn sentinel_value_should_not_be_converted() {
+        let _: BlockchainError = BlockchainTransactionError::SentinelVariant.into();
+    }
+
+    #[test]
+    fn display_for_blockchain_transaction_error() {
+        assert_eq!(
```
dnwiebe: Maybe this would be a better place for the `match` that alerts about new BlockchainTransactionErrors. Or maybe there should be a `match` in each of the tests that needs new code for a new error.  Hmm.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1534,4 +1753,117 @@ mod tests {
...
+        );
+        assert_eq!(
+            BlockchainTransactionError::SentinelVariant.to_string(),
+            String::from("This is sentinel for BlockchainTransactionError")
+        );
+    }
+
+    #[test]
+    fn carries_transaction_hash_works() {
+        assert_eq!(BlockchainError::InvalidUrl.carries_transaction_hash(), None);
```
dnwiebe: Another good place for a `match`

node/src/blockchain/test_utils.rs
```diff
@@ -130,8 +153,19 @@ impl BlockchainInterface for BlockchainInterfaceMock {
...
+        self.get_transaction_receipt_params
+            .lock()
+            .unwrap()
+            .push(hash);
+        self.get_transaction_receipt_results.borrow_mut().remove(0)
+    }
+
+    fn send_transaction_tools<'a>(
+        &'a self,
+        _backup_recipient: &dyn PaymentBackupRecipientWrapper,
```
dnwiebe: Leaving this untestable seems like a hole.

dnwiebe (commented):
Resume with tool_wrappers.rs

node/src/blockchain/tool_wrappers.rs
```diff
@@ -40,36 +52,86 @@ impl<'a, T: Transport + Debug> SendTransactionToolsWrapper
...
 
-impl SendTransactionToolsWrapper for SendTransactionToolsWrapperNull {
+impl SendTransactionToolWrapper for SendTransactionToolWrapperNull {
     fn sign_transaction(
         &self,
         _transaction_params: TransactionParameters,
         _key: &secp256k1secrets::key::SecretKey,
     ) -> Result<SignedTransaction, Web3Error> {
-        panic!("sing_transaction() for a null object - should never be called")
+        panic!("sing_transaction() should never be called on the null object")
```
dnwiebe: Sign, not sing

node/src/blockchain/tool_wrappers.rs
```diff
@@ -1,33 +1,45 @@
...
+        Self {
+            web3,
+            payment_backup_sub,
+        }
     }
 }
 
-impl<'a, T: Transport + Debug> SendTransactionToolsWrapper
-    for SendTransactionToolsWrapperReal<'a, T>
+impl<'a, T: Transport + Debug> SendTransactionToolWrapper
```
dnwiebe: Are there tests for this? I didn't see any. We were thinking there might be some in blockchain_bridge.rs, but none were immediately obvious.

node/src/database/db_initializer.rs
```diff
@@ -288,13 +292,38 @@ impl DbInitializerReal {
...
         conn.execute(
-            "create table if not exists receivable (
+            indoc!(
+                "create table if not exists receivable (
                 wallet_address text primary key,
                 balance integer not null,
                 last_received_timestamp integer not null
-            )",
+            )"
+            ),
```
dnwiebe: This table shouldn't have an index on wallet_address, since that's the primary key.

node/src/database/db_initializer.rs
```diff
@@ -639,6 +670,36 @@ mod tests {
...
+            create table pending_payments (
+                rowid integer primary key,
+                transaction_hash text not null,
+                amount integer not null,
+                payment_timestamp integer not null,
+                attempt integer not null,
+                process_error text null
+            )"
+        );
+        assert_on_part_of_the_schema(conn.as_ref(), expected_sql_used_to_create_this_table);
```
dnwiebe: Try asserting on the index as well

node/src/database/db_initializer.rs
```diff
@@ -723,15 +808,17 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "DB migration failed; the resulting records are still incorrect")]
+    #[should_panic(
+        expected = "DB migration failed, the resulting records are still incorrect; found 0 but target 1"
```
dnwiebe: It's not clear what 0 and 1 mean here. Perhaps "Found schema 0 but expecting schema 1?"

node/src/database/db_initializer.rs
```diff
@@ -893,14 +980,19 @@ mod tests {
...
-            bring_db_of_version_0_back_to_life_and_return_connection(
-                &updated_db_path_dir.join(DATABASE_FILE),
-            );
+            bring_db_0_back_to_life_and_return_connection(&updated_db_path_dir.join(DATABASE_FILE));
         }
         let subject = DbInitializerReal::default();
 
         let _ = subject
-            .initialize(&updated_db_path_dir, true, MigratorConfig::test_default())
+            .initialize(
```
dnwiebe, line 1009: See if this works with the old MigratorConfig::test_default(). If so, that would be simpler.

node/src/database/db_migrations.rs
```diff
@@ -250,6 +271,75 @@ impl DatabaseMigration for Migrate_2_to_3 {
...
+            warning!(declaration_utils.logger(),"Migration from 4 to 5: database belonging to the chain '{}'; \
+             we discovered possibly abandoned transactions that are said yet to be pending, these are: '{}'; \
+              continuing",declaration_utils.external_parameters().chain.rec().literal_identifier,unresolved_pending_transactions.join("', '") )
+        } else {
+            debug!(
+                declaration_utils.logger(),
+                "Migration from 4 to 5: no previous pending transactions found; continuing"
+            )
+        };
+        let statement_1 = "ALTER TABLE payable RENAME TO _payable_old";
```
dnwiebe: This should work, but if you want to save a little complexity, you could keep the old `payable` table in place and just use a couple of ALTER TABLE statements to drop the old column and add a new column. That way you wouldn't have to copy any data.

node/src/database/db_migrations.rs
```diff
@@ -261,7 +351,13 @@ impl DbMigratorReal {
     }
 
     fn list_of_migrations<'a>() -> &'a [&'a dyn DatabaseMigration] {
-        &[&Migrate_0_to_1, &Migrate_1_to_2, &Migrate_2_to_3]
+        &[
+            &Migrate_0_to_1,
+            &Migrate_1_to_2,
+            &Migrate_2_to_3,
+            &InterimMigrationPlaceholder(4),
```
dnwiebe: Don't forget to merge in the 3 -> 4 migration from GH-477

node/src/database/db_migrations.rs
```diff
@@ -1242,4 +1441,163 @@ mod tests {
...
+                )),
+            )
+            .unwrap();
+
+        assert_on_schema_5_was_adopted(conn_schema5.as_ref());
+        TestLogHandler::new().exists_log_containing("WARN: DbMigrator: Migration from 4 to 5: database belonging to the chain 'eth-ropsten'; \
+         we discovered possibly abandoned transactions that are said yet to be pending, these are: '0000000000000000000000000000000000000000000000000000000002b594d1', '00000000000000000000000000000000000000000000000000000000000f41d0'; continuing");
+    }
+
+    fn assert_on_schema_5_was_adopted(conn_schema5: &dyn ConnectionWrapper) {
```
dnwiebe, line 1853: Somewhere--especially if the rename-table-to-old-then-copy-data-to-new-table process is going to stay in place--I'd like to see assertions that old data was actually successfully copied to the new table and still exists.

node/src/accountant/mod.rs
bertllll: I fixed what suggested; but there is more I may want to think about...if I should represent the limit as time bounds or as number of blocks added to the blockchain meanwhile. I find it hard to think how to implement the latter. I may get back to this later.

node/src/blockchain/tool_wrappers.rs
bertllll: This is tested but yeah, not easy to find it. It's obfuscated under a helper method so it does not show the struct's name in the test.

dnwiebe (approved):

PR #99, opened 2022-01-14
zero-hop-hotfix: making past-neighbors configurable specially for zer… 

bertllll:
…o-hop mode


node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -100,7 +101,11 @@ impl NodeConfigurator<BootstrapperConfig> for NodeConfiguratorStandardUnprivileg
             &mut unprivileged_config,
             Some(persistent_config.as_mut()),
         )?;
-        configure_database(&unprivileged_config, persistent_config.as_mut())?;
+        configure_database(
```
dnwiebe: Rather than modifying the signature of this method for a special case, can you move the call to set_neighbors_in_zero_hop_if_supplied up into this method?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -301,6 +313,49 @@ pub fn configure_database(
     Ok(())
 }
 
+fn set_neighbors_in_zero_hop_if_supplied(
+    multi_config: &MultiConfig,
+    neighborhood_mode: NeighborhoodModeLight,
+    persistent_config: &mut dyn PersistentConfiguration,
+) -> Result<(), ConfiguratorError> {
+    fn split_descriptors(whole_input: String) -> Vec<NodeDescriptor> {
```
dnwiebe: Is this already implemented somewhere else? It seems like it ought to be. Can we use that implementation here?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1351,10 +1413,45 @@ mod tests {
...
             result,
-            Err(ConfiguratorError::required(
-                "neighborhood-mode",
-                "Node cannot run as --neighborhood-mode zero-hop if --neighbors is specified"
-            ))
+            Ok(NeighborhoodConfig {
+                mode: NeighborhoodMode::ZeroHop
+            })
+        )
+    }
```
dnwiebe: I think there should be an assertion here that the PersistentConfiguration was used with the password to set the past-neighbors value. Or maybe this test is now useless duplication.

dnwiebe (approved):

PR #102, opened 2022-02-02
GH-485: adding advanced configurable parameters 

bertllll:


USER-INTERFACE-INTERFACE.md
```diff
@@ -386,6 +405,66 @@ database password. If you want to know whether the password you have is the corr
   try to connect to when it starts up next time.  It's a secret, so if you don't supply the `dbPasswordOpt` in the
   request you won't see it.
 
+* `paymentCurves`: These are parameters that define the Node's regard to quality of financial records created during interaction with
+  other Nodes. The goal is to determine whether the records have crossed boarders meaning that a certain action needs to be taken
```
masqrauder: s/boarders/thresholds/

masqrauder: maybe `parameters that define thresholds to determine when and how much to pay other nodes for routing and exit services and the expectations the node should have for receiving payments from other nodes for routing and exit services.`

USER-INTERFACE-INTERFACE.md
```diff
@@ -386,6 +405,66 @@ database password. If you want to know whether the password you have is the corr
   try to connect to when it starts up next time.  It's a secret, so if you don't supply the `dbPasswordOpt` in the
   request you won't see it.
 
+* `paymentCurves`: These are parameters that define the Node's regard to quality of financial records created during interaction with
+  other Nodes. The goal is to determine whether the records have crossed boarders meaning that a certain action needs to be taken 
+  - either to make a payment to the touched Nodes in order to redeem our debts or put those Nodes under ban since they've too long 
+  ignored their liabilities toward us. These are ever present values, no matter if the user's set any value, they have defaults.
+
+* `balanceDecreasesForSec`: This interval begins after payment_suggested_after_sec for payables and after
```
masqrauder: `This time interval -- in seconds -- begins...` would help to explain the `sec` abbreviation.

USER-INTERFACE-INTERFACE.md
```diff
@@ -386,6 +405,66 @@ database password. If you want to know whether the password you have is the corr
...
 
+* `paymentCurves`: These are parameters that define the Node's regard to quality of financial records created during interaction with
+  other Nodes. The goal is to determine whether the records have crossed boarders meaning that a certain action needs to be taken 
+  - either to make a payment to the touched Nodes in order to redeem our debts or put those Nodes under ban since they've too long 
+  ignored their liabilities toward us. These are ever present values, no matter if the user's set any value, they have defaults.
+
+* `balanceDecreasesForSec`: This interval begins after payment_suggested_after_sec for payables and after 
+  payment_suggested_after_sec + payment_grace_before_ban for receivables. During the interval, the amount of a payable that is allowed
+  to remain unpaid, or a receivable that doesn't cause a ban, decreases linearly from the balance_to_decrease_from to
+  permanent_debt_allowed_gwei or unban-when_balance_below_gwei.
```
masqrauder: Is `unban-when_balance_below_gwei` supposed to be `unban_when_balance_below_qwei`

USER-INTERFACE-INTERFACE.md
```diff
@@ -372,7 +391,7 @@ database password. If you want to know whether the password you have is the corr
 * `earningWalletAddressOpt`: The wallet address for the earning wallet. This is not secret, so
   if you don't get this field, it's because it hasn't been set yet.
 
-* `gasPrice`: The Node will not pay more than this number of wei for gas to complete a transaction.
+* `gasPrice`: The Node will not pay more than this number of Gwei for gas to complete a transaction.
```
masqrauder, line 395: @kauri-hero you may want to review the changes in this file to make sure it's consistent with other documentation and how these same concepts are explained to users and GUI developers and testers.

masq/src/commands/configuration_command.rs
```diff
@@ -281,9 +358,24 @@ mod tests {
...
+|                                  Unban when balance below gwei:   12000\n\
+|Rate pack:                        \n\
+|                                  Routing byte rate:               8\n\
+|                                  Routing service rate:            9\n\
+|                                  Exit byte rate:                  12\n\
+|                                  Exit service rate:               14\n\
+|Scan intervals:                   \n\
+|                                  Pending payable sec:             150\n\
+|                                  Payable sec:                     155\n\
+|                                  Receivable sec:                  250\n"
```
masqrauder: I'm not sure people will be able to understand what `Balance decreases for sec` means from the description. it might also be good to spell out `seconds` or perhaps `secs` I've read this and the earlier documentation and I can't remember if `Balance decreases for sec` is the statement or if `Balance decreases for` in seconds is. If the latter placing the units of measure in parentheses may help.
e.g.: 
- `Balance decreases for (secs):`
- `Balance to decrease from (gwei):`
... etc.

USER-INTERFACE-INTERFACE.md
bertllll: my understanding is that "borders" is what should be there instead but I don't get your comment...

bertllll: Thanks, this is helpful.

masq_lib/src/combined_parameters.rs
```diff
@@ -0,0 +1,647 @@
...
+                    ),
+                ),
+                receivable_scan_interval: Duration::from_secs(
+                    CombinedParamsValueRetriever::get_value(
+                        &parsed_values,
+                        "receivable_scan_interval",
+                    ),
+                ),
+            })),
+            _ => panic!(
```
FinsaasGH: Do we want to panic here or just write to error log what happened and why?

masq_lib/src/constants.rs
```diff
@@ -1,7 +1,10 @@
 // Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
 
 use crate::blockchains::chains::Chain;
+use crate::combined_parameters::{PaymentCurves, RatePack, ScanIntervals};
 use const_format::concatcp;
+use lazy_static::lazy_static;
+use std::time::Duration;
 
 pub const DEFAULT_CHAIN: Chain = Chain::EthMainnet;
```
FinsaasGH: Do we want EthMainnet as default or PolygonMainnet?

USER-INTERFACE-INTERFACE.md
kauri-hero: yes we will provide some edits here to make it more understandable

masq_lib/src/constants.rs
kauri-hero: > Do we want EthMainnet as default or PolygonMainnet?

I would suggest we default to polygon mainnet @bertllll

USER-INTERFACE-INTERFACE.md
kauri-hero: yes that is good suggestion

masq/src/commands/configuration_command.rs
kauri-hero: After discussion with @bertllll we are suggesting these new names for parameters:

balanceDecreasesForSec => thresholdIntervalSec
balanceToDecreaseFromGwei => debtThresholdSec
paymentSuggestedAfterSec => maturityThresholdSec
paymentGraceBeforeBanSec => paymentGracePeriodSec
permanentDebtAllowedGwei => permanentDebtAllowedGwei
unbanWhenBalanceBelowGwei => unbanBelowGwei

USER-INTERFACE-INTERFACE.md
kauri-hero: yes agreed

USER-INTERFACE-INTERFACE.md
```diff
@@ -386,6 +405,66 @@ database password. If you want to know whether the password you have is the corr
...
+  pending; the payments were sent to pay our debts, the payable. The purpose of this process is to confirm that the payment was written
+  into the blockchain or failed.
+
+* `payableSec`: Amount of seconds between two sequential cycles of scanning aimed to find payable accounts of certain parameters; these 
+  accounts are tracked on behalf of our creditors. If they qualify by enough big significance of time or money amount we have the debt be
+  paid by sending a transaction to the creditor in question.
+
+* `receivableSec`: Amount of seconds between two sequential cycles of scanning for payments on the blockchain that have been sent
+  by our creditors to us, they pay for our services this way.
+
```
kauri-hero: My suggestions for editing these descriptions @bertllll 

Payment Thresholds: These are parameters that define thresholds to determine when and how much to pay other nodes for routing and exit services and the expectations the node should have for receiving payments from other nodes for routing and exit services. The thresholds are also used to determine whether to offer services to other Nodes or enact a ban since they have not paid mature debts. These are ever present values, no matter if the user's set any value, as they have defaults.

thresholdIntervalSec: This interval -- in seconds -- begins after maturityThresholdSec for payables and after maturityThresholdSec + paymentGracePeriodSec for receivables. During the interval, the amount of a payable that is allowed to remain unpaid, or a pending receivable that won’t cause a ban, decreases linearly from the debtThresholdGwei to permanentDebtAllowedGwei or unban- unbanBelowGwei.

debtThresholdGwei: Payables higher than this -- in Gwei of MASQ -- will be suggested for payment immediately upon passing the maturityThresholdSec age. Payables lower than this can stay unpaid longer. Receivables higher than this will be expected to be settled by other Nodes, but will never cause bans until they pass the maturityThresholdSec + paymentGracePeriodSec age. Receivables lower than this will survive longer without banning.

maturityThresholdSec: Large payables can get this old -- in seconds -- before the Accountant's scanner suggests that it be paid.

paymentGracePeriodSec: A large receivable can get as old as maturityThresholdSec + paymentGracePeriodSec -- in seconds -- before the Node that owes it will be banned.

permanentDebtAllowedGwei: Receivables this small and smaller -- in Gwei of MASQ -- will not cause bans no matter how old they get.

unbanBelowGwei: When a delinquent Node has been banned due to non-payment, the receivables balance must be paid below this level -- in Gwei of MASQ -- to cause them to be unbanned. In most cases, you'll want this to be set the same as permanentDebtAllowedGwei.

ratePack: These four parameters specify your rates that your Node will use for charging other Nodes for your provided services. They are currently denominated in Gwei of MASQ, but will be improved to allow denomination in Wei units. These are ever present values, no matter if the user's set any value, they have defaults.

exitByteRate: This parameter indicates an amount of MASQ demanded to process 1 byte of routed payload while the Node acts as the exit Node.

exitServiceRate: This parameter indicates an amount of MASQ demanded to provide services, unpacking and repacking 1 CORES package, while the Node acts as the exit Node.

routingByteRate: This parameter indicates an amount of MASQ demanded to process 1 byte of routed payload while the Node is a common relay Node.

routingServiceRate: This parameter indicates an amount of MASQ demanded to provide services, unpacking and repacking 1 CORES package, while the Node is a common relay Node.

scanIntervals: These three intervals describe the length of three different scan cycles running automatically in the background since the Node has connected to a qualified neighborhood that consists of neighbors enabling a complete 3-hop route. Each parameter can be set independently, but by default are all the same which currently is most desirable for the consistency of service payments to and from your Node. Technically, there doesn't have to be any lower limit for the minimum of time you can set; two scans of the same sort would never run at the same time but the next one is always scheduled not earlier than the end of the previous one (but the time elapsed may be a bit longer due to placing another, different task in between, because that could be created during the waiting period, and thus it factually goes before the second scan). These are ever present values, no matter if the user's set any value, they have defaults.

pendingPayableSec: Amount of seconds between two sequential cycles of scanning for payments that are marked as currently pending; the payments were sent to pay our debts, the payable. The purpose of this process is to confirm the status of the pending payment; either the payment transaction was written on blockchain as successful or failed.

payableSec: Amount of seconds between two sequential cycles of scanning aimed to find payable accounts of that meet the criteria set by the Payment Thresholds; these accounts are tracked on behalf of our creditors. If they meet the Payment Threshold criteria, our Node will send a debt payment transaction to the creditor in question.

receivableSec: Amount of seconds between two sequential cycles of scanning for payments on the blockchain that have been sent by our creditors to us, which are credited against receivables recorded for services provided.

masq_lib/src/messages.rs
```diff
@@ -501,9 +500,54 @@ pub struct UiConfigurationResponse {
     pub consuming_wallet_address_opt: Option<String>,
     #[serde(rename = "pastNeighbors")]
     pub past_neighbors: Vec<String>,
+    #[serde(rename = "paymentCurves")]
+    pub payment_curves: UiPaymentCurves,
+    #[serde(rename = "ratePack")]
+    pub rate_pack: UiRatePack,
+    #[serde(rename = "scanIntervals")]
+    pub scan_intervals: UiScanIntervals,
 }
```
kauri-hero: I propose we change the parameter set for the 'curves' to:

Payment Curves => Payment Thresholds

masq_lib/src/messages.rs
```diff
@@ -501,9 +500,54 @@ pub struct UiConfigurationResponse {
...
+    #[serde(rename = "balanceToDecreaseFromGwei")]
+    pub balance_to_decrease_from_gwei: i64,
+    #[serde(rename = "paymentGraceBeforeBanSec")]
+    pub payment_grace_before_ban_sec: i64,
+    #[serde(rename = "paymentSuggestedAfterSec")]
+    pub payment_suggested_after_sec: i64,
+    #[serde(rename = "permanentDebtAllowedGwei")]
+    pub permanent_debt_allowed_gwei: i64,
+    #[serde(rename = "unbanWhenBalanceBelowGwei")]
+    pub unban_when_balance_below_gwei: i64,
```
kauri-hero: proposed parameter names are:
balanceDecreasesForSec => thresholdIntervalSec
balanceToDecreaseFromGwei => debtThresholdGwei
paymentSuggestedAfterSec => maturityThresholdSec
paymentGraceBeforeBanSec => paymentGracePeriodSec
permanentDebtAllowedGwei => permanentDebtAllowedGwei
unbanWhenBalanceBelowGwei => unbanBelowGwei

kauri-hero (changes requested):
My review covers the updating of the UI interface document and description of the Payment Threshold params (previously Payment Curve), and suggested renaming of the parameters to be more understandable

node/src/bootstrapper.rs
```diff
@@ -2092,7 +2075,7 @@ mod tests {
             config: BootstrapperConfig,
             _actor_factory: Box<dyn ActorFactory>,
             _persist_config: &dyn PersistentConfiguration,
-            _tools: &dyn ActorSystemFactoryTools,
+            // _tools: &dyn ActorSystemFactoryTools,
```
FinsaasGH: Can we remove this?

node/src/daemon/setup_reporter.rs
```diff
@@ -78,10 +85,9 @@ impl SetupReporter for SetupReporterReal {
                     blanked_out_former_values.insert(v.name.clone(), former_value);
                 };
             });
-        //we had troubles at an attempt to blank out this parameter on an error resulting in a diverging chain from the data_dir
-        if blanked_out_former_values.get("chain").is_some() {
-            let _ = blanked_out_former_values.remove("chain");
-        }
+        //TODO investigate this, not sure if the right way to solve the issue
+        //answers an attempt to blank out chain beyond an error resulting in chain different from data_dir
+        let _ = blanked_out_former_values.remove("chain");
```
FinsaasGH, line 89: Seems reasonable. Would it make sense to move `blanked_out_former_values.remove("chain");` into the block above? EX:
`if let Some(former_value) = existing_setup.remove(&v.name) {
                    blanked_out_former_values.insert(v.name.clone(), former_value);
                    blanked_out_former_values.remove("chain");
                };`

node/src/daemon/setup_reporter.rs
```diff
@@ -1486,6 +1624,7 @@ mod tests {
     }
 
     #[test]
+    //TODO we should change the name of this test - these are including the requireds, maybe 'all but fundamentals'?
```
FinsaasGH: Maybe something like `get_required_setup_database_values_excluding_optional`?

masq_lib/src/combined_parameters.rs
bertllll: This event indicates that the programmer's design is wrong; it should never happen so we want a panic to get to know as soon as possible from anyone seeing this panic. If so, it would require to change the code. So I think this is okay.

node/src/daemon/setup_reporter.rs
bertllll: Well, the block above iterates in a loop and we would have to evaluate the expression every time, with it placed outside we know we do it just once.

masq_lib/src/shared_schema.rs
```diff
@@ -110,6 +110,61 @@ pub const REAL_USER_HELP: &str =
...
+     These are parameters that define thresholds to determine when and how much to pay other Nodes for routing and \
+     exit services and the expectations the Node should have for receiving payments from other Nodes for routing and \
+     exit services. The thresholds are also used to determine whether to offer services to other Nodes or enact a ban \
+     since they have not paid mature debts. These are ever present values, no matter if the user's set any value, as \
+     they have defaults. The parameters must be always supplied all together, delimited by vertical bars and in the right \
+     order.\n\n\
+     1. Debt Threshold Gwei: Payables higher than this -- in Gwei of MASQ -- will be suggested for payment immediately \
+     upon passing the Maturity Threshold Sec age. Payables lower than this can stay unpaid longer. Receivables higher than \
+     this will be expected to be settled by other Nodes, but will never cause bans until they pass the Maturity Threshold Sec \
+     + Payment Grace Period Sec age. Receivables lower than this will survive longer without banning.\n\n\
```
masqrauder: instead of `lower` perhaps `less` is more customary

masq_lib/src/shared_schema.rs
```diff
@@ -110,6 +110,61 @@ pub const REAL_USER_HELP: &str =
...
+     repacking 1 CORES package, while the Node acts as the exit Node.";
+pub const PAYMENT_THRESHOLDS_HELP: &str = "\
+     These are parameters that define thresholds to determine when and how much to pay other Nodes for routing and \
+     exit services and the expectations the Node should have for receiving payments from other Nodes for routing and \
+     exit services. The thresholds are also used to determine whether to offer services to other Nodes or enact a ban \
+     since they have not paid mature debts. These are ever present values, no matter if the user's set any value, as \
+     they have defaults. The parameters must be always supplied all together, delimited by vertical bars and in the right \
+     order.\n\n\
+     1. Debt Threshold Gwei: Payables higher than this -- in Gwei of MASQ -- will be suggested for payment immediately \
+     upon passing the Maturity Threshold Sec age. Payables lower than this can stay unpaid longer. Receivables higher than \
```
masqrauder: instead of `lower` perhaps `less` is more customary

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -0,0 +1,2335 @@
...
+        "rate-pack",
+        multi_config,
+        persist_config,
+        |str: &str| RatePack::try_from(str),
+        |pc: &dyn PersistentConfiguration| pc.rate_pack(),
+        |pc: &mut dyn PersistentConfiguration, rate_pack| pc.set_rate_pack(rate_pack),
+    )
+}
+
+//TODO untested explicit panics ... let's talk it through at a review
```
FinsaasGH: Not familiar with this, will investigate further.

PR #103, opened 2022-02-08
GH-563: HOT-FIX - update Mumbai test token contract

utkarshg6:


node/tests/contract_test.rs
```diff
@@ -0,0 +1,93 @@
...
+                "name": "",
+                "type": "uint8"
+            }
+        ],
+        "payable": false,
+        "stateMutability": "view",
+        "type": "function"
+    }]"#;
+    let abi = ethabi::Contract::load(min_abi_json.as_bytes()).unwrap();
+    let contract = web3::contract::Contract::new(web3.eth(), address, abi);
```
dnwiebe, line 47: This isn't critical, but I'd like to see a more descriptive name here. It's not really the contract, it's the pair of results from the two function calls. I understand that those two results are intimately connected with the contract, but they're not the same thing. Reading the code might be easier with a better name.

dnwiebe (approved):

PR #110, opened 2022-02-18
GH-571: db-password ineffectively distributed to the componnents needing it

bertllll:


node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1364,56 +1351,27 @@ mod tests {
...
-            &mut persistent_config,
-            &mut unprivileged_config,
-        )
-        .unwrap();
-
-        assert!(result.is_empty());
-    }
-
-    #[test]
-    fn get_past_neighbors_handles_unavailable_password() {
```
dnwiebe, line 1384: We eliminated this test. Was that a good idea? Investigate.

dnwiebe (approved):

PR #111, opened 2022-02-24
GH-569: Write Tests for Constants

utkarshg6:
This PR adds 
- The tests for constants
- Resolves the issue with the test `tls_through_node_test`.
- Changes due to version bump to **Rust 1.59.0**
- Add tests for `static ref`
- Add new constants add tests for `clap` code inside `masq/src/commands` directory
- Improve indentations
- Minor feedback changes


node/src/proxy_server/server_impersonator_tls.rs
```diff
@@ -42,6 +42,31 @@ const TLS_UNRECOGNIZED_NAME_ALERT: [u8; 7] = [
 mod tests {
     use super::*;
 
+    #[test]
+    fn constants_have_correct_values() {
+        const TLS_INTERNAL_ERROR_ALERT_EXPECTED: [u8; 7] = [
```
bertllll: Change these expected values to be simple variables not constants.

node/src/sub_lib/utils.rs
```diff
@@ -203,6 +203,11 @@ pub mod tests {
     use masq_lib::multi_config::CommandLineVcl;
     use masq_lib::test_utils::logging::{init_test_logging, TestLogHandler};
 
+    #[test]
```
bertllll: This is my mistake. I should have never brought this constant into life.

Accordingly, please delete your test and also inline the text of the constant in all affected places.

node/src/http_request_start_finder.rs
```diff
@@ -78,6 +78,16 @@ mod tests {
     use crate::sub_lib::http_packet_framer::ChunkProgressState;
     use crate::sub_lib::http_packet_framer::PacketProgressState;
 
+    #[test]
+    fn constants_have_correct_values() {
+        const METHODS_EXPECTED: &[&[u8]] = &[
```
bertllll: Make this a variable, not a constant.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
 
+    #[test]
+    fn constants_have_correct_values() {
+        assert_eq!(
+            BLOCKCHAIN_SERVICE_HELP,
+            "The Ethereum client you wish to use to provide Blockchain \
+     exit services from your MASQ Node (e.g. http://localhost:8545, \
+     https://ropsten.infura.io/v3/YOUR-PROJECT-ID, https://mainnet.infura.io/v3/YOUR-PROJECT-ID), \
+     https://polygon-mainnet.infura.io/v3/YOUR-PROJECT-ID"
+        );
```
bertllll, line 565: Consult these help messages with Dan, I don't feel easy to say no or yes.


My newer opinion is that it's a good idea to keep these assertions. So go ahead :).

node/src/apps.rs
```diff
@@ -92,3 +92,46 @@ const NODE_HELP_TEXT: &str = indoc!(
...
         3. Create the port forwarding entries in the router."
 );
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn constants_have_correct_values() {
+        assert_eq!(
```
bertllll, line 102: Same as the previous one; here is the question of testing help messages. But maybe that really is a good idea - given it's quite easy to leave an unintended char in text when not paying enough attention.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
+        assert_eq!(EARNING_WALLET_HELP, "An Ethereum wallet address. Addresses must begin with 0x followed by 40 hexadecimal digits \
+     (case-insensitive). If you already have a derivation-path earning wallet, don't supply this. \
+     If you have supplied an earning wallet address before, either don't supply it again or be \
+     careful to supply exactly the same one you supplied before.");
+        assert_eq!(IP_ADDRESS_HELP, "The public IP address of your MASQ Node: that is, the IPv4 \
+     address at which other Nodes can contact yours. If you're running your Node behind \
+     a router, this will be the IP address of the router. If this IP address starts with 192.168 or 10.0, \
+     it's a local address rather than a public address, and other Nodes won't be able to see yours. \
+     --ip is meaningless except in --neighborhood-mode standard.");
+        assert_eq!(LOG_LEVEL_HELP, "The minimum severity of the logs that should appear in the Node's logfile. You should probably not specify \
```
bertllll: Please just reconsider a better formatting for lines like this one so it doesn't stick out.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
 
+    use super::*;
     use crate::blockchains::chains::Chain;
     use crate::shared_schema::{common_validators, official_chain_names};
 
+    #[test]
+    fn constants_have_correct_values() {
+        assert_eq!(
+            BLOCKCHAIN_SERVICE_HELP,
+            "The Ethereum client you wish to use to provide Blockchain \
```
bertllll, line 537: If you can do it, get rid of the indents on the first line because this is unusual with regards to our general writing style. We usually use just a plain block of text without it.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
+     exit services from your MASQ Node (e.g. http://localhost:8545, \
+     https://ropsten.infura.io/v3/YOUR-PROJECT-ID, https://mainnet.infura.io/v3/YOUR-PROJECT-ID), \
+     https://polygon-mainnet.infura.io/v3/YOUR-PROJECT-ID"
+        );
+        assert_eq!(CHAIN_HELP, "The blockchain network MASQ Node will configure itself to use. You must ensure the \
+    Ethereum client specified by --blockchain-service-url communicates with the same blockchain network.");
```
bertllll: This is up to you but you could make the line aligned with the beginning of the other paragraph. I saw one more further down in this test.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
+     cannot maintain a constant IP address or stay constantly on the Network. --ip and --clandestine_port \
+     are incompatible with --neighborhood_mode consume-only.\n\n\
+     standard means that your Node will operate fully unconstrained, both originating and accepting \
+     connections, both consuming and providing services, and when you operate behind a router, it \
+     requires that you forward your clandestine port through that router to your Node's machine.");
+        assert_eq!(MAPPING_PROTOCOL_HELP, "The Node can speak three protocols to your router to make it allow outside Nodes to connect inward \
+    through it to your machine. These three protocols are pcp, pmp, and igdp. The Node can try them one \
+    by one to determine which your router supports, but if you happen to know already, you can supply the \
+    name of the protocol here. If you've taken care of port mapping in some other way, \
+    and you don't need Node to negotiate with your router, say 'none' here and be sure to specify your \
```
bertllll: You could put a single space in between the left border and the start of the text.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
 #[cfg(test)]
 mod tests {
 
+    use super::*;
     use crate::blockchains::chains::Chain;
     use crate::shared_schema::{common_validators, official_chain_names};
 
+    #[test]
+    fn constants_have_correct_values() {
```
bertllll, line 534: If you have time and determination you can read through all this tested text and see if there are any remarkable typos or stylistic imperfections etc. If so then its good to take this opportunity to make it right when we put these tests in.

masq_lib/src/shared_schema.rs
```diff
@@ -526,9 +526,110 @@ impl ConfiguratorError {
...
+    and you don't need Node to negotiate with your router, say 'none' here and be sure to specify your \
+    public IP address with the --ip parameter. If the Node communicates successfully with your router, \
+    it will remember the protocol it used, and on its next run it will try that protocol first, unless \
+    you specify a different protocol on the command line.");
+        assert_eq!(REAL_USER_HELP,"The user whose identity Node will assume when dropping privileges after bootstrapping. Since Node refuses to \
+     run with root privilege after bootstrapping, you might want to use this if you start the Node as root, or if \
+     you start the Node using pkexec or some other method that doesn't populate the SUDO_xxx variables. Use a value \
+     like <uid>:<gid>:<home directory>.");
+    }
+
```
bertllll, line 697: There are some files...I cannot point to them directly because the code is so far unchanged but it occurred to me that you should also take care of `static` global variables.

In a case like this:
```
lazy_static! {
    static ref UI_PORT_HELP: String = format!(
        "If the Daemon is listening for connections at some port other than {}, specify that port \
         here. Must be between {} and {}.",
        DEFAULT_UI_PORT, LOWEST_USABLE_INSECURE_PORT, HIGHEST_USABLE_PORT
    );
    static ref DEFAULT_UI_PORT_STRING: String = format!("{}", DEFAULT_UI_PORT);
}
```

Maybe not a good example, here you can fix from the second one to:
static ref DEFAULT_UI_PORT_STRING: String = DEFAULT_UI_PORT.to_string();

The first assertion could look like this: 
```   
    assert_eq!(UI_PORT_HELP, format!(
        "If the Daemon is listening for connections at some port other than {}, specify that port \
         here. Must be between {} and {}.",
        DEFAULT_UI_PORT, LOWEST_USABLE_INSECURE_PORT, HIGHEST_USABLE_PORT
    );
``` 
    
 Put differently, should look exactly like the real declaration. :)

bertllll, line 697: There is a place in `masq/src/schema.rs` where we're configuring the Clap's App object. We write the "author' in and also the version, probably a line of description too. You could bring in new constants for each of them and then also test them the way you test constants at other places. 

I'd like you to do the same with Clap code in `masq/src/commands.rs` where probably every file in there (except command_common.rs) contains a description - which is not tested in fact. 

And also in node/src/apps.rs. ;)

utkarshg6: @bertllll 
I would suggest you to have a look at this particular commit - https://github.com/MASQ-Project/Node/commit/e5cabf9de5bf7270b492ece00292103fb0c54b89

Here, I created way too many constants, do you think it is a good decision to create so many constants and then test them?

bertllll: You really did...we will need to clean up the mess 😄

utkarshg6: It is completed in **_GH-569 Review 2_**.

bertllll (approved):
I'm fine with this code finally. Long life to Utkarsh

PR #120, opened 2022-04-23
GH-597: correct AutomapProtocol db storage

bertllll:


automap/src/control_layer/automap_control.rs
```diff
@@ -232,6 +232,10 @@ impl AutomapControlReal {
     ) -> Result<ProtocolInfo<T>, AutomapError> {
         let mut transactors_ref_mut = self.transactors.borrow_mut();
         if let Some(usual_protocol) = self.usual_protocol_opt {
+            debug!(
+                self.logger,
+                "Using a previously discovered protocol: {}", usual_protocol
```
dnwiebe: This should either be reworded ("Trying a previously-discovered protocol" or something) or it should be moved down to where we know for sure that we really are using a previously-discovered protocol.

automap/src/mocks.rs
```diff
@@ -200,3 +219,227 @@ impl FindRoutersCommandMock {
...
+        result: Result<ChangeHandler, AutomapError>,
+    ) -> Self {
+        self.stop_housekeeping_thread_results
+            .borrow_mut()
+            .push(result);
+        self
+    }
+}
+
+pub fn pmp_protocol_scenario_for_actor_system_factory_test(
```
dnwiebe: Maybe this could be something like "prepare_test_automap_control" that accepted a bunch of parameters including mock Transactors and a change handler. Then you could prepare the mock Transactors in the distant test, and use this just for final assembly of the AutomapControlReal, without this function having specific knowledge of what the distant test requires.

node/ci/build.sh
```diff
@@ -3,5 +3,5 @@
 CI_DIR="$( cd "$( dirname "$0" )" && pwd )"
 
 pushd "$CI_DIR/.."
-cargo build --all --lib --bins --release --verbose --features masq_lib/no_test_share
+cargo build --all --lib --bins --release --verbose --features "masq_lib/no_test_share automap/no_test_share"
```
dnwiebe, line 7: Should this be in the build .yml file? If not, make a card for it.

node/src/bootstrapper.rs
```diff
@@ -78,6 +78,20 @@ pub fn alias_cryptde_ref<'a>() -> &'a dyn CryptDE {
     }
 }
 
+pub struct CryptdePair {
```
dnwiebe: This name should probably be CryptDEPair, with the DE capitalized.

dnwiebe (commented):
Continue with actor_system_factory.rs

node/src/actor_system_factory.rs
```diff
@@ -101,20 +102,19 @@ pub struct ActorSystemFactoryToolsReal {
...
         config: BootstrapperConfig,
+        persistent_config: &mut dyn PersistentConfiguration,
         actor_factory: Box<dyn ActorFactory>,
     ) -> StreamHandlerPoolSubs {
         let db_initializer = DbInitializerReal::default();
-        // make all the actors
         let (dispatcher_subs, pool_bind_sub) = actor_factory.make_and_start_dispatcher(&config);
         let proxy_server_subs =
-            actor_factory.make_and_start_proxy_server(main_cryptde, alias_cryptde, &config);
+            actor_factory.make_and_start_proxy_server(cryptdes.main, cryptdes.alias, &config);
```
dnwiebe: If you wanted, you could modify make_and_start_proxy_server to take two parameters instead of three.

node/src/actor_system_factory.rs
```diff
@@ -616,29 +648,29 @@ mod tests {
...
     }
 
     impl ActorSystemFactoryTools for ActorSystemFactoryToolsMock {
         fn prepare_initial_messages(
             &self,
-            main_cryptde: &'static dyn CryptDE,
-            alias_cryptde: &'static dyn CryptDE,
+            cryptdes: CryptdePair,
             config: BootstrapperConfig,
+            _persistent_config: &mut dyn PersistentConfiguration,
```
dnwiebe: I guess I'd like to see this Boxed and stored in the parameter list so that tests could verify that the PersistentConfiguration they sent in was the one that was used to call prepare_initial_messages().

node/src/actor_system_factory.rs
```diff
@@ -982,24 +1014,24 @@ mod tests {
...
+        let mut persistent_config =
             PersistentConfigurationMock::default().chain_name_result("eth-ropsten".to_string());
         Bootstrapper::pub_initialize_cryptdes_for_testing(
             &Some(main_cryptde()),
             &Some(alias_cryptde()),
         );
         let mut tools = ActorSystemFactoryToolsReal::new();
         tools.automap_control_factory = Box::new(
-            AutomapControlFactoryMock::new().make_result(
+            AutomapControlFactoryMock::new().make_result(Box::new(
```
dnwiebe, line 1064: Maybe this doesn't need to be Boxed? I dunno--you seemed to want me to mark this.

bertllll: Nope, I cannot remove it. A bit of disappointment here.

node/src/actor_system_factory.rs
bertllll: This is hard, there not opportunities to test it like you suggest anyway or not many. Also we know, after my change, can do that for prepare_initial_messages() but cannot do such a luxury for other methods like validate_database_chain.

dnwiebe (approved):

node/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod sub_lib;
 
 #[macro_use]
 extern crate masq_lib;
+extern crate core;
```
bertllll: Not needed; I have no idea about where it came from.

dnwiebe (approved):

PR #121, opened 2022-04-27
GH-595: better design of NotifyLaterHandle and NotifyHandle

bertllll:


node/src/sub_lib/utils.rs
```diff
@@ -139,60 +139,103 @@ fn crash_request_analyzer(
...
+    A: Actor<Context = Context<A>> + Handler<M>,
+{
     fn default() -> Self {
         Box::new(NotifyHandleReal {
             phantom: PhantomData::default(),
         })
     }
 }
 
+pub trait NLSpawnHandleWrapper {
```
dnwiebe: "Wrapper" implies that the wrapper type will have all the methods the delegate type does (and implement all the delegate's traits); if the job is just to hold and return an instance of the delegate type, a more appropriate name might be "Holder."

node/src/sub_lib/utils.rs
```diff
@@ -139,60 +139,103 @@ fn crash_request_analyzer(
     }
 }
 
-pub trait NotifyLaterHandle<T> {
-    fn notify_later<'a>(
-        &'a self,
+pub trait NotifyLaterHandle<T, A>
```
dnwiebe: Should T be M instead?

node/src/sub_lib/utils.rs
```diff
@@ -375,4 +421,116 @@ mod tests {
...
+    fn notify_handles_real_sends_their_messages_correctly() {
+        let (sender, receiver) = unbounded();
+        let test_actor = NotifyHandlesTestActor::new(sender);
+        let _ = test_actor.start();
+        let system = System::new("notify_handles_test");
+
+        system.run();
+
+        let mut data = Vec::new();
+        (0..2).for_each(|_| data.push(receiver.recv().unwrap()));
```
dnwiebe: If there's a version of .recv() that has a timeout in it, we can eliminate the danger of this hanging forever in Actions. The timeout can be fairly ridiculously long, as we don't expect it ever to time out unless there's a big problem.

node/src/test_utils/mod.rs
```diff
@@ -713,13 +715,15 @@ pub mod unshared_test_utils {
     }
 
     pub struct NotifyLaterHandleMock<T> {
-        notify_later_params: Arc<Mutex<Vec<(T, Duration)>>>, //I care only about the params; realize that it's hard to test self addressed messages if you cannot mock yourself
+        notify_later_params: Arc<Mutex<Vec<(T, Duration)>>>,
+        do_you_want_to_proceed_after: bool,
```
dnwiebe: When you wrote this, I bet it was really clear to you what "proceed" meant; but when another developer encounters this in a year, it won't be obvious to him.

Consider a name change to something involving sending the message on or not sending it on.

node/src/test_utils/mod.rs
```diff
@@ -729,30 +733,51 @@ pub mod unshared_test_utils {
...
             interval: Duration,
-            mut closure: Box<dyn FnMut(T, Duration) -> SpawnHandle + 'a>,
-        ) -> SpawnHandle {
+            ctx: &'a mut Context<A>,
+        ) -> Box<dyn NLSpawnHandleWrapper> {
             self.notify_later_params
                 .lock()
                 .unwrap()
                 .push((msg.clone(), interval));
             if !cfg!(test) {
```
dnwiebe: Pretty sure this is unnecessary.

node/src/test_utils/mod.rs
```diff
@@ -769,16 +794,25 @@ pub mod unshared_test_utils {
...
-    impl<T: Message + Clone> NotifyHandle<T> for NotifyHandleMock<T> {
-        fn notify<'a>(&'a self, msg: T, mut closure: Box<dyn FnMut(T) + 'a>) {
+    impl<M, A> NotifyHandle<M, A> for NotifyHandleMock<M>
+    where
+        M: Message + 'static + Clone,
+        A: Actor<Context = Context<A>> + Handler<M>,
+    {
+        fn notify<'a>(&'a self, msg: M, ctx: &'a mut Context<A>) {
             self.notify_params.lock().unwrap().push(msg.clone());
             if !cfg!(test) {
```
dnwiebe: ...again...

dnwiebe (approved):

dnwiebe (approved):

PR #122, opened 2022-04-28
GH-349: broadcasting to UIs about events through logs

bertllll:
Recreating this PR because it was contaminated by other works somehow.


USER-INTERFACE-INTERFACE.md
```diff
@@ -705,6 +705,32 @@ its database, in encrypted form, because it needs to withdraw money from the wal
...
+"logLevel": <String>
+```
+##### Description:
+This broadcast intends to give the user an immediate notification about a significant event that just occurred in
+the Node. At a bit lower level, this is a reaction to when the execution flow crosses any place with logging that bears
+a severity of `Info`, `Warn` or `Error`. Every such situation produces this broadcast along with the usual,
+long-adopted way of writing a log into a file and so both happen simultaneously. 
+
+Certain plans are to diverge between what is going to be printed to the log file and what is going to be given to
+the UI, while the letter in a prettier and user-friendlier form, but more preparation must go before we can implement
```
dnwiebe: latter instead of letter

masq/src/communications/broadcast_handler.rs
```diff
@@ -106,6 +107,8 @@ impl BroadcastHandlerReal {
                         stdout,
                         terminal_interface,
                     );
+                } else if let Ok((body, _)) = UiLogBroadcast::fmb(message_body.clone()) {
```
dnwiebe: It seems as though this will be 95% or 98% of the traffic here; if this parse were attempted first, then most of the time the others wouldn't need to be tried.

masq_lib/src/logger.rs
```diff
@@ -168,21 +242,403 @@ impl Logger {
...
+            let mut inner = self.received_messages.lock().unwrap();
+            inner.push(msg);
+            if inner.len() == self.expected_msg_count {
+                System::current().stop();
+            }
+        }
+    }
+
+    #[derive(Message)]
+    struct ScheduleStop {
```
dnwiebe: Consider using one message rather than two.

masq_lib/src/logger.rs
```diff
@@ -168,21 +242,403 @@ impl Logger {
...
+    fn overloading_function<C>(closure: C, msg: DoAllAtOnce)
+    where
+        C: Fn() + Send + 'static + Clone,
+    {
+        let barrier_arc = Arc::new(Barrier::new(msg.factor));
+        let mut join_handle_vector = Vec::new();
+        (0..msg.factor).for_each(|_| {
+            let barrier_arc_clone = Arc::clone(&barrier_arc);
+            let closure_clone = closure.clone();
+            join_handle_vector.push(thread::spawn(move || {
```
dnwiebe: We don't seem to be using this anywhere.

masq_lib/src/logger.rs
```diff
@@ -168,21 +242,403 @@ impl Logger {
...
+        let (template_before, template_after) = {
+            let before = SystemTime::now();
+            overloading_function(
+                move || {
+                    SENDER
+                        .lock()
+                        .unwrap()
+                        .as_ref()
+                        .unwrap()
+                        .clone()
```
dnwiebe: Is the clone() necessary?

masq_lib/src/logger.rs
```diff
@@ -168,21 +242,403 @@ impl Logger {
...
+            system.run();
+            let end = SystemTime::now();
+            (start, end)
+        };
+        let recording = recording_arc.lock().unwrap();
+        assert_eq!(recording.len(), number_of_sent_msgs_in_total);
+        let measured = actual_end.duration_since(actual_start).unwrap();
+        let safe_estimation = labour_time_example * 3;
+        eprintln!("measured {:?}, template {:?}", measured, safe_estimation);
+        assert!(measured < safe_estimation) //this should pass even on slow machines
```
dnwiebe: Can you put in a (fairly large) comment somewhere explaining the purpose of this test? I mean, the fact that it's comparing two intervals, for example, and exactly what those intervals are, and how they're measured.

masq_lib/src/test_utils/ui_connection.rs
```diff
@@ -124,10 +140,28 @@ impl UiConnection {
...
 
     pub fn shutdown(self) {
         self.client.shutdown().unwrap()
     }
 }
+
+pub enum ReceiveResult<T> {
+    Correct(T),
+    TransactionError((u64, String)),
+    MarshalError((NodeToUiMessage, UiMessageError, String)),
```
dnwiebe: You might not need the NodeToUiMessage here. (Or, you could use it; then you'd need it.)

node/src/actor_system_factory.rs
```diff
@@ -1172,6 +1236,7 @@ mod tests {
         };
         let make_params_arc = Arc::new(Mutex::new(vec![]));
         let mut subject = ActorSystemFactoryToolsReal::new();
+        subject.log_recipient_setter = Box::new(LogRecipientSetterNull::new());
```
dnwiebe: This is right on the border of being worth making a utility function for. Just a suggestion; you decide.

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -89,6 +90,7 @@ impl WebSocketSupervisorReal {
             next_client_id: 0,
             from_ui_message_sub,
             client_id_by_socket_addr: HashMap::new(),
+            socket_addr_by_client_id: Default::default(),
```
dnwiebe: These should both be either HashMap::new() or Default::default().

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -133,22 +135,40 @@ impl WebSocketSupervisorReal {
...
+
+    fn stc_errors(
+        errors: Vec<SendToClientWebsocketError>,
+        inner_arc: &Arc<Mutex<WebSocketSupervisorInner>>,
+    ) {
+        errors.into_iter().for_each(|e| match e {
+            SendToClientWebsocketError::FlushError((client_id, e)) => {
+                Self::handle_flush_error(e, inner_arc, client_id)
+            }
+            SendToClientWebsocketError::SendError(_) => (),
```
dnwiebe: Put in a comment here explaining why you're throwing away this error, even if it's a plaintive, helpless comment, like "I don't know what to do with this!"

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -388,23 +418,79 @@ impl WebSocketSupervisorReal {
...
+                        Ok(_) => None,
+                        Err(e) => Some(SendToClientWebsocketError::FlushError((client_id, e))),
+                    },
+                    Err(e) => {
+                        error!(
+                            Logger::new("WebSocketSupervisor"),
+                            "Error sending to client {}: {:?}", client_id, e
+                        );
+                        //TODO do we want to treat this error more in detail or to return None would be acceptable?
+                        Some(SendToClientWebsocketError::SendError((client_id, e)))
```
dnwiebe: If you can explore the client.send() code and conclude that SendError can't happen in real life in this context, then don't worry too much about handling it.  If it _could_ happen, but just _doesn't_ very much, then it should at least be logged.

dnwiebe (commented):
Continue starting with the tests for websocket_supervisor.rs.

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -539,6 +633,7 @@ mod tests {
         send_params: Arc<Mutex<Vec<OwnedMessage>>>,
         send_results: RefCell<Vec<Result<(), WebSocketError>>>,
         flush_results: RefCell<Vec<Result<(), WebSocketError>>>,
+        socket_results: RefCell<Vec<SocketAddr>>,
```
dnwiebe: This may not be necessary.

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -1027,16 +1174,85 @@ mod tests {
...
+        let inner = inner_arc.lock().unwrap();
+        assert_eq!(
+            inner.client_id_by_socket_addr.get(&socket_addr),
+            expected_assertion_result
+        );
+        assert_eq!(
+            inner.client_by_id.get(&1234).is_some(),
+            expected_assertion_result.is_some()
+        );
+    }
```
dnwiebe, line 1236: Do we need another assertion for the third HashMap?

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -1027,16 +1174,85 @@ mod tests {
         );
     }
 
+    fn flush_failure_test_body(
+        flush_error: WebSocketError,
+        expected_assertion_result: Option<&u64>,
```
dnwiebe: It seems as though this must always be either None or Some(&1234). If it were a boolean instead, not only would the domain of possible values be smaller and easier to understand, but the 1234 wouldn't be a tight coupling between test and utility function.

Possible name for boolean parameter: client_is_retained_after_error

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -1027,16 +1174,85 @@ mod tests {
...
+        );
+        assert_eq!(
+            inner.client_by_id.get(&1234).is_some(),
+            expected_assertion_result.is_some()
+        );
+    }
+
     #[test]
-    fn can_handle_flush_failure_after_send() {
+    fn can_handle_non_broken_pipe_flush_failure_after_send() {
```
dnwiebe: How about can_handle_non_fatal_flush_failure or something?

node/tests/ui_gateway_test.rs
```diff
@@ -68,3 +72,166 @@ fn ui_requests_something_and_gets_corresponding_response() {
...
+            read_buffer.clear();
+            MASQNode::wait_for_match_at_directory(
+                pattern_in_log(port_spec_ui).as_str(),
+                log_file_directory.as_path(),
+                &mut read_buffer,
+                Some(1500),
+            );
+            port_spec_ui.parse::<u16>().unwrap()
+        };
+    let assertion_lookup_pattern_1 = |port_spec_ui: &str| {
```
dnwiebe, line 167: Maybe reorder these definitions (reference the order in which they're used) to make them easier to understand?

node/tests/ui_gateway_test.rs
```diff
@@ -68,3 +72,166 @@ fn ui_requests_something_and_gets_corresponding_response() {
...
+            ("neighborhood-mode", Some("standard")),
+            ("log-level", Some("trace")),
+            ("data-directory", Some(&data_directory.to_str().unwrap())),
+        ]))
+        .unwrap();
+
+    let _: UiStartResponse = daemon_client.transact(UiStartOrder {}).unwrap();
+
+    let connect_and_disconnect_assertion =
+        |ordinary_number_of_the_occurrence: usize,
```
dnwiebe: I think this should be ordinal_number_of_the_occurrence.

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
+            limit_ms,
+        )
+        .unwrap()
+    }
+
+    fn wait_for_log_at_directory<T>(
+        pattern: &str,
+        path_to_logfile: &Path,
+        buffer: &mut String,
+        do_with_output: &dyn Fn(&String, &Regex) -> (Option<T>, bool),
```
dnwiebe: Can this closure ever return (Some(x), false) or (None, true)? If not, maybe just returning the Option is enough.

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
+            pattern,
+            logfile_path.as_ref(),
+            buffer,
+            &do_with_findings,
+            limit_ms,
+        )
+        .unwrap()
+    }
+
+    fn wait_for_log_at_directory<T>(
```
dnwiebe, line 219: Is there any way to give this a friendlier interface, or at least document it so that after a few months we'll still know how to use it?

!!!!!!!

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
+        buffer: &mut String,
+        groups: Vec<usize>,
+        required_number_of_captures: usize,
+        limit_ms: Option<u64>,
+    ) -> Vec<Vec<String>> {
+        let logfile_path = Self::path_to_logfile(logfile_dir);
+        let do_with_findings =
+            |buffer: &String, regex: &Regex| -> (Option<Vec<Vec<String>>>, bool) {
+                let captures = regex.captures_iter(&buffer[..]).collect::<Vec<Captures>>();
+                if captures.is_empty() {
```
dnwiebe: Looks like this isn't necessary

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
+                    return (None, false);
+                }
+                if captures.len() < required_number_of_captures {
+                    return (None, false);
+                }
+                let structured_captures = (0..captures.len())
```
dnwiebe: You might be able to call .flat_map() directly on captures, rather than making a range out of the length of captures, iterating on that, and using the result to index into captures.

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
+            &do_with_findings,
+            limit_ms,
+        )
+        .unwrap()
+    }
+
+    fn wait_for_log_at_directory<T>(
+        pattern: &str,
+        path_to_logfile: &Path,
+        buffer: &mut String,
```
dnwiebe: Possibly this buffer should be declared just outside the loop below, rather than being passed in here from the ultimate caller.

node/tests/utils.rs
```diff
@@ -159,16 +160,95 @@ impl MASQNode {
...
-                    if regex.is_match(&self.logfile_contents[..]) {
-                        break;
+                    buffer.clear();
+                    buffer.push_str(&contents);
+                    let (result_opt, finished) = do_with_output(&buffer, &regex);
+                    if finished {
+                        break result_opt;
                     }
                 }
                 Err(e) => {
```
dnwiebe, line 237: If the error here stays around regardless of how long you wait, then this is a potential infinite loop. Either it should immediately fail the test, or it should check the time and fail the test if the timeout has expired.

node/src/ui_gateway/websocket_supervisor.rs
bertllll: It's hard to find a clear answer. I have nothing with exact error examples because there are traits with so many impls and I got lost in there. However I found what I think could be sufficient for now even if it's quite general: 

_If the sink encounters an error other than being temporarily full, it
uses the `Err` variant to signal that error. In most cases, such errors
mean that the sink will permanently be unable to receive items._
    
The second sentence leads me to a thought that it may be allowed for me to simply discard that client the way I did with the flush error.

dnwiebe (approved):

PR #124, opened 2022-04-30
Finsaas gh cimatrix

FinsaasGH:


kauri-hero (approved):
Looks good - @dnwiebe @bertllll the change to the `ci-matrix.yml` modifies GH Actions to output artifacts containing ONLY the binary files
Hooray!

Second yml file detects a with a branch name prefixed by `releases/` and if all builds are successful it uploads those binaries into our AWS S3 bucket

PR #136, opened 2022-05-12
GH-574: Overall Connection Status of a Node with the MASQ Network

utkarshg6:


node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -560,14 +618,15 @@ impl GossipHandler for IntroductionHandler {
...
+        cpm_recipient: &Recipient<ConnectionProgressMessage>,
     ) -> GossipAcceptanceResult {
         if database.root().full_neighbor_keys(database).len() >= MAX_DEGREE {
             GossipAcceptanceResult::Ignored
         } else {
             let (introducer, introducee) = Self::identify_players(agrs, gossip_source)
                 .expect("Introduction not properly qualified");
             let introducer_key = introducer.inner.public_key.clone();
-            match self.update_database(database, cryptde, introducer) {
+            match self.update_database(database, cryptde, introducer.clone()) {
```
bertllll: There is a possibility not to have to clone it wholly. You could take just a copy of the ip address, e.g. creating it on a line before as a new variable, to use it in the connection progress message.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -576,6 +635,15 @@ impl GossipHandler for IntroductionHandler {
                     ));
                 }
             }
+            let connection_progess_message = ConnectionProgressMessage {
+                peer_addr: introducer.node_addr_opt.unwrap().ip_addr(),
```
bertllll: You need to avoid using unwrap() in production. Use expect with a decent small message instead please.
(I usually use expectv() but that's just my vain of an author.)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -576,6 +635,15 @@ impl GossipHandler for IntroductionHandler {
...
+                event: ConnectionProgressEvent::IntroductionGossipReceived(
+                    introducee.node_addr_opt.as_ref().unwrap().ip_addr(),
```
bertllll: Same here.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1991,6 +2214,165 @@ mod tests {
...
         );
+        System::current().stop();
+        assert_eq!(system.run(), 0);
+        let recording = recording_arc.lock().unwrap();
+        assert_eq!(recording.len(), 0);
+    }
+
+    #[test]
+    fn no_cpm_is_sent_in_case_full_neighborship_doesn_t_exist_and_cannot_be_created() {
+        // (false, false)
```
bertllll: Nobody will really know what this note was referring to. You can make it more obvious or just delete it.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1991,6 +2214,165 @@ mod tests {
...
+        let cryptde = main_cryptde();
+        let root_node = make_node_record(1111, true); // This is us
+        let mut root_db = db_from_node(&root_node);
+        // This node must not be neighbors with the root node.
+        let src_node = make_node_record(2222, true); // Full Neighbor
```
bertllll: The comment is misleading. That's why not writing comments is usually better.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1991,6 +2214,165 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn cpm_is_not_sent_in_case_full_neighborship_exists_and_is_destroyed() {
+        // (true, false)
+        let cryptde = main_cryptde();
+        let root_node = make_node_record(1111, true); // This is us
+        let mut root_db = db_from_node(&root_node);
+        let src_node = make_node_record(2222, true); // Full Neighbor
```
bertllll: I'd recommend to reconsider the use of comments like this one or if you want just remove it.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2540,27 +2922,218 @@ mod tests {
...
+                event: ConnectionProgressEvent::IntroductionGossipReceived(new_ip)
+            }
+        )
+    }
+
+    // 1. TCP ConnectionEstablished
+    // 2. Send the AskDebutMessage
+    // 3. Waiting Period [Introduction Gossip, Pass Gossip, Standard Gossip] - Update stage on the basis of gossip
+    // 4. Receive AskDebutMessage {handle whether gossip was received or not.}
+    // 5. Update the stage to Failed in case not received
```
bertllll: These notes may not be needed anymore.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2540,27 +2922,218 @@ mod tests {
...
+            PassHandler {
+                previous_pass_targets: RefCell::new(HashMap::new()),
+            }
+        );
+    }
+
     #[test]
     fn pass_is_properly_handled() {
+        // This test makes sure GossipAcceptor works correctly
+        // TODO: Make sure we test that PassHandler is called and not the others
```
bertllll: Probably, what I do is that I deleted the second line with the TODO only. I think the concern has been gone by now.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2540,27 +2922,218 @@ mod tests {
...
+        let final_timestamp = SystemTime::now();
+        match result {
+            GossipAcceptanceResult::Reply(_, _, _) => (),
+            other => panic!(
+                "Expected GossipAcceptanceResult::Reply but received {:?}",
+                other
+            ),
+        }
+        System::current().stop();
+        // System ran successfully
```
bertllll: Please, this is too much of information :)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2540,27 +2922,218 @@ mod tests {
...
+        let cryptde = main_cryptde();
+        let root_node = make_node_record(1234, true);
+        let mut db = db_from_node(&root_node);
+        let subject = PassHandler::new();
+        let (gossip, pass_target, gossip_source) = make_pass(2345);
+        let (cpm_recipient, recording_arc) = make_cpm_recipient();
+        let pass_target_ip_addr = pass_target.node_addr_opt().unwrap().ip_addr();
+        subject.previous_pass_targets.borrow_mut().insert(
+            pass_target_ip_addr,
+            SystemTime::now().sub(PASS_GOSSIP_EXPIRED_TIME.add(Duration::from_secs(1))),
```
bertllll: Maybe I'd recommend to create a variable, maybe even saying "excessive" or something to show clearly that the period of time is longer than PASS_GOSSIP_EXPIRED_TIME.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2774,7 +3348,7 @@ mod tests {
             .node(current_node.public_key(), false)
             .node(obsolete_node.public_key(), false)
             .build();
-        let subject = GossipAcceptorReal::new(main_cryptde());
+        let subject = make_subject(main_cryptde());
         let original_dest_db = dest_db.clone();
 
         let result = subject.handle(
```
bertllll, line 3343: This is not your fault but could you change the order of items in the first assertion. "result" belongs to the left.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2679,6 +3252,7 @@ mod tests {
         dest_db.add_arbitrary_full_neighbor(dest_node.public_key(), third_node.public_key());
         src_db.add_arbitrary_full_neighbor(dest_node.public_key(), third_node.public_key());
         src_db.add_arbitrary_full_neighbor(src_node.public_key(), third_node.public_key());
+        src_db.add_arbitrary_full_neighbor(src_node.public_key(), dest_node.public_key());
```
bertllll: I think this is wrong but cannot say why.
The sender probably shouldn't tell us we are a full neighbor if we had no knowledge a moment ago that the sender existed at all. We probably cannot jump from "no record" to "full neighbor" without tracking the node as a half neighbor before or sure, a full neighbor. :man_shrugging: 

Please reach the wise, Dan will help you.

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -1329,6 +1329,7 @@ mod tests {
 
     #[test]
     fn bad_dns_lookup_produces_log_and_sends_error_response() {
+        // TODO: This test fails sometimes for unknown reason, usually when a set of tests are run
```
bertllll: You can spend some time on recreating the error here but I don't want to force you. Feel free to leave the comments there and we may fix it at some other time...or...well...never really.

node/src/stream_handler_pool.rs
```diff
@@ -827,7 +858,7 @@ mod tests {
                 })
                 .unwrap();
 
-            sub_tx.send(subject_subs).expect("Internal Error");
+            sub_tx.send(subject_subs).expect("Neighborhood Unbound");
```
bertllll: Please change it to unwrap() it's more convenient.

node/src/sub_lib/neighborhood.rs
```diff
@@ -499,10 +526,27 @@ impl fmt::Display for GossipFailure_0v1 {
     }
 }
 
+pub struct NeighborhoodTools {
+    pub notify_later_ask_about_gossip:
+        Box<dyn NotifyLaterHandle<AskAboutDebutGossipMessage, Neighborhood>>,
+    // TODO: Should we change the above field to constant
```
bertllll: The comment actually should point to the field below. And it makes sense, you can create a constant meaning how long the interval for "ask about gossip" is. Then you can test the new constant in the test of all constants together.

node/src/sub_lib/neighborhood.rs
```diff
@@ -499,10 +526,27 @@ impl fmt::Display for GossipFailure_0v1 {
...
+    pub ask_about_gossip_interval: Duration,
+}
+
+impl Default for NeighborhoodTools {
+    fn default() -> Self {
+        Self {
+            notify_later_ask_about_gossip: Box::new(NotifyLaterHandleReal::new()),
+            ask_about_gossip_interval: Duration::from_secs(10),
```
bertllll: Here is the place where you'd put the constant. ;)

node/src/sub_lib/neighborhood.rs
```diff
@@ -531,6 +575,10 @@ mod tests {
                 exit_service_rate: 0,
             }
         );
+        assert_eq!(
+            NeighborhoodTools::default().ask_about_gossip_interval,
```
bertllll: Here you will be testing the constant itself instead of a field of a struct.

node/src/test_utils/neighborhood_test_utils.rs
```diff
@@ -274,3 +279,19 @@ impl From<&NodeRecord> for AccessibleGossipRecord {
...
+    (recipient, recording_arc)
+}
+
+pub fn make_node_to_ui_recipient() -> (Recipient<NodeToUiMessage>, Arc<Mutex<Recording>>) {
+    let (recorder, _, recording_arc) = make_recorder();
+    let addr = recorder.start();
+    let recipient = addr.recipient();
+
+    (recipient, recording_arc)
+}
```
bertllll, line 295: You could write:

```
fn make_recipient_and_recording<M:Message>()->(Recipient<M>,Arc<Mutex<Recording>)
{
    let (recorder, _, recording_arc) = make_recorder();
    let addr = recorder.start();
    let recipient = addr.recipient();

    (recipient, recording_arc)
}
```
Then you could stick it in two tiny functions like:
```
pub fn make_cpm_recipient() -> (Recipient<ConnectionProgressMessage>, Arc<Mutex<Recording>>) {
   make_recipient_and_recording()
}
```
...and the other, similarly...

bertllll (commented):
Remaining files to be reviewed:
node/src/neighborhood/mod.rs
node/src/neighborhood/overall_connection_status.rs

node/src/neighborhood/mod.rs
```diff
@@ -1565,14 +1626,500 @@ mod tests {
...
+        let (recipient, _) = make_node_to_ui_recipient();
+        subject.node_to_ui_recipient_opt = Some(recipient);
+        subject.tools.notify_later_ask_about_gossip = Box::new(
+            NotifyLaterHandleMock::default()
+                .notify_later_params(&notify_later_ask_about_gossip_params_arc),
+        );
+        subject.tools.ask_about_gossip_interval = Duration::from_millis(10);
+        let addr = subject.start();
+        let cpm_recipient = addr.clone().recipient();
+        let beginning_connection_progress = ConnectionProgress {
```
dnwiebe, line 1851: There seems to be a lot of stuff here in the Arrange section of the test that's really about asserting, and it might be clearer if it were moved into the Assert part of the test.

node/src/neighborhood/mod.rs
```diff
@@ -1565,14 +1626,500 @@ mod tests {
...
+            current_peer_addr: initial_desc_ip_addr,
+            connection_stage: ConnectionStage::TcpConnectionEstablished,
+        };
+        let addr = subject.start();
+        let recipient: Recipient<AskAboutDebutGossipMessage> = addr.clone().recipient();
+        let aadgrm = AskAboutDebutGossipMessage {
+            prev_connection_progress: beginning_connection_progress.clone(),
+        };
+        let system = System::new("testing");
+        let assertions = Box::new(move |actor: &mut Neighborhood| {
```
dnwiebe, line 1920: This is strictly assertion-related; it seems that it should be in the Assert part of the test rather than the Arrange part.

node/src/neighborhood/mod.rs
```diff
@@ -1565,14 +1626,500 @@ mod tests {
...
+                None,
+                "neighborhood_handles_connection_progress_message_with_tcp_connection_established",
+            ),
+        );
+        let (recipient, _) = make_node_to_ui_recipient();
+        subject.node_to_ui_recipient_opt = Some(recipient);
+        let addr = subject.start();
+        let cpm_recipient = addr.clone().recipient();
+        let system = System::new("testing");
+        let assertions = Box::new(move |actor: &mut Neighborhood| {
```
dnwiebe: It looks like this can also be moved down into the Assert section.

node/src/neighborhood/mod.rs
```diff
@@ -1565,14 +1626,500 @@ mod tests {
...
+        let cpm_recipient = addr.clone().recipient();
+        let system = System::new("testing");
+        // let new_pass_target = IpAddr::from_str("10.20.30.40").unwrap();
+        let assertions = Box::new(move |actor: &mut Neighborhood| {
+            assert_eq!(
+                actor.overall_connection_status.progress,
+                vec![ConnectionProgress {
+                    initial_node_descriptor: node_descriptor.clone(),
+                    current_peer_addr: node_ip_addr,
+                    connection_stage: ConnectionStage::Failed(DeadEndFound)
```
dnwiebe: "DeadEndFound" might be clearer as "PassLoopFound"...or at least something that mentioned pass loops.

node/src/neighborhood/mod.rs
```diff
@@ -1565,14 +1626,500 @@ mod tests {
...
+        subject.node_to_ui_recipient_opt = Some(recipient);
+        subject.overall_connection_status.update_connection_stage(
+            node_ip_addr,
+            ConnectionProgressEvent::TcpConnectionSuccessful,
+            subject.node_to_ui_recipient_opt.as_ref().unwrap(),
+        );
+        let addr = subject.start();
+        let cpm_recipient = addr.clone().recipient();
+        let system = System::new("testing");
+        // let new_pass_target = IpAddr::from_str("10.20.30.40").unwrap();
```
dnwiebe: Commented-out code

node/src/neighborhood/mod.rs
```diff
@@ -2648,11 +3195,9 @@ mod tests {
             payload: gossip.clone(),
             payload_len: 0,
         };
-        let system = System::new("");
+        let system = System::new("test");
         let addr: Addr<Neighborhood> = subject.start();
-        let peer_actors = peer_actors_builder().build();
-        addr.try_send(BindMessage { peer_actors }).unwrap();
-        let sub = addr.recipient::<ExpiredCoresPackage<Gossip_0v1>>();
+        let sub = addr.clone().recipient::<ExpiredCoresPackage<Gossip_0v1>>();
```
dnwiebe: Is this clone() necessary?

node/src/neighborhood/mod.rs
```diff
@@ -2847,10 +3392,13 @@ mod tests {
         let neighbor = make_node_record(1111, true);
         let mut subject: Neighborhood = neighborhood_from_nodes(&subject_node, Some(&neighbor));
         let replacement_database = subject.neighborhood_database.clone();
-        subject.gossip_acceptor = Box::new(DatabaseReplacementGossipAcceptor {
+        // TODO: Make sure we modify this test and didn't write a different one testing the same thing
```
dnwiebe: Leftover TODO

node/src/neighborhood/mod.rs
```diff
@@ -3533,15 +4082,16 @@ mod tests {
         let addr: Addr<Neighborhood> = subject.start();
         let peer_actors = peer_actors_builder().hopper(hopper).build();
         addr.try_send(BindMessage { peer_actors }).unwrap();
-
-        let sub = addr.recipient::<StartMessage>();
+        let sub = addr.clone().recipient::<StartMessage>();
```
dnwiebe: Unnecessary clone()?

node/src/neighborhood/mod.rs
```diff
@@ -4319,6 +4869,8 @@ mod tests {
         let mut subject = neighborhood_from_nodes(&root_node, Some(&neighbor_node));
         let persistent_config = PersistentConfigurationMock::new();
         subject.persistent_config_opt = Some(Box::new(persistent_config));
+        assert!(subject.gossip_acceptor_opt.is_none());
+        assert!(subject.gossip_producer_opt.is_none());
```
dnwiebe: Probably one of these isn't necessary, unless there's some side-effect stuff going on in the background, in which case we need either comments (suboptimal) or redesign (preferred)!

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+impl ConnectionProgress {
+    pub fn new(node_descriptor: &NodeDescriptor) -> Self {
+        let node_addr = node_descriptor.node_addr_opt.as_ref().unwrap_or_else(|| {
+            panic!(
+                "Unable to receive node addr for the descriptor {:?}",
+                node_descriptor
+            )
+        });
+        Self {
+            initial_node_descriptor: node_descriptor.clone(),
```
bertllll: I don't think this clone is necessary.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        let current_stage = usize::try_from(&self.connection_stage);
+        let new_stage = usize::try_from(&connection_stage);
+
+        if let (Ok(current_stage_num), Ok(new_stage_num)) = (current_stage, new_stage) {
+            if new_stage_num != current_stage_num + 1 {
+                panic!(
+                    "Can't update the stage from {:?} to {:?}",
+                    self.connection_stage, connection_stage
+                )
+            }
```
bertllll, line 79: Please consider to refactor this place by removing the sequential check because it has mostly a value for the developer not for the software itself in order to work properly. It's kind of code which otherwise would be found in a test but here it's checked at runtime of the production code. It's usually a bad decision. 

That would also allow you to get rid of the TryFrom<usize> implementation which has just a very shallow purpose even now and then it would be useless absolutely.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        self.connection_stage = ConnectionStage::StageZero;
+        self.current_peer_addr = new_pass_target;
+    }
+}
+
+#[derive(PartialEq, Debug, Copy, Clone)]
+enum OverallConnectionStage {
+    NotConnected = 0,        // Not connected to any neighbor.
+    ConnectedToNeighbor = 1, // Neighborship established. Same as No 3 hops route found.
+    ThreeHopsRouteFound = 2, // check_connectedness() returned true, data can now be relayed.
```
bertllll: Please check that you have a single test for what numbers are linked to which variant of the enum. Then research whether those numbers couldn't be eliminated in the declaration because I think enums have a special property that they can automatically be converted into a number without explicitly defining the numbers. But only if you want a sequence like 0,1,2... of course.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        };
+
+        self.connection_stage = ConnectionStage::StageZero;
+        self.current_peer_addr = new_pass_target;
+    }
+}
+
+#[derive(PartialEq, Debug, Copy, Clone)]
+enum OverallConnectionStage {
+    NotConnected = 0,        // Not connected to any neighbor.
```
bertllll: Consider removing the comments. The meaning is quite obvious at the enum.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+
+#[derive(PartialEq, Debug)]
+pub struct OverallConnectionStatus {
+    // Becomes true iff three hops route was found.
+    can_make_routes: bool,
+    // Stores one of the three stages of enum OverallConnectionStage.
+    stage: OverallConnectionStage,
+    // Corresponds to progress with node descriptors entered by the user
+    pub progress: Vec<ConnectionProgress>,
+}
```
bertllll, line 138: I think all three comments could easily go away. The programmer will figure it out from the context and names.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    }
+
+    pub fn get_connection_progress_by_ip(&mut self, peer_addr: IpAddr) -> &mut ConnectionProgress {
+        let connection_progress_to_modify = self
+            .progress
+            .iter_mut()
+            .find(|connection_progress| connection_progress.current_peer_addr == peer_addr)
+            .unwrap_or_else(|| {
+                panic!(
+                    "Unable to find the node in connections with IP Address: {}",
```
bertllll: Maybe I wouldn't say node with a small case n. Or some other formulation?

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    ) -> &ConnectionProgress {
+        let connection_progress = self
+            .progress
+            .iter()
+            .find(|connection_progress| {
+                &connection_progress.initial_node_descriptor == initial_node_descriptor
+            })
+            .unwrap_or_else(|| {
+                panic!(
+                    "Unable to find the node in connections with Node Descriptor: {:?}",
```
bertllll: Maybe here too?

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    }
+
+    pub fn update_connection_stage(
+        &mut self,
+        peer_addr: IpAddr,
+        event: ConnectionProgressEvent,
+        node_to_ui_recipient: &Recipient<NodeToUiMessage>,
+    ) {
+        let connection_progress_to_modify = self.get_connection_progress_by_ip(peer_addr);
+
```
bertllll, line 201: This is not mandatory but you could create a closure here to spare some space. 

let modify_connection_progress = | stage: ConnectionStage| connection_progress_to_modify.update_stage(stage); 

```
match event{
     ConnectionProgressEvent::TcpConnectionSuccessful  => modify_connection_progress(ConnectionStage::TcpConnectionEstablished),
```
     
You also can reimport ConnectionProgressEvent in your file like:
`use self::ConnectionProgressEvent::{TcpConnectionEstablished, TcpConnectionFailed,...etc.};     `

Then you can write the match statement as:

```
match event {
     TcpConnectionEstablished => <expression>,
```

It would also make the lines shorter and easier to read.

But again, it's not mandatory...you could try and decide by your liking

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        assert_eq!(subject.is_empty(), false);
+    }
+
+    #[test]
+    fn can_receive_mut_ref_of_connection_progress_from_peer_addr() {
+        let peer_1_ip = IpAddr::from_str("1.2.3.4").unwrap();
+        let peer_2_ip = IpAddr::from_str("5.6.7.8").unwrap();
+        let peer_3_ip = IpAddr::from_str("9.0.1.2").unwrap();
+
```
bertllll, line 424: blank lines

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        let desc_1 = make_node_descriptor_from_ip(peer_1_ip);
+        let desc_2 = make_node_descriptor_from_ip(peer_2_ip);
+        let desc_3 = make_node_descriptor_from_ip(peer_3_ip);
+
+        let initial_node_descriptors = vec![desc_1.clone(), desc_2.clone(), desc_3.clone()];
```
bertllll: Consider to go with just two instances. It would look more compact.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+            &mut ConnectionProgress::new(&desc_3)
+        );
+    }
+
+    #[test]
+    fn can_receive_connection_progress_from_initial_node_desc() {
+        let desc_1 = make_node_descriptor_from_ip(IpAddr::from_str("1.2.3.4").unwrap());
+        let desc_2 = make_node_descriptor_from_ip(IpAddr::from_str("5.6.7.8").unwrap());
+        let desc_3 = make_node_descriptor_from_ip(IpAddr::from_str("9.0.1.2").unwrap());
+
```
bertllll, line 457: As above?

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    #[test]
+    fn remove_deletes_descriptor_s_progress_and_returns_node_descriptor() {
+        let node_desc_1 = make_node_descriptor_from_ip(IpAddr::from_str("1.2.3.4").unwrap());
+        let node_desc_2 = make_node_descriptor_from_ip(IpAddr::from_str("1.2.3.5").unwrap());
+        let initial_node_descriptors = vec![node_desc_1.clone(), node_desc_2.clone()];
+        let mut subject = OverallConnectionStatus::new(initial_node_descriptors);
+
+        let removed_desc_1 = subject.remove(0);
+        let removed_desc_2 = subject.remove(0);
```
bertllll: Testing just one remove would suffice.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    #[test]
+    fn updates_the_connection_stage_to_tcp_connection_established() {
+        let node_ip_addr = IpAddr::from_str("1.2.3.4").unwrap();
+        let node_descriptor = make_node_descriptor_from_ip(node_ip_addr);
+        let initial_node_descriptors = vec![node_descriptor.clone()];
+        let (recipient, _) = make_node_to_ui_recipient();
+        let mut subject = OverallConnectionStatus::new(initial_node_descriptors);
+
+        subject.update_connection_stage(
+            node_descriptor.node_addr_opt.as_ref().unwrap().ip_addr(),
```
bertllll: You can also use your already created variable node_ip_addr.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+    #[test]
+    fn updates_the_connection_stage_to_failed_when_tcp_connection_fails() {
+        let node_ip_addr = IpAddr::from_str("1.2.3.4").unwrap();
+        let node_descriptor = make_node_descriptor_from_ip(node_ip_addr);
+        let initial_node_descriptors = vec![node_descriptor.clone()];
+        let (recipient, _) = make_node_to_ui_recipient();
+        let mut subject = OverallConnectionStatus::new(initial_node_descriptors);
+
+        subject.update_connection_stage(
+            node_ip_addr.clone(),
```
bertllll: This is actually Copy.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+            &recipient,
+        );
+
+        assert_eq!(
+            subject,
+            OverallConnectionStatus {
+                can_make_routes: false,
+                stage: OverallConnectionStage::NotConnected,
+                progress: vec![ConnectionProgress {
+                    initial_node_descriptor: node_descriptor.clone(),
```
bertllll: The .clone() call shouldn't be necessary.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+            &recipient,
+        );
+
+        assert_eq!(
+            subject,
+            OverallConnectionStatus {
+                can_make_routes: false,
+                stage: OverallConnectionStage::ConnectedToNeighbor,
+                progress: vec![ConnectionProgress {
+                    initial_node_descriptor: node_descriptor.clone(),
```
bertllll: Here too.

And then at couple of other tests. Just check them over.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+        )
+    }
+
+    #[test]
+    fn updates_the_connection_stage_to_neighborship_established() {
+        let node_ip_addr = IpAddr::from_str("1.2.3.4").unwrap();
+        let node_descriptor = make_node_descriptor_from_ip(node_ip_addr);
+        let initial_node_descriptors = vec![node_descriptor.clone()];
+        let mut subject = OverallConnectionStatus::new(initial_node_descriptors);
+        let new_node_ip_addr: IpAddr = Ipv4Addr::new(5, 6, 7, 8).into();
```
bertllll: It may look nicer if you create the IpAddr the same way each time.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+                    initial_node_descriptor: node_descriptor.clone(),
+                    current_peer_addr: node_ip_addr,
+                    connection_stage: ConnectionStage::NeighborshipEstablished
+                }],
+            }
+        )
+    }
+
+    #[test]
+    fn updates_the_connection_stage_to_neighborship_established_when_standard_gossip_is_received() {
```
bertllll, line 602: You may think of how to parametrize some of these tests (if you found just a small number of arguments needed to be supplied because of specifics of the given test), using common parts, reused.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+                can_make_routes: false,
+                stage: OverallConnectionStage::NotConnected,
+                progress: vec![ConnectionProgress {
+                    initial_node_descriptor: node_descriptor.clone(),
+                    current_peer_addr: new_node_ip_addr,
+                    connection_stage: ConnectionStage::StageZero
+                }],
+            }
+        )
+    }
```
bertllll, line 661: We probably lack a test proving that different tentacles cannot overwrite a more advanced stage achieved by the other one. I can imagine a threat with the reversed direction caused by answering a new pass gossip. That stimulates a step back to StageZero. Hopefully it's not possible but we also need a test making it obvious.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+
+    #[test]
+    #[should_panic(expected = "Unable to find the node in connections with IP Address: 5.6.7.8")]
+    fn panics_at_updating_the_connection_stage_if_a_node_is_not_a_part_of_connections() {
+        let node_ip_addr = IpAddr::from_str("1.2.3.4").unwrap();
+        let node_descriptor = make_node_descriptor_from_ip(node_ip_addr);
+        let initial_node_descriptors = vec![node_descriptor.clone()];
```
bertllll: This clone...

This shows up repeatedly at lines of the same purpose but in other tests.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+
+        subject.update_connection_stage(
+            node_ip_addr,
+            ConnectionProgressEvent::IntroductionGossipReceived(new_node_ip_addr),
+            &recipient,
+        );
+    }
+
+    #[test]
+    fn getter_fn_exists_for_boolean_can_make_routes() {
```
bertllll: maybe like "we_can_ask_about_can_make_routes()"? Just a suggestion...avoiding the phrase "getter_fn..."

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+                    initial_node_descriptor: node_descriptor.clone(),
+                    current_peer_addr: node_ip_addr,
+                    connection_stage: ConnectionStage::Failed(DeadEndFound)
+                }],
+            }
+        )
+    }
+
+    #[test]
+    fn updates_connection_stage_to_failed_when_no_gossip_response_is_received() {
```
bertllll, line 696: This part is frequently seen over here:

 ```
   let node_ip_addr = IpAddr::from_str("1.2.3.4").unwrap();
        let node_descriptor = make_node_descriptor_from_ip(node_ip_addr);
        let initial_node_descriptors = vec![node_descriptor.clone()];
        let mut subject = OverallConnectionStatus::new(initial_node_descriptors);
        let (recipient, _) = make_node_to_ui_recipient();
```
I believe you could put it inside an outboard method, a utility function.

File: node/src/neighborhood/overall_connection_status.rs [hunk 50]
```diff
@@ -0,0 +1,960 @@
...
+        let (recipient, _) = make_node_to_ui_recipient();
+        subject.update_can_make_routes(true);
+
+        subject.update_stage_of_overall_connection_status(&recipient);
+
+        assert_eq!(subject.stage, OverallConnectionStage::ThreeHopsRouteFound);
+    }
+
+    #[test]
+    fn updates_the_stage_to_three_hops_route_found_in_case_introduction_gossip_is_received_and_flag_is_true(
```
bertllll: These two tests below can be easily parametrized. That's one point. But we think we'd rather see four tests, one for each of four possible states in the relation between: can_make_routes (bolean) X ConnectionProgressEvent::{IntroductionGossipReceived, StandardGossipReceived}.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -0,0 +1,960 @@
...
+                body: UiConnectionChangeBroadcast {
+                    stage: UiConnectionChangeStage::ConnectedToNeighbor
+                }
+                .tmb(0)
+            }
+        );
+    }
+
+    #[test]
+    fn doesn_t_send_message_to_the_ui_in_case_gossip_is_received_but_stage_hasn_t_updated() {
```
bertllll: This one and the test below could be easily and heavily parametrized. Here I see it very straight. 

By a swift glimpse, this could take only two parameters, one is the stage in the subject, the second is the ConnectionProgressEvent passed into the act of the test.     

You may create variables for the two parameters to have a chance to give them names. The names could reveal what they stand for to the reader. Because a picture of a function with two inlined parameters without any further explanation is not so encouraging for me in the question of benefits of this solution.

node/src/proxy_client/stream_handler_pool.rs
utkarshg6: I wish we do it someday, I'm leaving it as it is for right now.

node/src/test_utils/neighborhood_test_utils.rs
utkarshg6: This is what the Rust compiler is accepting in the end:

```rust
pub fn make_recipient_and_recording_arc<M: 'static>() -> (Recipient<M>, Arc<Mutex<Recording>>)
where
    M: Message + std::marker::Send,
    <M as actix::Message>::Result: std::marker::Send,
    recorder::Recorder: actix::Handler<M>,
{
    let (recorder, _, recording_arc) = make_recorder();
    let addr = recorder.start();
    let recipient = addr.recipient::<M>();

    (recipient, recording_arc)
}

pub fn make_cpm_recipient() -> (Recipient<ConnectionProgressMessage>, Arc<Mutex<Recording>>) {
    make_recipient_and_recording_arc()
}

pub fn make_node_to_ui_recipient() -> (Recipient<NodeToUiMessage>, Arc<Mutex<Recording>>) {
    make_recipient_and_recording_arc()
}
```

Cons:
- The bounds are complex
- There is a `'static` lifetime

node/src/neighborhood/overall_connection_status.rs
utkarshg6: This one is required, because we can't move the value.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I can see the following two uses for this code:

1) It makes sure that transitions will only happen for the advancing stage.
2) It makes sure that we're not skipping any step, for example, going from `StageZero` to `NeighborshipEstablished` and skipping `TCPConnectionEstablished`.

According to me, it would be better if we keep it for now, and maybe add a TODO, saying

 "While working on regressing transitions, modify this function to allow both advancing and regressing transitions and also make sure no state is skipped during a transition"
 
 I would prefer if we open up this discussion, once the work on regressing transitions is completed, and we find these checks rudimentary at that point of time.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I searched for it, and every solution involves to explicitly enter the numbers.

Right now, we have a similar test, and I think since we only have three stages, this test is good enough.

```rust
    #[test]
    fn overall_connection_stage_can_be_converted_into_usize_and_can_be_compared() {
        assert!(
            OverallConnectionStage::ConnectedToNeighbor as usize
                > OverallConnectionStage::NotConnected as usize
        );
        assert!(
            OverallConnectionStage::ThreeHopsRouteFound as usize
                > OverallConnectionStage::ConnectedToNeighbor as usize
        );
    }
```

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I have updated the comments, I believe the comments have potential to revive our memory in the late future, and will help us and a new developer to understand "How the implementation works?".


I don't have enough courage to remove these comments as I believe that the way code is written, everything may seem obvious to now, may not be in the future and definitely not to a new developer.

I truly think, these comments offer a helping hand, when someone is reading this piece of code.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I believe the same is true for these comments too.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I think it would be insufficient, we need to prove that the `remove()` works in the order First In, First Out when the `0` is passed as an argument.

node/src/test_utils/neighborhood_test_utils.rs
bertllll: In general I don't mind to have those boundaries specified if they are needed. You can make the boundaries briefer like this:
 M: Message + Send,
 <M as actix::Message>::Result: Send,
'static

To be honest I don't understand this kind of boundary so please inspect if it is necessary...by temporarily removing it.    
Recorder: actix::Handler<M>,

node/src/neighborhood/overall_connection_status.rs
bertllll: You could, I guess, create just an instance of IpAddr instead of the whole NodeAddr because one is Copy and the second is just Clone.

You can see that we later need only IpAddr so you can grab less than you do now, it's unnecessarily too a big piece of data.  

Changing...
` let node_addr = node_descriptor.node_addr_opt.as_ref().unwrap_or_else(|| {...`
 
 into...

` let ip_addr = node_descriptor.node_addr_opt.as_ref().unwrap_or_else(|| {blabla}).ip_addr();`

node/src/neighborhood/overall_connection_status.rs
bertllll: I showed you this as an option to avoid a runtime check by moving the procedure to tests and making it their responsibility to guarantee that we cannot take two steps instead of one. 

It would probably be some kind of enum:
    Up,
    Down,
    ZeroStage,
    Failure

Where I think you can test all these "orders", which would be supplied to your function, and the exact outcome of what the order caused. You could prove like that going by one step always when you call Up or Down. 
At special cases, though, you will need something more powerful and that's why there are those other options allowing longer movement. 

Also I would add that this solution would allow eliminating your specialized case of `.handle_pass_gossip()` which is structurally different from those respective cases in `.update_connection_stage()`. You could use `.update_stage` also there. 

Please make some considerably small TODO out of these notes so that we can get back to it when implementing the regression.

node/src/neighborhood/overall_connection_status.rs
bertllll: Ok, I will leave you in peace. :))

node/src/neighborhood/overall_connection_status.rs
bertllll: Let's do it in your favor. But you know it works the other way too :D

node/src/neighborhood/overall_connection_status.rs
bertllll: I will look forward to the second review, you said it should be different then.

node/src/neighborhood/overall_connection_status.rs
bertllll: I still think, you may disagree though, that just one line would be enough. With one difference, 
`let removed_desc_1 = subject.remove(0);` could be rather `let removed_desc = subject.remove(1);`, proving that the index matters. With your assertion, you'll show that you got the right descriptor back and the first can be ignored in the test regarding assertions.
BUT the test would be stronger if you also assert that subject.is_empty() is false to prove that .`remove(number)` did not erase everything.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: Even after creating a new variable for `ip_addr`, it is not allowing me to move the `node_descriptor`. I guess, we'll have to keep a clone over here, I don't think we want to necessarily own the variable.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: A new utility fn is added `make_node_and_recipient()`.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: I don't think it's worth the struggle. Even though parameterising will make the size of these functions short, it would encapsulate the assertions, and will work for only two tests.

I agree that the differences are minor and it's possible to parameterise, but I believe that the trade off isn't worth it.

What I would recommend is to minimise the possible lines and not to go with the parameterisation.

node/src/neighborhood/overall_connection_status.rs
utkarshg6: This fn is now the owner of node_descriptor.

dnwiebe (approved):

PR #138, opened 2022-05-13
Docs/contributing plus cleanup

FinsaasGH:
Having an example workflow for dev vs repo admin may be helpful. The steps themselves are explained very clearly, only the responsibility of each party is unclear.

Dev: create new branch from master (GH-600 etc) -> PR completed card to master -> discuss with reviewer and/or repo admin if the merge warrants a version bump -> branch is merged into master

Repo admin: review/discuss card with dev and or reviewer -> if merge warrants version bump, create git tag with next version and instruct dev to increment cargo.tomls(if no version bump required, create a pre-release tag) -> complete merge and apply appropriate tag to merge commit


PRACTICE.md
```diff
@@ -157,5 +157,52 @@ Keep close track of both of these processes so that you can answer any questions
...
+
+Versioning follows basic [semver format](https://semver.org) - view the formal documentation and conventions for reference.
+
+The core specifications to be followed are:
+
+- Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version.
+- A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.–.
+
+### Steps Followed for Version Increment
```
FinsaasGH: Do we want to have non-release candidates tagged or just leave them as normal merges into master? Ex 3 cards merge into master, not one of the cards warrants a release by itself. Once the last card is going through PR can the cargo.tomls be incremented for that branch and tag the commit for release.
This will give a cleaner repo vs having additional pre-release.x tags, while also simplifying QA within React-app using "latest" binaries. If we want to start using versioned workflows within React-app this will require non-trivial testing.

kauri-hero: Well I guess the main reason for the tagging of non-release candidates is that its changing the master codebase each time.
so lets say something does need to be identified or rolled back (or even tested) then its easy to checkout the git tag of the master branch because every merge had a git tag, instead of searching based on commits and commit SHA.

But to be clear I didn't intend on the cargo.tomls being edited each time with the pre-release.x versions, its only for the git tag.

But to be honest this can be managed either way with clear and concise communication within the devs team and whoever is conducting the merges. Since master branch is protected it could be enforced by maintainers, which current are Dan, Bert and I.
before any merge this could be checked

PRACTICE.md
```diff
@@ -157,5 +157,52 @@ Keep close track of both of these processes so that you can answer any questions
...
+The determination of versioning will start with a discussion of the core developers and the Product Owner. This will begin towards the end of a cards engineering practice above, most likely during final review and QA steps.
+
+Versioning follows basic [semver format](https://semver.org) - view the formal documentation and conventions for reference.
+
+The core specifications to be followed are:
+
+- Once a versioned package has been released, the contents of that version MUST NOT be modified. Any modifications MUST be released as a new version.
+- A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.–.
+
```
bertllll, line 209: Why do we talk here about some general ways of the format if we later choose and talk just about our specific way. I find that a bit distracting if we put information here which we don't need to follow. I'd appreciate if it were more about us and less about all possibilities out there.

node/src/ui_gateway/README.md
```diff
@@ -5,8 +5,10 @@ User-interface endpoint for MASQ Node
 The purpose of `ui_gateway` is to provide a WebSockets server to which user-interface elements can 
 connect, and a sink for commands and source for status data from and to such elements.
 
-It probably isn't the most interesting place to begin digging into our code;
-[node](https://github.com/MASQ-Project/Node/tree/master/node)
-is a better place to start.
+### Documentation
```
bertllll: This is a bit misleading. The referred document is not to describe UIGateway (It speaks in detail of messages traveling between certain two points.), however you mention it in a read-me originally belonging only to this actor. 

Maybe "Further documentation, or Associated documentation..."?

PRACTICE.md
```diff
@@ -157,5 +157,52 @@ Keep close track of both of these processes so that you can answer any questions
...
+### Steps Followed for Version Increment
+Once the discussion around the development/feature card has completed, there will be a git tag added to the merge commit of the branch when it is merged to `master` by an admin user of the public repo.
+If the merged card is not determined to be included in an immediate release with version increment, then the git tag will be the current version number, denoted with `pre-release.1` or another suitable qualifier following the patch number, eg `v0.6.0-pre-release.1`. It may become common, that there are several cards being merged into `master` under pre-releases before the next version increment is formally tagged and released. In this case the git tag will be incremented with a dot separated number that increases each time, so there is never a duplicate git tag.
+
+If a merging card qualifies as a patch or minor release version, then prior to merge the developer will add a commit for review that adjusts all the version numbers in the `cargo.toml` files for all the major components within the codebase. This will be important for monitoring QA testing and logging of issues with different release versions in the wild.
```
bertllll: I think we rather want to adjust that with the final PR within the release period. Or put differently, with the last branch merged in the release. 

We certainly don't want to change these toml files with every minor commit. 

This time, by chance, I know Dan's already changed the cargo versioning in his card, merged by now. So even though it may have been considered in fact too early it doesn't harm anything really much. We will just not have to do it anymore for this upcoming release.

PRACTICE.md
```diff
@@ -157,5 +157,52 @@ Keep close track of both of these processes so that you can answer any questions
...
+
+### Steps Followed for Version Increment
+Once the discussion around the development/feature card has completed, there will be a git tag added to the merge commit of the branch when it is merged to `master` by an admin user of the public repo.
+If the merged card is not determined to be included in an immediate release with version increment, then the git tag will be the current version number, denoted with `pre-release.1` or another suitable qualifier following the patch number, eg `v0.6.0-pre-release.1`. It may become common, that there are several cards being merged into `master` under pre-releases before the next version increment is formally tagged and released. In this case the git tag will be incremented with a dot separated number that increases each time, so there is never a duplicate git tag.
```
bertllll: What I hoped for that I would read in this documentation is some sort of show of particular steps that you need to take in order to create the tag.

I thought a bit about it now and although I don't have very good knowledge of that I'd add something like this:

"About technical aspects of the tag. We think we need only the so-called lightweight tag, that is without an extend in the form of annotation. A possibly right way of taking care of the release tag is first to go to acquire an agreement from the product owner that your branch (in your open PR) is going to enclose the current, pending release. If this is the case you want to make sure your code in your branch appears errorless in Actions (it's passed all checks) and also it's satisfied all reviews and so is approved by the reviewer. 

At that point you create a tag to the HEAD commit of your branch with the previously agreed version number. 
If you do that locally on your machine you may want to use e.g.:

```
git tag v0.6.1
git push origin v0.6.1
```

The last step should cause a new trigger of a run in Actions, this time with a different workflow than usual, producing official release binaries at the end. 
 
Beside that, while you know the tag is in its place you're finally ready to let the merge button for your PR to master be pressed."

Maybe somewhere here I'm not accurate or even wrong in the steps so please think it over together with Finsaas. ;)

PRACTICE.md
kauri-hero: fair comment I'll make it specific for us

PRACTICE.md
kauri-hero: bert had a good conversation on this also @FinsaasGH 

we will forgo git tagging every commit, and only the commit to master that will have a release candidate.
if it needs to be a prerelease at that time, we can git tag it and have the release package labelled prerelease for testing if needed

PRACTICE.md
kauri-hero: I have adjusted this and finsaas is aligned with this approach as well

PRACTICE.md
kauri-hero: great I have added this with a few edits to your outline

Thanks @bertllll !

node/src/ui_gateway/README.md
kauri-hero: I have adjusted this now - thanks

bertllll (approved):

PR #150, opened 2022-07-04
GH-574: Review 5

utkarshg6:


masq_lib/src/logger.rs
```diff
@@ -381,25 +367,32 @@ mod tests {
...
-            .for_each(|handle| handle.join().unwrap());
+            .for_each(|handle| handle.join().unwrap())
     }
 
     lazy_static! {
         static ref SENDER: Mutex<Option<Sender<NodeToUiMessage>>> = Mutex::new(None);
     }
 
     #[test]
+    #[ignore]
```
bertllll: This is a step out. You need to delete the `ignore` attribute. As I remember the test worked okay on your machine in the end, right? If you remove it and it fails in Actions, consider to make the sleep between sending messages even a bit larger.

masq_lib/src/logger.rs
```diff
@@ -442,48 +434,56 @@ mod tests {
...
 
     fn prepare_test_environment<'a>() -> MutexGuard<'a, ()> {
-        let guard = TEST_LOG_RECIPIENT_GUARD.lock().unwrap();
-        LOG_RECIPIENT_OPT.lock().unwrap().take();
+        let guard = TEST_LOG_RECIPIENT_GUARD
+            .lock()
+            .expect("Unable to lock TEST_LOG_RECIPIENT_GUARD");
+        LOG_RECIPIENT_OPT
+            .lock()
+            .expect("Unable to lock LOG_ReCIPIENT_OPT")
```
bertllll: This is almost unworthy of mentioning but the lettering seems too clownish :)).

node/src/neighborhood/gossip.rs
```diff
@@ -97,37 +98,61 @@ impl TryFrom<&Value> for GossipNodeRecord {
...
         match NodeRecordInner_0v1::try_from(self) {
             Ok(nri) => {
-                human_readable.push_str("\n\tinner: NodeRecordInner_0v1 {");
-                human_readable.push_str(&format!("\n\t\tpublic_key: {:?},", &nri.public_key));
-                human_readable.push_str(&format!("\n\t\tnode_addr_opt: {:?},", self.node_addr_opt));
-                human_readable
-                    .push_str(&format!("\n\t\tearning_wallet: {:?},", nri.earning_wallet));
-                human_readable.push_str(&format!("\n\t\trate_pack: {:?},", nri.rate_pack));
-                human_readable.push_str(&format!(
+                write!(human_readable, "\n\tinner: NodeRecordInner_0v1 {{").expect("write failed");
```
bertllll, line 104: Sorry if you feel it otherwise but I myself would write a macro similar to short_writeln!() and replace these.

Well, or another option is that you would make a tiny card about it.

utkarshg6: I believe it's better to have a separate card for this. So, I created this new card - https://github.com/MASQ-Project/Node-issues/issues/615

PR #155, opened 2022-07-27
617 Optimize Routing Engine

dnwiebe:


node/src/neighborhood/mod.rs
```diff
@@ -1021,6 +997,32 @@ impl Neighborhood {
...
+                node_record.inner.rate_pack.routing_service_rate,
+            ),
+            LinkType::Exit => (
+                node_record.inner.rate_pack.exit_byte_rate,
+                node_record.inner.rate_pack.exit_service_rate,
+            ),
+            LinkType::Origin => (0, 0),
+        };
+        let rate_undesirability = per_cores as i64 + (per_byte as i64 * payload_size as i64);
+        let desirable_undesirability = if node_record.metadata.desirable {
```
bertllll: Make sure the desirable flag shouldn't be evaluated only for an exit Node.

node/src/neighborhood/mod.rs
```diff
@@ -1041,19 +1043,51 @@ impl Neighborhood {
...
+            &mut minimum_undesirability,
+        )
+        .into_iter()
+        .filter(|cr| cr.undesirability <= minimum_undesirability)
+        .map(|cr| cr.nodes)
+        .next()
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn complete_routes_inner<'a>(
```
bertllll: Maybe there is an alternative name without the word inner.

node/src/neighborhood/mod.rs
```diff
@@ -2429,41 +2426,122 @@ mod tests {
...
+        let before = Instant::now();
+
+        // All the target-designated routes from L to N
+        let route = subject
+            .best_route_segment(&l, Some(&n), 3, 10000, RouteDirection::Back)
+            .unwrap();
+
+        let after = Instant::now();
+        let northern_route = vec![&l, &g, &h, &i, &n];
+        let southern_route = vec![&l, &q, &r, &s, &n];
```
bertllll: We are no more deterministic and can say exactly which route will win.

node/src/neighborhood/mod.rs
```diff
@@ -2487,10 +2565,60 @@ mod tests {
...
+        node_record.inner.rate_pack = RatePack {
+            routing_byte_rate: 123456,
+            routing_service_rate: 234567,
+            exit_byte_rate: 345678,
+            exit_service_rate: 456789,
+        };
+
+        let result = subject.compute_undesirability(&node_record, 10000, LinkType::Origin);
+
+        assert_eq!(result, 0);
```
bertllll, line 2616: I think there should be a test that proves the steady undesirability gotten by a Node with the desirable flag as false.

node/src/neighborhood/mod.rs
```diff
@@ -4051,6 +4182,79 @@ mod tests {
...
+                target_component: Component::ProxyClient,
+                minimum_hop_count: 2,
+                return_component_opt: Some(Component::ProxyServer),
+                payload_size,
+            })
+            .unwrap();
+
+        let (over, back) = match response.expected_services {
+            ExpectedServices::OneWay(_) => panic!("Expecting RoundTrip"),
+            ExpectedServices::RoundTrip(o, b, _) => (o[1].clone(), b[1].clone()),
```
bertllll, line 4259: You can try to take those route segments by a reference, owning them might be unnecessary.

bertllll (approved):

node/src/neighborhood/mod.rs
dnwiebe: It _should_ be evaluated only for an exit Node. Nice catch.

node/src/neighborhood/mod.rs
dnwiebe: I fiddled around with them until I lost patience, but I couldn't find a pattern that worked. So I left them.

PR #156, opened 2022-07-28
GH-613: Combined ServiceConsumedMessages

bertllll:


node/src/proxy_server/mod.rs
```diff
@@ -717,56 +705,42 @@ impl ProxyServer {
...
+    ) -> Vec<RoutingServiceConsumed> {
+        let report_of_routing_services: Vec<RoutingServiceConsumed> = expected_services
             .iter()
             .filter_map(|service| match service {
                 ExpectedService::Routing(_, earning_wallet, rate_pack) => {
                     Some((earning_wallet, rate_pack))
                 }
                 _ => None,
             })
+            .map(|(earning_wallet, rate_pack)| RoutingServiceConsumed {
```
bertllll: you can do this operations already within the previous filter_map method

node/src/accountant/mod.rs
```diff
@@ -883,42 +868,52 @@ impl Accountant {
...
+            Some(Self::msg_id())
+        } else {
+            None
+        };
         debug!(
             self.logger,
-            "Accruing debt to wallet {} for consuming exit service {} bytes",
-            msg.earning_wallet,
-            msg.payload_size
+            "MsgId {}: Accruing debt to {} for consuming {} exited bytes",
```
bertllll, line 888: Make the id be computed out of the timestamp within the msg.

PR #158, opened 2022-08-04
GH-622: Add Connection Status Interrogation (Conversation Message) and introduce logs on connection stage update

utkarshg6:


masq/src/commands/connection_status_command.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::any::Any;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq)]
+pub struct ConnectionStatusCommand {}
+
+const CONNECTION_STATUS_ABOUT: &str =
+    "Returns the current stage of the connection status. (NotConnected, ConnectedToNeighbor \
+            or ThreeHopsRouteFound)";
+// TODO: Migrate the three below constants to a common file, so that it can be shared in this file as well as in notification/connection_change_notification.rs
```
dnwiebe: Reminder

masq/src/commands/connection_status_command.rs
```diff
@@ -0,0 +1,266 @@
...
+const NOT_CONNECTED_MSG: &str = "NotConnected: No external neighbor is connected to us.";
+const CONNECTED_TO_NEIGHBOR_MSG: &str =
```
dnwiebe, line 26: The language for these last two messages seems to imply that a change has just happened. That's valid for the broadcast announcements, but when you query the status, the result you get might have been the situation for hours or days.  This isn't a big deal, just something I noticed.

masq/src/commands/connection_status_command.rs
```diff
@@ -0,0 +1,266 @@
...
+            "ConnectedToNeighbor: Established neighborship with an external node."
+        );
+        assert_eq!(
+            THREE_HOPS_ROUTE_FOUND_MSG,
+            "ThreeHopsRouteFound: You can now relay data over the network."
+        )
+    }
+
+    #[test]
+    fn connection_status_works() {
```
dnwiebe: This same test is in command_factory.rs. It should probably be in one place and not the other. We haven't been very consistent about which place we put tests like this, so I won't force consistency on you either.

masq/src/commands/connection_status_command.rs
```diff
@@ -0,0 +1,266 @@
...
+        );
+        assert_eq!(
+            stderr_arc.lock().unwrap().get_string(),
+            "MASQNode is not running; therefore connection status cannot be displayed.\n"
+        );
+        assert_eq!(stdout_arc.lock().unwrap().get_string(), String::new());
+    }
+
+    #[test]
+    fn connection_status_command_happy_path_for_not_connected() {
```
dnwiebe, line 143: It might be worth factoring a big helper method out of these three tests and making them each just a few lines (maybe one?) long.

masq/src/notifications/connection_change_notification.rs
```diff
@@ -15,10 +15,13 @@ impl ConnectionChangeNotification {
     ) {
         let _lock = term_interface.lock();
         let output_string = match response.stage {
-            UiConnectionChangeStage::ConnectedToNeighbor => {
+            UiConnectionStage::NotConnected => {
+                todo!("Panic shouldn't happen after GH-623.")
```
dnwiebe: Can this message be clarified? When I first read it, I thought it meant that the panic happens frequently now, but after GH-623 was played, this section of code wouldn't be hit. Turns out the actual meaning is almost precisely opposite.

node/src/neighborhood/mod.rs
```diff
@@ -578,9 +584,9 @@ impl Neighborhood {
             &node_descriptor.encryption_public_key,
             node_addr,
         );
-        trace!(
+        debug!(
             self.logger,
-            "Sent Gossip: {}",
+            "Sent Debut Gossip: {}",
             debut_gossip.to_dot_graph(
```
dnwiebe: These dot graphs should be confined to TRACE logs. They can be pretty big (although I admit not in the Debut case) and shouldn't clutter up DEBUG-level logfiles.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -90,6 +99,13 @@ impl ConnectionProgress {
             )
         };
 
+        debug!(
+            logger,
+            "The node with IP address {:?} has passed to a new node with IP address {:?}.",
```
dnwiebe: This message should have something in it to make it clear that it has to do with Gossip. Note: the name of the Logger may actually take care of this; I don't know what it is.

dnwiebe (approved):

masq/src/commands/connection_status_command.rs
utkarshg6: Since, we are now using different messages for the different files, the todo that I wrote earlier has become obsolete.

masq/src/commands/connection_status_command.rs
utkarshg6: I have migrated this test to `command_factory.rs`. Perhaps in the future, we can migrate the similar tests of the other commands too.

PR #159, opened 2022-08-05
GH-574: Review 6

utkarshg6:


masq/src/communications/broadcast_handler.rs
```diff
@@ -334,6 +343,39 @@ mod tests {
...
+            &mut stdout,
+            &mut stderr,
+            &terminal_interface,
+        );
+
+        assert_eq!(result, true);
+        let stdout = handle.stdout_so_far();
+        assert_eq!(
+            stdout,
+            "\nConnectedToNeighbor: Established neighborship with an external node.\n\n"
```
bertllll, line 368: I'm thinking...don't we want to say something like..."Established locally originated neighborship with an external Node". To make it obvious in what direction the effort has gone. 
Because in the future you may want to say that some foreign Node initiated it instead of you.

masq/src/notifications/connection_change_notification.rs
```diff
@@ -0,0 +1,75 @@
...
+impl ConnectionChangeNotification {
+    pub fn handle_broadcast(
+        response: UiConnectionChangeBroadcast,
+        stdout: &mut dyn Write,
+        term_interface: &TerminalWrapper,
+    ) {
+        let _lock = term_interface.lock();
+        let output_string = match response.stage {
+            UiConnectionChangeStage::ConnectedToNeighbor => {
+                "\nConnectedToNeighbor: Established neighborship with an external node.\n"
```
bertllll, line 19: Similar question...?

masq/src/communications/broadcast_handler.rs
utkarshg6: Ummm.. As I understand it correctly, "locally originated" should be used because we want to specify, that it was us who sent the debut gossip and not the other person?

bertllll: That was my thought...to be explicit in that sense. But maybe I don't know if I'm saying crap :)

utkarshg6: Nah.. you're good. I was just confirming that's what you mean.

Although, my personal preference would be to keep it this way, as it is easier for the reader of the message to understand.

Apart from that once the implementation of 621 (https://github.com/MASQ-Project/Node-issues/issues/621https://github.com/MASQ-Project/Node-issues/issues/621) is finished, it won't matter whether our node is the debutee or debutant.

bertllll: Okay. Closing this discussion, friend ;)

bertllll (approved):

PR #160, opened 2022-08-09
GH 622 Review 1

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -585,9 +585,10 @@ impl Neighborhood {
             &node_descriptor.encryption_public_key,
             node_addr,
         );
-        debug!(
+        debug!(self.logger, "Debut Gossip sent to {:?}.", node_descriptor);
```
dnwiebe, line 588: Debug, not Debut

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -101,7 +101,8 @@ impl ConnectionProgress {
 
         debug!(
             logger,
-            "The node with IP address {:?} has passed to a new node with IP address {:?}.",
+            "Pass gossip received from node with IP Address {:?} to a node with IP Address {:?}. Hence, updating the connection \
+            stage of the new node node to StageZero.",
```
dnwiebe: We have "node node" toward the end of this string. Also, at some point it was a convention always to use "Node" instead of "node," because "node" can mean lots of different things, but "Node" in this context specifically means a MASQ Node.

dnwiebe (approved):

node/src/neighborhood/mod.rs
utkarshg6: I'm using Debut here to differentiate that this particular gossip is specifically debut and not of any other type. Also, debug will still be printed by the `debug!()` macro.

PR #166, opened 2022-08-25
GH-574 Review 7 Condensed

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -800,12 +804,28 @@ impl Neighborhood {
             payload_size: 10000,
         };
         if self.handle_route_query_message(msg).is_some() {
-            self.overall_connection_status.update_can_make_routes(true);
+            debug!(
+                &self.logger,
+                "The check_connectedness() found a 3 hops route for the first time."
```
dnwiebe, line 809: "The check_connectedness()" is a bit clumsy. Maybe "The connectivity check?" Also, and maybe more importantly, I'm not sure "for the first time" will always be correct. What if a Node gets fully connected, then loses enough neighbors to go back to just connected-to-neighbor, then gains neighbors again and goes back to full connectivity? Will this message ("for the first time") get fired again? If so, it'll be misleading the second time.

node/src/neighborhood/mod.rs
```diff
@@ -800,12 +804,28 @@ impl Neighborhood {
...
+                );
             self.connected_signal_opt
                 .as_ref()
                 .expect("Accountant was not bound")
                 .try_send(StartMessage {})
                 .expect("Accountant is dead")
+        } else {
+            debug!(
+                &self.logger,
+                "The check_connectedness() still can't found a good route."
```
dnwiebe, line 827: "found" is the wrong tense. Probably you want "find." Also, maybe find another way to say "The check_connectedness()."

node/src/neighborhood/mod.rs
```diff
@@ -3644,45 +3651,138 @@ mod tests {
...
+                    let relay_node = make_node_record(nonce(i), false);
+                    replacement_database.add_node(relay_node.clone()).unwrap();
+                    nodes.push(relay_node.public_key().clone());
+                }
+                i if i == hops => {
+                    let exit_node = make_node_record(nonce(i), false);
+                    replacement_database.add_node(exit_node.clone()).unwrap();
+                    nodes.push(exit_node.public_key().clone());
+                }
+                _ => panic!("The match statement should be exhausting."),
```
dnwiebe, line 3685: "Exhaustive," not "exhausting."

node/src/neighborhood/mod.rs
```diff
@@ -3644,45 +3651,138 @@ mod tests {
...
+    #[test]
+    fn neighborhood_starts_accountant_when_first_route_can_be_made() {
         let (accountant, _, accountant_recording_arc) = make_recorder();
-        let system = System::new("neighborhood_does_not_start_accountant_if_no_route_can_be_made");
+        let (ui_gateway, _, _) = make_recorder();
+        let mut subject = make_neighborhood_linearly_connected_with_nodes(3);
+        subject.node_to_ui_recipient_opt = Some(ui_gateway.start().recipient());
         let peer_actors = peer_actors_builder().accountant(accountant).build();
         bind_subject(&mut subject, peer_actors);
+        let system = System::new("neighborhood_does_not_start_accountant_if_no_route_can_be_made");
```
dnwiebe, line 3709: System name should either be something simple like "test" or the same as the test name.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -229,20 +226,30 @@ impl OverallConnectionStatus {
...
+                logger,
+                "The stage of OverallConnectionStatus has been changed \
+                from {:?} to {:?}. A message to the UI was also sent.",
+                prev_stage,
+                new_stage
+            );
+        } else {
+            trace!(
+                logger,
+                "There was an attempt to change the stage of OverallConnectionStatus \
```
dnwiebe, line 248: Maybe "regress" instead of "change" to make clear the reason why the attempt was discarded?

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -909,17 +906,21 @@ mod tests {
 
     fn assert_stage_and_node_to_ui_message(
         initial_stage: OverallConnectionStage,
-        can_make_routes: bool,
+        new_stage: OverallConnectionStage,
         test_name: &str,
+        // logger: &Logger,
```
dnwiebe, line 911: Remove commented-out code

node/src/ui_gateway/websocket_supervisor.rs
```diff
@@ -431,15 +431,19 @@ impl WebSocketSupervisorReal {
     ) -> Option<Vec<SendToClientWebsocketError>> {
         let errors: Vec<SendToClientWebsocketError> = clients
             .into_iter()
-            .flat_map(
-                |(client_id, client)| match client.send(OwnedMessage::Text(json.clone())) {
+            .flat_map(|(client_id, client)| {
+                debug!(
```
dnwiebe, line 435: Looks like debug code that shouldn't make it to production

PR #167, opened 2022-08-26
GH-613: cutting back on too many report consume messages jamming Accountant...version 2 with a fix

bertllll:


node/src/sub_lib/accountant.rs
```diff
@@ -116,17 +115,22 @@ pub struct ReportExitServiceProvidedMessage {
 }
 
 #[derive(Clone, PartialEq, Eq, Debug, Message)]
-pub struct ReportRoutingServiceConsumedMessage {
+pub struct ReportServicesConsumedMessage {
     pub timestamp: SystemTime,
+    pub exit: ExitServiceConsumed,
+    pub routing_payload_size: usize,
```
dnwiebe, line 119: Maybe this should go back inside RoutingServiceConsumed?

node/src/proxy_server/mod.rs
```diff
@@ -81,6 +83,7 @@ pub struct ProxyServer {
     logger: Logger,
     route_ids_to_return_routes: TtlHashMap<u32, AddReturnRouteMessage>,
     browser_proxy_sequence_offset: bool,
+    inbound_client_data_helper: Box<dyn ICDHelperFactory>,
```
dnwiebe: We traditionally call inbound client data IBCD. If you continued this here, it'd be easier to understand. (IB is for InBound.)

node/src/proxy_server/mod.rs
```diff
@@ -718,79 +585,74 @@ impl ProxyServer {
...
-        payload: &ClientRequestPayload_0v1,
-        logger: &Logger,
-    ) {
-        match expected_services
-            .iter()
-            .find_map(|expected_service| match expected_service {
-                ExpectedService::Exit(_, earning_wallet, rate_pack) => {
-                    Some((earning_wallet, rate_pack))
+    fn report_on_exit_service(
+        expected_services: &[ExpectedService],
```
dnwiebe, line 608: Can you make this a vector or the other one a slice, for consistency?

node/src/proxy_server/mod.rs
```diff
@@ -835,13 +698,23 @@ impl ProxyServer {
...
                 )
                 .expect("Key magically disappeared");
-                ProxyServer::report_routing_service(
-                    accountant_routing_sub,
-                    expected_services,
-                    timestamp,
-                    pkg.payload.len(),
-                    logger,
-                );
+                if !expected_services
```
dnwiebe: I think we never use the Proxy Server at all if we're in Zero-Hop mode. If that's the case (maybe some checking would be in order), then we don't have to do checks like this.

node/src/proxy_server/mod.rs
```diff
@@ -928,41 +801,255 @@ impl ProxyServer {
...
+        self.subs
+            .as_ref()
+            .expect("Accountant is unbound")
+            .accountant
+            .try_send(report_message)
+            .expect("Accountant is dead");
+    }
+}
+
+pub trait ICDHelperFactory {
```
dnwiebe: Would prefer IBCD...

dnwiebe (commented):
Resume at proxy_server/mod.rs:858 (ICDHelperFactory)

node/src/proxy_server/mod.rs
```diff
@@ -2660,9 +2634,51 @@ mod tests {
...
-            .await_log_containing("DEBUG: ProxyServer: No exit service requested.", 1000);
+            .await_log_containing("DEBUG: ProxyServer: No routing services requested.", 1000);
+
+        //report about consumed services is sent anyway, exit service is mandatory ever
+        accountant_awaiter.await_message_count(1)
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Exit service must go in each route while missing here: [Routing(cm91dGluZ19rZXlfMQ, \
```
dnwiebe: Wording is a little clumsy here. How about "Each route must demand an exit service, but this route has no such demand?"

node/src/proxy_server/mod.rs
```diff
@@ -3410,10 +3459,9 @@ mod tests {
                 Some(make_wallet("irrelevant")),
                 return_route_with_id(cryptde, 1234),
                 first_client_response_payload.into(),
-                0,
+                77,
```
dnwiebe: This (and the 99 below) are never asserted on and were originally 0. Change them back if you want, ignore them if you want....I don't care, but you told me to mark them.

node/src/accountant/mod.rs
```diff
@@ -883,42 +868,51 @@ impl Accountant {
...
-            msg.payload_size
-        );
-        self.record_service_consumed(
-            msg.service_rate,
-            msg.byte_rate,
-            msg.timestamp,
-            msg.payload_size,
-            &msg.earning_wallet,
-        );
+    fn msg_id(timestamp: SystemTime) -> String {
```
dnwiebe: You might want to write a short, temporary test where you call this method many times in a tight loop and record the resulting IDs in an array, then print them out. (Don't print them out inside the loop; it'll make the loop take too long.) See how many duplicates you get. (The granularity of the system clock is almost certainly bigger than one microsecond.) There might be a facility in Rust for atomically adding 1 to a number. If there is, that might be a better way to get guaranteed nonduplication here.

Note: Look at AtomicU64::fetch_add().

node/src/accountant/mod.rs
```diff
@@ -883,42 +868,51 @@ impl Accountant {
...
+        let lowest_3_digits = microseconds_since_epoch % 1000;
+        lowest_3_digits.to_string()
     }
 
-    fn handle_report_exit_service_consumed_message(
-        &mut self,
-        msg: ReportExitServiceConsumedMessage,
-    ) {
+    fn handle_report_services_consumed_message(&mut self, msg: ReportServicesConsumedMessage) {
+        let msg_id_opt = if self.logger.debug_enabled() {
```
dnwiebe: Would it make sense to put this check in Self::msg_id()? Maybe have it always return 0 if it fails? It looks from here like you're going to be doing this test everywhere you might call msg_id() anyway.

node/src/accountant/mod.rs
```diff
@@ -3291,91 +3149,237 @@ mod tests {
...
-    #[test]
-    fn report_exit_service_consumed_message_is_received_for_our_consuming_wallet() {
-        init_test_logging();
-        let consuming_wallet = make_wallet("own consuming wallet");
-        let config = bc_from_ac_plus_wallets(
-            make_accountant_config_null(),
-            consuming_wallet.clone(),
-            make_wallet("own earning wallet"),
-        );
+    fn do_we_ignore_own_wallets_when_recording_consumed_services(
```
dnwiebe: This name...can we have something that's easier to understand in context? This name makes it sound like the return type should be boolean.

node/src/accountant/mod.rs
```diff
@@ -3291,91 +3149,237 @@ mod tests {
...
+            do_we_ignore_own_wallets_when_recording_consumed_services(config, report_message);
+
+        let more_money_payable_params = more_money_payable_params_arc.lock().unwrap();
+        assert_eq!(
+            *more_money_payable_params,
+            //except processing the exit service there was no change in payables
+            vec![(timestamp, foreign_wallet, (45 + 10 * 1234))]
+        );
+        TestLogHandler::new().exists_log_containing(&format!(
+            "INFO: Accountant: Not recording service consumed to our wallet {}",
```
dnwiebe: This message is confusing. Perhaps a better one would be, "Declining to charge our wallet {} for service we provided."

dnwiebe (approved):

PR #172, opened 2022-09-07
Workflow changes

FinsaasGH:


kauri-hero (approved):
This looks good.

We can explore setting artifact expiry in a separate card.

PR #178, opened 2022-10-04
remove rustup check + added rust version override

FinsaasGH:


kauri-hero (approved):
This looks good for changes.

See Ubuntu build still running but looks like no runtime error.

Will merge when complete - nice work @FinsaasGH

Will ping you also when merged @utkarshg6

kauri-hero:
This is an interesting failure on Ubuntu...

PR #183, opened 2022-10-14
GH-633: Optimise routing to reduce DNS resolution 'mark'

utkarshg6:


node/src/sub_lib/neighborhood.rs
```diff
@@ -506,8 +509,9 @@ pub struct AskAboutDebutGossipMessage {
 }
 
 #[derive(Clone, Debug, Message, PartialEq, Eq)]
-pub enum NodeRecordMetadataMessage {
-    Desirable(PublicKey, bool),
+pub struct NodeRecordMetadataMessage {
```
bertllll, line 520: You need to think about the future of this message. We might want it to have a general character and keep it on.
I have in mind that the message could actually hold just one field which would be an enum. So far we probably know just about one situation of changing the metadata. But maybe it's going the direction where we will have also other requests on the change aside just the unreachable host name.

So I'd probably create an enum having just one variant. 

```
pub struct NodeRecordMetadataMessage {
    pub public_key: PublicKey,
    pub metadata_change: NRMetadataChange,
}
```

```
enum NRMetadataChange{
   AddUnreachableHost{host_name: String}
}
```

The important think here is to ask if we really need the Option. I see it logical to send no message if we don't know the host's name. (I'm going to return to some files I saw first...and talk more about it perhaps.)

node/src/proxy_server/mod.rs
```diff
@@ -278,17 +278,21 @@ impl ProxyServer {
                 })
                 .clone()
         };
+        let host_name = match &return_route_info.server_name {
+            Some(name) => name.clone(),
+            None => "<unspecified server>".to_string(),
```
bertllll: In my eyes, this part is really wrong. Because it seems like you want to go on by adding the `<unspecific server> `string into the list of unreachable hosts for this particular Node.

You probably shouldn't send any message if it's a DNS failure and you cannot get the server name, hopefully with some warning log. 

On the other hand, if I understood correctly your discussion with Dan, the host name is missing if we request to reach an already known IP address. If so then let's ask is it reasonable to process a DNS failure when we immediately read out that this request did not need DNS resolution? 

Maybe it's a more serious crack for the larger concept and an investigation should take place to attempt to know if it's technically possible to get the DNS failure message if there was no host name at the beginning.

node/src/proxy_server/mod.rs
```diff
@@ -278,17 +278,21 @@ impl ProxyServer {
...
                     .as_ref()
                     .expect("Neighborhood unbound in ProxyServer")
                     .update_node_record_metadata
-                    .try_send(NodeRecordMetadataMessage::Desirable(
-                        exit_public_key.clone(),
-                        false,
-                    ))
+                    .try_send(NodeRecordMetadataMessage {
+                        public_key: exit_public_key.clone(),
+                        unreachable_host_name_opt: Some(host_name),
```
bertllll: The same remark as near up here, I'd feel better if there were different options than the optional host name. 
More over, you violate it yourself just here when you have it optional but before this you made sure that the host name string will ever exist, using the sentinel `<unspecific server>`.

So even though I consider sending the message useless and filling the final hash set with garbage, the option at this field doesn't make sense either. (If you weren't forced to use it by other usages of the message, but it yet should lead to reconsideration of the message itself.)

node/src/neighborhood/mod.rs
```diff
@@ -5008,6 +5058,49 @@ mod tests {
...
+            let updated_node_record = actor
+                .neighborhood_database
+                .node_by_key(&public_key)
+                .unwrap();
+            assert!(updated_node_record
+                .metadata
+                .unreachable_hosts
+                .contains(&unreachable_host));
+            TestLogHandler::new().exists_log_matching(&format!(
+                "Host {unreachable_host} is marked unreachable for the node with public key {public_key}"
```
bertllll: Maybe I'd rather see log message have the values hardcoded, because they look can be anticipated to some good extend.

node/src/neighborhood/mod.rs
```diff
@@ -945,6 +951,7 @@ impl Neighborhood {
             msg.target_component,
             msg.payload_size,
             RouteDirection::Over,
+            msg.hostname_opt.clone(),
```
bertllll: I've seen the cascade of function calls down to the ultimate engine...
And I've got a feeling that despite the discomfort of implementing it, you could quite easily approach the host names just by reference from this place where you take it over from the Actix message. 

If you go by that way you might have to introduce a couple of lifetimes, or just somewhere, but they all will be pretty easy and, no trickiness, and the compiler also will tell you. 

If I counted correctly, you'll spare three or four clones on that, not having to be using in for the absence of the String type.   

I don't have the right nose for this but since this infrastructure is so critical it might matter.

node/src/neighborhood/node_record.rs
```diff
@@ -329,6 +329,7 @@ pub struct NodeRecordMetadata {
     pub desirable_for_exit: bool,
```
bertllll: This also cries as if it was wrong...
And I think it might be because to my understanding this boolean flag was exactly what we used to think should be responsible for the exit service reputation. Now, if we traverse to the numeric representation of the desirability and when we know how we want to deal with unreachable web sites, protecting the certain Node from a strict ban, we suddenly have no use for this flag and leaving the field in there isn't fine. 

I'm convinced it should basically work as replacement of that filed by your new one.

bertllll (commented):
I've got a couple of comments on your solution. Other than what I marked seemed good. 
I'd recommend to ask Dan to look at my comments with you (maybe me as well) and reason about their necessity.

multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -193,12 +201,27 @@ impl MASQMockNode {
...
         }
     }
 
+    pub fn absorb_configuration(&mut self, node_record: &NodeRecord) {
+        // Copy attributes from the NodeRecord into the MASQNode.
+        self.configurable_guts.earning_wallet = node_record.earning_wallet();
+        self.configurable_guts.rate_pack = *node_record.rate_pack();
+    }
+
+    pub fn finish_with_guts_pointer(self) -> MASQMockNode {
```
dnwiebe: Maybe should be an implementation of Into<MASQMockNode>? Or From<...> in the other direction?

multinode_integration_tests/src/masq_node_cluster.rs
```diff
@@ -77,38 +77,58 @@ impl MASQNodeCluster {
...
+            None => ConfigurableMASQMockNode::start(
+                ports,
+                index,
+                self.host_node_parent_dir.clone(),
+                self.chain,
+            ),
+        }
+    }
+
+    pub fn finish_mock_node(
```
dnwiebe: Could you name this something that makes it clear that what we're finishing is the _configuration_ of a mock Node? We may not be finished with the mock Node itself.

multinode_integration_tests/src/neighborhood_constructor.rs
```diff
@@ -98,7 +98,9 @@ fn make_mock_node_map(
         .map(|node| (node.main_public_key().clone(), node))
         .collect::<HashMap<PublicKey, MASQMockNode>>();
     additional_keys_to_mock.iter().for_each(|key| {
-        let mock_node = cluster.start_mock_node_with_public_key(vec![10000], key);
```
dnwiebe: Does this force every mock Node to start as configurable?
That seems...oppressive. At least could there be a function or method that does it behind the scenes so you don't have to see it?

dnwiebe (approved):

dnwiebe (approved):

PR #184, opened 2022-10-18
GH-633 Review 1

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -337,22 +337,23 @@ impl Handler<NodeRecordMetadataMessage> for Neighborhood {
...
+                let public_key = msg.public_key;
+                let node_record = self
+                    .neighborhood_database
+                    .node_by_key_mut(&public_key)
+                    .unwrap_or_else(|| panic!("No Node Record found for public_key: {public_key}"));
+                node_record
+                    .metadata
+                    .unreachable_hosts
+                    .insert(host_name.clone());
+                debug!(
```
bertllll, line 351: If you change the message slightly, saying for example "Marking host {host_name} unreachable for the Node with public key {public_key}", the information will be almost identical. When I consider it's a debug log and inserting the record is infallible it all leaves an impression that you can change the order of steps and so put the log before the state-changing operation. Benefits? You'll be free to remove the clone on the host name at line 350. :smiley_cat:

node/src/neighborhood/mod.rs
```diff
@@ -1494,9 +1496,9 @@ pub fn regenerate_signed_gossip(
 }
 
 #[derive(PartialEq, Eq)]
-enum LinkType {
+enum LinkType<'hostname> {
```
bertllll, line 1499: I might not have stressed the lifetime so much by giving it a full-word name - unless you have a good why for that. 
But I'm not to be strict on this, please listen to your heart. (I kinda like it too in a way hehe)

node/src/proxy_server/mod.rs
```diff
@@ -3911,11 +3914,71 @@ mod tests {
...
+                }
             }
         );
     }
 
+    #[test]
+    fn handle_dns_resolve_failure_does_not_sends_message_to_neighborhood_when_server_is_not_specified(
+    ) {
+        let system = System::new("test");
+        let (neighborhood_mock, _, neighborhood_log_arc) = make_recorder();
```
bertllll, line 3928: This is a formal pick but we are used to a different name at variable like this one. Instead of `neighborhood_log_arc` we would call it `neighborhood_recording_arc`.

node/src/proxy_server/mod.rs
```diff
@@ -3911,11 +3914,71 @@ mod tests {
...
+            .build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
+
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
+        subject_addr.try_send(expired_cores_package).unwrap();
+
+        System::current().stop();
+        system.run();
+        let neighborhood_recording = neighborhood_log_arc.lock().unwrap();
+        let record_opt = neighborhood_recording.get_record_opt::<NodeRecordMetadataMessage>(0);
```
bertllll, line 3978: Also a detail, sorry. But I'm a bit worried about this use of `get_record_opt`, to my knowledge, it was kind of a fiasco of mine that I wrote it and we kept it then. 
Please look at usages of it in the whole project and if its influence is only marginal then better use the standard syntax.
Oh, now I realized it can be even simpler, assuming no other message can come to Neighborhood in this test. I'd write myself:

        `assert_eq!(neighborhood_recording.len(), 0)`

node/src/proxy_server/mod.rs
```diff
@@ -278,23 +279,24 @@ impl ProxyServer {
...
+                    self.subs
+                        .as_ref()
+                        .expect("Neighborhood unbound in ProxyServer")
+                        .update_node_record_metadata
+                        .try_send(NodeRecordMetadataMessage {
+                            public_key: exit_public_key.clone(),
+                            metadata_change: NRMetadataChange::AddUnreachableHost { host_name },
+                        })
+                        .expect("Neighborhood is dead");
+                }
```
bertllll, line 299: This is still unpleasantly unclear. I understand what you're trying to solve but I'm nervous from our small knowledge of the wider mechanics. I have a hypothesis that the situation for which we are creating a new condition and an optional string might never happen in reality. The hypothesis is simply based on the logical reasoning that if something doesn't have a host name it then implies need of a substitute, the already workable IP address. With that in mind, we probably shouldn't get a notification about `DNSResolutionFailure` if it has been the pure IP address which traveled to the Exit Node and thus it shouldn't perform a the process of DNS resolution and thus if the process wasn't started it also cannot end with the DNS lookup failure. 

So I sought for the trajectory of the error backwards, from your ProxyServer across Hoper and ProxyClient whose part, the StreamHandlerPool actually, if I hasn't been mistaken, faces decision making in `fn make_stream_with_key(...)` function, where one of the options is to go to `Self::lookup_dns(...)` and that is the place, I believe, where a failure of the process brings about the initial impulse of handling the DNS lookup failure by firing `DnsResolveFailure_0v1` out.

I do think we should inspect carefully this place (very likely mainly `make_stream_with_key`) because I think it can blow off the doubts. If it's true that a hostname-less request can never produce this behavior then it means we treat it pointlessly and we have tests for all this might be confusing for our future work, so I'd prefer to have a serious focus on it now.      

I'm not able to say that it's 100% true because I've should have been sleeping already and my body is protesting, but I'd also like to think about this collectively. Best also with @dnwiebe

node/src/proxy_server/mod.rs
```diff
@@ -278,23 +279,24 @@ impl ProxyServer {
                 })
                 .clone()
         };
-        let host_name = match &return_route_info.server_name {
+        let host_name = match &return_route_info.server_name_opt {
             Some(name) => name.clone(),
-            None => "<unspecified server>".to_string(),
+            None => UNSPECIFIED_SERVER.to_string(),
```
bertllll, line 284: Regardless the outcome of the bigger scale issue below I also think this might be a mistake...
I think it doesn't make sense to make up our own string just to fill a void. 
Isn't that exactly why `Option` is a handy tool? 

Instead of:
```
 if host_name.ne(UNSPECIFIED_SERVER) {
                    self.subs
                        .as_ref()
                        ...
```                        
you can write better...
```
if let Some(host_name) = host_name_opt {
                    self.subs
                        .as_ref()
```                        
and it automatically cause that you won't need to care about the case if it's `None`. Literally, we ignore that case.

node/src/neighborhood/mod.rs
utkarshg6: I find it easy to understand.

node/src/proxy_server/mod.rs
utkarshg6: It was something that I started to write during my card GH-611. I'm unsure whether it is used anywhere else other than my own cards. I believe every developer has their style of writing code, so it's kinda my thing.. that I personally like.


I don't consider the two different methods discussed over here has a huge difference in terms of the computations performed. I believe it takes away the need to question ourselves which method is more computationally optimal.

I believe, the methodology that I have used helps us to check the presence of the message, using `Some` and `None`, that I find interesting, useful and easy to understand. Also, it's possible to annotate the type of message that we're expecting, which improves the code readability.

node/src/proxy_server/mod.rs
utkarshg6: The variable `host_name`, is used again in the same scope. It's more optimal according to me.

node/src/proxy_server/mod.rs
utkarshg6: The `match` statement explained below (it is being used to solve the problem of the current card) existed before I started working on this card.

```rust
        let host_name = match &return_route_info.server_name_opt {
            Some(name) => name.clone(),
            None => UNSPECIFIED_SERVER.to_string(),
        };
```

I was convinced that the author of this code considered a case when this might happen, and the `None` arm might execute. I'm unsure whether it was a necessity or a good to have thing.

When I started working on this card, I moved this `match` statement to few lines above so that it can be used at two places (use case of this card as well as its original use case).

TL;DR
The optional string was something that's been used way before this card was implemented. I'm unsure whether it's still desirable or not? This is the question that I consider others can answer better than me.

node/src/proxy_server/mod.rs
bertllll: Hmmm. I like you have your own opinion. 

However, besides that it is unusual (98% of our tests certainly use the other way of assertion) your assertion only proves that NodeRecordMetadataMessage did not come in Neighborhood...
My suggested assertion says a bit more, that no message, no matter of what kind, was detected to have come in that direction.

It's stronger. :man_shrugging: 

Also I still do remember me writing this. The reason was that I got to a situation where a couple of different messages were anticipated at one Actor (aka Recorder). It was different because for certain unique conditions it was practically possible that they wouldn't arrive in the same order every time (All I can come up with now is that they were sent from different Actors). So in order to make sure that I wasn't going to create another race condition, I rewrote it to allow the optionality and if I asked about one message at the first position in the queue and it wasn't **exactly** the right message type what I asked about, instead of blowing up, it allowed me to have another attempt. After I got None I just asked about the second's message type and it was clear then that it will fit right. Since then I don't remember to have seen a good usecase for that and even with the initial impulse I described here I'm not sure if I couldn't overcome the threat differently...would be nice if I could have.

node/src/proxy_server/mod.rs
bertllll: Oh. You might mean the error!() log at the very end of this function. Well, it still looks better to me to work as long as possible with Option which allows you to avoid the always clumsy situation when you need to fill in a void just to satisfy the compiler...
That happens also here, you are forced to create a string saying UNSPECIFIED_SERVER even though, it's unlikely you will have to use it. Then developers coming to this place may ask themselves what is the purpose of having the two words.
They will eventually figure it out but you can do better if you don't bother them with it at all because it has a simple solution.
Just use host_name_opt, most of time it will fit best. When it comes to the error!() macro, notice that it perfectly sticks with what you need just write something like `host_name_opt.unwrap_or("<unspecified server>".to_string())` inside the macro (yes, as its argument). This way you will spare people from thinking too much about the mystical string created up there. (To me it looks like it was already a suboptimal solution from the person writing this code before you, so the blame doesn't fall on you).  

Hoping that you are okay with this breakdown. :pray:

node/src/neighborhood/mod.rs
bertllll: Go ahead. I don't mind really.

node/src/proxy_server/mod.rs
utkarshg6: I see.

Although, I first started asserting on an `Option<T>` out of necessity in some of my own card. I'm delighted to see that you've also enjoyed the benefit of using it.

I find it easier to read and it also annotates and verifies the kind of message we are expecting inside the test.

In my opinion, here are the following benefits:
1) It'll help another developer to understand what's going on inside the test.
2) Provide a stronger assertion.
3) It uses the feature that most of the other languages fail to provide expect Rust.

I'm still more convinced that asserting on an `Option<T>` is more appropriate. 🤝

node/src/proxy_server/mod.rs
utkarshg6: Your argument over here is strong enough to convince me.

PR #188, opened 2022-10-30
GH-497: answering the second review

bertllll:


node/src/database/db_initializer.rs
```diff
@@ -86,16 +82,22 @@ impl DbInitializer for DbInitializerReal {
         match Connection::open_with_flags(database_file_path, flags) {
             Ok(conn) => {
                 eprintln!("Opened existing database at {:?}", database_file_path);
-                Self::conn_special_setup(&conn, &init_config)?;
+                Self::special_conn_configuration(&conn, &init_config)?;
                 let config = self.extract_configurations(&conn);
+                let found_schema_version = config.get("schema_version");
```
bertllll, line 87: I'm advising to hide the following match statement in a separate function. It takes too much space counter the length and focus of this big function.

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
 }
 ```
 ##### Description:
-This command can serve several purposes. The standard one is to request financial statistics from the Node. However,
-there are other options, like to ask for a view into the database (with multi-parameter customization), to be able to
-look at tracked debts of accounts recorded in respective tables. Such an option is available for both the payable and
-receivable table.
+This command can serve several purposes. The standard one is to request financial statistics from the Node. Besides
+that, there are other options, like to ask for a view into the database (with multi-parameter customization),
+to be able to look at tracked debts of accounts recorded in respective tables. This option is available for both
```
dnwiebe, line 623: Suggestion:

This command requests financial statistics from the Node. Without any parameters specified, it produces a brief summary of financial information; but the output can be customized to show more details.

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
+the payable and receivable table.
 
-One way to pursue it is a query of top N records, ordered by one's choice, either by balance or age. This command setup
-automatically returns results for the tables in a pair.
+One way to proceed is a special query of top N records, ordered by one's choice, either by balance or age. This command
+setup automatically returns two sets of results covering both tables.
```
dnwiebe, line 627: "covering both tables" -- I took out references to the tables above, so if you accept my changes, this should be something like "covering both payable and receivable debts."

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
-together that will be put on the returned records.
-Regarding possible values, the utmost extremes supported by this query derive from the limits of a signed 64-bit
-integer. Bigger values will be caught and answered with an error.
+The user also can make up a more customized query, even separately, one table a time (but it works as easy as it does
+for both at once). The user must input two ranges, for age and balance. Those form restrictions together that are
+followed by the returned records. As for possible values, the utmost extremes supported in the query match 
+9223372036854775807 and -9223372036854775808 respectively. Bigger values will be caught and answered with an error.
 
-An important fact is that the two different modes described above exclude each other, only one can go a time.
+Another important fact is that both modes described above exclude the other one.
```
dnwiebe, line 634: Not "both modes exclude," but "each mode excludes."

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
+While statistics has been considered the spinal feature of this command, it reports back structured information about
+the Node's historical financial operations. This will include services ordered from other Nodes as well as the opposite,
+services provided for other Nodes, represented by monetary values owed and paid to and from this Node's wallets. 
 
-`statsRequested` works for signalization of a refusal of the statistics and so, if those yet wanted, this boolean flag
-should stay as `true`. However, using `false` without any attached query is forbidden as it would lead to a pointless
-request.
+`statsRequired` should be true if historical total balances (see below) are required, false if they are
+not necessary. However, if topRecordsOpt and customQueriesOpt are both null, statsRequired must not be false, since
+it would produce an empty response works for signalization of a refusal of the statistics and so, if those yet wanted,
```
dnwiebe, line 642: I think there's a mistake here somewhere: "produce an empty response works for signalization" You thought everything from "works" to the end of the paragraph should disappear.

Also, "signalization" is a bit clumsy; "indication" might be better.

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
+this boolean flag should stay as `true`. However, using `false` without any attached query is forbidden as it would
+lead to a pointless request.
 
-`topRecordsOpt` initiates a request for a given number of top records in the tables. An optional feature. It is good to
-know that only accounts with positive balances can be displayed in this mode which means some receivables theoretically
-might be omitted given their nature of possible negative numbers.
+`topRecordsOpt` initiates the request for a given number of top records in the tables. An optional feature. It is good to
+know that only accounts with positive balances can be displayed in this mode which means some receivables might stay
+out of scope given their nature of possibly negative numbers.
```
dnwiebe, line 648: ..."out of scope since in some cases receivable balances can be negative."

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
 
-`count` is number of records to be returned if they exist. It should be a number bigger than zero. The other side limit
-is made by unsigned 16-bit integer.
+`count` is the number of records to be returned if so many exist. It should be a number bigger than zero. The other
```
dnwiebe, line 650: How about: "It should be a number between 1 and 65535, inclusive."?

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
+side's limit is defined by the number 65535.
 
-`orderedBy` allows a choice of ordering, while each one goes downwards. Possible values: `Balance` and  `Age`.
+`orderedBy` orderedBy allows you to choose whether the results are sorted in descending order by `Balance` or by `Age`.
```
dnwiebe, line 653: "orderedBy" is repeated here

USER-INTERFACE-INTERFACE.md
```diff
@@ -618,60 +618,61 @@ field will be null or absent.
...
-The only difference is in value limits over the balance parameters where for payables you can supply values from 0
-to the maximum of signed 64-bit integer and for receivables the limits are from the minimum of signed 64-integer to
-up to its maximum.   
+The following subfields, standing as parameters of the two previously described fields, seem mostly the same for both.  
+The only difference lies in value limits at the parameters of balance where for payables you can supply values from 0
+to the maximum 9223372036854775807 and for receivables from -9223372036854775808 up to 9223372036854775807.   
 
 `minAgeS` is measured in seconds from the present time and represents a constraint for the accounts we will be 
-searching over; this is the lower limit of the debt's age, respectively how long has it been since the last payment.
+searching over; this is the lower limit of the debt's age, respectively how long it has been since the last payment.
```
dnwiebe, line 672: "respectively" doesn't really fit here. How about "or?" Also below for maxAgeS.

USER-INTERFACE-INTERFACE.md
```diff
@@ -715,53 +716,50 @@ their balances.
...
-In case of use of the "topRecords" variation, those records will be sorted downwards and limited by their maximum count
-while their ordering depends on the second parameter that participated in the request (either by Balance or Age).
-If the table contains fewer records than requested by the count, a smaller amount is sent back.
-With the more customized query, a limiting range of both age and balances can be specified by the user and these two
-will determine the subset of records being returned. The results are always sorted by balance here. As well as with
-the previous, this kind of request produces an empty array if no records could be retrieved. 
-It's of note that these requests are optional and also because it is possible to request only one table view, not of
-both tables together, Null might be returned in various places, either at the level of individual tables or in the
-place of the whole thing. 
+In case of use of the "topRecords" version, those records will be sorted downwards and limited in their number by
```
dnwiebe, line 736: Maybe "If the `topRecords` parameter is used,"?

USER-INTERFACE-INTERFACE.md
```diff
@@ -715,53 +716,50 @@ their balances.
...
+or `Age`. If the table contains fewer records than requested, the smaller amount comes back.
 
-`payableOpt` is the part referring to payable records if any exist, as explained before, Null or an empty array are
-also possible.
+With `customQueryOpt`, the limiting ranges of both age and balance depend on the user and both will determine the subset
+of records being returned. The results are always ordered by balance in this mode. Likewise as the previous, this kind
+of request produces an empty array if no records can be retrieved. 
 
-`wallet` is a wallet of the Node that we owe to.
+It's of note that because this request is optional, and also it is possible to request an only table view, not both, null
```
dnwiebe, line 744: "an only table view" is probably wrong. Maybe "a single-table view?"

USER-INTERFACE-INTERFACE.md
```diff
@@ -715,53 +716,50 @@ their balances.
...
 payment was later also confirmed on the blockchain.
 
 `balanceGwei` is a number of Gwei we owe to this particular Node.
 
-`pendingPayableHashOpt` is present only sporadically, but says that we've recently approached paying our debt,
-but the payment wasn't confirmed yet by our confirmation detector. If the value is present, it gives you a transaction
-hash of the so-believed pending transaction. 
+`pendingPayableHashOpt` is present only sporadically, but says that we've recently sent a payment to the blockchain,
+but the payment wasn't confirmed yet by our detector for confirmations. If the value is present, it consists of 
+a transaction hash of the so-believed pending transaction.
```
dnwiebe, line 758: Try removing "so-believed" and see how it feels.

USER-INTERFACE-INTERFACE.md
```diff
@@ -715,53 +716,50 @@ their balances.
...
-but the payment wasn't confirmed yet by our confirmation detector. If the value is present, it gives you a transaction
-hash of the so-believed pending transaction. 
+`pendingPayableHashOpt` is present only sporadically, but says that we've recently sent a payment to the blockchain,
+but the payment wasn't confirmed yet by our detector for confirmations. If the value is present, it consists of 
+a transaction hash of the so-believed pending transaction. 
 
 `receivable` is the part of top records that concerns receivable records if any exist.
 
-`wallet` is a wallet of the Node that owes money to us for services we pursued to them in the past.
+`wallet` is the wallet of the Node that owes money to us for services we consumed from them in the past.
```
dnwiebe, line 762: Other Nodes don't owe us money for services we consumed. Not sure which direction this should go, but some direction other than this one.

masq/src/commands/financials_command.rs
```diff
@@ -5,52 +5,58 @@ use crate::commands::commands_common::{
...
-const NO_STATS_ARG_HELP: &str = "Disables statistics that displays by default, containing totals of paid and unpaid money from the perspective of debtors and creditors. This argument is not accepted alone and must be placed before other arguments";
+const TOP_ARG_HELP: &str = "Returns a subset of the top N records (or fewer, if their total count is smaller) from both payable and receivable. By default, the ordering is done by balance but can be changed with the additional \
+ '--ordered' argument.";
+const PAYABLE_ARG_HELP: &str = "Forms a detailed query about payable records by specifying two ranges, one for their age in seconds and another for their balance in MASQs (decimal numbers are supported, allowing Gwei precision). \
+ The required format consists of two ranges separated by '|' like here: <MIN-AGE>-<MAX-AGE>|<MIN-BALANCE>-<MAX-BALANCE>. Leaving out MAX-BALANCE, including the dash before, defaults to maximum (2^64 - 1).";
```
dnwiebe, line 34: You said you thought the single quotes were unnecessary and distracting.

dnwiebe (commented):
Resume at financials_command.rs:395 green.

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
+    fn convert_masq_from_gwei_and_dress_well<T>(gwei: T) -> String
+    where
+        T: Display + PartialEq + From<u32> + Div<Output = T> + PartialOrd<T>,
+        <T as Div<T>>::Output: PartialEq<T> + PartialOrd<T> + Display,
+    {
+        let (affected_part, is_positive) = match Self::preformatting_math_analysis(gwei) {
+            Left(result) => return result,
+            Right(tuple) => tuple,
+        };
+        let minimal_padding = if is_positive { 3 } else { 4 };
```
dnwiebe, line 441: Two things here: first, it looks like this line is designed to leave extra space for a minus sign in case of a negative number, but it seems to be leaving extra space in case of a positive number instead. Second, I'm curious about why the formatting is being done with leading zeros. I would have expected to see something like `{:>count$}` to left-pad with spaces.

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
-            preprocessed_values[2],
-            preprocessed_values[3],
-            ordinal_num_width = optimal_widths[0],
-            wallet_width = WALLET_ADDRESS_LENGTH,
-            age_width = optimal_widths[2],
-            balance_width = optimal_widths[3],
-            hash_width = optimal_widths[4]
-        )
+            "\nFinancial status totals in {}\n",
+            &Self::gwei_or_masq_balance(gwei)[9..13]
```
dnwiebe, line 477: Couple things here.
First, as you pointed out, `is_gwei` would be a better name than `gwei`.
Second, it would be clearer to have the function that produced either "Gwei" or "MASQ" just return one of those strings, and have `gwei_or_masq_balance(bool)` call it. Then you could call that function here, and its operation would be obvious.

dnwiebe (commented):
Resume at financials_command.rs:481 green

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
+        fn handle_captures_of_number_components(captures: Captures) -> [Option<String>; 3] {
+            let fetch_group = |idx: usize| -> Option<String> {
+                FinancialsCommand::single_capture(&captures, idx)
+            };
             [
                 fetch_group(1),
                 fetch_group(2),
-                fetch_group(5).or_else(|| fetch_group(3)),
+                //takes 5 or says None if 6 or 3 is present, but not both
+                fetch_group(5).or_else(|| fetch_group(6).xor(fetch_group(3))),
```
dnwiebe, line 524: Would a multiline comment like this be clearer than the single-line comment?
```
5   6   3   Result
------------------
Y   ?   ?   Some(5)
N   N   N   None
N   N   Y   Some(3)
N   Y   N   Some(6)
N   Y   Y   None
```

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
-        user_ranges: &UsersOriginalLiteralRanges,
+        user_ranges: &UserOriginalTypingOfRanges,
     ) -> (String, String) {
-        fn handle_captures(captures: Captures) -> [Option<String>; 3] {
-            let fetch_group = |idx: usize| FinancialsCommand::single_capture(&captures, idx);
+        fn handle_captures_of_number_components(captures: Captures) -> [Option<String>; 3] {
+            let fetch_group = |idx: usize| -> Option<String> {
+                FinancialsCommand::single_capture(&captures, idx)
+            };
             [
```
dnwiebe, line 520: An additional comment here telling us what groups 1, 2, 3, 5, and 6 are would be helpful.

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
         }
 
-        let simplified_syntax_extractor =
-            Regex::new("^(-?)0*(\\d+?)(((\\.\\d*[1-9])0*$)|$)|^$").expect("wrong regex");
-        let apply_care = |remembered_user_input: &str| {
-            simplified_syntax_extractor
-                .captures(remembered_user_input)
-                .map(handle_captures)
+        let simplifying_extractor =
+            Regex::new("^(-?)0*(\\d+?)(((\\.\\d*[1-9])0*$|(\\.0))|$)").expect("wrong regex");
```
dnwiebe, line 546: For group 2, you probably want `\\d*` instead of `\\d+?`. Also, I don't think group 4 is necessary. If you remove group 4, then group 3 is just a grouping group instead of a capturing group, because of that pesky `0*` outside the parentheses; so you're left with groups 1, 2, and (4, 5, or nothing). Finally, you said you don't think the regex has to match on an empty string.

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
         }
         fn compose_ranges(mut numbers: VecDeque<String>) -> (String, String) {
-            fn resolve_upper_bound(val: String) -> String {
-                if val.is_empty() {
-                    "UNLIMITED".to_string()
-                } else {
-                    val
-                }
-            }
+            if numbers.get(3).expectv("fourth element").is_empty() {
```
dnwiebe, line 531: We need at _least_ a comment here.

Can you do something like numbers[3] = "UNLIMITED"? (A test suggests that you can.)
Or, as you suggested, remove the last element and push on a different one?

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
+                let _ = mem::replace(
+                    numbers.iter_mut().last().expect("fourth element"),
+                    "UNLIMITED".to_string(),
+                );
+            };
             let mut pop = |param: &str| numbers.pop_front().expectv(param);
-
+            let mut str_range = |tuple: (&str, &str)| format!("{}-{}", pop(tuple.0), pop(tuple.1));
```
dnwiebe, line 538: Would this be clearer as a closure with two `&str` parameters?

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
 
-        let simplified_syntax_extractor =
-            Regex::new("^(-?)0*(\\d+?)(((\\.\\d*[1-9])0*$)|$)|^$").expect("wrong regex");
-        let apply_care = |remembered_user_input: &str| {
-            simplified_syntax_extractor
-                .captures(remembered_user_input)
-                .map(handle_captures)
+        let simplifying_extractor =
+            Regex::new("^(-?)0*(\\d+?)(((\\.\\d*[1-9])0*$|(\\.0))|$)").expect("wrong regex");
+        let apply_care = |cached_user_input: &str| {
```
dnwiebe, line 547: Could you put an explicit return type here? I think that'd help me understand.

masq/src/commands/financials_command.rs
```diff
@@ -402,319 +377,183 @@ impl FinancialsCommand {
...
             stdout,
             "Specific {} query: {} sec {} MASQ\n",
-            distinguished,
+            table_type,
             age_range,
             balance_range
         )
     }
 
     fn correct_users_writing_if_needed(
```
dnwiebe, line 513: It occurs to me that the fact that you have one function here containing a lot of inner/internal functions may indicate that this hunk of code should be extracted somehow: maybe as a class, maybe as a module, maybe as another file. I don't have any firm opinions on the matter; I'm just offering something to think about.

masq/src/commands/financials_command.rs
```diff
@@ -725,64 +564,120 @@ impl FinancialsCommand {
...
+        if leading_dump {
+            short_writeln!(stdout)
+        } else {
+            short_writeln!(stdout, "\n\n")
+        }
+    }
+
+    fn no_records_found(stdout: &mut dyn Write, headings: &[String]) {
+        let mut headings_widths = Self::widths_of_str_values(headings);
+        let _ = mem::replace(&mut headings_widths[1], WALLET_ADDRESS_LENGTH);
```
dnwiebe, line 582: Pretty sure this can be `headings_widths[1] = WALLET_ADDRESS_LENGTH`

masq/src/commands/financials_command.rs
```diff
@@ -725,64 +564,120 @@ impl FinancialsCommand {
...
+        accounts: Vec<A>,
+        headings: &HeadingsHolder,
+    ) {
+        let preformatted_subset = &accounts
+            .iter()
+            .enumerate()
+            .map(|(idx, account)| account.string_values(idx + 1, headings.is_gwei))
+            .collect::<Vec<_>>();
+        let optimal_widths = Self::width_precise_calculation(headings, preformatted_subset);
+        let zipped_inputs = &Self::zip_them(headings.words.as_slice(), &optimal_widths);
```
dnwiebe, line 615: Maybe a better name for this variable? Maybe headings_and_widths or something similar?

masq/src/commands/financials_command.rs
```diff
@@ -725,64 +564,120 @@ impl FinancialsCommand {
...
+
+    fn write_column_formatted(
+        stdout: &mut dyn Write,
+        preprocessed_segmented_account_values: &[(&String, &usize)],
+    ) {
+        let column_count = preprocessed_segmented_account_values.len();
+        preprocessed_segmented_account_values
+            .iter()
+            .enumerate()
+            .for_each(|(idx, (value, opt_width))| {
```
dnwiebe, line 638: "opt" makes me say "optional."

masq/src/commands/financials_command.rs
```diff
@@ -797,84 +692,89 @@ impl FinancialsCommand {
...
     where
         N: FromStr<Err = ParseIntError> + TryFrom<i64> + Mul<Output = N> + CheckedMul + Display,
         <N as TryFrom<i64>>::Error: Debug,
     {
-        let (first, second_opt) =
-            Self::extract_individual_masq_values(range_str, Self::masq_values_range_regex())?;
-        let first_as_num = Self::process_optionally_fragmentary_number(&first)?;
-        let second_as_num = if let Some(second) = second_opt.as_ref() {
-            Self::process_optionally_fragmentary_number(second)?
+        let regex = Regex::new("^((-?\\d+\\.?\\d*)\\s*-\\s*(-?\\d+\\.?\\d*))|(-?\\d+\\.?\\d*)$")
```
dnwiebe, line 735: Isn't there a facility in Rust where you can specify a string like `r#"`...blah...`"` and not have to put in double backslashes?

dnwiebe (commented):
Resume at financials_command.rs:752 green

masq/src/commands/financials_command.rs
```diff
@@ -916,39 +822,47 @@ impl StringValuesOfAccount for UiPayableAccount {
...
+    })?;
     let (min_age, max_age) = FinancialsCommand::parse_time_params(&time_range)?;
     let (min_amount, max_amount, _, _): (N, N, _, _) =
         FinancialsCommand::parse_masq_range_to_gwei(separate_ranges[1])?;
+    //There is no sense in trying to check an exact age because of its all time moving nature
+    //In money, it is less nonsensical but still quite unlikely that you would really need it;
+    //the backend engine carries the search out always on the Gwei precision
     if min_age >= max_age || min_amount >= max_amount {
-        Err("Both ranges must be ascending".to_string())
+        Err(format!("Both ranges '{}' must be ascendant", two_ranges))
```
dnwiebe, line 865: Maybe something simpler, like "must be low to high?"

masq/src/commands/financials_command.rs
```diff
@@ -1034,24 +962,31 @@ mod tests {
...
+                    custom_queries_opt: None
+                }
+                .tmb(0),
+                STANDARD_COMMAND_TIMEOUT_MILLIS
+            )]
+        );
     }
 
     #[test]
     fn command_factory_everything_demanded_with_top_records() {
```
dnwiebe, line 984: I may delete this comment in a moment, but would it be a better test if you demanded sorting by age, since we've already tried balance?

masq/src/commands/financials_command.rs
```diff
@@ -1648,43 +1585,43 @@ mod tests {
...
         assert_eq!(
             result,
-            ("456500000-354815".to_string(), "33330-454".to_string())
+            ("456500-35481533".to_string(), "-500--45".to_string())
         )
     }
 
     #[test]
-    fn correct_users_writing_if_needed_threats_0_followed_decimal_numbers_gently() {
+    fn correct_users_writing_if_needed_threats_zero_followed_decimal_numbers_gently() {
```
dnwiebe, line 1624: Should be "treats," not "threats."

masq/src/commands/financials_command.rs
```diff
@@ -1777,11 +1715,40 @@ mod tests {
...
+    ) {
+        let gwei: u64 = 9999999;
+
+        let result = FinancialsCommand::convert_masq_from_gwei_and_dress_well(gwei);
+
+        assert_eq!(result, "< 0.01")
+    }
+
+    #[test]
+    fn convert_masq_from_gwei_and_dress_well_handles_values_bigger_than_one_hundredth_of_masq_and_smaller_than_zero(
```
dnwiebe, line 1734: "Minus" or "negative" should be somewhere in this test name

masq/src/commands/financials_command.rs
```diff
@@ -2384,7 +2357,7 @@ mod tests {
                     UiReceivableAccount {
                         wallet: "0xaa22968a5263f165F014d3F21A443f10a116EDe0".to_string(),
                         age_s: 566668,
-                        balance_gwei: -550,
+                        balance_gwei: 550,
```
dnwiebe, line 2360: I'd be interested to see what happens if you change this back to negative. Does the test fail? If not, that's probably an issue.

masq/src/commands/financials_command.rs
```diff
@@ -2662,56 +2630,88 @@ mod tests {
...
+    }
+
     #[test]
     fn validate_two_ranges_with_misused_central_delimiter() {
         let result = validate_two_ranges::<i64>("45-500545-006".to_string());
 
         assert_eq!(
             result,
-            Err("Central vertical delimiter misused".to_string())
+            Err("Central vertical delimiter '|' misused in '45-500545-006'".to_string())
```
dnwiebe, line 2653: Hmm. Maybe "Vertical delimiter '|' should be used between balance and age ranges and only there. Example: '1234-2345|3456-4567', not  '45-500545-006'." That's pretty complicated, but shouldn't require any more code than what you have. Or just show the clap help message; that might work too.

masq/src/commands/financials_command.rs
```diff
@@ -2662,56 +2630,88 @@ mod tests {
...
     }
 
     #[test]
     fn validate_two_ranges_with_misused_range_delimiter() {
         let result = validate_two_ranges::<i64>("45+500|545+006".to_string());
 
-        assert_eq!(result, Err("First range is formatted wrong".to_string()))
+        assert_eq!(
+            result,
+            Err("First range '45+500' is formatted wrong".to_string())
```
dnwiebe, line 2663: Maybe "age range" instead of "first range?"

masq/src/commands/financials_command.rs
```diff
@@ -2720,34 +2720,37 @@ mod tests {
...
     }
 
     #[test]
     fn validate_two_ranges_non_numeric_value_for_second_range() {
         let result = validate_two_ranges::<i64>("1000-1234|7878.0-a lot".to_string());
 
-        assert_eq!(result, Err("Second range in improper format".to_string()))
+        assert_eq!(
+            result,
+            Err("Second range '7878.0-a lot' in improper format".to_string())
```
dnwiebe, line 2733: Maybe 'balance range' instead of 'second range?'

node/src/accountant/big_int_db_processor.rs
```diff
@@ -20,67 +19,120 @@ use std::marker::PhantomData;
 use std::ops::Neg;
 
 #[derive(Debug)]
-pub struct BigIntDbProcessor<T: DAOTableIdentifier> {
+pub struct BigIntDbProcessor<T: TableNameDAO> {
+    overflow_handler: Box<dyn UpdateOverflowHandler<T>>,
     phantom: PhantomData<T>,
```
dnwiebe, line 24: You might not need this anymore.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -20,67 +19,120 @@ use std::marker::PhantomData;
...
+impl<T: TableNameDAO> UpdateOverflowHandler<T> for UpdateOverflowHandlerReal<T> {
+    fn update_with_overflow<'a>(
         &self,
         conn: Either<&dyn ConnectionWrapper, &Transaction>,
         config: BigIntSqlConfig<'a, T>,
     ) -> Result<(), BigIntDbError> {
         let select_sql = config.select_sql();
-        let mut select_stm = Self::prepare_statement(conn, &select_sql);
+        let mut select_stm = BigIntDbProcessor::<T>::prepare_statement(conn, &select_sql);
         match select_stm.query_row([], |row| {
```
dnwiebe, line 109: Could you use a temporary result here so that we don't have a closure definition in the middle of a match? It's kind of hard to read this way. Maybe either declare the closure separately, or call .query_row() and get a result, then use match on the result?

node/src/accountant/big_int_db_processor.rs
```diff
@@ -20,67 +19,120 @@ use std::marker::PhantomData;
...
-                        & 0x7FFFFFFFFFFFFFFF) as i64
-                        - low_bytes;
-                    let update_sql = config.prepare_update_sql();
-                    let mut update_stm = Self::prepare_statement(conn, &update_sql);
+            let high_bytes_result = row.get::<usize, i64>(0);
+            let low_bytes_result = row.get::<usize, i64>(1);
+            match [high_bytes_result, low_bytes_result] {
+                [Ok(previous_high_bytes), Ok(previous_low_bytes)] => {
+                    let requested_wei_change = &config.params.wei_change_params;
+                    let (high_bytes_correction, low_bytes_correction) = Self::correct_bytes(
```
dnwiebe, line 115: "Correction" implies that this is a factor that can be combined with the wrong value to correct into the right value. But this is the complete correct value; perhaps it could be named something like `xxx_bytes_corrected`?

dnwiebe (commented):
Resume at big_int_db_processor.rs:188 green.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -197,11 +233,12 @@ impl<'a, T: DAOTableIdentifier> BigIntSqlConfig<'a, T> {
             .skip_while(|char| char.is_whitespace())
             .take_while(|char| !char.is_whitespace())
             .collect::<String>();
-        match (keyword.trim(), self.update_clause_opt.is_some()) {
-            ("insert", true) => "upsert".to_string(),
-            ("update", false) => keyword,
+        match keyword.trim() {
```
dnwiebe, line 236: Given the `skip_while` and `take_while` above, you may be able to remove this `trim`.

dnwiebe (commented):
Resume at big_int_db_processor.rs: 541 green

node/src/accountant/big_int_db_processor.rs
```diff
@@ -669,19 +719,53 @@ mod tests {
...
             params_except_wei_change: vec![],
         }
     }
 
     #[test]
     fn determine_command_works_for_upsert() {
         let subject: BigIntSqlConfig<'_, DummyDao> = BigIntSqlConfig {
-            main_sql: "insert into table (a,b) values ('a','b')",
-            update_clause_opt: Some(|table| format!("side clause {}", table)),
+            main_sql:
```
dnwiebe, line 766: Have a look whether you take into account the word "update."

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
+
+        let result = analyse_sql_commands_execution_without_details_of_overflow(
             "update_alone_works_for_addition_with_overflow",
-            "update test_table set balance_high_b = balance_high_b + :balance_high_b, balance_low_b = balance_low_b + :balance_low_b where name = :name",
-            None,
-            Some((&"Joe",47,i64::MAX - 12)),
-            Addition("balance", 25578),
-            true);
+            STANDARD_EXAMPLE_OF_UPDATE_CLAUSE,
+            "",
```
dnwiebe, line 1020: I think you said that this really should be populated to show what it would look like in real life, but this test is limited and doesn't get far enough to use it.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
+        );
+
+        assert_eq!(BigIntDivider::deconstruct(wei_change), (-1, 4));
+        assert_eq!(
+            result,
+            ConventionalUpsertUpdateAnalysisData {
+                was_update_with_overflow: false,
+                final_database_values: ReadFinalRow {
+                    high_bytes: 54,
+                    low_bytes: 9223372036854775806,
```
dnwiebe, line 1059: Is this `i64::MAX - 9`?  If it is, it'd be easier to read that way, I think.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
+    fn main_sql_clause_error_handled() {
+        let conn = initiate_simple_connection_and_test_table(
             "big_int_db_processor",
-            "select_stm_in_update_with_overflow_gets_also_well_along_numeric_key_value_and_different_table_and_substitution_param_name"
+            "main_sql_clause_error_handled",
+        );
+        let subject = BigIntDbProcessor::<DummyDao>::default();
+        let balance_change = Addition("balance", 4879898145125);
+        let config = BigIntSqlConfig::new(
+            "insert into test_table (name, balance_high_b, balance_low_b) values (:name, :balance_a, :balance_b) on conflict (name) do \
```
dnwiebe, line 1314: You didn't like :balance_a and :balance_b. It sounds like you wanted something like :balance_wrong_a and :balance_wrong_b.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
 
         assert_eq!(
             result,
-            Err(BigIntDbError(
-                //sadly, I wasn't able to get a nicer error case with a more obvious relation to the tested requirements
-                "Wei change: error after invalid upsert command for test_table of 5555 Wei to Joe with error 'NOT NULL constraint failed: test_table.balance_low_b'"
-                    .to_string()
-            ))
+            UpdateWithOverflowSummary {
+                balance_change_decomposed: (3, 9223372036854775804),
```
dnwiebe, line 1407: This might be more illuminating in hexadecimal. I think they'd be 0x3 (or 0b11 in binary) and 0xFFFFFFFFFFFFFFFC.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
+                final_read_high_bytes: 4555 + 3 + SINGLE_UNIT_FROM_OVERFLOW,
+                final_read_low_bytes: 9223372036854775804 - 55 - SINGLE_UNIT_FROM_OVERFLOW,
```
dnwiebe, line 1409: Yeeeahhh...I don't understand the subtraction of SINGLE_UNIT_FROM_OVERFLOW here. Seems like something's weird.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -890,261 +961,508 @@ mod tests {
...
-        let balance_change = Addition("balance", 4879898145125);
-        let config = BigIntSqlConfig::new(
-            "insert into test_table (name,balance_high_b,balance_low_b) values (:name,:balance_a,:balance_b) on conflict (name) do",
-            Some(|table_name |format!("update {} set balance_high_b = balance_high_b + 5, balance_low_b = balance_low_b + 10 where name = :name",table_name)),
-            SQLParamsBuilder::default()
-                .key("name", ":name", &"Joe")
-                .wei_change(balance_change)
-                .build(),
+
+        assert_eq!(
```
dnwiebe, line 1430: This strangeness could be overlooked if you could do the 128-bit math right here in the test and show that the result of the 128-bit math was the same as what you get when you reconstitute the data from the database.

Like this:
let big_initial = 803247859038902890349457i128;
let big_addend = 89438948030294886894302i128;
let big_sum = big_initial + big_addend;
let (initial_high_bytes, initial_low_bytes) = BigIntDivider::decompose(big_initial);
...send through your process...
let actual_sum = BigIntDivider::reconstitute(resulting_high_bytes, resulting_low_bytes);
assert_eq (actual_sum, big_sum);

Something similar for subtraction would be good too.

node/src/accountant/big_int_db_processor.rs
```diff
@@ -1541,26 +1867,28 @@ mod tests {
         assert_eq!(high_bytes_error,
                    SqliteFailure(
                        rusqlite::ffi::Error{ code: ErrorCode::Unknown, extended_code: 1 },
-                       Some("Error from biginthigh: None negative slope, while designed only for use with negative one: 5656.23".to_string())
+                       Some("Error from slope_drop_high_bytes: Nonnegative slope '5656.23', \
```
dnwiebe, line 1870: I may have mentioned a rewording here already, but in case I didn't:
"Nonnegative slope 5656.23; delinquency slope must be negative, since debts must become more delinquent over time."

node/src/accountant/dao_utils.rs
```diff
@@ -295,6 +292,36 @@ fn to_age(timestamp: SystemTime) -> u64 {
...
+    conn: &dyn ConnectionWrapper,
+    table: &str,
+    param_name: &str,
+    value_creation: fn(&mut usize, &Row) -> rusqlite::Result<i128>,
+) -> i128 {
+    let mut row_counter = 0;
+    let select_stm = format!("select {param_name}_high_b, {param_name}_low_b from {table}");
+    conn.prepare(&select_stm)
+        .expect("select stm error")
+        .query_map([], |row| value_creation(&mut row_counter, row))
```
dnwiebe, line 319: Requiring the caller to provide a function that increments the row count seems a little impolite. Could you fix things here so that the `value_creation` function only has to read the row count, rather than incrementing it?

node/src/accountant/mod.rs
```diff
@@ -1458,73 +1396,69 @@ pub struct ThresholdUtils {}
...
-        debt_threshold: i64,
-        maturity_threshold_sec: i64,
-    ) -> f64 {
-        let m = m.abs();
-        let x_interception = (debt_threshold as f64 / m) + maturity_threshold_sec as f64;
-        m * x_interception
-    }
-
-    pub fn calculate_sloped_threshold_by_time(
+    fn calculate_sloped_threshold_by_time(
```
dnwiebe, line 1409: Maybe "calculate_finite_debt_limit_by_age"?

node/src/accountant/mod.rs
```diff
@@ -1458,73 +1396,69 @@ pub struct ThresholdUtils {}
...
         payment_thresholds: &PaymentThresholds,
         time: u64,
```
dnwiebe, line 1411: This parameter would be easier to understand if it were named something like "age_in_sec".

node/src/accountant/mod.rs
```diff
@@ -1458,73 +1396,69 @@ pub struct ThresholdUtils {}
...
     ) -> f64 {
+        if Self::test_if_lies_after_the_slope(time, payment_thresholds) {
```
dnwiebe, line 1413: Name suggestion: qualifies_for_permanent_debt_limit

node/src/accountant/mod.rs
dnwiebe, line 1409: There's a fair amount of integer / floating-point conversion going on here. We might be able to keep this in the integer domain by using a big factor (perhaps WEIS_OF_GWEI) to jack up the slope so that it's not so small that floating-point is necessary. We might run into small rounding errors, but as long as they're small, nobody will care, since this isn't about stored debt but rather about delinquency detection.

Here's what I wrote during the review. My counsel would be to spend a small amount of time doing this on an experimental branch, and if it causes more than a little trouble, make a separate card out of it, because it is an optimization that can be applied at any time later.

```
slope: f64 = dY as f64 / dX as f64

let FACTOR = 1_000_000_000;

int_slope: i64 = (dY * FACTOR) / dX

int_y = slope * x + (b * FACTOR);

y = int_y / FACTOR
```

node/src/accountant/mod.rs
```diff
@@ -4234,26 +4172,131 @@ mod tests {
...
+            lower_corner_point <= ideal_template_lower + allowed_imprecision
+                && ideal_template_lower - allowed_imprecision <= lower_corner_point,
+            "ideal: {}, real: {}",
+            ideal_template_lower,
+            lower_corner_point
         )
     }
 
+    #[test]
+    fn testing_granularity_calculate_sloped_threshold_by_time() {
```
dnwiebe, line 4225: It would be nice to run the gap_tester on the real PaymentThresholds, just so that if somebody modified them to be stupid, we'd know immediately.

node/src/accountant/mod.rs
```diff
@@ -4301,23 +4376,25 @@ mod tests {
...
+                + payment_thresholds.threshold_interval_sec
+                + 1,
         ) as u64;
-        let seeming_lower_overflow = ThresholdUtils::calculate_sloped_threshold_by_time(
+        let a_certain_distance_further = ThresholdUtils::calculate_sloped_threshold_by_time(
             &payment_thresholds,
             payment_thresholds.maturity_threshold_sec
                 + payment_thresholds.threshold_interval_sec
-                + 33,
+                + 333,
```
dnwiebe, line 4389: You said this number doesn't make sense.

dnwiebe (commented):
Resume at accountant/mod.rs:4407 green

node/src/accountant/mod.rs
```diff
@@ -5273,8 +5350,26 @@ mod tests {
         )
     }
 
+    macro_rules! extract_ages_from_accounts {
+        ($account_type: ident, $main_structure: expr, $account_specific_field_opt: ident) => {{
```
dnwiebe, line 5354: I don't see $account_type used anywhere in this macro. Am I missing something?

node/src/accountant/mod.rs
```diff
@@ -5670,60 +5711,16 @@ mod tests {
...
-            vec![CustomQuery::RangeQuery {
-                min_age_s: 2000,
-                max_age_s: 200000,
-                min_amount_gwei: 0,
-                max_amount_gwei: 150000000000,
-                timestamp: actual_timestamp
-            }]
-        );
-        assert!(before <= actual_timestamp && actual_timestamp <= after)
+        let (response, _) = UiFinancialsResponse::fmb(result).unwrap();
```
dnwiebe, line 5717: You're replacing lots of expensive assertions with these two because you're already asserting the complete behavior of the `financials` command elsewhere, yes?

dnwiebe (commented):
Resume at payable_dao.rs:207 green

node/src/accountant/payable_dao.rs
```diff
@@ -198,13 +199,16 @@ impl PayableDao for PayableDaoReal {
...
         .collect()
     }
 
     fn custom_query(&self, custom_query: CustomQuery<u64>) -> Option<Vec<PayableAccount>> {
-        let variant_top = TopStmConfig::new("last_paid_timestamp asc");
-
+        let variant_top = TopStmConfig{
+            limit_clause: "limit :limit_count",
+            gwei_min_resolution_clause: "where (balance_high_b > 0) or ((balance_high_b = 0) and (balance_low_b >= 1000000000))",
+            age_param_name: "last_paid_timestamp asc",
```
dnwiebe, line 210: This isn't just a name anymore.

node/src/accountant/payable_dao.rs
```diff
@@ -815,43 +794,72 @@ mod tests {
         assert_eq!(result, None)
     }
 
-    fn common_setup_of_accounts_for_tests_of_top_records(
+    type InsertPayableHelperFn<'b> =
+        &'b dyn for<'a> Fn(&'a dyn ConnectionWrapper, &'a str, i128, i64, Option<i64>);
+
+    macro_rules! simplified_insert {
```
dnwiebe, line 800: This broke Dan's brain.

node/src/accountant/payable_dao.rs
```diff
@@ -981,54 +989,48 @@ mod tests {
...
         let now = now_time_t();
-        let timestamp1 = now - 70_000;
-        let timestamp2 = now - 55_120;
-        let timestamp3 = now - 200_001;
-        let timestamp4 = now - 19_999;
-        let timestamp5 = now - 30_786;
-        let timestamp6 = now - 100_401;
-        let timestamp7 = now - 80_333;
-        let main_setup = |insert: &dyn Fn(&str, i128, i64, Option<i64>)| {
+        let main_setup = |conn: &dyn ConnectionWrapper, insert: InsertPayableHelperFn| {
```
dnwiebe, line 992: `insert` is already a closure, and might be able to close around the connection so that another layer of closure wasn't needed.

node/src/accountant/payable_dao.rs
```diff
@@ -1077,18 +1079,21 @@ mod tests {
 
     #[test]
     fn range_query_does_not_display_values_from_below_1_gwei() {
-        let timestamp = now_time_t() - 5000;
-        let main_setup = |insert: &dyn Fn(&str, i128, i64, Option<i64>)| {
+        let timestamp_1 = now_time_t() - 11_001;
+        let timestamp_2 = now_time_t() - 5000;
```
dnwiebe, line 1083: I notice you're calling now_time_t() twice. It'd be better to call it once and use the result twice.

node/src/accountant/payable_dao.rs
```diff
@@ -1232,32 +1225,13 @@ mod tests {
...
-            ];
-            eprintln!("timestamp before: {}", timestamp);
-            conn
-                .prepare("insert into payable (wallet_address, balance_high_b, balance_low_b, last_paid_timestamp, pending_payable_rowid) values (?,?,?,?,?)")
-                .unwrap()
-                .execute(params)
-                .unwrap();
-        };
-        main_setup_fn(&insert);
+        main_setup_fn(conn.as_ref(), &insert_record);
```
dnwiebe, line 1234: You suggested a name change to `insert_record_fn`. I think that's a great idea.

dnwiebe (commented):
Resume at receivable_dao.rs:390 green

node/src/lib.rs
```diff
@@ -6,6 +6,7 @@ pub mod sub_lib;
 
 #[macro_use]
 extern crate masq_lib;
+extern crate core;
```
dnwiebe, line 9: Is this necessary?

dnwiebe (commented):
Resume at related_to_financials.rs:64 green

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+pub fn check_fit_from_zero_to_i64max_for_u64<T>(num: &T) -> bool
+where
+    T: Ord + Copy + TryFrom<u64>,
+    u64: TryFrom<T>,
+    <T as TryFrom<u64>>::Error: Debug,
+{
+    match u64::try_from(*num) {
+        Ok(num) => num > i64::MAX as u64,
```
dnwiebe, line 71: Maybe this should be <=? Also, this function may not need to be public.

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+        max_age_s,
+        min_amount_gwei,
+        max_amount_gwei,
+        ..
+    } = query
+    {
+        match (
+            min_age_s > &(i64::MAX as u64),
+            max_age_s > &(i64::MAX as u64),
+            check_fit_from_zero_to_i64max_for_u64(min_amount_gwei),
```
dnwiebe, line 118: Might be able to use this function for all four tuple elements

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+            "Range query for payable: Min age requested \
+         too big. Should be less than or equal to 9223372036854775807, not: 9223372036854775808",
+        )
+    }
+
+    #[test]
+    fn check_query_is_within_tech_limits_catches_error_at_age_max() {
+        let query = CustomQuery::RangeQuery {
+            min_age_s: 32656,
+            max_age_s: i64::MAX as u64 + 3,
```
dnwiebe, line 184: Would be better as +1

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+         too big. Should be less than or equal to 9223372036854775807, not: 9223372036854775810",
+        )
+    }
+
+    #[test]
+    fn check_query_is_within_tech_limits_catches_error_at_amount_min() {
+        let query = CustomQuery::RangeQuery {
+            min_age_s: 32656,
+            max_age_s: 4545555,
+            min_amount_gwei: i64::MAX as u64 + 9,
```
dnwiebe, line 202: Would be better as +1

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+        )
+    }
+
+    #[test]
+    fn check_query_is_within_tech_limits_catches_error_at_amount_max() {
+        let query = CustomQuery::RangeQuery {
+            min_age_s: 32656,
+            max_age_s: 4545555,
+            min_amount_gwei: 144,
+            max_amount_gwei: i64::MAX as u64 + 11,
```
dnwiebe, line 220: ...again...

node/src/accountant/related_to_financials.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn check_query_is_within_tech_limits_works_for_negative_values() {
+        let query = CustomQuery::RangeQuery {
+            min_age_s: 32656,
+            max_age_s: 4545555,
+            min_amount_gwei: -500000,
+            max_amount_gwei: -500,
+            timestamp: SystemTime::now(),
+        };
+
+        let result = check_query_is_within_tech_limits(&query, "payable", 789);
```
dnwiebe, line 241: Would be better as "receivable"

node/src/actor_system_factory.rs
```diff
@@ -1826,49 +1825,91 @@ mod tests {
...
+                _ => low(high(previous)),
+            }
+        }
+
+        let file_path = current_dir()
+            .unwrap()
+            .join("src")
+            .join("accountant")
+            .join("receivable_dao.rs");
+        let file = File::open(file_path).unwrap();
```
dnwiebe, line 1875: Make sure that if we're failing here because the tests are running in an environment that doesn't contain the source code, it's easy for the victim to figure out what's going on. If there's no source code available, it may not be reasonable to fail this test. (Or it may; I dunno.)

node/src/db_config/persistent_configuration.rs
```diff
@@ -299,56 +309,55 @@ impl PersistentConfiguration for PersistentConfigurationReal {
...
-                .expect("ever-supplied value is missing: neighborhood-mode; database is corrupt!")
-                .as_str(),
+        let primitive_decoder = NULL_DECODER;
+        let complex_type_translator = |val_opt: Option<String>| {
+            NeighborhoodModeLight::from_str(&val_opt.expectv("checked non optional value"))
+                .map_err(PersistentConfigError::UninterpretableValue)
+        };
+        self.normalized_get(
+            "neighborhood_mode",
+            primitive_decoder,
```
dnwiebe, line 351: Why not NULL_DECODER here?

node/src/db_config/persistent_configuration.rs
```diff
@@ -6,10 +6,11 @@ use crate::blockchain::bip32::Bip32ECKeyProvider;
 use crate::blockchain::bip39::{Bip39, Bip39Error};
```
dnwiebe, line 6: If you decide to put this back the way it was, I won't be insulted.

I understand the attraction of cleverness in code, but simplicity is better.

node/src/database/db_migrations.rs
```diff
@@ -535,17 +532,16 @@ impl DatabaseMigration for Migrate_6_to_7 {
         migration_carrier.retype_table(
             "pending_payable",
             "amount",
-            "transaction_hash text not null,
+            "rowid integer primary key,
+                              transaction_hash text not null,
```
dnwiebe, line 536: Indentation is a little strange

node/src/database/db_migrations.rs
```diff
@@ -649,42 +649,71 @@ impl<'a> Migrate_6_to_7_carrier<'a> {
...
-                    })
-                    .collect::<Vec<i64>>(),
-            ))),
-        };
-        self.statements.push(Box::new(statement_with_params));
+    fn fill_compensatory_table(&mut self, all_big_int_values_found: Vec<(i64, i64)>, table: &str) {
+        let sql_stm = format!(
+            "insert into compensatory_{} (old_rowid, high_bytes, low_bytes) values {}",
+            table,
+            (0..all_big_int_values_found.len() - 1)
```
dnwiebe, line 656: This hunk of code is a lot confusing.

node/src/database/db_initializer.rs
dnwiebe: Granted.

node/src/database/db_initializer.rs
```diff
@@ -1157,136 +1138,97 @@ mod tests {
...
-                .unwrap();
-            let tx = conn.transaction().unwrap();
-            let by_default = ask_pragma(tx.deref());
-            let mut conn = modifier().unwrap();
-            let test_tx = conn.transaction().unwrap();
-            ending_assertion(test_tx.deref(), by_default)
-        }
+    const PRAGMA_CASE_SENSITIVE: &str = "case_sensitive_like";
+
+    fn assert_case_sensitiveness_has_been_switched_turned_on(
```
dnwiebe, line 1146: You only need one of "switched" and "turned"

node/src/database/db_initializer.rs
```diff
@@ -1476,28 +1412,56 @@ mod tests {
         let _ = bring_db_0_back_to_life_and_return_connection(&data_dir.join(DATABASE_FILE));
         let subject = DbInitializerReal::default();
 
-        let _ = subject.initialize(
-            &data_dir,
-            false,
+        let _ = subject.initialize(&data_dir, DbInitializationConfig::panic_on_migration());
+    }
+
+    fn test_if_new_database_was_created(home_dir: &Path) {
```
dnwiebe, line 1418: How about "assert_new_database_was_not_created?"

node/src/database/db_initializer.rs
```diff
@@ -1476,28 +1412,56 @@ mod tests {
...
+        let mut flags = OpenFlags::empty();
+        flags.insert(OpenFlags::SQLITE_OPEN_READ_ONLY);
+        let result = Connection::open_with_flags(home_dir.join(DATABASE_FILE), flags);
+        match result.err().unwrap() {
+            Error::SqliteFailure(_, _) => (),
+            x => panic!("Expected SqliteFailure, got {:?}", x),
+        }
+    }
+
+    fn common_body_for_testing_db_creation_under_unfavorable_conditions(data_dir: &Path) {
```
dnwiebe, line 1428: assert_that_database_is_not_created_by_certain_initialization_configs

node/src/database/db_initializer.rs
```diff
@@ -1578,24 +1539,24 @@ mod tests {
...
+            special_conn_configuration: pointers,
+            init_mode: InitializationMode::CreationBannedMigrationPanics,
         }
     }
 
     fn make_config_with_specific_migration_arguments() -> DbInitializationConfig {
         DbInitializationConfig {
-            special_conn_setup: vec![|_: &_| {
+            special_conn_configuration: vec![|_: &_| {
                 5;
```
dnwiebe, line 1550: This 5; should probably go away

node/src/test_utils/database_utils.rs
```diff
@@ -176,65 +165,59 @@ pub fn assert_table_created_as_strict(conn: &dyn ConnectionWrapper, table_name:
...
-    expected
-        .iter()
-        .find(|hash_set_expected| {
-            single_actual_hash_set
-                .symmetric_difference(&hash_set_expected)
-                .collect_vec()
-                .is_empty()
-        })
-        .is_some()
+fn contains_particular_list_of_key_words(input: (&SQLLinesChoppedIntoWords, &[String])) {
```
dnwiebe, line 188: maybe two parameters?

node/tests/financials_test.rs
```diff
@@ -97,12 +90,17 @@ fn financials_command_retrieves_payable_and_receivable_records() {
     assert_eq!(receivable[0].balance_gwei, amount_receivable_1 as i64);
     assert_eq!(receivable[1].wallet, wallet_receivable_2.to_string());
     assert_eq!(receivable[1].balance_gwei, amount_receivable_2 as i64);
+    let act_phase_time_period = after.duration_since(before).unwrap().as_secs() + 1;
     let age_payable = payable[0].age_s;
-    assert!(678 <= age_payable && age_payable <= 678 + 5);
+    assert!(678 >= age_payable && age_payable <= (age_payable + act_phase_time_period));
```
dnwiebe, line 95: Probably should be 678 <= age_payable?

node/tests/financials_test.rs
```diff
@@ -97,12 +90,17 @@ fn financials_command_retrieves_payable_and_receivable_records() {
...
     let age_receivable_1 = receivable[0].age_s;
-    assert!(age_receivable_1 <= 0 + 5);
+    assert!(
+        0 >= age_receivable_1 && age_receivable_1 <= (age_receivable_1 + act_phase_time_period)
```
dnwiebe, line 98: Again: <= instead of >=

node/tests/financials_test.rs
```diff
@@ -97,12 +90,17 @@ fn financials_command_retrieves_payable_and_receivable_records() {
...
+    );
     let age_receivable_2 = receivable[1].age_s;
-    assert!(1111 <= age_receivable_2 && age_receivable_2 <= 1111 + 5);
+    assert!(
+        1111 >= age_receivable_2 && age_receivable_2 <= (age_receivable_2 + act_phase_time_period)
```
dnwiebe, line 102: Again: <= instead of >=

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -1972,6 +1988,52 @@ mod tests {
...
+            configure_default_persistent_config(ACCOUNTANT_CONFIG_PARAMS | MAPPING_PROTOCOL)
+                .set_payment_thresholds_result(Ok(()));
+
+        let result = configure_accountant_config(
+            &multi_config,
+            &mut bootstrapper_config,
+            &mut persistent_config,
+        );
+
+        let expected_msg = "Value of DebtThresholdGwei (19999) must be bigger than of PermanentDebtAllowedGwei (20000)";
```
dnwiebe, line 2009: Maybe this error message could be improved. Either take out the "of" or precede it with "the value."

PR #192, opened 2022-11-22
GH 633 adolescent Review

utkarshg6:


node/src/proxy_server/mod.rs
```diff
@@ -3911,11 +3911,69 @@ mod tests {
...
-                unreachable_host_name_opt: Some("server.com".to_string())
+                metadata_change: NRMetadataChange::AddUnreachableHost {
+                    host_name: "server.com".to_string()
+                }
             }
         );
     }
 
+    #[test]
+    fn handle_dns_resolve_failure_does_not_sends_message_to_neighborhood_when_server_is_not_specified(
```
bertllll: ...eh...maybe "does not send message" instead of "sends"?

node/src/proxy_server/mod.rs
```diff
@@ -3911,11 +3911,69 @@ mod tests {
...
+            alias_cryptde(),
+            true,
+            Some(STANDARD_CONSUMING_WALLET_BALANCE),
+            false,
+        );
+        let stream_key = make_meaningless_stream_key();
+        let socket_addr = SocketAddr::from_str("1.2.3.4:5678").unwrap();
+        subject
+            .keys_and_addrs
+            .insert(stream_key.clone(), socket_addr.clone());
```
bertllll: I guess socket address is Copy. You can remove the .clone().

node/src/proxy_server/mod.rs
```diff
@@ -4055,7 +4113,12 @@ mod tests {
         system.run();
 
         TestLogHandler::new().exists_log_containing(
-            format!("Discarding DnsResolveFailure message for <unspecified server> from an unrecognized stream key {:?}", stream_key).as_str());
+            format!(
+                "Discarding DnsResolveFailure message for <unspecified_server> from an unrecognized stream key {:?}",
+                stream_key
+            )
+            .as_str(),
```
bertllll: We usually use the ampersand because it's small and decent... but if you insist on this...

bertllll (approved):
There are some light suggestions but it would be nice if you go through them anyway.
Otherwise from me, this is good enough to approve it. But I don't like that the wild tests are not happening.
 
Because they discover something changes will have to be submitted... and maybe new review needs to be done...

PR #193, opened 2022-11-28
GH-600: Add get_block_number

masqrauder:
* Introduce the ability to get the current block number of a chain.
* Utilize batch transport to request block number and transaction logs within the same blockchain service provider API batch request.
* Incrementally scan the blockchain for payments and use the upper block number bound as required.
* Handle block number range errors reported by blockchain service providers to allow recovery and catch up to the current block if the node falls behind.

### DIFF
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -336,7 +336,7 @@ impl BlockchainBridge {
             Err (e) => panic! ("Cannot retrieve start block from database; payments to you may not be processed: {:?}", e)
         };
         let max_block_count = match self.persistent_config.max_block_count() {
-            Ok(mbc) => mbc,
+            Ok(Some(mbc)) => mbc,
             _ => u64::MAX,
         };
         let end_block = match self.blockchain_interface.get_block_number() {
@@ -393,7 +393,7 @@ impl BlockchainBridge {
                 if let Some(max_block_count) = self.extract_max_block_count(e.clone()) {
                     debug!(self.logger, "Writing max_block_count({})", max_block_count);
                     self.persistent_config
-                        .set_max_block_count(max_block_count)
+                        .set_max_block_count(Some(max_block_count))
                         .map_or_else(
                             |_| {
                                 warning!(self.logger, "{} update max_block_count to {}. Scheduling next scan with that limit.", e, max_block_count);
@@ -1325,7 +1325,7 @@ mod tests {
             )))
             .get_block_number_result(LatestBlockNumber::Ok(U64::from(1234u64)));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(100_000))
+            .max_block_count_result(Ok(Some(100_000)))
             .start_block_result(Ok(5)); // no set_start_block_result: set_start_block() must not be called
         let mut subject = BlockchainBridge::new(
             Box::new(blockchain_interface),
@@ -1620,7 +1620,7 @@ mod tests {
             )));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1711,7 +1711,7 @@ mod tests {
             .get_block_number_result(latest_block_number);
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1778,7 +1778,7 @@ mod tests {
             .get_block_number_result(Ok(0u64.into()));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1861,7 +1861,7 @@ mod tests {
         let persistent_config = PersistentConfigurationMock::new()
             .start_block_result(Ok(1234))
             .set_start_block_result(Err(PersistentConfigError::TransactionError))
-            .max_block_count_result(Ok(10000u64));
+            .max_block_count_result(Ok(Some(10000u64)));
         let blockchain_interface = BlockchainInterfaceMock::default()
             .get_block_number_result(Ok(0u64.into()))
             .retrieve_transactions_result(Ok(RetrievedBlockchainTransactions {
@@ -2045,8 +2045,6 @@ mod tests {
 
     #[test]
     fn extract_max_block_range_from_pokt_error_response() {
-        // [{"jsonrpc":"2.0","id":4,"result":"0x2400ee1","error":{}},{"jsonrpc":"2.0","id":5,"error":{"code":-32001,"message":"Relay request failed validation: invalid relay request: eth_getLogs block range limit (100000 blocks) exceeded"}}]
-        // BlockchainError::QueryFailed("Rpc(Error { code: ServerError(-32001), message: \"Relay request failed validation: invalid relay request: eth_getLogs block range limit (100000 blocks) exceeded\", data: None })"
         let result = BlockchainError::QueryFailed("Rpc(Error { code: ServerError(-32001), message: \"Relay request failed validation: invalid relay request: eth_getLogs block range limit (100000 blocks) exceeded\", data: None })".to_string());
         let subject = BlockchainBridge::new(
             Box::new(BlockchainInterfaceMock::default()),
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -366,7 +366,7 @@ impl BlockchainBridge {
         match retrieved_transactions {
             Ok(transactions) => {
                 debug!(
-                    logger,
+                    self.logger,
                     "Write new start block: {}", transactions.new_start_block
                 );
                 if let Err(e) = self
@@ -521,7 +521,7 @@ impl BlockchainBridge {
 
     pub fn extract_max_block_count(&self, error: BlockchainError) -> Option<u64> {
         let regex_result =
-            Regex::new(r".* (max:|more than) (?P<max_block_count>\d+)(.*| blocks at once\.)")
+            Regex::new(r".* (max:|more than|allowed for your plan:|limited to) (?P<max_block_count>\d+)(.*| blocks at once\.)")
                 .expect("Invalid regex");
         let max_block_count = match error {
             BlockchainError::QueryFailed(msg) => match regex_result.captures(msg.as_str()) {
@@ -532,7 +532,10 @@ impl BlockchainBridge {
                     },
                     _ => None,
                 },
-                _ => None,
+                None => match msg.as_str() {
+                    "Got invalid response: Expected batch, got single." => Some(1000),
+                    _ => None,
+                },
             },
             _ => None,
         };
@@ -2042,6 +2045,99 @@ mod tests {
         let max_block_count = subject.extract_max_block_count(result);
 
         assert_eq!(Some(100000u64), max_block_count);
+    }
+    /*
+        POKT (Polygon mainnet and mumbai)
+        {"jsonrpc":"2.0","id":7,"error":{"message":"You cannot query logs for more than 100000 blocks at once.","code":-32064}}
+    */
+    /*
+        Ankr
+        {"jsonrpc":"2.0","error":{"code":-32600,"message":"block range is too wide"},"id":null}%
+    */
+    #[test]
+    fn extract_max_block_range_for_ankr_error_response() {
+        let result = BlockchainError::QueryFailed("RPC error: Error { code: ServerError(-32600), message: \"block range is too wide\", data: None }".to_string());
+        let subject = BlockchainBridge::new(
+            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(PersistentConfigurationMock::default()),
+            false,
+            None,
+        );
+        let max_block_count = subject.extract_max_block_count(result);
+
+        assert_eq!(None, max_block_count);
+    }
+
+    /*
+    MaticVigil
+    [{"error":{"message":"Blockheight too far in the past. Check params passed to eth_getLogs or eth_call requests.Range of blocks allowed for your plan: 1000","code":-32005},"jsonrpc":"2.0","id":7},{"error":{"message":"Blockheight too far in the past. Check params passed to eth_getLogs or eth_call requests.Range of blocks allowed for your plan: 1000","code":-32005},"jsonrpc":"2.0","id":8}]%
+    */
+    #[test]
+    fn extract_max_block_range_for_matic_vigil_error_response() {
+        let result = BlockchainError::QueryFailed("RPC error: Error { code: ServerError(-32005), message: \"Blockheight too far in the past. Check params passed to eth_getLogs or eth_call requests.Range of blocks allowed for your plan: 1000\", data: None }".to_string());
+        let subject = BlockchainBridge::new(
+            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(PersistentConfigurationMock::default()),
+            false,
+            None,
+        );
+        let max_block_count = subject.extract_max_block_count(result);
+
+        assert_eq!(Some(1000), max_block_count);
+    }
+
+    /*
+    Blockpi
+    [{"jsonrpc":"2.0","id":7,"result":"0x21db466"},{"jsonrpc":"2.0","id":8,"error":{"code":-32602,"message":"eth_getLogs is limited to 1024 block range. Please check the parameter requirements at  https://docs.blockpi.io/documentations/api-reference"}}]
+    */
+    #[test]
+    fn extract_max_block_range_for_blockpi_error_response() {
+        let result = BlockchainError::QueryFailed("RPC error: Error { code: ServerError(-32005), message: \"eth_getLogs is limited to 1024 block range. Please check the parameter requirements at  https://docs.blockpi.io/documentations/api-reference\", data: None }".to_string());
+        let subject = BlockchainBridge::new(
+            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(PersistentConfigurationMock::default()),
+            false,
+            None,
+        );
+        let max_block_count = subject.extract_max_block_count(result);
+
+        assert_eq!(Some(1024), max_block_count);
+    }
+
+    /*
+    blastapi - completely rejected call on Public endpoint as won't handle eth_getLogs method on public API
+    [{"jsonrpc":"2.0","id":2,"error":{"code":-32601,"message":"Method not found","data":{"method":""}}},{"jsonrpc":"2.0","id":1,"error":{"code":-32600,"message":"Invalid Request","data":{"message":"Cancelled due to validation errors in batch request"}}}] (edited)
+    [8:50 AM]
+    */
+
+    #[test]
+    fn extract_max_block_range_for_blastapi_error_response() {
+        let result = BlockchainError::QueryFailed("RPC error: Error { code: ServerError(-32601), message: \"Method not found\", data: \"'eth_getLogs' is not available on our public API. Head over to https://docs.blastapi.io/blast-documentation/tutorials-and-guides/using-blast-to-get-a-blockchain-endpoint for more information\" }".to_string());
+        let subject = BlockchainBridge::new(
+            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(PersistentConfigurationMock::default()),
+            false,
+            None,
+        );
+        let max_block_count = subject.extract_max_block_count(result);
+
+        assert_eq!(None, max_block_count);
+    }
+
+    #[test]
+    fn extract_max_block_range_for_expected_batch_got_single_error_response() {
+        let result = BlockchainError::QueryFailed(
+            "Got invalid response: Expected batch, got single.".to_string(),
+        );
+        let subject = BlockchainBridge::new(
+            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(PersistentConfigurationMock::default()),
+            false,
+            None,
+        );
+        let max_block_count = subject.extract_max_block_count(result);
+
+        assert_eq!(Some(1000), max_block_count);
     }
 
     #[test]
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -391,10 +391,19 @@ impl BlockchainBridge {
             }
             Err(e) => {
                 if let Some(max_block_count) = self.extract_max_block_count(e.clone()) {
-                    info!(self.logger, "Writing max_block_count({})", max_block_count);
+                    debug!(self.logger, "Writing max_block_count({})", max_block_count);
                     self.persistent_config
                         .set_max_block_count(max_block_count)
-                        .map_or_else(|_| Ok(()), |_| Ok(()))
+                        .map_or_else(
+                            |_| {
+                                warning!(self.logger, "{} update max_block_count to {}. Scheduling next scan with that limit.", e, max_block_count);
+                                Err(format!("{} updated max_block_count to {}. Scheduling next scan with that limit.", e, max_block_count))
+                            },
+                            |e| {
+                                warning!(self.logger, "Writing max_block_count failed: {:?}", e);
+                                Err(format!("Writing max_block_count failed: {:?}", e))
+                            },
+                        )
                 } else {
                     warning!(
                         self.logger,
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -31,17 +31,17 @@ use masq_lib::blockchains::chains::Chain;
 use masq_lib::logger::Logger;
 use masq_lib::messages::ScanType;
 use masq_lib::ui_gateway::NodeFromUiMessage;
+use regex::Regex;
 use std::path::PathBuf;
 use std::time::SystemTime;
 use web3::transports::Http;
-use web3::types::{TransactionReceipt, H256};
-use web3::Transport;
+use web3::types::{BlockNumber, TransactionReceipt, H256};
 
 pub const CRASH_KEY: &str = "BLOCKCHAINBRIDGE";
 
-pub struct BlockchainBridge<T: Transport = Http> {
+pub struct BlockchainBridge {
     consuming_wallet_opt: Option<Wallet>,
-    blockchain_interface: Box<dyn BlockchainInterface<T>>,
+    blockchain_interface: Box<dyn BlockchainInterface>,
     logger: Logger,
     persistent_config: Box<dyn PersistentConfiguration>,
     sent_payable_subs_opt: Option<Recipient<SentPayables>>,
@@ -331,21 +331,49 @@ impl BlockchainBridge {
     }
 
     fn handle_retrieve_transactions(&mut self, msg: RetrieveTransactions) -> Result<(), String> {
-        let start_block = match self.persistent_config.start_block() {
+        let start_block_nbr = match self.persistent_config.start_block() {
             Ok (sb) => sb,
             Err (e) => panic! ("Cannot retrieve start block from database; payments to you may not be processed: {:?}", e)
         };
-        let retrieved_transactions = self
-            .blockchain_interface
-            .retrieve_transactions(start_block, &msg.recipient);
-
+        let max_block_count = match self.persistent_config.max_block_count() {
+            Ok(mbc) => mbc,
+            _ => u64::MAX,
+        };
+        let end_block = match self.blockchain_interface.get_block_number() {
+            Ok(eb) => {
+                if u64::MAX == max_block_count {
+                    BlockNumber::Number(eb)
+                } else {
+                    BlockNumber::Number(eb.as_u64().min(start_block_nbr + max_block_count).into())
+                }
+            }
+            Err(e) => {
+                info!(
+                    self.logger,
+                    "Using 'latest' block number instead of a literal number. {:?}", e
+                );
+                if max_block_count == u64::MAX {
+                    BlockNumber::Latest
+                } else {
+                    BlockNumber::Number((start_block_nbr + max_block_count).into())
+                }
+            }
+        };
+        let start_block = BlockNumber::Number(start_block_nbr.into());
+        let retrieved_transactions =
+            self.blockchain_interface
+                .retrieve_transactions(start_block, end_block, &msg.recipient);
         match retrieved_transactions {
             Ok(transactions) => {
+                debug!(
+                    logger,
+                    "Write new start block: {}", transactions.new_start_block
+                );
                 if let Err(e) = self
                     .persistent_config
                     .set_start_block(transactions.new_start_block)
                 {
-                    panic! ("Cannot set start block in database; payments to you may not be processed: {:?}", e)
+                    panic! ("Cannot set start block {} in database; payments to you may not be processed: {:?}", transactions.new_start_block, e)
                 };
                 if transactions.transactions.is_empty() {
                     debug!(self.logger, "No new receivable detected");
@@ -361,10 +389,23 @@ impl BlockchainBridge {
                     .expect("Accountant is dead.");
                 Ok(())
             }
-            Err(e) => Err(format!(
-                "Tried to retrieve received payments but failed: {:?}",
-                e
-            )),
+            Err(e) => {
+                if let Some(max_block_count) = self.extract_max_block_count(e.clone()) {
+                    self.persistent_config
+                        .set_max_block_count(max_block_count)
+                        .map_or_else(|_| Ok(()), |_| Ok(()))
+                } else {
+                    warning!(
+                        self.logger,
+                        "Attempted to retrieve received payments but failed: {:?}",
+                        e
+                    );
+                    Err(format!(
+                        "Attempted to retrieve received payments but failed: {:?}",
+                        e
+                    ))
+                }
+            }
         }
     }
 
@@ -477,6 +518,26 @@ impl BlockchainBridge {
             .as_ref()
             .expect("Accountant unbound")
     }
+
+    pub fn extract_max_block_count(&self, error: BlockchainError) -> Option<u64> {
+        let regex_result =
+            Regex::new(r".* (max:|more than) (?P<max_block_count>\d+)(.*| blocks at once\.)")
+                .expect("Invalid regex");
+        let max_block_count = match error {
+            BlockchainError::QueryFailed(msg) => match regex_result.captures(msg.as_str()) {
+                Some(captures) => match captures.name("max_block_count") {
+                    Some(m) => match m.as_str().parse::<u64>() {
+                        Ok(value) => Some(value),
+                        Err(_) => None,
+                    },
+                    _ => None,
+                },
+                _ => None,
+            },
+            _ => None,
+        };
+        max_block_count
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
@@ -495,7 +556,7 @@ mod tests {
     use crate::blockchain::bip32::Bip32EncryptionKeyProvider;
     use crate::blockchain::blockchain_interface::ProcessedPayableFallible::Correct;
     use crate::blockchain::blockchain_interface::{
-        BlockchainError, BlockchainTransaction, RetrievedBlockchainTransactions,
+        BlockchainError, BlockchainTransaction, LatestBlockNumber, RetrievedBlockchainTransactions,
     };
     use crate::blockchain::test_utils::{make_tx_hash, BlockchainInterfaceMock};
     use crate::db_config::persistent_configuration::PersistentConfigError;
@@ -1245,10 +1306,14 @@ mod tests {
             .system_stop_conditions(match_every_type_id!(ScanError))
             .start()
             .recipient();
-        let blockchain_interface = BlockchainInterfaceMock::default().retrieve_transactions_result(
-            Err(BlockchainError::QueryFailed("we have no luck".to_string())),
-        );
-        let persistent_config = PersistentConfigurationMock::new().start_block_result(Ok(5)); // no set_start_block_result: set_start_block() must not be called
+        let blockchain_interface = BlockchainInterfaceMock::default()
+            .retrieve_transactions_result(Err(BlockchainError::QueryFailed(
+                "we have no luck".to_string(),
+            )))
+            .get_block_number_result(LatestBlockNumber::Ok(U64::from(1234u64)));
+        let persistent_config = PersistentConfigurationMock::new()
+            .max_block_count_result(Ok(100_000))
+            .start_block_result(Ok(5)); // no set_start_block_result: set_start_block() must not be called
         let mut subject = BlockchainBridge::new(
             Box::new(blockchain_interface),
             Box::new(persistent_config),
@@ -1273,12 +1338,12 @@ mod tests {
             &ScanError {
                 scan_type: ScanType::Receivables,
                 response_skeleton_opt: None,
-                msg: "Tried to retrieve received payments but failed: QueryFailed(\"we have no luck\")".to_string()
+                msg: "Attempted to retrieve received payments but failed: QueryFailed(\"we have no luck\")".to_string()
             }
         );
         assert_eq!(recording.len(), 1);
         TestLogHandler::new().exists_log_containing(
-            "WARN: BlockchainBridge: Tried to retrieve \
+            "WARN: BlockchainBridge: Attempted to retrieve \
          received payments but failed: QueryFailed(\"we have no luck\")",
         );
     }
@@ -1509,6 +1574,100 @@ mod tests {
     }
 
     #[test]
+    fn handle_retrieve_transactions_uses_latest_block_number_upon_get_block_number_error() {
+        init_test_logging();
+        let retrieve_transactions_params_arc = Arc::new(Mutex::new(vec![]));
+        let system = System::new(
+            "handle_retrieve_transactions_uses_latest_block_number_upon_get_block_number_error",
+        );
+        let (accountant, _, accountant_recording_arc) = make_recorder();
+        let earning_wallet = make_wallet("somewallet");
+        let amount = 42;
+        let amount2 = 55;
+        let expected_transactions = RetrievedBlockchainTransactions {
+            new_start_block: 8675309u64,
+            transactions: vec![
+                BlockchainTransaction {
+                    block_number: 7,
+                    from: earning_wallet.clone(),
+                    wei_amount: amount,
+                },
+                BlockchainTransaction {
+                    block_number: 9,
+                    from: earning_wallet.clone(),
+                    wei_amount: amount2,
+                },
+            ],
+        };
+        let blockchain_interface_mock = BlockchainInterfaceMock::default()
+            .retrieve_transactions_params(&retrieve_transactions_params_arc)
+            .retrieve_transactions_result(Ok(expected_transactions.clone()))
+            .get_block_number_result(LatestBlockNumber::Err(BlockchainError::QueryFailed(
+                "Failed to read the latest block number".to_string(),
+            )));
+        let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
+        let persistent_config = PersistentConfigurationMock::new()
+            .max_block_count_result(Ok(10000u64))
+            .start_block_result(Ok(6))
+            .set_start_block_params(&set_start_block_params_arc)
+            .set_start_block_result(Ok(()));
+
+        let subject = BlockchainBridge::new(
+            Box::new(blockchain_interface_mock),
+            Box::new(persistent_config),
+            false,
+            Some(make_wallet("consuming")),
+        );
+
+        let addr = subject.start();
+        let subject_subs = BlockchainBridge::make_subs_from(&addr);
+        let peer_actors = peer_actors_builder().accountant(accountant).build();
+        send_bind_message!(subject_subs, peer_actors);
+        let retrieve_transactions = RetrieveTransactions {
+            recipient: earning_wallet.clone(),
+            response_skeleton_opt: Some(ResponseSkeleton {
+                client_id: 1234,
+                context_id: 4321,
+            }),
+        };
+        let before = SystemTime::now();
+
+        let _ = addr.try_send(retrieve_transactions).unwrap();
+
+        System::current().stop();
+        system.run();
+        let after = SystemTime::now();
+        let set_start_block_params = set_start_block_params_arc.lock().unwrap();
+        assert_eq!(*set_start_block_params, vec![8675309u64]);
+        let retrieve_transactions_params = retrieve_transactions_params_arc.lock().unwrap();
+        assert_eq!(
+            *retrieve_transactions_params,
+            vec![(
+                BlockNumber::Number(6u64.into()),
+                BlockNumber::Number(10006u64.into()),
+                earning_wallet
+            )]
+        );
+        let accountant_received_payment = accountant_recording_arc.lock().unwrap();
+        assert_eq!(accountant_received_payment.len(), 1);
+        let received_payments = accountant_received_payment.get_record::<ReceivedPayments>(0);
+        check_timestamp(before, received_payments.timestamp, after);
+        assert_eq!(
+            received_payments,
+            &ReceivedPayments {
+                timestamp: received_payments.timestamp,
+                payments: expected_transactions.transactions,
+                response_skeleton_opt: Some(ResponseSkeleton {
+                    client_id: 1234,
+                    context_id: 4321
+                }),
+            }
+        );
+
+        TestLogHandler::new().exists_log_containing("INFO: BlockchainBridge: Using 'latest' block number instead of a literal number. QueryFailed(\"Failed to read the latest block number\")");
+    }
+
+    #[test]
     fn handle_retrieve_transactions_sends_received_payments_back_to_accountant() {
         let retrieve_transactions_params_arc = Arc::new(Mutex::new(vec![]));
         let system =
@@ -1532,11 +1691,14 @@ mod tests {
                 },
             ],
         };
+        let latest_block_number = LatestBlockNumber::Ok(1024u64.into());
         let blockchain_interface_mock = BlockchainInterfaceMock::default()
             .retrieve_transactions_params(&retrieve_transactions_params_arc)
-            .retrieve_transactions_result(Ok(expected_transactions.clone()));
+            .retrieve_transactions_result(Ok(expected_transactions.clone()))
+            .get_block_number_result(latest_block_number);
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
+            .max_block_count_result(Ok(10000u64))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1565,9 +1727,16 @@ mod tests {
         system.run();
         let after = SystemTime::now();
         let set_start_block_params = set_start_block_params_arc.lock().unwrap();
-        assert_eq!(*set_start_block_params, vec![1234]);
+        assert_eq!(*set_start_block_params, vec![1234u64]);
         let retrieve_transactions_params = retrieve_transactions_params_arc.lock().unwrap();
-        assert_eq!(*retrieve_transactions_params, vec![(6, earning_wallet)]);
+        assert_eq!(
+            *retrieve_transactions_params,
+            vec![(
+                BlockNumber::Number(6u64.into()),
+                BlockNumber::Number(1024u64.into()),
+                earning_wallet
+            )]
+        );
         let accountant_received_payment = accountant_recording_arc.lock().unwrap();
         assert_eq!(accountant_received_payment.len(), 1);
         let received_payments = accountant_received_payment.get_record::<ReceivedPayments>(0);
@@ -1592,9 +1761,11 @@ mod tests {
             .retrieve_transactions_result(Ok(RetrievedBlockchainTransactions {
                 new_start_block: 7,
                 transactions: vec![],
-            }));
+            }))
+            .get_block_number_result(Ok(0u64.into()));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
+            .max_block_count_result(Ok(10000u64))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1651,10 +1822,12 @@ mod tests {
         expected = "Cannot retrieve start block from database; payments to you may not be processed: TransactionError"
     )]
     fn handle_retrieve_transactions_panics_if_start_block_cannot_be_read() {
+        let blockchain_interface =
+            BlockchainInterfaceMock::default().get_block_number_result(Ok(0u64.into()));
         let persistent_config = PersistentConfigurationMock::new()
             .start_block_result(Err(PersistentConfigError::TransactionError));
         let mut subject = BlockchainBridge::new(
-            Box::new(BlockchainInterfaceMock::default()),
+            Box::new(blockchain_interface),
             Box::new(persistent_config),
             false,
             None, //not needed in this test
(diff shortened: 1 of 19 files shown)


node/src/blockchain/blockchain_interface.rs
```diff
@@ -254,12 +276,22 @@ where
             )
             .build();
 
-        let log_request = self.web3.eth().logs(filter);
+        let end_block_number = match Option::from(end_block) {
```
bertllll: You're wrapping the value so that you take the covering away just a moment later. Instead, you can ignore the Option and just write:

       let end_block_number = match end_block {
            _ => start_block + 1,
        };

Also, related to the suggestion at args, this name should probably change a bit too, so that it better fits to the choice for the name of the argument up there.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -232,6 +253,7 @@ where
     fn retrieve_transactions(
         &self,
         start_block: u64,
+        end_block: BlockNumber,
```
bertllll, line 323: I do recommend to use a clearer name here, because variables like this one are mentioned a few times through this function. It's not so easy to read and understand. 

My suggestions: `minimal_end_block`, `null_end_block`, `smallest_possible_end_block`...I'm quite sure you will be better at finding an appropriate name.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -254,12 +276,22 @@ where
...
+        let log_request = self.web3_batch.eth().logs(filter);
+
         let logger = self.logger.clone();
-        log_request
-            .then(|logs| {
-                debug!(logger, "Transaction retrieval completed: {:?}", logs);
-                future::result::<RetrievedBlockchainTransactions, BlockchainError>(match logs {
+        match self.web3_batch.transport().submit_batch().wait() {
+            Ok(_) => {
+                let block_number = match block_request.wait() {
```
bertllll: Maybe a name like: `possibly_updated_block_number`...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2169,168 @@ mod tests {
...
+            transport.clone(),
+            make_fake_event_loop_handle(),
+            TEST_DEFAULT_CHAIN,
+        );
+
+        match subject.get_block_number() {
+            Ok(latest_block_number) => assert_eq!(latest_block_number, U64::from(31_682_662u64)),
+            Err(e) => assert_eq!(
+                e,
+                BlockchainError::QueryFailed("unexpected failure".to_string())
```
bertllll: This is a strange way of handling an unfavorable result. 
Instead of an assertion which tends to be used for an expected behavior, it's more appropriate to use panic!(), printing the error, whatever it is, with a simple msg. 

I' making the point because I'm assuming this test focuses on a success on this method call...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2169,168 @@ mod tests {
...
+            ),
+        }
+
+        let mut prepare_params = prepare_params_arc.lock().unwrap();
+        let (method_name, actual_arguments) = prepare_params.remove(0);
+        let actual_arguments: Vec<String> = actual_arguments
+            .into_iter()
+            .map(|arg| serde_json::to_string(&arg).unwrap())
+            .collect();
```
bertllll: Given the params of the prepare method are hard-coded, that is an empty vector, and so you will always get the same result for such a param assertion I'd welcome a simpler way of going through it. 

You can skip the step above, trying to iterate over an empty vector, leaving:

        let (method_name, actual_arguments) = prepare_params.remove(0);
        assert_eq!(method_name, "eth_blockNumber".to_string());
        let expected_arguments: Vec<Value> = vec![];
        assert_eq!(actual_arguments, expected_arguments);

The same should be preferably done with all the following tests too.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2169,168 @@ mod tests {
...
+            .prepare_params(&prepare_params_arc)
+            .send_result(Value::Null);
+        let subject = BlockchainInterfaceNonClandestine::new(
+            transport.clone(),
+            make_fake_event_loop_handle(),
+            TEST_DEFAULT_CHAIN,
+        );
+
+        match subject.get_block_number() {
+            Ok(_) => unimplemented!("Expected an error"),
```
bertllll: panic!() is preferred to unimplemented!() as unimplemented!() usually refers to something yet to be done.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2169,168 @@ mod tests {
...
+
+    #[test]
+    fn non_clandestine_can_fetch_latest_block_number_successfully() {
+        let prepare_params_arc: Arc<Mutex<Vec<(String, Vec<Value>)>>> =
+            Arc::new(Mutex::new(vec![]));
+        let transport = TestTransport::default()
+            .prepare_params(&prepare_params_arc)
+            .send_result(json!("0x1e37066"));
+        let subject = BlockchainInterfaceNonClandestine::new(
+            transport.clone(),
```
bertllll: You use unnecessary clones at this place (for the transport) in all of these tests from here down below.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2169,168 @@ mod tests {
...
+            .map(|arg| serde_json::to_string(&arg).unwrap())
+            .collect();
+
+        assert_eq!(method_name, "eth_blockNumber");
+        let expected_arguments: Vec<String> = vec![];
+        assert_eq!(actual_arguments, expected_arguments);
+    }
+
+    #[test]
+    fn non_clandestine_can_handle_latest_true_block_number_error() {
```
bertllll: To be honest I've got quite a problem with all these last three tests. On one hand, I don't see sense in testing situations which very likely cannot happen at all...

We hardly can expect to receive a boolean, neither of its two values. 

So I'm suggesting to write a single test aiming at only one case of an error. Because all other cases are covered automatically. Their handling depends on foreign code, it has nothing to do with ours. With just one error covered there is no other path to be left untested.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -841,8 +903,15 @@ mod tests {
             TEST_DEFAULT_CHAIN,
         );
 
-        let result = subject
-            .retrieve_transactions(42, &Wallet::new("0x3f69f9efd4f2592fd70beecd9dce71c472fc"));
+        let end_block = match subject.get_block_number() {
+            Ok(block_nbr) => BlockNumber::Number(block_nbr),
+            Err(_) => BlockNumber::Latest,
+        };
```
bertllll: Why do we query this if we can simply create the value ourselves randomly since it has no effect on the act of the test?

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -266,13 +266,18 @@ impl BlockchainBridge {
     }
 
     fn handle_retrieve_transactions(&mut self, msg: &RetrieveTransactions) -> Result<(), String> {
+        let end_block = match self.blockchain_interface.get_block_number() {
+            Ok(eb) => BlockNumber::Number(eb),
+            Err(_) => BlockNumber::Latest,
```
bertllll: I'd appreciate at least a log here, if we have to play down all possible errors...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -254,12 +276,22 @@ where
             )
             .build();
 
-        let log_request = self.web3.eth().logs(filter);
+        let end_block_number = match Option::from(end_block) {
+            Some(BlockNumber::Number(eb)) => eb.as_u64(),
+            _ => start_block + 1,
```
bertllll: I'm quite sure one of these two paths is not ever tested... the second probably...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -841,8 +903,15 @@ mod tests {
             TEST_DEFAULT_CHAIN,
         );
 
-        let result = subject
```
bertllll, line 1037: Ups, I just realized this whole test is a garbage. 
Not your fault but look at the header of it. It say it expects a panic. Then look at the end, it has an assertion there, which is a nonsense. More over, the way this test is built-up is silly because it really doesn't test anything on the code defined in this file. 

It only achieves to blow up within the Wallet object when .address() is called on this wrong and so uninitialized wallet. This could happen anywhere not just here and it would also end up looking exactly the same. With an all-demolishing panic.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -799,14 +857,18 @@ mod tests {
...
             .map(|request| serde_json::from_slice(&request.body()).unwrap())
             .collect();
+        let bodies: Vec<String> = bodies
+            .into_iter()
+            .map(|b| serde_json::to_string(&b).unwrap())
+            .collect();
         assert_eq!(
-            format!("\"0x000000000000000000000000{}\"", &to[2..]),
-            bodies[0]["params"][0]["topics"][2].to_string(),
+            vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])],
```
bertllll: You could use the `r#"abc"#` syntax, allowing you to remove the backslashes.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -254,12 +276,22 @@ where
...
         let logger = self.logger.clone();
-        log_request
-            .then(|logs| {
-                debug!(logger, "Transaction retrieval completed: {:?}", logs);
-                future::result::<RetrievedBlockchainTransactions, BlockchainError>(match logs {
+        match self.web3_batch.transport().submit_batch().wait() {
+            Ok(_) => {
+                let block_number = match block_request.wait() {
+                    Ok(block_nbr) => block_nbr.as_u64(),
+                    Err(_) => end_block_number,
```
bertllll: Again, it looks like one of these one of these two cases doesn't have a test for it.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -266,13 +266,18 @@ impl BlockchainBridge {
...
+        };
```
bertllll, line 361: I'm reluctant to believe that both of those branches of the match are adequately tested. I miss one test.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -254,12 +276,22 @@ where
...
+                };
```
bertllll, line 375: Also, I don't like that we retry this method call all the time as if we can just overlook if it failed first time and it means nothing. Now, if we want to be pulling our luck for the second time (it happened firstly behind the entrance to this function ) it seems awkward and not much trustful.

But the worst situation would be if all those attempts fail ...then we could end up with `start_block + 1` even though we are preparing some found transactions to be send away to Accountant (which means that we're risking to scan them again next time). I know, this situation is rather sci-fi but making it look even slightly possible seems bad.

node/src/blockchain/blockchain_interface.rs
masqrauder: What name do you recommend instead of `end_block_number`? It's really a fallback value that is used in the event there's an error in retrieving the latest block number from the chain. This value may also be changed after GH-585 where the blockchain service might provide a maximum block count delta from the start block as a limitation we have to use on subsequent calls.

node/src/blockchain/blockchain_interface.rs
masqrauder: Yes, this working value name could be better as `response_block_number` because it's in response to our block number request call. This value may be impacted by GH-585 error handling for error responses providing a max block range value.

node/src/blockchain/blockchain_interface.rs
masqrauder: You're right. I should think more in terms of production vs test code.

node/src/blockchain/blockchain_interface.rs
masqrauder: ok. The point of this part of the test was to confirm that there are no arguments passed to the method. I was keeping the tests consistent so that you don't have to mentally switch gears back and forth. I guess this had the opposite effect. If there's a better way to ensure the test breaks appropriately if the arguments change, then I'd like to follow that. I was too focused on trying to get data type held by the vector to be consistent I didn't think of what you have.

node/src/blockchain/blockchain_interface.rs
masqrauder: You're right. Probably some leftovers from how things were before your changes on GH-629. I compared my branch to yours to see if I was way off track.

node/src/blockchain/blockchain_interface.rs
masqrauder: I don't remember if there was a reason for this. I made the change you suggested.

node/src/blockchain/blockchain_interface.rs
masqrauder: I take your point. I'll delete the last 2 tests dealing with booleans and just have the one to cover invalid data. I do think with GH-585 and handling some of the cases that Kauri and Syther have encountered testing with certain Blockchain API endpoint providers we will need to handle receiving a single error response where now we are expecting a batch response. It seems not all the implementations behave exactly the same. I hope to ensure the code is resilient to gracefully handle even the most impossible or ridiculous responses. Stranger things have happened on networks.

node/src/blockchain/blockchain_interface.rs
masqrauder: ok. Deleted.

node/src/blockchain/blockchain_interface.rs
masqrauder: The path with `_ => start_block + 1,` was tested when end_block is passed `BlockNumber::Latest` but I deleted that test from the feedback in the next comment. I added `blockchain_interface_non_clandestine_retrieve_transactions_uses_block_number_latest_as_fallback_start_block_plus_one` to cover this path.

node/src/blockchain/blockchain_interface.rs
masqrauder: I don't think I can still use `format!()` to add the wallet address and the raw rust string, can I?

I can hardcode the address if the escapes are too distracting.

node/src/blockchain/blockchain_interface.rs
masqrauder: this should be covered by `blockchain_interface_non_clandestine_retrieve_transactions_uses_block_number_latest_as_fallback_start_block_plus_one` now.

node/src/blockchain/blockchain_interface.rs
masqrauder: If we encounter some transactions then the latest block number will increase from each transaction and no longer be `start_block + 1`. The next block number will be the max block number found plus 1. I think you're noticing that we still have a little bit of the old non-batch way we requested block number and transactions separately. When I incorporate GH-585 error handling into this, I'll look at only passing `BlockNumber::Latest` if we don't have a range limit imposed by error responses. This way we can avoid the initial `get_block_number()` call outside the batch.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -245,7 +267,7 @@ where
         let filter = FilterBuilder::default()
             .address(vec![self.contract_address()])
             .from_block(BlockNumber::Number(ethereum_types::U64::from(start_block)))
-            .to_block(BlockNumber::Latest)
+            .to_block(end_block)
```
dnwiebe, line 338: Should the debug! message above have the new end_block in it too?

node/src/blockchain/blockchain_interface.rs
dnwiebe: I also notice that `start_block` and `end_block` are of different types. That's surprising to somebody reading the code for the first time; is there an easy way to make it clear why they're different?

dnwiebe (commented):
Resume at blockchain_interface:287 green

node/src/blockchain/blockchain_interface.rs
```diff
@@ -275,40 +307,51 @@ where
                             let transactions: Vec<BlockchainTransaction> = logs
```
dnwiebe: This hunk of code seems like something that could live in an outboard method and make the caller clearer.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -275,40 +307,51 @@ where
...
-                                transactions.iter().fold(start_block, |so_far, elem| {
-                                    if elem.block_number > so_far {
-                                        elem.block_number
-                                    } else {
-                                        so_far
-                                    }
-                                });
+                            // transactions, in which case use end_block, unless get_latest_block()
+                            // was not successful.
+                            let transaction_max_block_number = if transactions.is_empty() {
```
dnwiebe: This hunk of code seems like something that could live in an outboard method and make the caller clearer.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -275,40 +307,51 @@ where
...
+                                        if elem.block_number > so_far {
+                                            elem.block_number
+                                        } else {
+                                            so_far
+                                        }
+                                    })
+                            };
                             Ok(RetrievedBlockchainTransactions {
-                                new_start_block: last_transaction_block + 1,
+                                new_start_block: transaction_max_block_number,
```
dnwiebe: Also, as long as our request wasn't rejected by the blockchain, don't we want to use the current head block + 1 as the new start block in all cases? Suppose we scan 1000 blocks and find a few dozen transactions, with the last one in the 800th block we scan. If that happens, we've already scanned the last 200 blocks in that 1000-block range, and we know there are no transactions there. Why would we want to scan them again next time? Using the current head block in all cases (except where the request fails) might even make things simpler.

node/src/blockchain/blockchain_interface.rs
masqrauder: I could potentially wrap `start_block` and `end_block` in a new block range datatype if you think it warrants it. The web3 interface uses the `BlockNumber` enum for `from_block` and `to_block.` For our use case, if we were to use `BlockNumber` for `start_block` some of the enum values don't make sense like `BlockNumber::Pending` We could interpret `BlockNumber::Earliest` as the MASQ smart contract block number, but that would not fit the true web3 meaning. But then again for the most part we want to use `BlockNumber::Latest` for someone starting node for the first time with a new wallet address.

node/src/blockchain/blockchain_interface.rs
masqrauder: It could have `end_block` in the `debug!` I'll have to confirm how it will handle the variations and should add more tests to confirm.

node/src/blockchain/blockchain_interface.rs
masqrauder: Rescanning blocks would matter less if we change the accountant to keep track of transaction hashes so it knows if it's seen a transaction before. However, the case I'm trying to avoid is the error we receive if we exceed the bounds of `BlockNumber::Latest` meaning calling web3 with the literal block number result from a call with `BlockNumber::Latest` and adding 1 will generate an error if the block number doesn't exist yet. There's a short time period where that may be true. It could be a longer time period if the blockchain service provider is having an issue and the block number isn't increasing. I can try to refine this a bit and confirm we avoid rescanning blocks.

node/src/blockchain/blockchain_interface.rs
bertllll: Good point. You probably can leave it as it is then.

node/src/blockchain/blockchain_interface.rs
bertllll: Actually `fallback_end_block_number` or somehow shorter looks good to me. (I know you've got better sense for the language, so I'm not suggesting nothing but that the word `fallback` would be nice to have in there). What do you think? It might stop making people wonder about what that is.

node/src/blockchain/blockchain_interface.rs
bertllll: Yeah, in general, I'm all for naming variables with more meaningful names, if they are somehow specific or different from other ones. So I think it'd be an improvement.

node/src/blockchain/blockchain_interface.rs
bertllll: You know, for now the assertion testing an empty vector is good enough. If it changes it will definitely breaks. But again, that parameter is so-far constant. One could point out it could change with a different version of the library. 

For clarity, I didn't complain about the fact you want to assert on this parameter I just didn't see a point in converting the potential values to a different, maybe better assertable data types. That step just prolongs the test even though it makes no difference. The vector is empty at any case.

node/src/blockchain/blockchain_interface.rs
bertllll: Maybe you talk from experience which I don't have... In that sense, it's hard for me to evaluate. If the library changes in a bad direction maybe even your tests for wider range of technically allowed inbound data would detect a problem :man_shrugging. But if we were to think of every hypothetical thing we all would go mad.

node/src/blockchain/blockchain_interface.rs
bertllll: Thank you :+1:

node/src/blockchain/blockchain_interface.rs
bertllll: Yeah, my comment was wholly wrong. Badly stated and also wrong in its reasoning. I'm sorry for that :slightly_smiling_face: 

I started to fully understand after reading your additional explanation.

node/src/blockchain/blockchain_interface.rs
bertllll: Would storing the older transactions be safe? I've always thought the answer is no and so that's why we've always tried to find a way not to be obligated to do that. 

But maybe there just hasn't been a good enough discussion about this problem. In the end, all transactions are stored on the blockchain anyway. But they would have to first figure out what is our consuming wallet. Which should be too hard for them.

In other words, yes, it is a security issue to have an unprotected storage of previous transaction in your machine because they could enable uncovering your identity by the repeated consuming wallet in them. So we would probably have to encrypt these entries in the storage. Would it be still workable? Probably not if there were plenty of them.  What do you think would it be sufficient if we stored only the hash of the last transaction we saw? Could we use it to double-check if we hold the right start block?

I don't know. I didn't think much about these cases.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -384,10 +412,58 @@ where
...
+impl<T> BlockchainInterfaceNonClandestine<T>
+where
+    T: 'static + BatchTransport + Debug,
+{
+    fn extract_transactions_from_logs(&self, logs: Vec<Log>) -> Vec<BlockchainTransaction> {
+        logs.iter()
+            .filter_map(|log: &Log| match log.block_number {
+                None => None,
+                Some(block_number) => {
+                    to_gwei(U256::from(log.data.0.as_slice())).map(|gwei_amount| {
```
bertllll: Maybe this idea is silly but what if you put the value `log.data.0` on a different line and give it a name through using a variable. Readability of the code might get better. I'll let you decide if it's worth it. But you know, only the person who will go to the source code of the library will really know what this is exactly.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -384,10 +412,58 @@ where
...
+                        BlockchainTransaction {
+                            block_number: block_number.as_u64(),
+                            from: Wallet::from(log.topics[1]),
+                            gwei_amount,
```
bertllll: One important thing here, by the time of this writing we've got already "in-wei counting" system implemented. That means that the other Nodes will probably pay you something different than just clear gwei values. These numbers moved between Nodes are going to be much more accurate. So you don't want to lose the accuracy here by conversion to gwei. You'll have to change the field's name and the data type to u128, plus get rid of the conversion. 

It might be the case that it will be automatically taken care of by doing the merge (I might have adjusted this code already...well, given your method is now concise and placed alone...the merge might not affect this code...but its mostly the same so maybe yes.). Please check it out anyway.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -640,13 +716,13 @@ mod tests {
             std::env::set_var("SIMPLESERVER_THREADS", "1");
             let (tx, rx) = unbounded();
             let _ = thread::spawn(move || {
-                let bodies_arc = Arc::new(Mutex::new(bodies));
+                let bodies_arc: Arc<Mutex<Vec<Vec<u8>>>> = Arc::new(Mutex::new(bodies));
```
bertllll: This type definition shouldn't be necessary but if you want it to stay, okay.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -640,13 +716,13 @@ mod tests {
...
                 Server::new(move |req, mut rsp| {
                     if req.headers().get("X-Quit").is_some() {
                         panic!("Server stop requested");
                     }
                     tx.send(req).unwrap();
-                    let body = bodies_arc.lock().unwrap().remove(0);
+                    let body: Vec<u8> = bodies_arc.lock().unwrap().remove(0);
```
bertllll: Similarly...
We usually don't do this if the compiler knows the truth and you as a human can read it out of the combination of the function signature and the code in the body. I'm saying that just for your better knowledge. I don't find it serious though and won't request a change.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -719,14 +798,18 @@ mod tests {
...
             .collect();
+        let bodies: Vec<String> = bodies
+            .into_iter()
+            .map(|b| serde_json::to_string(&b).unwrap())
+            .collect();
         assert_eq!(
-            format!("\"0x000000000000000000000000{}\"", &to[2..]),
-            bodies[0]["params"][0]["topics"][2].to_string(),
+            vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])],
+            bodies //bodies[0]["params"][0]["topics"][1].to_string(),
```
bertllll: We have a policy of putting the value being asserted on the left side of the `assert_eq!()` macro.

bertllll: Also this whole structure: 
`vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])]`, 
which you use in the assertion as the template is equal to the variable `bodies`. 

That's why the comment `//bodies[0]["params"][0]["topics"][1].to_string(),` is pointless what more, it's confusing. I'm asking you to remove it.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -799,14 +883,18 @@ mod tests {
...
             .collect();
+        let bodies: Vec<String> = bodies
+            .into_iter()
+            .map(|b| serde_json::to_string(&b).unwrap())
+            .collect();
         assert_eq!(
-            format!("\"0x000000000000000000000000{}\"", &to[2..]),
-            bodies[0]["params"][0]["topics"][2].to_string(),
+            vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])],
+            bodies,
```
bertllll: Please also reverse the asserted value with the asserting one.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -799,14 +883,18 @@ mod tests {
...
         );
         assert_eq!(
             result,
             RetrievedBlockchainTransactions {
-                new_start_block: 0x4be663 + 1,
+                new_start_block: 0x4be663,
```
bertllll: Is this block going to be included in the next scan? Because if yes then it is the same as the block of the last transaction we found and so we would pick that transaction again. Please verify.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -826,12 +914,14 @@ mod tests {
...
+    fn blockchain_interface_non_clandestine_retrieve_transactions_returns_an_error_if_a_response_with_too_few_topics_is_returned(
     ) {
-        let port = 8545;
+        let port = find_free_port();
+        let _test_server = TestServer::start (port, vec![
+            br#"[{"jsonrpc":"2.0","id":2,"result":"0x400"},{"jsonrpc":"2.0","id":3,"result":[{"address":"0xcd6c588e005032dd882cd43bf53a32129be81302","blockHash":"0x1a24b9169cbaec3f6effa1f600b70c7ab9e8e86db44062b49132a4415d26732a","blockNumber":"0x4be663","data":"0x0000000000000000000000000000000000000000000000056bc75e2d63100000","logIndex":"0x0","removed":false,"topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"],"transactionHash":"0x955cec6ac4f832911ab894ce16aa22c3003f46deff3f7165b32700d2f5ff0681","transactionIndex":"0x0"}]}]"#.to_vec()
+        ]);
         let (event_loop_handle, transport) = Http::with_max_parallel(
-            &format!("http://{}:{}", &Ipv4Addr::LOCALHOST, port),
+            &format!("http://{}:{}", &Ipv4Addr::LOCALHOST.to_string(), port),
```
bertllll: `.to_string()` inside the format!() macro is redundant. Format will do it itself.

bertllll (commented):
Continue at blockchain_interface.rs: blockchain_interface_non_clandestine_can_transfer_tokens

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1181,8 +1288,19 @@ mod tests {
...
+        let (method_name, actual_arguments) = send_params.remove(0);
+        let actual_arguments: Vec<String> = actual_arguments
+            .into_iter()
+            .map(|arg| serde_json::to_string(&arg).unwrap())
+            .collect();
+        assert_eq!(method_name, "eth_sendRawTransaction".to_string());
+        assert_eq!(
+            actual_arguments,
+            vec![String::from(
+                r#""0xf8a901851bf08eb00082dbe894384dec25e03f94931767ce4c3556168468ba24c380b844a9059cbb00000000000000000000000000000000000000000000000000626c61683132330000000000000000000000000000000000000000000000000000082f79cd900029a0d4ecb2865f6a0370689be2e956cc272f7718cb360160f5a51756264ba1cc23fca005a3920e27680135e032bb23f4026a2e91c680866047cf9bbadee23ab8ab5ca2""#
```
bertllll: Not sure what the seemingly blank line (or really white spaces?) at the beginning of the line 1301 means. I'd feel better if you had a look at it to check out for some strange formatting and to confirm it's false alarm :slightly_smiling_face:

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1822,15 +1942,22 @@ mod tests {
...
-        transport.assert_request(
-            "eth_getTransactionCount",
-            &[
+        assert_eq!(result, Ok(U256::from(1)));
+        let mut prepare_params = prepare_params_arc.lock().unwrap();
+        let (method_name, actual_arguments) = prepare_params.remove(0);
+        assert!(prepare_params.is_empty());
+        let actual_arguments: Vec<String> = actual_arguments
+            .into_iter()
+            .map(|arg| serde_json::to_string(&arg).unwrap())
```
bertllll: You do this as often that it would have some positive effect if you created a helper function (used only this test module) which would take the a parameter with the type of `actual_arguments` and it would return a `Vec<String>`.

It will take care of duplication and make the tests slightly shorter.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2199,88 @@ mod tests {
 
         assert_eq!(check, vec![11, 22, 33, 44, 55, 66, 77])
     }
+
+    #[test]
+    fn non_clandestine_can_fetch_latest_block_number_successfully() {
+        let prepare_params_arc: Arc<Mutex<Vec<(String, Vec<Value>)>>> =
```
bertllll: The explicit type should not be necessary to understand the test and the compiler will not complain too.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2199,88 @@ mod tests {
...
+        );
+
+        match subject.get_block_number() {
+            Ok(latest_block_number) => assert_eq!(latest_block_number, U64::from(31_682_662u64)),
+            Err(e) => panic!("Unexpected failure {}", e.to_string()),
+        }
+
+        let mut prepare_params = prepare_params_arc.lock().unwrap();
+        let (method_name, actual_arguments) = prepare_params.remove(0);
+        assert_eq!(method_name, "eth_blockNumber".to_string());
```
bertllll, line 2679: It's good practice to add another assertion testing that the container for the params doesn't have more items in it. 

`assert!(prepare_params.is_empty())`

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2199,88 @@ mod tests {
...
+            .prepare_params(&prepare_params_arc)
+            .send_result(json!("0x1e37066"));
+        let subject = BlockchainInterfaceNonClandestine::new(
+            transport,
+            make_fake_event_loop_handle(),
+            TEST_DEFAULT_CHAIN,
+        );
+
+        match subject.get_block_number() {
+            Ok(latest_block_number) => assert_eq!(latest_block_number, U64::from(31_682_662u64)),
```
bertllll: Could you please interpret this expected number also as a hexadecimal number? The understanding is gonna be easier thanks to the fast comparison of those two values, even though as different data types. 

`U64::from(0x1e37066u64)`

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2199,88 @@ mod tests {
...
+            Ok(_) => panic!("Expected an error"),
+            Err(e) => assert_eq!(
+                e,
+                BlockchainError::QueryFailed("Decoder error: Error(\"invalid type: null, expected a 0x-prefixed hex string with length between (0; 16]\", line: 0, column: 0)".to_string())
+            ),
+        }
+
+        let mut prepare_params = prepare_params_arc.lock().unwrap();
+        let (method_name, actual_arguments) = prepare_params.remove(0);
+        assert_eq!(method_name, "eth_blockNumber".to_string());
```
bertllll, line 2705: Please add the additional assertion as I asked above.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2072,4 +2199,88 @@ mod tests {
...
+                e,
+                BlockchainError::QueryFailed(
+                    "Decoder error: Error(\"0x prefix is missing\", line: 0, column: 0)"
+                        .to_string()
+                )
+            ),
+        }
+
+        let mut prepare_params = prepare_params_arc.lock().unwrap();
+        let (method_name, actual_arguments) = prepare_params.remove(0);
```
bertllll, line 2734: Also here.

node/src/blockchain/test_utils.rs
```diff
@@ -132,86 +201,37 @@ impl BlockchainInterfaceMock {
...
     asserted: usize,
-    requests: Rc<RefCell<Vec<(String, Vec<rpc::Value>)>>>,
-    responses: Rc<RefCell<VecDeque<rpc::Value>>>,
+    prepare_params: Arc<Mutex<Vec<(String, Vec<rpc::Value>)>>>,
+    send_params: Arc<Mutex<Vec<(RequestId, rpc::Call)>>>,
+    send_results: RefCell<VecDeque<rpc::Value>>,
+    send_batch_params: Arc<Mutex<Vec<Vec<(RequestId, rpc::Call)>>>>,
+    send_batch_results: RefCell<Vec<Vec<Result<rpc::Value, web3::Error>>>>,
+    //to check if we hold a true descendant of a certain initial instance
+    reference_counter_opt: Option<Arc<()>>,
```
bertllll: You probably don't need the two last lines. I implemented them but your version of the codebase doesn't need it.

It would stay idle for a while and it's better not to add it in now.

node/src/blockchain/test_utils.rs
```diff
@@ -221,31 +241,80 @@ impl Transport for TestTransport {
...
+
+    pub fn send_batch_result(
+        self,
+        batched_responses: Vec<Result<rpc::Value, web3::Error>>,
+    ) -> Self {
+        self.send_batch_results.borrow_mut().push(batched_responses);
+        self
+    }
+
+    pub fn initiate_reference_counter(mut self, reference_arc: &Arc<()>) -> Self {
```
bertllll: Also please delete this method. Your card doesn't need it and I don't want to let it be in master until there is actual usage for it. :slightly_smiling_face:

node/src/blockchain/blockchain_interface.rs
masqrauder: So you're saying `actual` then `expected`? I think these escaped the initial policy switch. I think this code was originally written with `expected` then `actual` order.

bertllll: I see. Yeah, sometimes we still come across some old tests where it's reversed. 

Confirming that this is the preferred style: `assert_eq!(actual, expected)`

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -270,13 +270,25 @@ impl BlockchainBridge {
     }
 
     fn handle_retrieve_transactions(&mut self, msg: &RetrieveTransactions) -> Result<(), String> {
+        let end_block = match self.blockchain_interface.get_block_number() {
+            Ok(eb) => BlockNumber::Number(eb),
+            Err(e) => {
+                warning!(
+                    self.logger,
+                    "Using 'latest' block number instead of a literal number. {:?}",
```
dnwiebe: Is it going to be clear from context what the latest block number is being used for, and what the consequences of that will be?

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1113,6 +1127,95 @@ mod tests {
...
+            .retrieve_transactions_result(Ok(expected_transactions.clone()))
+            .get_block_number_result(LatestBlockNumber::Err(BlockchainError::QueryFailed(
+                "Failed to read the latest block number".to_string(),
+            )));
+        let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
+        let persistent_config = PersistentConfigurationMock::new()
+            .start_block_result(Ok(6))
+            .set_start_block_params(&set_start_block_params_arc)
+            .set_start_block_result(Ok(()));
+
```
dnwiebe, line 1627: If you removed these blank lines but left the ones around the Act section, it'd be quicker to visually locate the Act...

node/src/blockchain/test_utils.rs
dnwiebe: ?

node/src/blockchain/test_utils.rs
```diff
@@ -221,31 +241,80 @@ impl Transport for TestTransport {
...
 
-        let (m, p) = self
-            .requests
-            .borrow()
+        let (m, c) = self
+            .prepare_params
+            .lock()
+            .unwrap()
             .get(idx)
             .expect("Expected result.")
```
dnwiebe: Not your fault, but this message isn't particularly helpful. If a more helpful message isn't forthcoming, this is test code, which means you can use an `.unwrap()` if you want...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -149,7 +157,8 @@ impl BlockchainInterface for BlockchainInterfaceClandestine {
 
     fn retrieve_transactions(
         &self,
-        _start_block: u64,
+        _start_block: BlockNumber,
+        _latest_block: BlockNumber,
```
dnwiebe: Why is this called `_latest_block` here and `_end_block` above?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -383,10 +416,57 @@ where
...
+{
+    fn find_largest_transaction_block_number(
+        &self,
+        response_block_number: u64,
+        transactions: &Vec<BlockchainTransaction>,
+    ) -> u64 {
+        if transactions.is_empty() {
+            response_block_number
+        } else {
+            transactions
```
dnwiebe: Is there already a maximum function somewhere in the Rust standard library (or perhaps in `itertools`) that could make this simpler?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -718,14 +801,18 @@ mod tests {
             .into_iter()
             .map(|request| serde_json::from_slice(&request.body()).unwrap())
             .collect();
+        let bodies: Vec<String> = bodies
```
dnwiebe: Why bother to `.collect()` the first iterator? Why not just have two adjacent `.map()` calls and one `.collect()`? That way you wouldn't have two differently-typed `bodies` variables and it wouldn't be as confusing to read.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -718,14 +801,18 @@ mod tests {
...
+            .into_iter()
+            .map(|b| serde_json::to_string(&b).unwrap())
+            .collect();
         assert_eq!(
-            format!("\"0x000000000000000000000000{}\"", &to[2..]),
-            bodies[0]["params"][0]["topics"][2].to_string(),
+            bodies,
```
dnwiebe: Is there any danger that this assert will be disrupted by JSON formatting? Say, if somehow we get spaces after the commas, or newlines, or indentation, or whatever? If so, it might be better to compare these two as parsed objects rather than JSON strings.

From Bert: you could eliminate a lot of the backslashes if you used `r#"` quotes.

dnwiebe (commented):
Resume at blockchain_interface:822 green.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -840,54 +934,59 @@ mod tests {
...
-        assert_eq!(
-            result.expect_err("Expected an Err, got Ok"),
-            BlockchainError::InvalidResponse
-        );
+        assert_eq!(result, Err(BlockchainError::InvalidResponse));
     }
 
     #[test]
-    fn blockchain_interface_non_clandestine_retrieve_transactions_returns_an_error_if_a_response_with_data_that_is_too_long_is_returned(
+    fn blockchain_interface_non_clandestine_retrieve_transactions_ignores_transaction_logs_that_have_no_block_number(
```
dnwiebe: I understand that transaction logs without block numbers are probably not a pressing real-world problem, and that you have to handle them only because we're not allowed to crash because of something that comes in from outside; but just in case it does happen, would it be worth writing an error log?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -926,15 +1033,23 @@ mod tests {
...
 
+        let start_block = BlockNumber::Number(42u64.into());
         let result = subject.retrieve_transactions(
-            42,
+            start_block,
+            BlockNumber::Latest,
             &Wallet::from_str("0x3f69f9efd4f2592fd70be8c32ecd9dce71c472fc").unwrap(),
         );
 
+        let new_start_block = if let BlockNumber::Number(start_block_nbr) = start_block {
```
dnwiebe: It would have helped me understand this code if this variable had been called something like `expected_fallback_start_block`.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1906,20 +2039,6 @@ mod tests {
         assert_eq!(converted_wei, U256::from_dec_str("1000000000").unwrap());
     }
 
-    #[test]
```
dnwiebe, line 2338: I understand getting rid of the part of this test that tests the behavior of the `U256` type; but why get rid of the part of it that verifies the value of the `GWEI` constant?

node/src/blockchain/blockchain_interface.rs
dnwiebe: +1

Also, the test would be easier to read if you untangled the Assert from the Act.

node/src/blockchain/blockchain_bridge.rs
masqrauder: Maybe it should be informational instead of a warning? It's not meant to sound bad or dangerous; only that we couldn't fetch the current block number and so now we're just going to query using BlockNumber::Latest.

node/src/blockchain/blockchain_interface.rs
masqrauder: Good question. I'll make it the same.

node/src/blockchain/blockchain_interface.rs
masqrauder: Hopefully the change is more readable

node/src/blockchain/blockchain_interface.rs
masqrauder: ok I can combine them.

node/src/blockchain/blockchain_interface.rs
masqrauder: If this breaks because of json formatting I suspect that means the block chain service has changed and we'd want the broken test to warn us of that. The raw jsonrpc response is how we see the data and is very compact without extra spaces and new lines. Do you have another suggestion to do instead?

I saw Bert's comment regarding the `format!()` I could remove the format to eliminate the escaping but this means the `&to[2..]` address would have to be hardcoded in an additional place. Do you prefer it hardcoded?

node/src/blockchain/blockchain_interface.rs
masqrauder: ok, I used that name to take advantage of `new_start_block: new_start_block` initialization shorthand. `expected_fallback_start_block` probably makes the test more readable to see where it's used.

node/src/blockchain/test_utils.rs
masqrauder: Changed to `.unwrap()`

node/src/blockchain/blockchain_interface.rs
masqrauder: I don't know. It's just the formatter adding the indentation.

node/src/blockchain/test_utils.rs
masqrauder: Deleted

node/src/blockchain/blockchain_interface.rs
masqrauder: Do we still need the conversion? Bert said the amounts are in `wei` now. Regardless I restored the test.

node/src/blockchain/blockchain_bridge.rs
dnwiebe: Yeah, it sounds like something that ought to be an `info!()` instead of a `warning!()`.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -297,20 +298,29 @@ where
...
+                        let logs_len = logs.len();
                         if logs
                             .iter()
                             .any(|log| log.topics.len() < 2 || log.data.0.len() > 32)
                         {
                             warning!(
                                 logger,
                                 "Invalid response from blockchain server: {:?}",
-                                logs
+                                logs.clone()
```
dnwiebe: Bert wonders: does the macro really take ownership here?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -325,10 +335,13 @@ where
                             })
                         }
                     }
-                    Err(_) => Ok(RetrievedBlockchainTransactions {
-                        new_start_block: fallback_end_block_number,
-                        transactions: vec![],
-                    }),
+                    Err(e) => {
+                        error!(self.logger, "Retrieving transactions: {}", e.to_string());
```
dnwiebe: I think the `.to_string()` is unnecessary here. Doesn't the `{}` placeholder automatically call the `Display` implementation, which is what `.to_string()` does?

(Bert says, "Hell, yes!")

node/src/blockchain/blockchain_interface.rs
```diff
@@ -785,37 +791,34 @@ mod tests {
...
 
         let requests = test_server.requests_so_far();
-        let bodies: Vec<Value> = requests
+
+        let bodies: Vec<String> = requests
             .into_iter()
             .map(|request| serde_json::from_slice(&request.body()).unwrap())
-            .collect();
-        let bodies: Vec<String> = bodies
             .into_iter()
```
dnwiebe: Pretty sure you don't need this `.into_iter()`; the output of `.map` is already an iterator.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -785,37 +791,34 @@ mod tests {
...
-            .map(|b| serde_json::to_string(&b).unwrap())
+            .map(|b: Value| serde_json::to_string(&b).unwrap())
             .collect();
         assert_eq!(
-            bodies,
-            vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])],
+            bodies[0],
```
dnwiebe: I liked the strength of the assertion that `bodies` was equal to `vec[`-just one thing-`]`, because it included the requirement that there could only be one element in the vector.

Also, you could DRY that back up, if you wanted, by doing something like

`let expected_body_prefix = `-complicated `#r` string ending in zeros-`;`
`let expected_body = format!("{}{}", expected_body_prefix, &to[2..]);`
`assert_eq!(bodies, vec[expected_body]);`

node/src/blockchain/blockchain_interface.rs
```diff
@@ -882,17 +885,14 @@ mod tests {
...
-            .collect();
-        let bodies: Vec<String> = bodies
-            .into_iter()
-            .map(|b| serde_json::to_string(&b).unwrap())
+            .map(|b: Value| serde_json::to_string(&b).unwrap())
             .collect();
         assert_eq!(
-            bodies,
-            vec![format!("[{{\"id\":0,\"jsonrpc\":\"2.0\",\"method\":\"eth_blockNumber\",\"params\":[]}},{{\"id\":1,\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{{\"address\":\"0x384dec25e03f94931767ce4c3556168468ba24c3\",\"fromBlock\":\"0x2a\",\"toBlock\":\"0x400\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x000000000000000000000000{}\"]}}]}}]", &to[2..])],
+            bodies[0],
```
dnwiebe: ...see above...

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1011,6 +1010,10 @@ mod tests {
                 transactions: vec![]
             })
         );
+        let test_log_handler = TestLogHandler::new();
+        test_log_handler.assert_logs_contain_in_order(vec![
```
dnwiebe: This looks as though it would work fine, but why `.assert_logs_contain_in_order` instead of `.exists_log_containing`?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2223,19 +2236,21 @@ mod tests {
             TEST_DEFAULT_CHAIN,
         );
 
-        match subject.get_block_number() {
+        let expected_error = match subject.get_block_number() {
```
dnwiebe: I think you can do `subject.get_block_number().unwrap_err()` here and ditch the `match`, if you want.

dnwiebe (approved):
I noticed a few things, but nothing that would cause real trouble, so I'll approve. It'd be nice if you'd take a look through the comments anyway, though.

node/src/blockchain/blockchain_interface.rs
masqrauder: It doesn't. I removed the `.clone()`

node/src/blockchain/blockchain_interface.rs
masqrauder: I was able to remove the `.into_iter()` between `.map()` calls.

node/src/blockchain/blockchain_interface.rs
masqrauder: I used your recommended changes but could not do so without introducing an `expected_body_suffix` for the closing JSON syntax. It would be nice if `format!()` could work with raw Strings.

node/src/blockchain/blockchain_interface.rs
masqrauder: same as above

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2026,7 +2026,11 @@ mod tests {
 
         assert_eq!(
             actual_error,
-            BlockchainError::QueryFailed(format!("Transport error: Error(Connect, Os {{ code: {}, kind: ConnectionRefused, message: \"Connection refused\" }})", if cfg!(target_os = "macos") { 61 } else { 111 }).to_string())
+            BlockchainError::QueryFailed(format!("Transport error: Error(Connect, Os {{ code: {}, kind: ConnectionRefused, message: \"{}\" }})",
+                                                 if cfg!(target_os = "windows") { 10061 } else if cfg!(target_os = "macos") { 61 } else { 111 },
+                                                 if cfg!(target_os = "windows") { "No connection could be made because the target machine actively refused it." }
+                                                 else { "Connection refused" }).to_string(),
+            )
```
bertllll: By our standard, we'd check out the message probably like this:

    match result {
        Err(BlockchainError::QueryFailed(err_message)) => assert!(
            err_message.contains("Transport error: Error(Connect, Os {{ code:") &&
            err_message.contains("kind: ConnectionRefused"),
            "we got this error msg: {}" but expected specific error fragments, see in the assertion: ,
            err_message
        ),
        Err(e) => panic!("we expected ConnectionRefused error but got this one: {}", e),
        Ok(x) => panic!("we expected an error, but got: {:?}", x),
    };
...to avoid the dynamical, os specific parts.

bertllll (approved):
I'd like to see a change making the assertion simpler and obvious at what is important in the error message. Also without if statements involved that aren't becoming to assertions in general.

Hopefully you will agree :pray:

node/src/blockchain/blockchain_interface.rs
bertllll: I wouldn't give the life for being right on this but I think it would allow you to use `format!()` also with `r#"...text..."...text..."#` literals that contain double quotes and it even can have a placeholder in it like `{}` or `{:?}` and it should still work as long as you supply that many arguments as how many placeholders you put in there. You can try but maybe not worth it.

bertllll (approved):

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -270,13 +270,24 @@ impl BlockchainBridge {
...
         };
-        let retrieved_transactions = self
-            .blockchain_interface
-            .retrieve_transactions(start_block, &msg.recipient);
+        let retrieved_transactions =
+            self.blockchain_interface
+                .retrieve_transactions(start_block, end_block, &msg.recipient);
         match retrieved_transactions {
             Ok(transactions) => {
                 if let Err(e) = self
```
bertllll, line 372: Please put a debug log here announcing the newly picked start block before you write it into the db. It's going to make debugging a lot easier for the future.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -251,13 +275,30 @@ where
...
+
         let logger = self.logger.clone();
-        log_request
-            .then(|logs| {
-                debug!(logger, "Transaction retrieval completed: {:?}", logs);
-                future::result::<RetrievedBlockchainTransactions, BlockchainError>(match logs {
+        match self.web3_batch.transport().submit_batch().wait() {
+            Ok(_) => {
+                let response_block_number = match block_request.wait() {
+                    Ok(block_nbr) => block_nbr.as_u64(),
```
bertllll: Also here, this would be such good information for a debugging person. Please add the debug log saying what current head block was reported from the blockchain.

kauri-hero (approved):
adding approval for PR merge

PR #196, opened 2022-11-29
GH-631:  Gossip Storm Avoidance

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -396,6 +396,10 @@ impl AccessibleGossipRecord {
         self.signed_gossip = signed_gossip;
         self.signature = signature;
     }
+
+    pub fn get_all_inner_neighbors(&self) -> HashSet<&PublicKey> {
```
dnwiebe: I need to look at more code, but at the moment I'm suspicious about this duplication.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -108,6 +108,13 @@ impl NeighborhoodDatabase {
         self.has_half_neighbor(from, to) && self.has_half_neighbor(to, from)
     }
 
+    pub fn get_all_half_neighbors(&self) -> HashSet<&PublicKey> {
```
dnwiebe: Instead of iterating through the entire database, you can just get the root node and ask for all its half neighbors.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -681,6 +688,44 @@ mod tests {
...
 
+    #[test]
+    fn can_retrieve_all_full_and_half_neighbors() {
+        let root_node = make_node_record(1111, true);
+        let full_neighbor_1 = make_node_record(2222, true);
+        let full_neighbor_2 = make_node_record(3333, true);
+        let full_neighbor_3 = make_node_record(4444, true);
+        let half_neighbor_1 = make_node_record(5555, false);
+        let half_neighbor_2 = make_node_record(6666, false);
+        let some_node = make_node_record(7777, false);
```
dnwiebe: Can you call this maybe `not_a_neighbor` or something to clarify its role?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -926,6 +927,26 @@ impl GossipHandler for StandardGossipHandler {
     ) -> GossipAcceptanceResult {
         let initial_neighborship_status =
             StandardGossipHandler::check_full_neighbor(database, gossip_source.ip());
+
+        let agrs_by_key = agrs
```
dnwiebe: Try pulling this logic out into an outboard method to avoid mixing levels of abstraction. Something like this might work:
```
fn compute_patch (
    agrs: ...,
    database: &NeighborhoodDatabase
) -> HashSet<PublicKey> {
        let agrs_by_key = agrs
            .iter()
            .map(|agr| (&agr.inner.public_key, agr))
            .collect::<HashMap<&PublicKey, &AccessibleGossipRecord>>();

        let mut patch: HashSet<PublicKey> = HashSet::new();
        self.compute_patch_recursive( // you'd need to change the name of this function
            &mut patch,
            database.root().public_key(),
            &agrs_by_key,
            DEFAULT_MINIMUM_HOP_COUNT,
            database,
        );
        patch
}
```

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -926,6 +927,26 @@ impl GossipHandler for StandardGossipHandler {
...
+        let mut patch: HashSet<PublicKey> = HashSet::new();
+        self.compute_patch(
+            &mut patch,
+            database.root().public_key(),
+            &agrs_by_key,
+            DEFAULT_MINIMUM_HOP_COUNT,
+            database,
+        );
+
+        let agrs = agrs
```
dnwiebe: There's an argument for moving this logic out and giving it a good name as well.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +983,42 @@ impl StandardGossipHandler {
...
+    fn compute_patch(
+        &self,
+        patch: &mut HashSet<PublicKey>,
+        node: &PublicKey,
+        agrs: &HashMap<&PublicKey, &AccessibleGossipRecord>,
+        hops_remaining: usize,
+        database: &NeighborhoodDatabase,
+    ) {
+        patch.insert(node.clone());
+        if hops_remaining > 0 {
```
dnwiebe: I see an early return below. If you're not opposed to early returns, you could save a level of indentation by doing an early return here: `if hops_remaining == 0 {return}`

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +983,42 @@ impl StandardGossipHandler {
...
+            let neighbors = if patch.len() == 1 {
```
dnwiebe: It occurs to me that this might be easier to read (and not require that comment) if the condition were `node == database.root()` instead of `patch.len() == 1`. (The syntax probably isn't exactly right, but you get the drift.)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +983,42 @@ impl StandardGossipHandler {
...
+                // Root Node
+                database.get_all_half_neighbors()
+            } else {
+                match agrs.get(node) {
+                    Some(agr) => agr.get_all_inner_neighbors(),
```
dnwiebe: I think you can probably handle a BTreeSet being returned from here, because all you need for `neighbors` is something that can be iterated over, and both `HashSet` and `BTreeSet` implement `IntoIterator`.

There's the additional question of whether `NodeRecord::half_neighbor_keys` needs to convert a `BTreeSet` into a `HashSet` as well, but that's not your fault and you're not responsible for it unless you want to be.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,19 +2289,248 @@ mod tests {
...
+        node_a_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_d.public_key(), node_e.public_key());
+        let gossip = GossipBuilder::new(&node_a_db)
+            .node(node_b.public_key(), false)
+            .node(node_c.public_key(), false)
+            .node(node_d.public_key(), false)
+            .node(node_e.public_key(), false)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
```
dnwiebe, line 2352: Bert doesn't like this blank line.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,19 +2289,248 @@ mod tests {
...
+        node_b_db.add_node(node_y.clone()).unwrap();
+        node_b_db.add_node(node_c.clone()).unwrap();
+        node_b_db.add_node(node_d.clone()).unwrap();
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_a.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_y.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_c.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+
+        let gossip = GossipBuilder::new(&node_b_db)
+            .node(node_b.public_key(), false)
```
dnwiebe: Having a `true` here isn't necessary for the test, but it would be more realistic and less surprising.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,19 +2289,248 @@ mod tests {
...
+        node_a_db.add_arbitrary_full_neighbor(node_b.public_key(), node_c.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+        let gossip = GossipBuilder::new(&node_a_db)
+            .node(node_b.public_key(), false)
+            .node(node_c.public_key(), false)
+            .node(node_d.public_key(), false)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
+        let hashmap = agrs
```
dnwiebe: This is so common it might be profitable to put it in a utility function.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2252,7 +2560,7 @@ mod tests {
         assert_eq!(system.run(), 0);
         let recording = recording_arc.lock().unwrap();
         assert_eq!(recording.len(), 0);
-        assert_eq!(result, GossipAcceptanceResult::Accepted);
+        assert_eq!(result, GossipAcceptanceResult::Ignored);
```
dnwiebe, line 2566: Put a comment in here explaining that this is Standard Gossip, even though it looks like a Debut, because it's specifically handled by a `StandardGossipHandler` instead of the `GossipAcceptor` (which would identify it as a Debut), so the test is unrealistic. Also that the Gossip is ignored because Node B isn't in Node A's patch, which matters to a `StandardGossipHandler`.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3249,6 +3557,26 @@ mod tests {
...
+        N1---Dest---Third
+             /  \
+            N4  N5
+
+        Source Database ==>
+        Src---Third---Dest
+
+        Disc
+
+        Expected Destination Database ==>
```
dnwiebe, line 3576: This diagram may be incorrect.

dnwiebe (commented):
Resume at gossip_acceptor.rs:3558 green.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3297,10 +3625,8 @@ mod tests {
         let mut expected_dest_db = src_db.clone();
         expected_dest_db.add_arbitrary_half_neighbor(dest_node.public_key(), src_node.public_key());
         expected_dest_db
-            .node_by_key_mut(disconnected_node.public_key())
```
dnwiebe, line 3300: There's a nontrivial problem here. We need to demonstrate, either in this test or another, that the GossipAcceptor filters out the NodeAddr of incoming Gossip where the NeighborhoodDatabase should not know that NodeAddr. That was the original purpose of the `disconnected_node`. Now the `disconnected_node` cannot serve that purpose because it's being removed altogether by the patch-analysis algorithm.

We need to modify this test so that it has a connected Node that A) is far enough from the receiving Node that it shouldn't have a NodeAddr, and B) has a NodeAddr anyway, or we need to remove the `disconnected_node` from this test (since it serves no purpose now) and add an assertion for this behavior to another test, or we need to write an entirely new test that shows this behavior.

node/src/neighborhood/gossip_acceptor.rs
utkarshg6: I made you left this comment. After checking one more time, I realised that the diagram is accurately depicting what we're doing in the test.

bertllll (approved):

PR #199, opened 2022-12-05
fix: removed failfast false and attempting single vm upload to s3

FinsaasGH:


kauri-hero (approved):
This all looks good
I'm assuming now with this change to be a single file, this will also stop the false s3-release uploading when the v* condition is met!

will be able to test properly in the next release.

kauri-hero:
Strangely 3 status checks and sitting 'waiting for status to be reported', so forcing merge (6 other status checks on the workflows were pass)

PR #202, opened 2022-12-11
GH-631: Review 1

utkarshg6:


node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -984,21 +968,44 @@ impl StandardGossipHandler {
...
         patch.insert(node.clone());
-        if hops_remaining > 0 {
-            let neighbors = if patch.len() == 1 {
-                // Root Node
-                database.get_all_half_neighbors()
+        if hops_remaining == 0 {
+            return;
+        } else {
+            let neighbors = if node == root_node.public_key() {
+                root_node.inner.neighbors.clone()
```
dnwiebe: Maybe these `clone()`s aren't necessary, since we're using `&neighbors` below instead of `neighbors`.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1011,14 +1018,30 @@ impl StandardGossipHandler {
...
+                        agrs,
+                        hops_remaining - 1,
+                        root_node,
+                    )
                 }
             }
         }
     }
 
+    fn filter_agrs(
```
dnwiebe: Maybe a better name here?

dnwiebe (approved):

PR #209, opened 2022-12-16
GH-611: Review 1

utkarshg6:


node/src/accountant/mod.rs
```diff
@@ -200,7 +195,7 @@ impl Handler<ScanForPayables> for Accountant {
     type Result = ();
 
     fn handle(&mut self, msg: ScanForPayables, ctx: &mut Self::Context) -> Self::Result {
-        self.handle_scan_for_payable_request(msg.response_skeleton_opt);
+        self.handle_request_for_scan_for_payable(msg.response_skeleton_opt);
```
bertllll: This is my bad. I know I suggested this, but the mid "for" stands a bit clumsy. Maybe "handle_request_of_scan_for_payable"?

As well as for the two other cases:

handle_request_for_scan_for_pending_payable
handle_request_for_scan_for_receivable

node/src/accountant/mod.rs
```diff
@@ -371,37 +366,31 @@ impl Handler<NodeFromUiMessage> for Accountant {
...
             scan_intervals,
-            suppress_initial_scans_opt: config.suppress_initial_scans_opt,
+            suppress_initial_scans: config.suppress_initial_scans,
             consuming_wallet: config.consuming_wallet_opt.clone(),
             earning_wallet: Rc::clone(&earning_wallet),
             payable_dao: dao_factories.payable_dao_factory.make(),
             receivable_dao: dao_factories.receivable_dao_factory.make(),
             pending_payable_dao: dao_factories.pending_payable_dao_factory.make(),
-            crashable: config.crash_point == CrashPoint::Message,
             scanners: Scanners::new(
```
bertllll: You misunderstood me :) 

I wish you could do this for me:

Instead of:

```
        Accountant {
            scan_intervals,
            suppress_initial_scans_opt: config.suppress_initial_scans_opt,
            suppress_initial_scans: config.suppress_initial_scans,
            consuming_wallet: config.consuming_wallet_opt.clone(),
            earning_wallet: Rc::clone(&earning_wallet),
            payable_dao: dao_factories.payable_dao_factory.make(),
            receivable_dao: dao_factories.receivable_dao_factory.make(),
            pending_payable_dao: dao_factories.pending_payable_dao_factory.make(),
            crashable: config.crash_point == CrashPoint::Message,
            scanners: Scanners::new(
                dao_factories,
                payment_thresholds,
                Rc::new(payment_thresholds),
                Rc::clone(&earning_wallet),
                when_pending_too_long_sec,
                config.when_pending_too_long,
                Rc::clone(&financial_statistics),
            ),
            crashable: config.crash_point == CrashPoint::Message,
            notify_later: NotifyLaterForScanners::default(),
            financial_statistics: Rc::clone(&financial_statistics),
            report_accounts_payable_sub_opt: None,
            retrieve_transactions_sub: None,
            request_transaction_receipts_subs_opt: None,
            report_new_payments_sub: None,
            report_sent_payments_sub: None,
            ui_message_sub: None,
            message_id_generator: Box::new(MessageIdGeneratorReal::default()),
            logger: Logger::new("Accountant"),
        }
    }
```

Rather this:

```
        let scanners = Scanners::new(
                dao_factories,
                payment_thresholds,
                Rc::new(payment_thresholds),
                Rc::clone(&earning_wallet),
                when_pending_too_long_sec,
                config.when_pending_too_long,
                Rc::clone(&financial_statistics),
            ); 

        Accountant {
            scan_intervals,
            suppress_initial_scans_opt: config.suppress_initial_scans_opt,
            suppress_initial_scans: config.suppress_initial_scans,
            consuming_wallet: config.consuming_wallet_opt.clone(),
            earning_wallet: Rc::clone(&earning_wallet),
            payable_dao: dao_factories.payable_dao_factory.make(),
            receivable_dao: dao_factories.receivable_dao_factory.make(),
            pending_payable_dao: dao_factories.pending_payable_dao_factory.make(),
            crashable: config.crash_point == CrashPoint::Message,
            scanners,
            crashable: config.crash_point == CrashPoint::Message,
            notify_later: NotifyLaterForScanners::default(),
            financial_statistics: Rc::clone(&financial_statistics),
            report_accounts_payable_sub_opt: None,
            retrieve_transactions_sub: None,
            request_transaction_receipts_subs_opt: None,
            report_new_payments_sub: None,
            report_sent_payments_sub: None,
            ui_message_sub: None,
            message_id_generator: Box::new(MessageIdGeneratorReal::default()),
            logger: Logger::new("Accountant"),
        }
    }
```

I believe that looks better.

node/src/accountant/mod.rs
```diff
@@ -371,37 +366,31 @@ impl Handler<NodeFromUiMessage> for Accountant {
...
                 dao_factories,
-                payment_thresholds,
+                Rc::new(payment_thresholds),
                 Rc::clone(&earning_wallet),
-                when_pending_too_long_sec,
+                config.when_pending_too_long,
```
bertllll: Please add the ending "sec" back to it, it's handy to know sometimes.

node/src/accountant/mod.rs
```diff
@@ -1246,29 +1195,31 @@ mod tests {
...
-        subject_addr.try_send(ui_message).unwrap();
-        let ui_message = NodeFromUiMessage {
-            client_id: 1234,
-            body: UiScanRequest {
-                scan_type: ScanType::PendingPayables,
-            }
-            .tmb(4321),
-        };
+        subject_addr
+            .try_send(NodeFromUiMessage {
```
bertllll: I probably drag you into this somehow but what I'd like here actually is if you reassigned the two messages to two different variables that would have names like: 

let primary_msg = ...

let repeated_msg = ...

You can gain low but still some benefits from playing with the names ;)



Secondly, I think what I really didn't like and want you to change is that you bring mess the area in the test where the messages are set and put to the queue (by saying try_send), instead, you could keep it off from variable assignments. Simply:

1) define all variables later needed for sending with try_send()

2) just write an uninterrupted sequence of `subject_addr.try_from()` calls...I counted up to three of them.  



Thirdly, you can spare some place by defining a closure or a function for the NodeFromUiMessages.
```
let ui_scan_request = ||NodeFromUiMessage {
                client_id: 1234,
                body: UiScanRequest {
                    scan_type: ScanType::PendingPayables,
                }
                .tmb(4321),
            };
```

Maybe I should've elaborated this more for you, sorry if still confusing.

node/src/accountant/mod.rs
```diff
@@ -1729,6 +1677,9 @@ mod tests {
         system.run();
         let begin_scan_params = begin_scan_params_arc.lock().unwrap();
         let notify_later_receivable_params = notify_later_receivable_params_arc.lock().unwrap();
+        TestLogHandler::new().exists_log_matching(&format!(
```
bertllll: These logs don't need to be `.exists_log_matching()`, you don't don any regex actually. `.exists_log_containing` is enough.

node/src/accountant/mod.rs
```diff
@@ -371,37 +366,31 @@ impl Handler<NodeFromUiMessage> for Accountant {
...
-        let payment_thresholds = Rc::new(
-            config
-                .payment_thresholds_opt
-                .take()
-                .expectv("Payment thresholds"),
-        );
-        let scan_intervals = config.scan_intervals_opt.take().expectv("Scan Intervals");
-        let when_pending_too_long_sec = config
-            .when_pending_too_long_opt
+    pub fn new(mut config: BootstrapperConfig, dao_factories: DaoFactories) -> Accountant {
```
bertllll: This might be confusing at first but think about this carefully. Look at the type of config. It is a fully owned structure and that means that you can do whatever you want to it. It can also dismantles into pieces and nobody will be in the right to stop you.

So... I'm thinking... Look also at the operations here right below the comment. There are things like `take()`, in particular, used twice. If you start to discover why it is optional you might not come to a satisfiable reason for that. The truth is that the reason is past or maybe using the optionality here has never had one. 

I think this was made optional so that it could be split up from the rest of the BootstrapperConfig. 
However, by Rust rules, you should be guaranteed to be able to take that piece anyway, without worrying about any additional construction as this `Option` is. As long as you don't want to refer to that field again or you don't want to move the whole config, it's completely okay to _"tear it out and let the config with a hole after it"_.

That can be done for both:
```
     let payment_thresholds = config
            .payment_thresholds_opt
            .take()
            .expectv("Payment thresholds");
```
and 

`let scan_intervals = config.scan_intervals_opt.take().expectv("Scan Intervals");`

Do that; and so change all attached code.

node/src/accountant/mod.rs
```diff
@@ -776,7 +725,7 @@ impl Accountant {
     }
 
     fn financial_statistics(&self) -> FinancialStatistics {
-        self.financial_statistics.as_ref().borrow().clone()
+        self.financial_statistics.borrow().clone()
```
bertllll: The clone feels itchy to me. I thought it out from some directions. This is a special situation, though, the layers of Rc<Refcell<T>> make it hard to be predicted. 

Well, in the end, I think I find what I want to believe is right: I think nobody forces you actually to do the `.clone()`, 
just use this method to return `&RefCell<FinancialStatistics>` which should be easy to be done by writing 

```
  fn financial_statistics(&self) -> &RefCell<FinancialStatistics> {
        self.financial_statistics.borrow()
```
That allows you to leave the decision of cloning anything for later and you also can decide what filed of it. (This is hypothetical because there aren't expensive fields now).

At the place of usage of this function I'd simply call another borrow() which will get you the same struct to manipulate with (but behind a reference (which makes no difference because we are going to `copy` numbers at any case.) 

I considered also:
1) continuing as &Rc<RefCell<T>>
2) cloning Rc<RefCell<T>>.clone() instead of your clone of the inner value. (should be also cheaper than your deep copy)
3) letting this decision for future coders; now, the structure of `FinancialsStatistics` is very cheap, just a struct with two integers (regardless the future need of i128 and u128 values); but it is very likely that this place will be widened at some point with also other parameters. Then the clone might happen to be an unwanted element.

I think this is an academic debate and so I don't want to request any change directly. Instead, I'd like to let Dan to make the final decision whatever it will be. @dnwiebe

node/src/accountant/mod.rs
```diff
@@ -1858,21 +1813,24 @@ mod tests {
...
+        assert_eq!(begin_scan_params.len(), 2);
         assert_eq!(
             *notify_later_payables_params,
             vec![
                 (
                     ScanForPayables {
                         response_skeleton_opt: None
                     },
-                    Duration::from_millis(97)
+                    Duration::from_millis(100)
```
bertllll: I'm so sorry, this is my fault. Do you see these few test at which I demanded a change from you on these scan intervals values? I think I now see why they were picked out so, it's because they are made on purpose different so we can easily conclude they matches to the settings of the particular interval.  

If this one belongs with payable scanning than all other types should stay 100 and probably payable scan interval should be like 99.

node/src/accountant/mod.rs
```diff
@@ -1882,32 +1840,43 @@ mod tests {
...
         System::current().stop();
-        assert_eq!(system.run(), 0);
+        system.run();
+        let payable_begin_scan_params = payable_begin_scan_params_arc.lock().unwrap();
+        let pending_payable_begin_scan_params =
+            pending_payable_begin_scan_params_arc.lock().unwrap();
+        let receivable_begin_scan_params = receivable_begin_scan_params_arc.lock().unwrap();
+        assert_eq!(payable_begin_scan_params.is_empty(), true);
+        assert_eq!(pending_payable_begin_scan_params.is_empty(), true);
+        assert_eq!(receivable_begin_scan_params.is_empty(), true);
```
bertllll: Ehm...

It's not easy for me to evaluate the optimum in a test like this one but I do think this test would be just good enough and much shorter if you simply omitted all the params assertions. Yes, that means to delete for example:

```
  let payable_begin_scan_params_arc = Arc::new(Mutex::new(vec![]));
...
       let payable_scanner = ScannerMock::new().begin_scan_params(&payable_begin_scan_params_arc);
...
 subject.scanners.payable = Box::new(payable_scanner);
...
let payable_begin_scan_params = payable_begin_scan_params_arc.lock().unwrap();
...
 assert_eq!(payable_begin_scan_params.is_empty(), true);
```

All the above would go away three times as for each scanner type once. 

Please think of it as:
What is the objective of the test...
...it's to prove that scanners don't run under certain conditions...
...if DAO methods are not called we know the scanners don't run...
...if we don't supply prepared results for those DAO methods they cannot be called without causing a panic...
... if we can prove that no panic occurred we can make the test strong enough.  
(By the way, notice of the comment down here it talks about exactly that)

Just remember one important thing. To make this trick work properly you need to make sure that the system didn't encounter a panic and I think for that you need to prove the function call of `system.run()` needs to end politely. That's wy we use `assert_eq!(system.run(), 0)`. So you're going to need it too.

node/src/accountant/mod.rs
```diff
@@ -1684,6 +1623,14 @@ mod tests {
...
         assert_eq!(paid_delinquencies_params.len(), 1);
         assert_eq!(paid_delinquencies_params[0], PaymentThresholds::default());
+        let tlh = TestLogHandler::new();
+        tlh.exists_log_containing("INFO: Accountant: Scanning for payables");
+        tlh.exists_log_containing("INFO: Accountant: Scanning for pending payable");
+        tlh.exists_log_containing(&format!(
+            "INFO: Accountant: Scanning for receivables to {}",
+            earning_wallet
+        ));
+        tlh.exists_log_containing("INFO: Accountant: Scanning for delinquencies");
```
bertllll, line 1626: Two things here:

The way you're phrasing the logs here, almost none of them brings any value because exactly the same messages ca be produced by different tests in this file too which would then satisfy these assertions despite this test is not passing alone. If needed, you ca use the trick with a logger with a specific name given it only in that single test. However...

This is going to be a broader idea but maybe it's good to think it through comprehensively.
I was looking at the three INFO messages and I was thinking how much value do they put up? Imagine the UI notifications - regardless the initial state of things, where they don't work right. 

With scanning coming, we will see three INFO messages like these in the assertions. I can see some value in it, however I was thinking further and asked myself if we see the beginning of the process what will we get the the end of it? I'm not sure but I think it will not of a severity INFO, probably only DEBUG while it talks about the results of the scanning. 
My opinion on this is that the letter case is bloody more valuable for the user than to see just the start point.

**My suggestions:**
I know that I myself am an extremist in this and many people would be pushing so far but I'll tell you what I'd do. 

I'd try to bring in conditions under which you could abstract from the three INFO logs about starting messages into a single one. The problem is that we would have to know how to detect the next round. 
I'd probably:

_a) write a test ensuring that the individual scanners run always in the considered order (for example: pending_payable, payable, receivable)._ 

_b) I'd modify the first scan so that it would log a summary log in the INFO level which would say something like:_

`INFO: Accountant: periodical scanning for pending payable, payable and receivable is starting`
_The issue is how could we be 100% sure about the moment. 
You would have to check for the requester of the first scan. If the_ `response_skeleton_opt` _is empty we can print it if it's not we won't do it._

_Also, this would require some fiddling with what place to choose for the logger and the logic around; it would probably require to be outside of the main functionality of the first scanning process as it doesn't touch only the matter of this particular scanner but it touches all of them as a group._ 

_c) I'd convert the other INFO logs, currently marking the start, into debug or trace logs._

_d) and finally, I'd drive in new significant logs (using INFO) for printing the results, in a simple manner, of each of those scanning processes._ 

After that I'd be happier. I want you to think of these ideas and decide what could be done to improve the user experience. Theoretically you could write a new card for it but I think this is quite closely related to what you've been doing in this card so either approach seems fine.

node/src/accountant/scanners.rs
```diff
@@ -97,6 +100,56 @@ pub enum BeginScanError {
...
+            BeginScanError::CalledFromNullScanner => {
+                if !cfg!(test) {
+                    panic!("Null Scanner shouldn't be running inside production code.")
+                }
+            }
+        }
+    }
+
+    fn timestamp_as_string(timestamp: &SystemTime) -> String {
+        const TIME_FORMATTING_STRING: &str =
```
bertllll: In this place there are two possible issues. 

First I wish you could find the similar code, for formatting time like this, which we use in the logger's general setup. (We teach the logger to write the message with starting with the timestamp) If you find it make your best to make the code sharable and reuse it here. 

If impossible, but I doubt that's the case, you can for example share only the formatting literal constant. 

And if so, and I would've needed to rewrite the functionality, then at least get rid of those `unwrap()`s you left down here.

node/src/accountant/scanners.rs
```diff
@@ -97,6 +100,56 @@ pub enum BeginScanError {
...
+            "[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:3]";
+
+        let offset_date_time = OffsetDateTime::from(*timestamp);
+        offset_date_time
+            .format(&parse(TIME_FORMATTING_STRING).unwrap())
+            .unwrap()
+    }
+
```
bertllll: I know you don't have much liking for this... so this is rather a discussion... 

If I'm considering division of code with call that will be obviously used just at one place but it would still separate functionality in the sake of better readability I do it usually by writing an inner function. (Because I don't want it to be accessible from outside and also I like how close it is to the principal part of the bigger-scope, or "parental" function)

node/src/accountant/scanners.rs
```diff
@@ -140,21 +193,23 @@ impl Scanner<ReportAccountsPayable, SentPayable> for PayableScanner {
...
                 Err(BeginScanError::NothingToProcess)
             }
-            false => Ok(ReportAccountsPayable {
-                accounts: qualified_payables,
-                response_skeleton_opt,
-            }),
+            false => {
+                info!(
+                    logger,
+                    "Chose {} qualified debts to pay",
```
bertllll, line 237: Yeah... this is truly debatable... not sure how much info we should give the user in the standard course of things...maybe more than just a quantity? 

But I agree with the way it is too.

node/src/accountant/test_utils.rs
```diff
@@ -889,8 +889,10 @@ impl PendingPayableDaoFactoryMock {
     }
 }
 
-impl Default for PayableScanner {
-    fn default() -> Self {
+pub type PayableScannerBuilder = PayableScanner;
```
bertllll: Nope. I cannot swallow this bite. 

This is like laying. Plus confusing. :man_shrugging: 

Please just accept that Default is not a method that should be utilized in the testing area. You can do it some other way, just drop the effort to implement this trait for any cost. It's not worth it, trust me.

Default can be put on pure mock objects. You can have a default on something like `PayableScannerMock` (I know you don't have an object like that).  I don't care there.

In this specific case, where you manipulate the real object you most likely want to have simple helper methods like `make_payable_scanner_with_mocks`, `make_receivable_scanner_with_mocks`,...

Or you hopefully will devise a better name. The good principle we have is to use the `make_xxx` format.

node/src/accountant/test_utils.rs
```diff
@@ -911,8 +913,10 @@ impl PayableScanner {
     }
 }
 
-impl Default for PendingPayableScanner {
-    fn default() -> Self {
+pub type PendingPayableScannerBuilder = PendingPayableScanner;
```
bertllll: Also wrong...

node/src/accountant/test_utils.rs
```diff
@@ -935,8 +939,10 @@ impl PendingPayableScanner {
     }
 }
 
-impl Default for ReceivableScanner {
-    fn default() -> Self {
+pub type ReceivableScannerBuilder = ReceivableScanner;
```
bertllll: Also wrong... not solving the issue... just masquerading it

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1021,7 +1021,7 @@ mod tests {
     }
 
     #[test]
-    fn handle_request_transaction_receipts_short_circuits_on_failure_of_the_first_payment_and_it_does_not_send_any_message_just_aborts_and_logs(
+    fn handle_request_transaction_receipts_short_circuits_on_failure_of_the_first_payment_and_it_sends_a_message_with_empty_vector_and_logs(
```
bertllll, line 1024: I was asking myself here: should we send an error mark along with the message we send back to Accountant? 
At the moment, I think, Accountant cannot do anything if it sensed that there has been an error. We don't retry or so. That means, I think, we are not obliged to tell Accountant about it, as long as we tell it at least that the scan ended.

However, you should check what will be the reaction of Accountant after it receives this empty vector...
Will it say something like: "No transaction receipts were found in this scan"? 
Hmm...that'd sound acceptable.

My point is that you should check if it won't continue with a message which together the error situation does not look good or doesn't make much sense. Otherwise no change needed.

node/src/accountant/scanners_tools.rs
```diff
@@ -277,42 +267,64 @@ pub(crate) mod receivable_scanner_tools {
...
 
-    const TIME_FORMATTING_STRING: &str =
-        "[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:3]";
-
-    pub fn timestamp_as_string(timestamp: &SystemTime) -> String {
-        let offset_date_time = OffsetDateTime::from(*timestamp);
-        offset_date_time
-            .format(&parse(TIME_FORMATTING_STRING).unwrap())
-            .unwrap()
+    pub fn update_timestamp_and_log(
```
bertllll: This function is nice but its job doesn't correspond to its name very well. It'd be better to say like: 
"erase_timestamp..." or "zero_timestamp_out_and_log".

node/src/accountant/scanners_tools.rs
```diff
@@ -277,42 +267,64 @@ pub(crate) mod receivable_scanner_tools {
...
+        time_initiated_opt: &mut Option<SystemTime>,
```
bertllll: Also I'm not fully satisfied with this usage of a mutable reference given this function is really specialized a lot and the first argument at your function is general - it can come from wherever - which isn't safe at a function that changes a state as its side effect

A more suitable solution would be to make this simply a method of the ScannerCommon` struct in `scanners.rs`. 

I will elaborate that further in there.

node/src/accountant/scanners_tools.rs
```diff
@@ -531,9 +518,52 @@ mod tests {
...
+        assert_eq!(err, vec![error.clone()]);
+        TestLogHandler::new().exists_log_containing(&format!(
+            "WARN: {}: Outbound transaction failure due to '{:?}",
+            test_name, error
+        ));
+    }
+
+    #[test]
+    #[should_panic(expected = "Unable to calculate elapsed time for the scan.")]
+    fn update_timestamp_and_log_panics_if_timestamp_is_wrong() {
```
bertllll: We don't test the panics from `expect()`. This test doesn't need to be here.

node/src/accountant/scanners_tools.rs
```diff
@@ -531,9 +518,52 @@ mod tests {
...
+
+        update_timestamp_and_log(
+            &mut Some(time_in_future),
+            ScanType::Payables,
+            &Logger::new("test"),
+        );
+    }
+
+    #[test]
+    fn update_timestamp_and_log_if_timestamp_is_correct() {
```
bertllll: change names of these tests if you change the name of the function as suggested

node/src/accountant/scanners_tools.rs
```diff
@@ -531,9 +518,52 @@ mod tests {
...
+        init_test_logging();
+        let test_name = "update_timestamp_and_log_if_timestamp_is_correct";
+        let time_in_past = SystemTime::now().sub(Duration::seconds(10));
+        let logger = Logger::new(test_name);
+
+        update_timestamp_and_log(&mut Some(time_in_past), ScanType::Payables, &logger);
+
+        TestLogHandler::new().exists_log_matching(&format!(
+            "INFO: {test_name}: The Payables scan ended in \\d+ms."
```
bertllll: You can do asserting even better than this. You should prove that the time is with the before and after interval.

You can interpret the `time_in_past` value and require most of it in the assertion with some milliseconds added. But that is dangerous because of special cases where the seconds for example just overflows or something...

If you could find a way of writing a regex checking a number within a range of values it would be awesome to have it here. And as I've gone through many other places of that kind, you could make it a part also of them. I wish there was something like that but because i myself don't know how to do that I leave it up to you, with a little friendly booster of your will to do it.

node/src/accountant/mod.rs
```diff
@@ -1791,23 +1743,26 @@ mod tests {
 
         system.run();
         let begin_scan_params = begin_scan_params_arc.lock().unwrap();
-        assert_eq!(begin_scan_params.len(), 2);
         let notify_later_pending_payable_params =
             notify_later_pending_payable_params_arc.lock().unwrap();
+        TestLogHandler::new().exists_log_matching(&format!(
```
bertllll: No need for `exists_log_matching()` !! It uses regex and is unnecessarily expensive. Use the simpler method asserting only on the string directly.

node/src/accountant/mod.rs
```diff
@@ -1858,21 +1813,24 @@ mod tests {
         //the second attempt is the one where the queue is empty and System::current.stop() ends the cycle
         let begin_scan_params = begin_scan_params_arc.lock().unwrap();
         let notify_later_payables_params = notify_later_payables_params_arc.lock().unwrap();
-        assert_eq!(*begin_scan_params, vec![(), ()]);
+        TestLogHandler::new().exists_log_matching(&format!(
```
bertllll: Again. Stop using these expensive methods for simply-to-achieve assertions.

node/src/accountant/scanners.rs
```diff
@@ -1683,8 +1731,12 @@ mod tests {
         let is_scan_running = subject.scan_started_at().is_some();
         assert_eq!(message_opt, None);
         assert_eq!(is_scan_running, false);
-        TestLogHandler::new().exists_log_matching(&format!(
-            "INFO: {test_name}: The Pending Payable scan ended in \\d+ms."
+        let tlh = TestLogHandler::new();
+        tlh.exists_log_matching(&format!(
```
bertllll: No need for the `matching` assertion. `Existing` is the right fit.

node/src/accountant/scanners.rs
```diff
@@ -97,6 +100,56 @@ pub enum BeginScanError {
...
+        const TIME_FORMATTING_STRING: &str =
+            "[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:3]";
+
+        let offset_date_time = OffsetDateTime::from(*timestamp);
+        offset_date_time
+            .format(&parse(TIME_FORMATTING_STRING).unwrap())
+            .unwrap()
+    }
+
+    fn log(logger: &Logger, log_message: String, is_externally_triggered: bool) {
```
bertllll: To be honest, I find the code easier to read if this piece of code is just inlined. Please do it, since it's only a match statement without any special functionality and anybody sees immediately that it's about logging so giving it a name also doesn't help.

node/src/accountant/scanners.rs
```diff
@@ -32,6 +33,8 @@ use std::cell::RefCell;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 use std::time::SystemTime;
+use time::format_description::parse;
```
bertllll, line 36: This doesn't relate to imports :). I just needed a general place to put my comment...  :man_shrugging: 

Those three last methods of this trait extremely repeat through this file.
You will do right if you use a macro. To justify myself:

a) the methods are very simple, so they can be read easily even from a macro,

b) they really appear in each implementation without any difference 

c) the space (and so the attention) they take is just distraction; the reader should instead pay attention to more meaningful functions such as `begun_scan()` and `end_scan()`. 

I wrote the macro for you, though it's pretty simple. 
```

macro_rules! time_marking_methods {
    ($scan_type_variant: ident) => {
        fn scan_started_at(&self) -> Option<SystemTime> {
            self.common.initiated_at_opt
        }

        fn mark_as_started(&mut self, timestamp: SystemTime) {
            self.common.initiated_at_opt = Some(timestamp);
        }

        fn mark_as_ended(&mut self, logger: &Logger) {
            self.common.zero_timestamp_out_and_log(
                ScanType::$scan_type_variant,
                logger,
            );
        }
    }
} 
```

node/src/accountant/scanners.rs
```diff
@@ -97,6 +100,56 @@ pub enum BeginScanError {
...
+        match is_externally_triggered {
+            true => info!(logger, "{}", log_message),
+            false => debug!(logger, "{}", log_message),
+        }
+    }
+}
+
 pub struct ScannerCommon {
     initiated_at_opt: Option<SystemTime>,
     pub payment_thresholds: Rc<PaymentThresholds>,
```
bertllll, line 148: Again, I cannot write this comment to the right place but please move the fn `update_timestamp_and_log` from `scanner_tools.rs` over here. 
Plus rename it, plus change the signature slightly. All is shown here:
```

impl ScannerCommon {
    fn new(payment_thresholds: Rc<PaymentThresholds>) -> Self {
        Self {
            initiated_at_opt: None,
            payment_thresholds,
        }
    }

    fn zero_timestamp_out_and_log(
        &mut self,
        scan_type: ScanType,
        logger: &Logger,
    ) {
        match self.initiated_at_opt.take() {
            Some(timestamp) => {
                let elapsed_time = SystemTime::now()
                    .duration_since(timestamp)
                    .expect("Unable to calculate elapsed time for the scan.")
                    .as_millis();
                info!(
                    logger,
                    "The {:?} scan ended in {}ms.", scan_type, elapsed_time
                );
            }
            None => {
                error!(
                    logger,
                    "Called scan_finished() for {:?} scanner but timestamp was not found",
                    scan_type
                );
            }
        };
    }
}
```

node/src/accountant/scanners.rs
```diff
@@ -189,20 +244,11 @@ impl Scanner<ReportAccountsPayable, SentPayable> for PayableScanner {
...
-                    .as_millis();
-                info!(logger, "The Payable scan ended in {elapsed_time}ms.");
-                self.common.initiated_at_opt = None;
-            }
-            None => error!(
-                logger,
-                "The scan_finished() was called for Payable scanner but timestamp was not found"
-            ),
-        };
+        update_timestamp_and_log(
```
bertllll: Use the macro. You'll be fine :)

node/src/accountant/scanners.rs
```diff
@@ -357,20 +403,11 @@ impl Scanner<RequestTransactionReceipts, ReportTransactionReceipts> for PendingP
...
-                        .as_millis();
-                    info!(
-                        logger,
-                        "The Pending Payable scan ended in {elapsed_time}ms."
-                    );
-                    self.common.initiated_at_opt = None;
-                }
-                None => error!(logger, "The scan_finished() was called for Pending Payable scanner but timestamp was not found"),
-            };
+        update_timestamp_and_log(
```
bertllll: Use the macro to replace the three methods.

node/src/accountant/scanners.rs
```diff
@@ -621,20 +642,11 @@ impl Scanner<RetrieveTransactions, ReceivedPayments> for ReceivableScanner {
...
-                    .as_millis();
-                info!(logger, "The Receivable scan ended in {elapsed_time}ms.");
-                self.common.initiated_at_opt = None;
-            }
-            None => error!(
-                logger,
-                "The scan_finished() was called for Receivable scanner but timestamp was not found"
-            ),
-        };
+        update_timestamp_and_log(
```
bertllll: Use the macro here too.

node/src/accountant/scanners_tools.rs
```diff
@@ -369,8 +381,10 @@ mod tests {
     fn payable_above_threshold_values_is_marked_qualified_and_returns_threshold() {
         let now = SystemTime::now();
         let payment_thresholds = PaymentThresholds::default();
-        let debt = payment_thresholds.permanent_debt_allowed_gwei + 1_000_000_000;
-        let time = to_time_t(now) - payment_thresholds.maturity_threshold_sec - 1;
+        let debt = payment_thresholds.debt_threshold_gwei - 1;
+        let time = payment_thresholds.maturity_threshold_sec
+            + payment_thresholds.threshold_interval_sec
+            - 1;
```
bertllll, line 351: These values are way safe! If I were you I'd check if there are still tests that test the effect of being above or under the declining part of the payment threshold somehow more seriously.

node/src/accountant/scanners_tools.rs
```diff
@@ -330,7 +342,7 @@ mod tests {
     }
 
     #[test]
-    fn payable_with_low_debt_is_marked_unqualified() {
+    fn payable_with_debt_smaller_than_permanent_debt_allowed_is_marked_unqualified() {
         let now = SystemTime::now();
         let payment_thresholds = PaymentThresholds::default();
         let unqualified_debt = payment_thresholds.permanent_debt_allowed_gwei - 1;
```
bertllll, line 312: Not here but the name of the test `payable_with_low_payout_threshold_is_marked_unqualified` seems cheap. It doesn't say the important thing, that is testing of a debt situated below the declining part of the payment thresholds. It's an important difference from the other tests around here.

node/src/accountant/scanners_tools.rs
```diff
@@ -330,7 +342,7 @@ mod tests {
     }
 
     #[test]
-    fn payable_with_low_debt_is_marked_unqualified() {
+    fn payable_with_debt_smaller_than_permanent_debt_allowed_is_marked_unqualified() {
```
bertllll: Some of these test might be deprecated later when you have the master merged in. I spent a lot of time on testing the thresholds really in deep. So don't hesitate to call me over and I can navigate you through the merge. I'd be tempered if we lost the exacter tests I wrote.

node/src/accountant/scanners.rs
```diff
@@ -331,13 +373,17 @@ impl Scanner<RequestTransactionReceipts, ReportTransactionReceipts> for PendingP
...
         logger: &Logger,
     ) -> Option<NodeToUiMessage> {
-        debug!(
-            logger,
-            "Processing receipts for {} transactions",
-            message.fingerprints_with_receipts.len()
-        );
-        let statuses = self.handle_pending_transaction_with_its_receipt(&message, logger);
-        self.process_transaction_by_status(statuses, logger);
+        if message.fingerprints_with_receipts.len() != 0 {
```
bertllll, line 394: I find `!message.fingerprints_with_receipts.is_empty()` nicer and clearer but that's your decision...

node/src/accountant/scanners.rs
```diff
@@ -966,7 +1023,7 @@ mod tests {
         let (_, _, all_non_pending_payables) = make_payables(now, &PaymentThresholds::default());
         let payable_dao =
             PayableDaoMock::new().non_pending_payables_result(all_non_pending_payables);
-        let mut subject = PayableScanner::default().payable_dao(payable_dao);
+        let mut subject = PayableScannerBuilder::default().payable_dao(payable_dao);
```
bertllll: Yeah...the bad taste from this still is in my mouth. Hoping you will discard the obfuscating new types. Then I will smile again :slightly_smiling_face: .

bertllll (commented):
Please proceed with care.

node/src/accountant/mod.rs
utkarshg6: Actually, we can't pass `dao_factories` too early. Your approach will pass it to the Scanners and the other DAOs inside the Accountant constructors won't be able to make any connections.

node/src/accountant/mod.rs
utkarshg6: Aye Aye captain 🫡

node/src/accountant/mod.rs
utkarshg6: Please refer to this function `merge_unprivileged` and also consider discussing It with Dan. In my opinion I prefer to wrap it with an `Option`, because of the certain functions that are populating its value.

node/src/accountant/mod.rs
utkarshg6: I can do as you say, but it'll raise other problems like this:

```bash
    |
603 |         let total_paid_payable = financial_statistics.total_paid_payable;
    |                                                       ^^^^^^^^^^^^^^^^^^ unknown field
```

I can fix that using `into_inner()`, but I question myself, is this complexity really worth it?

Since, it is an easy structure now, we can use `clone()` without any trouble.

node/src/accountant/mod.rs
utkarshg6: Haha! I feel like a tennis ball. I know it can happen with anyone. Don't worry I'll fix it, but recheck again, because I may miss some places.

node/src/accountant/mod.rs
utkarshg6: Again I feel like a 🎾, I thought you wanted that, that's why I added it. Are you sure, that it's really that bad that I should remove it?

node/src/accountant/mod.rs
utkarshg6: I created a new card, because I'm expecting more changes like this - https://github.com/MASQ-Project/Node-issues/issues/661

node/src/accountant/scanners.rs
utkarshg6: The function that is a parent of this function is used in the test module. Due to that, I only imported the constant that we talked about in the previous comment.

node/src/accountant/scanners.rs
utkarshg6: I've added it to the newly created card - https://github.com/MASQ-Project/Node-issues/issues/661

node/src/accountant/test_utils.rs
utkarshg6: I truly think it's safe to use, since it's a builder type, it doesn't matter whether we use default or some function. I think you're confusing that Default must always be used in production and never in test.

node/src/accountant/test_utils.rs
utkarshg6: I've a completely different opinion.

node/src/accountant/test_utils.rs
utkarshg6: Again! I have a completely different opinion.

node/src/blockchain/blockchain_bridge.rs
utkarshg6: It is nicely taken care in a generalised manner.

node/src/accountant/scanners_tools.rs
utkarshg6: I really liked this change. I'm glad that you pointed out.

node/src/accountant/scanners_tools.rs
utkarshg6: I don't get it. Is there a different name you want to suggest? If yes, what name seems good to you?

node/src/accountant/scanners.rs
utkarshg6: I find new types as a cleaner implementation. Since, its definition and implementation resided inside the test module, it won't get used anywhere in the production code.

Since, I have a different opinion so I'd like to discuss it with @dnwiebe, and to know his opinion.

node/src/accountant/scanners_tools.rs
utkarshg6: I see. Still, I don't know which values will get to the corner case.

node/src/accountant/scanners_tools.rs
utkarshg6: I think we should keep it. In the end, it's just another safety measure.

node/src/accountant/scanners_tools.rs
utkarshg6: The `time` is tested in other tests. Here's one example which accurately tests the time `payable_scanner_can_initiate_a_scan`. Hence, I don't feel the need to add an approximate assertion to this test.

In the end I believe it's better to limit the test to assertions for which it was written on the first place, instead of making it over complicated by adding tighter assertions.

node/src/accountant/mod.rs
bertllll: I see. Still, nothing should prevent you from writing:

```
        let payable_dao =  dao_factories.payable_dao_factory.make();
        let receivable_dao = dao_factories.receivable_dao_factory.make(),
        let pending_payable_dao = dao_factories.pending_payable_dao_factory.make();

        let scanners = Scanners::new(
                dao_factories,
                payment_thresholds,
                Rc::new(payment_thresholds),
                Rc::clone(&earning_wallet),
                when_pending_too_long_sec,
                config.when_pending_too_long,
                Rc::clone(&financial_statistics),
            ); 

        Accountant {
            scan_intervals,
            suppress_initial_scans_opt: config.suppress_initial_scans_opt,
            suppress_initial_scans: config.suppress_initial_scans,
            consuming_wallet: config.consuming_wallet_opt.clone(),
            earning_wallet: Rc::clone(&earning_wallet),
            payable_dao,
            receivable_dao,
            pending_payable_dao,
            crashable: config.crash_point == CrashPoint::Message,
            scanners,
            crashable: config.crash_point == CrashPoint::Message,
            notify_later: NotifyLaterForScanners::default(),
            financial_statistics: Rc::clone(&financial_statistics),
            report_accounts_payable_sub_opt: None,
            retrieve_transactions_sub: None,
            request_transaction_receipts_subs_opt: None,
            report_new_payments_sub: None,
            report_sent_payments_sub: None,
            ui_message_sub: None,
            message_id_generator: Box::new(MessageIdGeneratorReal::default()),
            logger: Logger::new("Accountant"),
        }
```

However, you think about it. If you still like more your initial version I don't want to argue because of that. Of my liking, I think this more structuralized lay out I'm proposing will make reading easier.

node/src/accountant/mod.rs
bertllll: Sorry, mate. I wouldn't ask you if it were too bothering. ;)

node/src/accountant/scanners.rs
bertllll: I like it :handshake:

node/src/accountant/scanners_tools.rs
bertllll: Hmm...then could say there plenty of untested places  because there are places like these, even after the same fn call, which would then deserve it too. 
But I don't think it deserves it. For me it is a spring of confusement because one's must ask why this case is different from others :man_shrugging: 

I myself wouldn't hesitate and would delete it straight away. For me, it just takes space. The added value is so small that I don't see it actually.

What shall I do with you?

node/src/accountant/scanners_tools.rs
bertllll: The closer the assertion is to where the new timestamp is generated the better, probably. Yeah, no big deal...
...mostly because, as I noticed this test is gone...

However it makes me sad that I missed that this test was just half done. 
I were testing the printed message but what you maybe should have test in the first place was what happened to the value of 'time_in_past`. An assertion on the value was obviously missing. 

Printing message does not mean the same as setting a state of an outer value. So you didn't provide a view to understand what happened inside the black box, that function. The test was saying only printing a message was going on.

node/src/accountant/scanners_tools.rs
bertllll: Sure. Still talking about the test that lies below this one. 
I think its name should be like: "debt_under_the_slope_is_marked_unqualified".

As well as another test `payable_above_threshold_values_is_marked_qualified_and_returns_threshold` should say rather:
`debt_above_the_slope_is_qualified_and_the_threshold_point_is_returned`

node/src/accountant/scanners_tools.rs
bertllll: Yeah... I have serious concerns here. I will get into them...
As for your test it struck my mind that instead of the variable name `threshold` and welcome `point_on_the_slope`.

However, now the important part. I have reasons to think that during your merge with master some test were lost. These tests in fact used to provide the tight assertion which I've been demanding from you. Do you see my previous comment about the tested value picked too far from the threshold? 

I will mention them (I afraid I could miss some :frowning: potentially). These are either tests or functions which are in master but not in your code and I don't see a reason why they shouldn't be considered belonging with there anymore (I actually did really big effort to document the thresholds with tests because we had lacked any good idea if it worked wrong or right): 

`payables_debug_summary_stays_still_if_no_qualified_payments`,
`payout_sloped_segment_in_payment_thresholds_goes_along_proper_line`,
`gap_tester`,
`assert_on_height_granularity_with_advancing_time`,
`testing_granularity_calculate_sloped_threshold_by_time`,
`checking_chosen_values_for_the_payment_thresholds_defaults_on_height_values_granularity`,
`slope_has_loose_enough_limitations_to_allow_work_with_number_bigger_than_masq_token_max_supply`,
`slope_after_its_end_turns_into_permanent_debt_allowed`,
`is_innocent_age_works_for_age_smaller_than_innocent_age`,
`is_innocent_age_works_for_age_equal_to_innocent_age`,
`is_innocent_age_works_for_excessive_age`
`is_innocent_balance_works_for_balance_smaller_than_innocent_balance`,
`is_innocent_balance_works_for_balance_equal_to_innocent_balance`,
`is_innocent_balance_works_for_excessive_balance`,
`payable_is_found_innocent_by_age_and_returns`,
`payable_is_found_innocent_by_balance_and_returns`,
`threshold_calculation_depends_on_user_defined_payment_thresholds`

Aside those above, which are all a must, I'd like to see you use the master version of :
`payables_debug_summary` (ignoring the fact I prefer functional programming to imperative one, the structure of the function answers the extra tools I created for testing the threshold mechanism and so it requires `self` of Accountant and so on...). I'd really appreciate if you simply pick it up from master and put it into your branch. 
Together with this function there are also two tests which don't have an alternative in your code. They are:
`payables_debug_summary_prints_pretty_summary `,
`payables_debug_summary_stays_still_if_no_qualified_payments` 

I also wish you to answer this comment in the sense of "done" and then let it open so that I can compare it with the changes submitted for the next iteration. 

Maybe I will have to have a look at the commits where you were doing the merge of master. Just to be sure there aren't more places where you overlooked that there was new piece of code added.

node/src/accountant/mod.rs
bertllll: :+1:

node/src/accountant/scanners_tools.rs
utkarshg6: 1) `payables_debug_summary_stays_still_if_no_qualified_payments` -> `qualified_payables_can_be_filtered_out_from_non_pending_payables_along_with_their_summary`

2) `payout_sloped_segment_in_payment_thresholds_goes_along_proper_line` -> `New Code`

3) `gap_tester` -> `New Code`

4) `assert_on_height_granularity_with_advancing_time` -> `New Code`

5) `testing_granularity_calculate_sloped_threshold_by_time` -> `New Code`

6) `checking_chosen_values_for_the_payment_thresholds_defaults_on_height_values_granularity` -> `New Code`

7) `slope_has_loose_enough_limitations_to_allow_work_with_number_bigger_than_masq_token_max_supply` -> `New Code`

8) `slope_after_its_end_turns_into_permanent_debt_allowed` -> `New Code`

9) `is_innocent_age_works_for_age_smaller_than_innocent_age` -> `New Function - is_payable_qualified()`

10) `is_innocent_age_works_for_age_equal_to_innocent_age` -> `New Function - is_payable_qualified()`

11) `is_innocent_age_works_for_excessive_age` -> `New Function - is_payable_qualified()`

12) `is_innocent_balance_works_for_balance_smaller_than_innocent_balance` -> `New Function - is_payable_qualified()`

13) `is_innocent_balance_works_for_balance_equal_to_innocent_balance` -> `New Function - is_payable_qualified()`

14) `is_innocent_balance_works_for_excessive_balance` -> `New Function - is_payable_qualified()`

15) `payable_is_found_innocent_by_age_and_returns` -> `New Function - is_payable_qualified()`

16) `payable_is_found_innocent_by_balance_and_returns` -> `New Function - is_payable_qualified()`

17) `threshold_calculation_depends_on_user_defined_payment_thresholds` -> `Removed due to the change in the way things are used`


The `payables_debug_summary` was merged into `qualified_payables_and_summary` after the discussion that the three of us had. At that time, it was decided due to the fact that we're doing the same thing two time and it is better to do it one go.

bertllll: You're right on that :point_up:. I just wanna add that our preferred state is to keep the next tests as simple as possible and therefore don't use any higher-level function in the tests but only these bare functions whose names stand in the names of the tests: 

    is_innocent_age_works_for_age_smaller_than_innocent_age

    is_innocent_age_works_for_age_equal_to_innocent_age

    is_innocent_age_works_for_excessive_age

    is_innocent_balance_works_for_balance_smaller_than_innocent_balance

    is_innocent_balance_works_for_balance_equal_to_innocent_balance

    is_innocent_balance_works_for_excessive_balance

node/src/accountant/mod.rs
bertllll: I see your point and it's a good one. I need to add, "sadly", because it would be nice if we didn't have to use the options to initiate values sequentially (we do exactly that; leaving in a None until we get to a chance to create the values belonging with that place).

I've found an improvement that makes me a bit happier. Given the rules I mentioned in my first comment you will be allowed to leave out the calls of `.take()` ... you'll be taking the value without replacing it with None...in fact you will tear it out from the config. Which is okay though. After applying my suggestion; notice of those missing `.take()`s: 

     pub fn new(mut config: BootstrapperConfig, dao_factories: DaoFactories) -> Accountant {
        let payment_thresholds = config
            .payment_thresholds_opt
            .expectv("Payment thresholds");
        let scan_intervals = config.scan_intervals_opt.expectv("Scan Intervals");
        let earning_wallet = Rc::new(config.earning_wallet.clone());
        let financial_statistics = Rc::new(RefCell::new(FinancialStatistics::default()));

        Accountant {

node/src/accountant/mod.rs
bertllll: Okay. This will be a bit more complex regarding an explanation of the alternative. I speculate that I forgot to complete the original comment I wrote and so an adequate example was not attached. 

I'm aware of a feasible alternative to your code with .clone(). Mine files lighter, will prevent cloning in the future and will not hit one's eyes as potentially unidiomatic code. 
Let me explain how you could implement it. 

This is the getter method. Ref<_,_> is a carrier provided with the RefCell and acquired by the `.borrow()` method. (Here carefully. If you happened to get a new import in accidentally and the import was `Borrow` that is not what we want. That is a trait and a different thing, ours comes from RefCell as its method. So if you try using my code snippet and get weird compilation issues look around for this scenario, that might be it. Delete the unwanted import and all should start working just fine.)

    fn financial_statistics(&self) -> Ref<'_,FinancialStatistics> {
        self.financial_statistics.borrow()
    }

Next, at the place in the production code of using this method nothing will need to be changed. However, a lot of new issues arise in the test tree. I suggest to solve it easily, simply use an additional `.clone()` attached to that function. It will give you the wanted owned instance instead of Ref<_,_>. In tests, I don't care if the cloning needs to be there but I don't like the obligatory use of it in the production code. :man_shrugging:

node/src/accountant/mod.rs
bertllll: Hmm...

I felt guilty for what you are referring me to. So I went to check what I actually instructed in the review 0 (well, this one is called 1 but is 2 correctly). I didn't find there anything that should have led you to rearrangement which was done later on. I don't know where the idea comes from.

However, I think that the scanners are probably (or should be) asserted with the params somewhere else where it is more needed. I suppose we could see it where the cycles of scanning are repeating and we test that we went through at least two cycles. 

This kind of test, which we discuss here, doesn't require such heavy tooling though. It really will be as much provable as it is without the param assertions. 

Again, the trick lies in on-purpose undelivered results for the used mocks. If you do that, plus if you make it obvious (sometimes we write a comment, which you can see, by the way, in the older version, written by Dan if remember). So you can extend the comment or adjust a little but otherwise the principle is just the same. 
We run it, no new scans are requested, the Actix system doesn't need to continue running and stops. However it stops clear, without a panic (and so the assertion `assert_eq!(system.run(), 0)` should be in place). 

I'm sorry if you feel injustice in what I've suggested but I do think my reasons are grounded. :pray:

node/src/accountant/scanners.rs
bertllll: I'm okay with how it looks now. :+1:  No need to go on in this conversation.

node/src/accountant/scanners_tools.rs
utkarshg6: Ok!

node/src/accountant/scanners_tools.rs
utkarshg6: Done!

node/src/accountant/mod.rs
utkarshg6: Ok, I did that.

node/src/accountant/mod.rs
utkarshg6: Right from start, I was never against the idea of improving optimisation and I followed my way of using `clone()` in the hope that it will increase readability. In the end, I find the second method, that you mentioned above as _balanced between the trade-off of __readability__ and __optimisation___. 

You may close this conversation as the intended change is done.

node/src/accountant/mod.rs
utkarshg6: This change was the by-product of a different comment. The idea was to assert on the begin scan params. 
You may not be able to find the specific comment for this test, because this change is the side-effect.

I named the PR Review 1, because this contains the changes of first review that you did on GH-611 (the Parent PR). So, please read it as "the changes of first review on GH-611", and you'll perform your next phase of review on top of it.

I don't felt injustice.

Now this is completely unrelated to this comment. I just felt that it would be better if we try to keep the review changes specific to the changes I made in this card rather than review on some old code (there are exceptions, I understand.. again I just want to minimise it). I don't have the problem in modifying the code, just not on the current card. This may be a bit problematic for you, in case you see a possible improvement. I would suggest a todo in case you think that a certain change has the possibility of making the current PR unstable.

The reason I feel that way, is that it makes the code changes volatile. That's just my opinion, in the end, we'll end up finding a middle ground every time.


I removed the assertions as you mentioned and I feel like, it made assertions weaker and is now requiring a comment instead of making it easier for the developer to figure out what is expected.

The expectation is that begin_scan is never called for any scanner. And I think, this change is doing the right job. 

I'm preferring to keep the assertions intact.

I'm sorry if you feel the other way is right. 😞

node/src/accountant/mod.rs
utkarshg6: I find this easy to read. That's why I have made the change that you mentioned in the second iteration.

You may close this conversation as the intended change is done.

node/src/accountant/mod.rs
bertllll: Oh yeah. 

I think I will stop abusing you in the concern of this card :slightly_smiling_face: 

However, I still think there is no big difference between my [and it's not just mine; it was actually Dan who invented that and I just found it cool] and your approach. Yours is more explicit but also wordy.
 
At least I'd like you to accept what I suggested into your dev toolset, because I'm positive we will continue using it in some tests, me and Dan, and I don't see why you should turn this method into the unacceptable. It's simply legitimate. Dan sometimes says "_Hmm, no, we can do even better_...which means he's planning to remove the assertions and leave the test being based on the implicit thought construct. 

Just to make sure you understood me, this is how it works: 
If you have a test where you use mocks and the mocks are constructed the standard way (the should be), it means that if you want to call any possible method from the mock's trade, you will have put up results for it as long as you want to avoid panicking on calling the method and internally calling .remove(0) on an empty queue of results. 

This principle is the core of the idea that at some tests you don't need to use `assertxx` macros. 
(Then it's only politeness to tell about it in a comment to just navigate the reader a bit but I think he could understand even alone).

If I'm to generalize on this, this method is useful in all those places where we try to prove that _something didn't run_. That is also this test. And you could see that Dan also initially wrote the test the way without explicit assertions.

I won't be angry at you, but I'm saying I wouldn't hesitate to remove it. I thought a lot about potential benefits of your assertions and found none except it's literal. I just want to encourage you to rather stop using it in the future. 

---

That was first part. However, we need to deal with the comment as I don't like there are your assertions and the comment together. I think you should delete it.

Also change the log assertion to assert on the name of the log. Make the name of the logger sound the same as the name of the test please.

PR #210, opened 2022-12-19
GH-497: seventh review, better file system arrangement 

bertllll:


masq/tests/subcommand_help_test_integration.rs
```diff
@@ -22,15 +23,16 @@ fn assure_that_all_subcommands_hang_on_the_help_root_integration() {
...
+            let entry_name_long = entry_name_path.file_name().unwrap().to_str().unwrap();
+            if !entry_name_long.starts_with("mod")
+                && !entry_name_long.starts_with("commands_common")
             {
-                let mut short_name = String::from("    ");
-                let suffix = file_name_long
-                    .trim_end_matches("_command.rs")
-                    .replace("_", "-");
-                short_name.push_str(&suffix);
+                let regex = Regex::new(r#"(.+)_command(\.rs|$)"#)
```
dnwiebe, line 31: How hard would it be to check that all xxx_command.rs files are represented in the help? That would force somebody who invents a new command to at least mention it in the help...

bertllll: It's already worked like that so the test can stay as it is. I just wasn't able to address your concern at the moment when you were writing this.

node/src/accountant/big_int_processing/big_int_divider.rs
```diff
@@ -0,0 +1,579 @@
...
+        let g = -4567;
+        let h = (i64::MIN + 1) as i128;
+        let i = i64::MIN as i128;
+        let j = i64::MIN as i128 - 1;
+        let k = i64::MIN as i128 * 32;
+        let vec = vec![b, c, d, e, f, g, h, i, j, k];
+
+        let _ = vec.into_iter().enumerate().fold(
+            a,
+            |previous, current: (usize, i128)| {
```
dnwiebe, line 283: You could name the two components of `current` here, which would make this a bit easier to read.

node/src/accountant/big_int_processing/test_utils.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+pub(in crate::accountant::big_int_processing) mod mod_restricted {
```
dnwiebe, line 3: Find a term or phrase which could be used repeatedly in this position. ("mod_restricted")

PR #213, opened 2022-12-22
GH-647: HOTFIX: Prevent duplication of records of connection-progress

utkarshg6:


node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -63,6 +63,7 @@ trait GossipHandler: NamedType + Send /* Send because lazily-written tests requi
         database: &mut NeighborhoodDatabase,
         agrs: Vec<AccessibleGossipRecord>,
         gossip_source: SocketAddr,
+        peer_addrs: &[IpAddr],
```
dnwiebe: Can you think of a better name than `peer_addrs`? When you're thinking about Gossip, you're probably not thinking about connection progress messages, and it's difficult to understand what a peer might be in this context. There's `cpm_recipient` right below it; maybe these two parameters could be named similarly to make their association clear?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3403,6 +3487,40 @@ mod tests {
...
 
+    #[test]
+    fn handles_pass_target_that_is_a_part_of_a_different_connection_progress() {
+        let cryptde = main_cryptde();
+        let root_node = make_node_record(1234, true);
+        let mut db = db_from_node(&root_node);
+        let subject = PassHandler::new();
+        let (gossip, pass_target, gossip_source) = make_pass(2345);
+        let pass_target_ip_addr = pass_target.node_addr_opt().unwrap().ip_addr();
+        let system = System::new("handles_pass_target_that_is_not_yet_expired");
```
dnwiebe: Missed

node/src/neighborhood/mod.rs
```diff
@@ -266,38 +266,37 @@ impl Handler<ConnectionProgressMessage> for Neighborhood {
...
-                _ => (),
             }
-        } else {
-            trace!(
-                self.logger,
-                "An unnecessary ConnectionProgressMessage received from IP Address: {:?}",
-                msg.peer_addr
-            );
+            Err(e) => {
+                trace!(self.logger, "{}", e);
```
dnwiebe: Is this error specific enough to be able to track down if we see it in the log? Should an identifying string be added so that a codebase text search can lead us here?

node/src/neighborhood/mod.rs
```diff
@@ -1835,11 +1840,58 @@ mod tests {
...
+            peer_addr: peer_2,
+            event: ConnectionProgressEvent::PassGossipReceived(peer_1),
+        };
+
+        cpm_recipient.try_send(cpm).unwrap();
+
+        System::current().stop();
+        assert_eq!(system.run(), 0);
+        TestLogHandler::new().exists_log_containing(&format!(
+            "TRACE: Neighborhood: We've been passed to a peer with IP Address: {:?} that's already a part of different connection progress.",
```
dnwiebe: This message implies that the pass actually happened. Can you reword so that it's clear that the pass was attempted but rejected?

node/src/neighborhood/mod.rs
```diff
@@ -3678,6 +3731,48 @@ mod tests {
...
+        let peer_2_socket_addr: SocketAddr = peer_2.metadata.node_addr_opt.unwrap().into();
+        let pass_gossip = GossipBuilder::new(&peer_2_db)
+            .node(peer_1.public_key(), true)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = pass_gossip.try_into().unwrap();
+
+        subject.handle_agrs(agrs, peer_2_socket_addr);
+
+        TestLogHandler::new()
+            .exists_log_containing(&format!("Gossip from {} ignored", peer_2_socket_addr));
```
dnwiebe, line 3781: If we use this same SocketAddr in a different test, the logs might collide and provide a false positive. Make sure the value of `peer_2_socket_addr` is unique to this test (in this file).

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -224,6 +226,44 @@ impl OverallConnectionStatus {
...
+    pub fn get_connection_progress_to_modify(
+        &mut self,
+        msg: &ConnectionProgressMessage,
+    ) -> Result<&mut ConnectionProgress, String> {
+        if let ConnectionProgressEvent::PassGossipReceived(pass_target) = msg.event {
+            // Check if Pass Target can potentially create a duplicate ConnectionProgress
+            let is_duplicate = self
+                .progress
+                .iter()
+                .map(|connection_progress| connection_progress.current_peer_addr)
```
dnwiebe: Maybe you can use `get_peer_addrs` here.

node/src/neighborhood/overall_connection_status.rs
```diff
@@ -467,6 +507,21 @@ mod tests {
...
+        let peer_2 = make_ip(2);
+        let peer_3 = make_ip(3);
+
+        let subject = OverallConnectionStatus::new(vec![
+            make_node_descriptor(peer_1),
+            make_node_descriptor(peer_2),
+            make_node_descriptor(peer_3),
+        ]);
+
+        assert_eq!(subject.get_peer_addrs(), vec![peer_1, peer_2, peer_3]);
```
dnwiebe: `subject.get_peer_addrs()` is the Act part of this test. It should appear on a line by itself.

node/src/neighborhood/gossip_acceptor.rs
utkarshg6: I used the name `connection_progress_peers`. I think it's conveying the correct idea.

node/src/neighborhood/gossip_acceptor.rs
utkarshg6: Done!

node/src/neighborhood/mod.rs
utkarshg6: I also wanted to add some string to this trace log before raising the PR but somehow, I didn't. Later, when you added this comment, I got a bit excited to rephrase these logs. 

I've used this prefix - `"Found unnecessary connection progress message"`.

Now, it will be easier to find the log inside our codebase.

node/src/neighborhood/mod.rs
utkarshg6: I have reworded it.

dnwiebe (approved):

PR #214, opened 2022-12-23
GH-647: Review 1

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -5513,7 +5518,7 @@ mod tests {
             database: &mut NeighborhoodDatabase,
             agrs: Vec<AccessibleGossipRecord>,
             gossip_source: SocketAddr,
-            _peer_addrs: &[IpAddr],
+            _connection_progress_peers: &[IpAddr],
```
dnwiebe: This field shouldn't be ignored by the mock, and whatever tests assert on these parameters should also assert on this one.

dnwiebe (commented):
Almost!

PR #217, opened 2023-01-05
GH-611: Review 2

utkarshg6:


node/src/accountant/mod.rs
```diff
@@ -195,7 +195,7 @@ impl Handler<ScanForPayables> for Accountant {
     type Result = ();
 
     fn handle(&mut self, msg: ScanForPayables, ctx: &mut Self::Context) -> Self::Result {
-        self.handle_request_for_scan_for_payable(msg.response_skeleton_opt);
+        self.handle_request_of_scan_for_payable(msg.response_skeleton_opt);
```
bertllll, line 198: This doesn't touches the place next to the comment. 
I just realized that in the constructor of Accountant there is a line:

        let earning_wallet = Rc::new(config.earning_wallet.clone());

Take please the `.clone()` away as it has no sensible function there.

node/src/accountant/scanners.rs
```diff
@@ -162,6 +155,45 @@ impl ScannerCommon {
             payment_thresholds,
         }
     }
+
+    fn remove_timestamp_and_log(&mut self, scan_type: ScanType, logger: &Logger) {
```
bertllll: To be honest, the name doesn't fit to my expectations. 

At least, I think, I'd prefer to cut off the part `_and_log` as we log very often in various functions and we don't mention it in their names. 

However, a detail. I will swallow it in either way.

node/src/accountant/scanners.rs
```diff
@@ -162,6 +155,45 @@ impl ScannerCommon {
...
+                    .as_millis();
+                info!(
+                    logger,
+                    "The {:?} scan ended in {}ms.", scan_type, elapsed_time
+                );
+            }
+            None => {
+                error!(
+                    logger,
+                    "Called scan_finished() for {:?} scanner but timestamp was not found",
```
bertllll, line 174: Also details but I have a feeling that the function cited in the log message is not around anymore. Should you change it to the up-to-date name? Is that supposed to be `.mark_as_ended()?`

node/src/accountant/scanners.rs
```diff
@@ -1776,9 +1793,10 @@ mod tests {
             .new_delinquencies_result(vec![])
             .paid_delinquencies_result(vec![]);
         let earning_wallet = make_wallet("earning");
-        let mut receivable_scanner = ReceivableScannerBuilder::default()
+        let mut receivable_scanner = ReceivableScannerBuilder::new()
             .receivable_dao(receivable_dao)
-            .earning_wallet(earning_wallet.clone());
+            .earning_wallet(earning_wallet.clone())
```
bertllll: I think the clone is redundant.

node/src/accountant/scanners.rs
```diff
@@ -1805,22 +1823,25 @@ mod tests {
...
         let banned_dao = BannedDaoMock::new()
             .ban_list_result(vec![])
             .ban_parameters(&ban_parameters_arc)
             .unban_parameters(&unban_parameters_arc);
-        let mut receivable_scanner = ReceivableScannerBuilder::default()
+        let mut receivable_scanner = ReceivableScannerBuilder::new()
             .receivable_dao(receivable_dao)
             .banned_dao(banned_dao)
-            .payment_thresholds(payment_thresholds.clone());
+            .payment_thresholds(payment_thresholds.clone())
```
bertllll: Here, and also on another line lower, you can remove the clone as payment thresholds should be just copy.

node/src/accountant/scanners.rs
```diff
@@ -1922,4 +1945,35 @@ mod tests {
             "INFO: receivable_scanner_handles_received_payments_message: The Receivables scan ended in \\d+ms.",
         );
     }
+
+    #[test]
+    fn remove_timestamp_and_log_if_timestamp_is_correct() {
+        init_test_logging();
+        let test_name = "update_timestamp_and_log_if_timestamp_is_correct";
```
bertllll: This string doesn't correspond to the test name.

node/src/accountant/scanners.rs
```diff
@@ -1922,4 +1945,35 @@ mod tests {
...
+
+        TestLogHandler::new().exists_log_matching(&format!(
+            "INFO: {test_name}: The Payables scan ended in \\d+ms."
+        ));
+    }
+
+    #[test]
+    fn remove_timestamp_and_log_if_timestamp_is_not_found() {
+        init_test_logging();
+        let test_name = "update_timestamp_and_log_if_timestamp_is_not_found";
```
bertllll: The same thing as above. 

However, if you change the name of the tested function you make sure it will propagate also here. ;)

node/src/accountant/test_utils.rs
```diff
@@ -889,90 +889,131 @@ impl PendingPayableDaoFactoryMock {
...
-
-impl Default for PayableScannerBuilder {
-    fn default() -> PayableScanner {
-        PayableScanner::new(
-            Box::new(PayableDaoMock::new()),
-            Box::new(PendingPayableDaoMock::new()),
-            Rc::new(PaymentThresholds::default()),
-        )
-    }
+pub struct PayableScannerBuilder {
```
bertllll, line 892: Hmm. I like these builders. What I don't like is the name. If you see it used somewhere you will not be told that it creates: a PayableScanner stuffed with mock objects. It might be confused with a builder for a production-code-like structure. I'd rather rename slightly each of these three builders to say something as these (probably not best) examples:

```
PayableScannerTestBuilder,
PayableScannerWithMocksBuilder,
PayableScannerMockedBuilder,
BuilderOfPayableScannerWithMocks,
```
I'd vote for the second option, but admitting that these all sound more or less clumsy. Will you think up a better name?

node/src/accountant/test_utils.rs
```diff
@@ -889,90 +889,131 @@ impl PendingPayableDaoFactoryMock {
...
+    pub fn build(self) -> PayableScanner {
+        PayableScanner::new(
+            Box::new(self.payable_dao),
+            Box::new(self.pending_payable_dao),
+            Rc::new(self.payment_thresholds),
         )
     }
 }
 
+pub struct PendingPayableScannerBuilder {
```
bertllll, line 929: I wish you did the same here.

node/src/accountant/test_utils.rs
```diff
@@ -889,90 +889,131 @@ impl PendingPayableDaoFactoryMock {
...
+            Box::new(self.payable_dao),
+            Box::new(self.pending_payable_dao),
+            Rc::new(self.payment_thresholds),
+            self.when_pending_too_long_sec,
+            Rc::new(RefCell::new(self.financial_statistics)),
         )
     }
 }
 
+pub struct ReceivableScannerBuilder {
```
bertllll, line 969: And also here...

node/src/accountant/test_utils.rs
```diff
@@ -889,90 +889,131 @@ impl PendingPayableDaoFactoryMock {
...
 
     pub fn banned_dao(mut self, banned_dao: BannedDaoMock) -> Self {
-        self.banned_dao = Box::new(banned_dao);
+        self.banned_dao = banned_dao;
         self
     }
 
     pub fn payment_thresholds(mut self, payment_thresholds: PaymentThresholds) -> Self {
-        self.common.payment_thresholds = Rc::new(payment_thresholds);
+        self.payment_thresholds = payment_thresholds;
```
bertllll, line 999: Just curious. Why do you need this method at this scanner while you didn't have a setter for payment thresholds in the previous (mocked) scanner, even though it also contains it. You used the default only. Which is strange...as if somewhere in some test an assertion on exact values for the thresholds might be missing.

bertllll (commented):
Done :slightly_smiling_face: 

Please let's continue in the same policy; comment if you feel a need to discuss further.

node/src/accountant/scanners.rs
utkarshg6: The `scan_finished()` is a parent function of `mark_as_ended()`. While reading the logs the focal point will be the `scan_finished()`, that's why I kept it this way.

node/src/accountant/test_utils.rs
utkarshg6: Since we are calling the builder as `AccountantBuilder` and using it only for tests. I think it's appropriate to use this name.

I personally prefer short names, whenever possible :)

node/src/accountant/test_utils.rs
utkarshg6: It's something I just did because we were using the custom payment thresholds for a particular test named `receivable_scanner_scans_for_delinquencies`. There's nothing special about it.

PR #219, opened 2023-01-10
GH-633: For bert

utkarshg6:


masq_lib/src/utils.rs
```diff
@@ -357,6 +357,13 @@ where
     fn helper_access(&mut self) -> &mut Option<T>;
 }
 
+pub fn to_hex(data: &[u8]) -> String {
```
bertllll, line 360: All these changes should be discarded. We've already been using a library `rustc-hex` that provides this simple functionality. It's better not to duplicate code this way. (Yes, I know Dan wrote it, not you).

Now, this is a procedure with with you can replace the usages.

This unwanted code is used as a naked function. Instead, you're gonna have to use a trait method from the cited library. 

So...example like this:
`write!(f, "{}", to_hex(self.data.as_slice()))`

...should be reworked to be like this:
`write!(f, "{}", self.data.as_slice().to_hex::<String>())`

Please don't be shy and delete all the green code in this file.

multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -29,20 +30,18 @@ use node_lib::test_utils::data_hunk_framer::DataHunkFramer;
...
 use std::net::{IpAddr, Shutdown};
 use std::ops::Add;
-use std::rc::Rc;
 use std::thread;
 use std::time::{Duration, Instant};
 
 pub struct MASQMockNode {
     control_stream: RefCell<TcpStream>,
-    guts: Rc<MASQMockNodeGuts>,
+    guts: MASQMockNodeGuts,
```
bertllll, line 44: This is the bloomer. 
I know...
My commit is not included in this PR ;)

multinode_integration_tests/tests/bookkeeping_test.rs
```diff
@@ -39,6 +40,9 @@ fn provided_and_consumed_services_are_recorded_in_databases() {
     // get all payables from originating node
     let payables = non_pending_payables(&originating_node);
 
+    // Waiting until the serving nodes have finished accruing their receivables
+    thread::sleep(Duration::from_secs(4));
```
bertllll, line 44: This is probably too big. However, I would fear a bit to experiment with it. Maybe 3, maybe 2? Let's try 3. :wink:

multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -107,3 +122,129 @@ fn neighborhood_notified_of_newly_missing_node() {
         dot_graph
     );
 }
+
+#[test]
+fn dns_resolution_failure_no_longer_blacklists_exit_node_for_all_hosts() {
+    let mut cluster = MASQNodeCluster::start().unwrap();
```
bertllll, line 128: I just don't understand why this line isn't below the figure. Would make more sense to me...

multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -107,3 +122,129 @@ fn neighborhood_notified_of_newly_missing_node() {
...
+        .transmit_package(
+            relay1_mock.port_list()[0],
+            dns_fail_pkg,
+            &masquerader,
+            originating_node.main_public_key(),
+            originating_node.socket_addr(PortSelector::First),
+        )
+        .unwrap();
+    let _dns_error_response: Vec<u8> = client.wait_for_chunk();
+    // TODO: make sure it's the expected error response
```
bertllll, line 195: This yet remains to be done. 

I advise printing the message first and begin with eye-check. If it's still not clear whether the response is reasonable then you should go to the place where in the Node the DNS failure is handled to be later handed to the client. There should be evidences sufficing for reconstruction of how the response should look. We need to compare it with the response from the test.

To be honest, I like the idea to add an assertion on that response so that we would be able to detect in the future if it's changed.

multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -107,3 +122,129 @@ fn neighborhood_notified_of_newly_missing_node() {
...
+        .unwrap();
+    let (_, intended_exit_public_key) =
+        CryptDENull::extract_key_pair(normal_exit_key.len(), &live_cores_package.payload);
+    assert_eq!(intended_exit_public_key, normal_exit_key);
+
+    // // Now request a different host; it should be routed through cheap_exit because it's cheaper
+    // client.send_chunk("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n".as_bytes());
+    // let (_, _, live_cores_package) = relay1_mock.wait_for_package(&masquerader, Duration::from_secs(2)).unwrap();
+    // let (_, intended_exit_public_key) = CryptDENull::extract_key_pair(cheap_exit_key.len(), &live_cores_package.payload);
+    // assert_eq! (intended_exit_public_key, cheap_exit_key);
```
bertllll, line 210: Ohhhh...

Does this mean that the test was never completed?
Maybe so...we'll have to check it out, running the test also together with these commented-out lines!

If it fails then we will have to repair the test :man_shrugging:

multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -107,3 +122,129 @@ fn neighborhood_notified_of_newly_missing_node() {
...
+        .stream_key;
+    let dns_fail_vdata = VersionedData::new(
+        &node_lib::sub_lib::migrations::dns_resolve_failure::MIGRATIONS,
+        &DnsResolveFailure_0v1 { stream_key },
+    );
+    IncipientCoresPackage::new(
+        destination_alias_cryptde,
+        route,
+        MessageType::DnsResolveFailed(dns_fail_vdata),
+        destination_alias_cryptde.public_key(),
```
bertllll, line 247: If you are brave Utkarsh, you can try verifying that both `destination_alias_cryptde` and `destination_alias_cryptde.public_key()` need to be passed in both. Of course it looks like a nonsense, because one can be got from the other one, however, sometimes our code is not abode by the optimal principles, usually because of test purposes. 

I'd probably dive in myself but I don't want to pressurize you. You would theoretically find out that it is not necessary but only if you check most of usages this constructor and you see, in tests, that there aren't pairs of two different cryptdes.

node/src/actor_system_factory.rs
```diff
@@ -1166,13 +1167,14 @@ mod tests {
         check_start_message(&recordings.neighborhood, 2);
         let hopper_config = Parameters::get(parameters.hopper_params);
         check_cryptde(hopper_config.cryptdes.main);
-        assert_eq!(hopper_config.per_routing_service, 102);
+        assert_eq!(hopper_config.per_routing_service, 300);
```
bertllll, line 1170: Check please why these rates are different if the rest of the test remained the same.  We need a formal check that it's all right. In other words, where this change has been sourced. 

I suspect that it could have something to do with some changed test utils or with the default (production) rate pack being different.

node/src/hopper/routing_service.rs
```diff
@@ -1402,18 +1394,17 @@ mod tests {
...
                 Ok(cryptdata) => cryptdata,
                 Err(e) => panic!("Couldn't encode hop: {:?}", e),
             })
             .collect();
-
         let route = Route { hops };
-
-        let icp = IncipientCoresPackage::new(main_cryptde, route, payload, &public_key).unwrap();
+        let icp =
+            IncipientCoresPackage::new(main_cryptde, route, payload, main_cryptde.public_key())
```
bertllll, line 1406: It's a detail that can stay but I'd prefer to use the variable `public_key` since we use it a lot in the above part of the test.

node/src/hopper/routing_service.rs
```diff
@@ -641,23 +626,30 @@ mod tests {
...
+            Box::new(CryptDEReal::new(TEST_DEFAULT_CHAIN));
+    }
+
     #[test]
     fn logs_and_ignores_message_that_cannot_be_decrypted() {
         init_test_logging();
-        let main_cryptde = main_cryptde();
-        let alias_cryptde = alias_cryptde();
-        let rogue_cryptde = CryptDENull::new(TEST_DEFAULT_CHAIN);
-        let route = route_from_proxy_client(&main_cryptde.public_key(), main_cryptde);
```
bertllll, line 655: I'm convinced that you can use the proposed code:

```
 #[test]
    fn logs_and_ignores_message_that_cannot_be_decrypted() {
        init_test_logging();
        let (main_cryptde, alias_cryptde) = {
            //initialization to real CryptDEs 
            let pair = Bootstrapper::pub_initialize_cryptdes_for_testing(&None, &None);
            (pair.main, pair.alias)
        };
        let rogue_cryptde = CryptDEReal::new(TEST_DEFAULT_CHAIN);
        .
        .
```

...and you can delete the lazy_static! implementation up from the test because we can do it with already existing utils. Moreover these `static` variables are referred from this only test. Let's pull them out.

node/src/hopper/routing_service.rs
```diff
@@ -1869,6 +1860,113 @@ mod tests {
...
+        let subject = RoutingService::new(
+            make_cryptde_pair(),
+            RoutingServiceSubs {
+                proxy_client_subs_opt: peer_actors.proxy_client_opt,
+                proxy_server_subs: peer_actors.proxy_server,
+                neighborhood_subs: peer_actors.neighborhood,
+                hopper_subs: peer_actors.hopper,
+                to_dispatcher: peer_actors.dispatcher.from_dispatcher_client,
+                to_accountant_routing: peer_actors.accountant.report_routing_service_provided,
+            },
```
bertllll, line 1879: You know... I'd like to see a utility function that would simply take `peer_actors` as the only argument and return complete RoutingServiceSubs.
You could use it here as well as at many other places in this file, speaking of tests. I do think it would be a good think to do. It takes too much interest for no big reason. 

`make_routing_service_subs_from_peer_actors()` or a better name...

node/src/hopper/routing_service.rs
```diff
@@ -1869,6 +1860,113 @@ mod tests {
...
+                )),
+            )
+            .unwrap()
+        };
+        route_data_to_peripheral_component_uses_proper_key_on_payload_for_component(
+            payload_factory,
+            Component::Neighborhood,
+        );
+    }
+
```
bertllll, line 1969: Why is a test for the component being the Hopper missing? :thinking:

node/src/hopper/routing_service.rs
```diff
@@ -1869,6 +1860,113 @@ mod tests {
...
+        // CryptDENull panics when you try decrypting with the wrong key; no panic means test passes
+    }
+
+    #[test]
+    fn route_data_to_peripheral_component_uses_alias_key_on_payload_for_proxy_client() {
+        let payload_factory = |cryptdes: &CryptDEPair| {
+            encodex(
+                cryptdes.alias,
+                cryptdes.main.public_key(),
```
bertllll, line 1903: Why are those two supplied cryptDEs each of a different kind? I wouldn't have expected to see the alias cryptDE used anywhere here, that is on the route over...

node/src/hopper/routing_service.rs
```diff
@@ -1869,6 +1860,113 @@ mod tests {
...
+            Component::ProxyClient,
+        );
+    }
+
+    #[test]
+    fn route_data_to_peripheral_component_uses_alias_key_on_payload_for_proxy_server() {
+        let payload_factory = |cryptdes: &CryptDEPair| {
+            encodex(
+                cryptdes.alias,
+                cryptdes.alias.public_key(),
```
bertllll, line 1932: Also I really hate these two distinguished arguments as by logic they should allow a merge of them into one. 
I know there might be a catch somewhere out here but I would like to at least try if it is reasonable or just a silly idea that we keep alive.

bertllll (commented):
These files are yet to be reviewed (fully unchecked now):

node/src/neighborhood/gossip_acceptor.rs
node/src/neighborhood/mod.rs
node/src/proxy_client/mod.rs
node/src/sub_lib/cryptde_null.rs
node/src/sub_lib/neighborhood.rs
node/src/sub_lib/route.rs
node/src/test_utils/mod.rs

node/src/neighborhood/mod.rs
```diff
@@ -78,7 +78,8 @@ use neighborhood_database::NeighborhoodDatabase;
 use node_record::NodeRecord;
 
 pub const CRASH_KEY: &str = "NEIGHBORHOOD";
-pub const UNDESIRABLE_FOR_EXIT_PENALTY: i64 = 100_000_000;
+pub const UNREACHABLE_HOST_PENALTY: i64 = 100_000_000;
+pub const RESPONSE_UNDESIRABILITY_FACTOR: usize = 1_000; // assumed response length is request * this
```
bertllll, line 82: What is meant by the word `request`? Is it its payload size? Is the length regarded as an amount of bytes?

I'd expect the comment to be less ambiguous (as every comment should be anyway) if it is to stay.

Another acceptable option is just to delete the comment. (Could be understood from the code actually)

node/src/neighborhood/mod.rs
```diff
@@ -525,6 +529,7 @@ impl Neighborhood {
     }
 
     fn handle_route_query_message(&mut self, msg: RouteQueryMessage) -> Option<RouteQueryResponse> {
+        // TODO: Don't do this work unless log level is Debug or Trace
```
bertllll, line 532: Yep. Follow the TODO, it's a good one. You should bring in a condition like (naively) `logger.debug(||format!("{:?}", msg))`.

node/src/neighborhood/mod.rs
```diff
@@ -1127,29 +1134,42 @@ impl Neighborhood {
...
-                node_record.inner.rate_pack.exit_byte_rate,
-                node_record.inner.rate_pack.exit_service_rate,
-            ),
-            LinkType::Origin => (0, 0),
-        };
-        let rate_undesirability = per_cores as i64 + (per_byte as i64 * payload_size as i64);
-        if let LinkType::Exit(Some(hostname)) = link_type {
+        let mut rate_undesirability = match undesirability_type {
+            UndesirabilityType::Relay => {
+                (node_record.inner.rate_pack.routing_byte_rate * payload_size as u64)
```
bertllll, line 1144: I wonder if it were better to change the `payload_size` argument to take on straightly u64, and so the need for repeating `as u64` would evaporate.

node/src/neighborhood/mod.rs
```diff
@@ -1127,29 +1134,42 @@ impl Neighborhood {
...
+                (node_record.inner.rate_pack.exit_byte_rate * payload_size as u64)
+                    + node_record.inner.rate_pack.exit_service_rate
+            }
+            UndesirabilityType::ExitAndRouteResponse => {
+                (node_record.inner.rate_pack.exit_byte_rate * payload_size as u64)
+                    + node_record.inner.rate_pack.exit_service_rate
+                    + (node_record.inner.rate_pack.routing_byte_rate * payload_size as u64)
+                    + node_record.inner.rate_pack.routing_service_rate
+            }
+        } as i64;
```
bertllll, line 1157: I also don't understand why we change a number that is apparently always positive to a signed one. I understand it will have to happen eventually but I myself would've done it outside this function. By doing so, you would make it easy for the coder that what this function does can always end up just with a positive number.

node/src/neighborhood/mod.rs
```diff
@@ -1127,29 +1134,42 @@ impl Neighborhood {
...
-            ),
-            LinkType::Exit(_) => (
-                node_record.inner.rate_pack.exit_byte_rate,
-                node_record.inner.rate_pack.exit_service_rate,
-            ),
-            LinkType::Origin => (0, 0),
-        };
-        let rate_undesirability = per_cores as i64 + (per_byte as i64 * payload_size as i64);
-        if let LinkType::Exit(Some(hostname)) = link_type {
+        let mut rate_undesirability = match undesirability_type {
```
bertllll, line 1142: Unnecessary mutability! This situation can be easily solved by `rate_undesirability + UNREACHABLE_HOST_PENALTY` instead of `rate_undesirability += UNREACHABLE_HOST_PENALTY`.

node/src/neighborhood/mod.rs
```diff
@@ -1127,29 +1134,42 @@ impl Neighborhood {
...
+            }
+        } as i64;
+        if let UndesirabilityType::ExitRequest(Some(hostname)) = undesirability_type {
             if node_record.metadata.unreachable_hosts.contains(hostname) {
-                return rate_undesirability + UNDESIRABLE_FOR_EXIT_PENALTY;
+                trace!(
+                    logger,
+                    "To {:?} ({:?}): unreachable host {}; undesirability {} + {} = {}",
+                    node_record.public_key(),
+                    undesirability_type,
```
bertllll, line 1164: By this moment you know very well what the type is as you've satisfied this narrowly intended if statement.  
I can imagine you can do some refactoring to get rid of this variable and hard-code the information instead.

node/src/neighborhood/mod.rs
```diff
@@ -1190,20 +1210,23 @@ impl Neighborhood {
...
-            hostname_opt,
-        )
-        .into_iter()
-        .filter(|cr| cr.undesirability <= minimum_undesirability)
-        .map(|cr| cr.nodes)
-        .next()
+        let result = self
+            .routing_engine(
+                vec![source],
+                self.compute_initial_undesirability(source, payload_size, direction),
```
bertllll, line 1216: I might like this more if it were assigned to a variable instead of putting it in the function invocation directly. 

Not required.

node/src/neighborhood/mod.rs
```diff
@@ -1190,20 +1210,23 @@ impl Neighborhood {
...
+                target_opt,
+                minimum_hops,
+                payload_size,
+                direction,
+                &mut minimum_undesirability,
+                hostname_opt,
+            )
+            .into_iter()
+            .filter(|cr| cr.undesirability <= minimum_undesirability)
+            .map(|cr| cr.nodes)
```
bertllll, line 1226: These two lines could be replaced by a single method and so done in a single operation:

a) filter_map(||),
b) flat_map(||)

With same guts:

```
|cr| match cr.undesirability <= minimum_undesirability {
        true => Some(cr.nodes),
        false => None
}
```

bertllll (commented):
Continue at neighborhood/mod.rs on `compute_new_desirability()`

node/src/neighborhood/mod.rs
```diff
@@ -953,12 +958,14 @@ impl Neighborhood {
             hostname_opt,
         )?;
         debug!(self.logger, "Route over: {:?}", over);
+        // Estimate for routing-undesirability calculations
```
bertllll, line 961: I don't think this is clear and the comment is not helping. 
Please elaborate what's the matter here within the comment after you study why the multiplication is required.

(Dan made the comment so feel free to ask him for help on improving it)

node/src/neighborhood/mod.rs
```diff
@@ -1317,15 +1361,22 @@ impl Neighborhood {
...
-            (RouteDirection::Over, _) => LinkType::Relay,
-            (RouteDirection::Back, Some(target)) if &node_record.inner.public_key == target => {
-                LinkType::Origin
+        let undesirability_type = match (direction, target_opt) {
+            (RouteDirection::Over, None) if hops_remaining == 0 => {
+                UndesirabilityType::ExitRequest(hostname_opt)
             }
-            (RouteDirection::Back, _) => LinkType::Relay,
+            (RouteDirection::Over, _) => UndesirabilityType::Relay,
+            (RouteDirection::Back, _) if undesirability == 0 => {
```
bertllll, line 1369: Please look carefully into how a situation like this (when undesirability equals to 0) could happen because I tried to trace down places where we have the value 0 (which is really a special number!) and I found almost none. The only place I found that could set undesirability to 0 is in `compute_initial_undesirability()` and it belongs under a condition stating this: `direction == RouteDirection::Over`. 

That surprised me a lot because in the match here above my comment the situation talks about `(RouteDirection::Back, _)`, in other words, it doesn't allow for any possibility of a route directed over. In that case, if my concerns are proved, we'll get an unreachable branch in this statement. (Not possible to see such conditions).

The consequences could be various. The main one is that I don't think we can ever account the exit services correctly. Please take time and check if my concern is vain. :pray:   

Another suggestion: If true, then why tests did not prove it wrong? Are those tests wrong too?

node/src/neighborhood/mod.rs
```diff
@@ -3261,75 +3312,145 @@ mod tests {
...
+            None,
         );
 
-        assert_eq!(result, 1 + (1 * 10000) + UNDESIRABLE_FOR_EXIT_PENALTY);
+        let rate_pack = node_record.rate_pack();
+        assert_eq!(
+            new_undesirability,
+            1_000_000 // existing undesirability
+                + rate_pack.routing_charge (1_000) as i64 // charge to route response
+        );
```
bertllll, line 3453: The main issue here, I think, is a lack of a test or assertions or some existing tests are written wrongly. It should exercise the (I know, complicated) `routing_engine()` more closely. The test needs to prove that the code within `compute_new_desirability()` is right or broken. Now I suspect it with a dead branch in it (so far cannot be reached). It's this part:

```
            .
            .
            (RouteDirection::Back, _) if undesirability == 0 => {
                UndesirabilityType::ExitAndRouteResponse
            }  
            .
            .
```

You can begin by putting a panic-invoking piece of code to that match arm. But be demanding, don't be okay with only tests that inspect directly `compute_new_desirability()` and at the same time they artificially supply `0` for the ` undesirability: i64` argument. Such a test cannot reveal the full truth, we need to hit the panic with a test that runs the routing engine. Actually it's also not enough. We need a test calling at least `find_best_route_segment()`. Why? Because here we get a combination of factors that make the reach of `UndesirabilityType::ExitAndRouteResponse` impossible. This when proved would require an intervention...

node/src/proxy_client/mod.rs
```diff
@@ -1036,8 +1068,8 @@ mod tests {
         );
         assert_eq!(accountant_recording.len(), 2);
         let tlh = TestLogHandler::new();
-        tlh.exists_log_containing(format!("ERROR: ProxyClient: Received InboundServerData from 1.2.3.4:5678: stream +dKB2Lsh3ET2TS/J/cexaanFQz4, sequence 1236, length {}; but no such known stream - ignoring", data.len()).as_str());
-        tlh.exists_log_containing(format!("ERROR: ProxyClient: Received InboundServerData (last_data) from 1.2.3.4:5678: stream +dKB2Lsh3ET2TS/J/cexaanFQz4, sequence 1237, length {}; but no such known stream - ignoring", data.len()).as_str());
+        tlh.exists_log_containing(format!("ERROR: ProxyClient: Received InboundServerData from 1.2.3.4:5678: stream {}, sequence 1236, length {}; but no such known stream - ignoring", stream_key, data.len()).as_str());
```
bertllll, line 1071: I don't generally like this because we cannot see how the stream key looks in the finalized string. What if it's printed somehow unreadable... I know, probably not... but there were situations which I encountered when the printed thing didn't match to what one would expect and we had to fix it a long time after the code was called done... we simply realized that the messages contained useless (uninterpretable) characters. 

This is likely not the case and no big danger. So decide yourself...

node/src/sub_lib/cryptde_null.rs
```diff
@@ -94,7 +103,12 @@ impl CryptDE for CryptDENull {
         signature: &CryptData,
         public_key: &PublicKey,
     ) -> bool {
-        let claimed_hash = match Self::decode_with_key_data(public_key.as_slice(), signature) {
+        let key_data = public_key.as_slice();
```
bertllll, line 106: What about renaming this to `public_key_data` or something containing a mention of `public`.

node/src/sub_lib/cryptde_null.rs
```diff
@@ -94,7 +103,12 @@ impl CryptDE for CryptDENull {
...
+        let (k, _) = Self::key_and_data(key_data.len(), signature);
```
bertllll, line 107: `k` would look better if named as `private_key_data` or similar...

node/src/sub_lib/neighborhood.rs
```diff
@@ -58,6 +58,16 @@ pub struct RatePack {
...
 }
 
+impl RatePack {
+    pub fn routing_charge(&self, payload_len: usize) -> u64 {
+        self.routing_service_rate + (self.routing_byte_rate * payload_len as u64)
+    }
+
+    pub fn exit_charge(&self, payload_len: usize) -> u64 {
+        self.exit_service_rate + (self.exit_byte_rate * payload_len as u64)
+    }
```
bertllll, line 68: I've got a suspicion that these methods have been created but never used in the production code. It seems to me that they aren't needed for any tests too.

Either find usage for them in the production code, by looking for places where this kind of an operation performs, or discard them because they would act as latent (referred from their tests) dead code.

node/src/sub_lib/route.rs
```diff
@@ -374,21 +373,6 @@ mod tests {
...
-
-        let subject = Route {
-            hops: vec![Route::encrypt_return_route_id(42, &cryptde1)],
-        };
-
-        assert_eq!(
-            subject.id(&cryptde2),
-            Err("DecryptionError(OpeningFailed)".to_string())
-        );
-    }
```
bertllll, line 390: I might be happier if this test stayed but was modified to await a panic instead.

node/src/sub_lib/route.rs
```diff
@@ -783,9 +767,9 @@ mod tests {
...
+Encrypted with AQIDBA: LiveHop { public_key: 0x02030405, payer: Some(Payer { wallet: Wallet { kind: Address(0x71d0fc7d1c570b1ed786382b551a09391c91e33d) }, proof: Signature { v: 0, r: "8b663e5a10f40c3307e6fb5340482a5e11df78dafc619ceff97f11fa79fea320", s: "8b663e5a10f40c3307e6fb5340482a5e11df78dafc619ceff97f11fa79fea320" } }), component: Hopper }
+Encrypted with AgMEBQ: LiveHop { public_key: 0x03040506, payer: Some(Payer { wallet: Wallet { kind: Address(0x71d0fc7d1c570b1ed786382b551a09391c91e33d) }, proof: Signature { v: 1, r: "63be72962f19dda7802220ed48c0d8199d510b45608a3789c50f61912b98a15b", s: "63be72962f19dda7802220ed48c0d8199d510b45608a3789c50f61912b98a15b" } }), component: Hopper }
+Encrypted with AwQFBg: LiveHop { public_key: 0x, payer: Some(Payer { wallet: Wallet { kind: Address(0x71d0fc7d1c570b1ed786382b551a09391c91e33d) }, proof: Signature { v: 1, r: "9ca23557adf96d7aed407a06ce96851a4184e947a7b29b6c3872eef902fcba1e", s: "9ca23557adf96d7aed407a06ce96851a4184e947a7b29b6c3872eef902fcba1e" } }), component: Neighborhood }
```
bertllll, line 772: I'm wondering why the third key is rendered as empty. It's a mystery for me. I'd appreciate inspecting it somewhat. 

Also, we should unify our thoughts of what's the preferred way of graphical expression of the keys. 
Why do we have to combine two different styles? It says `Encrypted with AQIDBA: LiveHop {...` but later we use `public_key: 0x03040506`. That makes a false impression that we're looking at fundamentally different things.
...but I guess using both is also acceptable... because somebody might find the Base64 encoding better readable.

node/src/test_utils/mod.rs
```diff
@@ -181,12 +182,22 @@ pub fn make_meaningless_message_type() -> MessageType {
     DnsResolveFailure_0v1::new(make_meaningless_stream_key()).into()
 }
 
+pub fn make_route(public_keys: Vec<&PublicKey>) -> Route {
```
bertllll, line 185: What about a name change showing that the route is ever oriented `over`? Because of the target component being a ProxyClient?

node/src/test_utils/mod.rs
```diff
@@ -210,8 +221,12 @@ pub fn make_meaningless_wallet_private_key() -> PlainData {
     )
 }
 
-pub fn route_to_proxy_client(key: &PublicKey, cryptde: &dyn CryptDE) -> Route {
-    shift_one_hop(zero_hop_route_response(key, cryptde).route, cryptde)
+// TODO: The three functions below should use only one argument, cryptde
```
bertllll, line 224: You can commit all your active stuff and try revamping these functions...will be an ugly job though... 
Feel free to leave it for somebody from the future.

node/src/test_utils/mod.rs
```diff
@@ -210,8 +221,12 @@ pub fn make_meaningless_wallet_private_key() -> PlainData {
...
+pub fn route_to_proxy_client(main_key: &PublicKey, main_cryptde: &dyn CryptDE) -> Route {
+    shift_one_hop(
+        zero_hop_route_response(main_key, main_cryptde).route,
+        main_cryptde,
+    )
 }
 
 pub fn route_from_proxy_client(key: &PublicKey, cryptde: &dyn CryptDE) -> Route {
```
bertllll, line 232: This function seems wrong as well as `route_to_proxy_server()`. 
I understand it as:
`route_to_proxy_client` demonstrates [over->ProxyClient], [ProxyClient->back][ProxyServer->back] 
`route_from_proxy_client` should rightly be like: [ProxyClient->Back][ProxyServer->back],
`route_to_proxy_server` was probably meant like [ProxyServer->back]
 
I know this is simplification. However, I find it enough confusing that it concerns me. I'd appreciate if the next steps would be to create an appropriate card about these imperfections that could theoretically hide issues because the tests using this have unreal instructions within the `Route`.  

I advise looking into that together with Dan, because it's not trivial making the right decision here.

node/src/test_utils/mod.rs
```diff
@@ -223,6 +238,28 @@ pub fn route_to_proxy_server(key: &PublicKey, cryptde: &dyn CryptDE) -> Route {
     shift_one_hop(route_from_proxy_client(key, cryptde), cryptde)
```
bertllll, line 238: If not anything else, changing the arguments saying `key` and `cryptde` into `main_key` and `main_cryptde`.  I do believe I described the true situation by stating so. 

It ought to be done for both functions above.

node/src/test_utils/mod.rs
```diff
@@ -223,6 +238,28 @@ pub fn route_to_proxy_server(key: &PublicKey, cryptde: &dyn CryptDE) -> Route {
...
+//     eprintln!("Route before drain: {:?}", route.hops);
+//     let remaining_hops = route
+//         .hops
+//         .drain(hops_to_remove..)
+//         .collect::<Vec<CryptData>>();
+//     Route {
+//         hops: remaining_hops,
+//     }
+// }
+
```
bertllll, line 262: Deleting these lines make as much sense as moving it to the new card mentioned in one of the previous comments, acting as inspiration. Leaving it here is the worst case; please don't.

node/src/test_utils/mod.rs
```diff
@@ -293,13 +332,13 @@ pub fn rate_pack_routing_byte(base_rate: u64) -> u64 {
     base_rate + 1
 }
 pub fn rate_pack_routing(base_rate: u64) -> u64 {
-    base_rate + 2
+    base_rate + 200
```
bertllll, line 335: I barely remember that but I think I posed a question before if you can check why some values changed by *100; so this might be the answer.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -926,6 +927,10 @@ impl GossipHandler for StandardGossipHandler {
     ) -> GossipAcceptanceResult {
         let initial_neighborship_status =
             StandardGossipHandler::check_full_neighbor(database, gossip_source.ip());
+
+        let patch = self.compute_patch(&agrs, database.root());
+        let agrs = self.filter_agrs_from_patch(agrs, patch);
```
bertllll, line 932: I would be tempted to apply a more specific name for this variable; saying clearly that these agrs are not anymore those agrs passed into the `handle` function.

`patch_agrs` ??
`filtered_agrs` ??
`relevant_agrs` ??

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
         StandardGossipHandler { logger }
     }
 
+    fn compute_patch(
+        &self,
+        agrs: &[AccessibleGossipRecord],
+        root_node: &NodeRecord,
+    ) -> HashSet<PublicKey> {
```
bertllll, line 974: I can imagine a HashSet of `&PublickKey`.
I think it's worth a try.
You aren't gonna need the owned keys actually; you use them as "identification marks" to say which Nodes should be in the patch in the end but then you drop these keys on the floor. I think you could achieve it without having the ownership, meaning you won't need to do cloning.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
...
+
+    fn compute_patch_recursive(
+        &self,
+        patch: &mut HashSet<PublicKey>,
+        node: &PublicKey,
+        agrs: &HashMap<&PublicKey, &AccessibleGossipRecord>,
+        hops_remaining: usize,
+        root_node: &NodeRecord,
+    ) {
+        patch.insert(node.clone());
```
bertllll, line 1000: This clone could be probably eliminated :point_up:

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
...
+            root_node,
+        );
+
+        patch
+    }
+
+    fn compute_patch_recursive(
+        &self,
+        patch: &mut HashSet<PublicKey>,
+        node: &PublicKey,
```
bertllll, line 995: I'd appreciate a more telling name. Give the `node` an adjective. Also consider be clear on that this is a key; from seeing `node` one could thing it's much more than a simple key.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
...
+        };
+
+        for neighbor in neighbors {
+            if !patch.contains(neighbor) {
+                self.compute_patch_recursive(patch, neighbor, agrs, hops_remaining - 1, root_node)
+            }
+        }
+    }
+
+    fn filter_agrs_from_patch(
```
bertllll, line 1028: maybe `filter_agrs_by_patch` instead of using `from`...it evokes that the patch is a set of agrs from which you filter some other agrs...

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
...
+        patch: HashSet<PublicKey>,
+    ) -> Vec<AccessibleGossipRecord> {
+        agrs.into_iter()
+            .filter(|agr| patch.contains(&agr.inner.public_key))
+            .collect::<Vec<AccessibleGossipRecord>>()
+    }
+
     fn identify_and_add_non_introductory_new_nodes(
         &self,
         database: &mut NeighborhoodDatabase,
```
bertllll, line 1040: Here somebody could have also done this better. Look at:

```
        let all_keys = database
            .keys()
            .into_iter()
            .cloned()
            .collect::<HashSet<PublicKey>>();
```
 
I'm convinced that also here the clone of all the keys is not necessary. You can simply collect references. Please consider refactoring it. Could be better for performance.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -962,6 +967,74 @@ impl StandardGossipHandler {
...
+        &self,
+        agrs: Vec<AccessibleGossipRecord>,
+        patch: HashSet<PublicKey>,
+    ) -> Vec<AccessibleGossipRecord> {
+        agrs.into_iter()
+            .filter(|agr| patch.contains(&agr.inner.public_key))
+            .collect::<Vec<AccessibleGossipRecord>>()
+    }
+
     fn identify_and_add_non_introductory_new_nodes(
```
bertllll, line 1038: I don't like this function name. When I read it I cannot figure out what it's supposed to do even though the name is long and so it should reveal a lot of information. Could you come up with a better name? Let your imagination play. :slightly_smiling_face: 

Oh, I later realized that... because we passed the gossip qualification stages of Debut, Introduction, and Pass and so we are, of course, the Standard type, then we know that if there is a new Node it's not a participant of an active introduction...the opposite, we get to know about it from a share which other Nodes send us, without our initiative at all.

Still hard to come up with the explanation for yourself in a quick move...

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -926,6 +927,10 @@ impl GossipHandler for StandardGossipHandler {
     ) -> GossipAcceptanceResult {
         let initial_neighborship_status =
             StandardGossipHandler::check_full_neighbor(database, gossip_source.ip());
+
+        let patch = self.compute_patch(&agrs, database.root());
+        let agrs = self.filter_agrs_from_patch(agrs, patch);
+
         let mut db_changed =
             self.identify_and_add_non_introductory_new_nodes(database, &agrs, gossip_source);
         db_changed = self.identify_and_update_obsolete_nodes(database, agrs) || db_changed;
```
bertllll, line 936: Please ask Dan if he thinks that the order of

`self.identify_and_update_obsolete_nodes(database, agrs) || db_changed`

or 

`db_changed || self.identify_and_update_obsolete_nodes(database, agrs)`

matters?

My puzzle is whether the function needs to be evaluated (executed) if the `db_changed` variable stands first and its value is `true` or whether it must happen at any case.

bertllll, line 936: The line below...
Function called `handle_root_node()` could bear also a better name. How about saying what it does in its name? Could you just try? Don't rack your brain by it too long...

bertllll (commented):
Continue with tests in gossip_acceptor.rs.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            .node(node_c.public_key(), false)
+            .node(node_d.public_key(), false)
+            .node(node_e.public_key(), false)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
+        let result = subject.compute_patch(&agrs, node_a_db.root());
+
+        let expected_hashset = vec![
+            node_a.public_key().clone(),
```
bertllll, line 2370: These clones don't appear being needed. It should work without them as well as this.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
         assert_eq!(system.run(), 0);
         let recording = recording_arc.lock().unwrap();
         assert_eq!(recording.len(), 0);
     }
 
+    #[test]
+    fn standard_gossip_handler_can_compute_patch() {
+        /*
+            Over here, root node is A and patch contains [A, B, C, D].
+                                  A---B---C---D---E
```
bertllll, line 2339: I was thinking a while if adding more than the node E would make the test stronger. I'm still not sure. Usually, I add another item to prove that the algorithm doesn't stop at the first excluded item but have an effect on all items matching the conditions. 

That's a meet of two different attitudes (or philosophies) and I can also grasp the attitude of my opponent without sticking to mine.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            .node(node_c.public_key(), false)
+            .node(node_d.public_key(), false)
+            .node(node_y.public_key(), false)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
+        let patch = subject.compute_patch(&agrs, node_a_db.root());
+
+        let expected_hashset = vec![
+            node_a.public_key().clone(),
```
bertllll, line 2423: Also here, unnecessary clones across the whole vector.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+    #[test]
+    fn standard_gossip_handler_computes_neighbors_from_database_and_neighbors_of_neighbors_from_agrs(
+    ) {
+        /*
+            Over here, root node is A and patch contains [A, B, Y, C, D].
+                                  A---B---C---D
+                                      |
+                                      Y
+            What does this test proves:
+            1) To find neighbors, we'll look into the root node's database. (For Example, A---B).
```
bertllll, line 2389: I think the part `(For example, A---B)` is misleading. There are no other neighbors before the gossip arrives. The root database contains exactly just A---B.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            2) To find neighbors of neighbors, we'll look into the AGRs. (For Example, B---Y, B---C, and C---D).
```
bertllll, line 2390: Also here, it's not a good place for "For example", you enumerate all of them, not just some :wink:

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            node_d.public_key().clone(),
+            node_y.public_key().clone(),
+        ]
+        .into_iter()
+        .collect::<HashSet<PublicKey>>();
+        assert_eq!(patch, expected_hashset);
+    }
+
+    #[test]
+    fn standard_gossip_handler_can_handle_node_for_which_agr_is_not_found_while_computing_patch() {
```
bertllll, line 2435: I don't understand this test. I feel worried from what I see here.

If my interpretation is right then:
We as Node A have two immediate neighbors and somebody from outside will send as a Gossip (surely the Node B is suspected on that). The gossip doesn't contain Node X and so we will blindly obey and remove X from our patch.

How can somebody else than us know better than us that we lost a neighbor. Shouldn't come such information first from ourselves? Not sure what is the cause. Either this scenario is unrealistic. Or we have holes in our algorithms if they allow things like this. 

I would be open to a debate, in a similar matter, about the following schema:

          A---B---C---D
          |
          X
          |
          Y

Claiming that Y disappeared. 

However, I see there another worry. Let's see if I can explain somehow. 

Here is a root A with two immediate neighbors oppositely oriented. 

Z---Y---X---A---B---C---D

The problem is that Node Z is too far from B (it's more than 3 hops) but it still should be visible for the Node A.
Now let's think of what happens if Node B sends a Gossip. There is no way how this Node could see as far as the Node Z which means this Node will necessarily be missing in the Gossip. 

According to your scenario, we're receiving the Gossip and are going to decide that Node Z is not among the received agrs, meaning the Node will be dropped from the patch. I say it's wrong, because Z can be easily still up and fully connected and it, by general metrics, qualifies as a Node belonging to your patch. So B caused unnecessary loss of one (or possibly more) Node in our patch. 

Leaving Y---X---A---B---C---D in the A's database.

Another consequence, even sillier, is that if we continue in this experiment, it could be X who sends a Gossip next time where you'll happily add Z back in (because it was missing) and will decide to remove D for the same reasons why Z was lost before.

Do you think my reasoning is wrong? I wouldn't bet my life that this is so. I just have a suspicion. We can discuss with Dan more in detail.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            .node(node_b.public_key(), false)
+            .node(node_c.public_key(), false)
+            .node(node_d.public_key(), false)
+            .build();
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
+        let patch = subject.compute_patch(&agrs, node_a_db.root());
+
+        let expected_hashset = vec![
+            node_a.public_key().clone(),
```
bertllll, line 2472: Clones...

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+            "TRACE: {}: While computing patch no AGR record found for public key {:?}",
+            test_name,
+            node_x.public_key()
+        ));
+    }
+
+    #[test]
+    fn standard_gossip_handler_will_ignore_gossips_from_outside_the_patch() {
+        /*
+            Over here, root node is A and patch contains [A, B, C, D].
```
bertllll, line 2490: Isn't saying something like `...and its original patch contains...` better?

I found the figures a bit confusing. I'd expect you to draw the initial stage first:

A---B---C---D

And then to continue with the coming change:

A <---- A---B---C---D---E---F

However, I understand it even your way. I just thought that adding a few words more specifying various things in the text might be  somewhat helpful.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+        let mut node_a_db = db_from_node(&node_a);
+        node_a_db.add_node(node_b.clone()).unwrap();
+        node_a_db.add_node(node_c.clone()).unwrap();
+        node_a_db.add_node(node_d.clone()).unwrap();
+        node_a_db.add_node(node_x.clone()).unwrap();
+        node_a_db.add_arbitrary_full_neighbor(node_a.public_key(), node_b.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_a.public_key(), node_x.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_b.public_key(), node_c.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+        let gossip = GossipBuilder::new(&node_a_db)
```
bertllll, line 2462: I don't understand why sometimes you add `.node(node_a.public_key(), true)`, in other words a key of the Node to which the Gossip is meant to go. However, here you didn't. :thinking:

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+        let mut node_a_db = db_from_node(&node_a);
+        node_a_db.add_node(node_b.clone()).unwrap();
+        node_a_db.add_node(node_c.clone()).unwrap();
+        node_a_db.add_node(node_d.clone()).unwrap();
+        node_a_db.add_node(node_e.clone()).unwrap();
+        node_a_db.add_arbitrary_full_neighbor(node_a.public_key(), node_b.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_b.public_key(), node_c.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+        node_a_db.add_arbitrary_full_neighbor(node_d.public_key(), node_e.public_key());
+        let gossip = GossipBuilder::new(&node_a_db)
```
bertllll, line 2359: Here you also didn't add the node_a's public key and still it works...weird...does it mean it has got no effect whether you do or not?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+        let mut node_b_db = db_from_node(&node_b);
+        node_b_db.add_node(node_a.clone()).unwrap();
+        node_b_db.add_node(node_y.clone()).unwrap();
+        node_b_db.add_node(node_c.clone()).unwrap();
+        node_b_db.add_node(node_d.clone()).unwrap();
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_a.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_y.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_b.public_key(), node_c.public_key());
+        node_b_db.add_arbitrary_full_neighbor(node_c.public_key(), node_d.public_key());
+
```
bertllll, line 2411: An extra blank line.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+    }
+
     #[test]
     fn no_cpm_is_sent_in_case_full_neighborship_doesn_t_exist_and_cannot_be_created() {
         // Received gossip from a node we couldn't make a neighbor {Degree too high or malefactor banned node} (false, false)
+        // This is Standard Gossip, even though it looks like a Debut,
+        // because it's specifically handled by a StandardGossipHandler
+        // instead of the GossipAcceptor (which would identify it as a Debut),
+        // so the test is unrealistic. Also that the Gossip is ignored because
+        // Node B isn't in Node A's patch, which matters to a StandardGossipHandler.
```
bertllll, line 2553: Can you elaborate on why this, as you say yourself, unrealistic situation was made. What leads you to conviction that writing this test like so (even though unrealistic) is better than writing it some other, maybe more realistic way?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2210,22 +2319,238 @@ mod tests {
...
+        let agrs: Vec<AccessibleGossipRecord> = gossip.try_into().unwrap();
+
+        let result = subject.handle(cryptde, &mut node_a_db, agrs, gossip_source, &cpm_recipient);
+
+        assert_eq!(result, GossipAcceptanceResult::Ignored);
+    }
+
     #[test]
     fn no_cpm_is_sent_in_case_full_neighborship_doesn_t_exist_and_cannot_be_created() {
         // Received gossip from a node we couldn't make a neighbor {Degree too high or malefactor banned node} (false, false)
```
bertllll, line 2548: You're stating that the reason of ignoring this Node is that it might have full capacity for neighbors or might be malefactor banned. However I see no evidences of such conditions in the test's setup and therefore I don't really understand the test...

I've got bad feeling from this test. It looks like you crippled a standard Gossip and force the StandardGossipHandler to process it while it finds that the Gossip looks somewhat weird and so it drops it on the floor. However, you also claim that there two situation where this could happen `{Degree too high or malefactor banned node}` and I sadly cannot see either of them be set up in this test... So how can we be sure that it tests the mechanism that's supposed to deal with the cited situations and on contrary, that it doesn't exercise completely different piece of code whose job is simply conclude that the received Gossip is malformed and so let's not care about it (ignoring).

I really don't like this test...

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3249,6 +3574,26 @@ mod tests {
     #[test]
     fn initial_standard_gossip_does_not_produce_neighborship_if_destination_degree_is_already_full()
```
bertllll, line 3575: Actually...

Tell me, is this talking about a hypothetical neighborship between Dest and the disconnected Node or what the test name really refers to? 

Plus, the formulation about a possibility that standard Gossip produces neighborship sounds strange too. I think, in the most of situations, standard Gossip doesn't have any competence to moderate a connection of other until-now-not-known Nodes to the destination Node. 

There might be really only a special case, talking about the response to a Debut?

--------------------

Anyway...even though, as I said, I don't grasp this fully, and the test might need to change or might not...I can have some comments to the figures. I found two things in them which made me feel even more that I don't understand.

First thing is the Node depicted as `Disc`. I had looked at the picture many times but only after some time passed I realized what the word there tries to remark. I did not consider it a part of the middle depicted situation because it stands somehow too much aside and the letters `Disc` really didn't make any sense to me first. When you're reading the test from top down, you don't see any mention of a disconnected Node. What more, it cannot be understood from the test's title at all. 

Then, when I realized that it shows a disconnected Node I started thinking how much it is important that the Node is disconnected and why there was no clear mention about this fact...

Secondly, I find the third figure confusing in its part where it shows :
```
      Third
-----|
      Src
```
I couldn't decide easily what this schema might mean. 

In the end the most probably explanation is that it describes connections to the Dest Node of both `Third` and `Src`. However it looks so suspicious to me and I don't know what to thing about it because it raised the number of neighbors up to 7 and I thought that wouldn't be possible. Even fewer than that would look strange. 

One could argue that most of the Neighbors are just half Neighbors and they might not count in. If that's the point though then at least the test name is misleading if it talks about destination degree already full. 

I'm aware that I'm probably wrong on many things here but just this fact alone means that the test isn't much clear if I wasn't able to help myself resolve it.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3297,10 +3642,8 @@ mod tests {
         let mut expected_dest_db = src_db.clone();
         expected_dest_db.add_arbitrary_half_neighbor(dest_node.public_key(), src_node.public_key());
         expected_dest_db
-            .node_by_key_mut(disconnected_node.public_key())
-            .unwrap()
-            .metadata
-            .node_addr_opt = None;
+            .remove_neighbor(disconnected_node.public_key())
+            .unwrap();
         for idx in 0..MAX_DEGREE {
```
bertllll, line 3647: This whole for loop is located in this test exactly twice, could you please deduplicate it?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3249,6 +3574,26 @@ mod tests {
     #[test]
     fn initial_standard_gossip_does_not_produce_neighborship_if_destination_degree_is_already_full()
     {
+        /*
+        Destination Database ==>
+            N2  N3
+             \  /
+        N1---Dest---Third
```
bertllll, line 3581: Also confusing: using the name `Third` while you also using numeric indexes for the unimportant Nodes and so the Node means `N3` in the same picture. Instead of `Third` it could be named, for example, `Middle`.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3264,10 +3609,10 @@ mod tests {
             dest_db.add_arbitrary_half_neighbor(dest_node.public_key(), failed_node_key);
```
bertllll, line 3609: I'd also appreciate if we looked finally reliably into the comments left over here. There are two. One about a revealed Ip and another about half neighbors and degree calculations. 

Please, understand, I don't want to let this uttered uncertainty about the Gossip effects live on too long. This will serve indirectly as documentation of the processes and we should make it as accurate as possible. Probably Dan again needed for his assistance here.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3297,10 +3642,8 @@ mod tests {
         let mut expected_dest_db = src_db.clone();
         expected_dest_db.add_arbitrary_half_neighbor(dest_node.public_key(), src_node.public_key());
```
bertllll, line 3643: Oh...I suppose now that this might have something to do with the goal of the test. As if we wanted to prove that we couldn't make the src_node a full_neighbor because the degree is too high? 

However, I don't understand why the further `Src` Node had to be linked by the `Dest`. Who wanted them to try being close neighbors? Shouldn't this be a job of Debut, Introduction or Pass? I'll be happy for some explanation.

bertllll:
Please let's continue in the strategy with you commenting on my objections in case you've got a different opinion or if you don't understand my arguments. Okay?

There are some fundamental places with lack of clarity in how it works, whether it works right, whether we've succeeded in making it out for others. I really wish we had a good look at them and did our best to improve the quality of the information they radiate or to fix what isn't correct or set just accordingly to the real use case.

bertllll:
Please let's continue in the strategy with you commenting on my objections in case you've got a different opinion or if you don't understand my arguments. Okay?

There are some fundamental places with lack of clarity in how it works, whether it works right, whether we've succeeded in making it out for others. I really wish we had a good look at them and did our best to improve the quality of the information they radiate or to fix what isn't correct or set just accordingly to the real use case.

That said, I think some of those places should be checked and verified in Dan's company.

PR #222, opened 2023-01-13
GH-611: review to be done by Utkarsh

bertllll:


node/src/accountant/dao_utils.rs
```diff
@@ -343,15 +342,81 @@ pub fn sum_i128_values_from_table(
...
+        .threshold_interval_sec over 10^9 would mean continuing to declare debts delinquent after
+        more than 31 years.
+
+        If payment_thresholds are ever configurable by the user, these validations should be done
+        on the values before they are accepted.
+        */
+
+        (gwei_to_wei::<i128, _>(payment_thresholds.permanent_debt_allowed_gwei)
+            - gwei_to_wei::<i128, _>(payment_thresholds.debt_threshold_gwei))
+            / payment_thresholds.threshold_interval_sec as i128
```
utkarshg6, line 370: Is there any risk associated with typecasting the number once we receive it from the function.

I mean `gwei_to_wei(payment_thresholds.permanent_debt_allowed_gwei) as i128`.

Is doing this the same or different to what we're doing in the code?

node/src/accountant/dao_utils.rs
```diff
@@ -545,4 +610,191 @@ mod tests {
...
+        expected_height_change_wei: u64,
+    ) {
+        const WE_EXPECT_ALWAYS_JUST_ONE_SECOND_TO_CHANGE_THE_HEIGHT: u64 = 1;
+        let (seconds_needed_for_smallest_change_in_height, absolute_height_change_wei) =
+            gap_tester(&payment_thresholds);
+        assert_eq!(seconds_needed_for_smallest_change_in_height, WE_EXPECT_ALWAYS_JUST_ONE_SECOND_TO_CHANGE_THE_HEIGHT,
+                   "while testing {} we expected that these thresholds: {:?} will require only {} s until we see the height change but computed {} s instead",
+                   description_of_given_pt, payment_thresholds, WE_EXPECT_ALWAYS_JUST_ONE_SECOND_TO_CHANGE_THE_HEIGHT, seconds_needed_for_smallest_change_in_height);
+        assert_eq!(absolute_height_change_wei, expected_height_change_wei,
+                   "while testing {} we expected that these thresholds: {:?} will cause a height change of {} wei as a result of advancement in time by {} s but the true result is {}",
```
utkarshg6: I can see it can be formatted better, please mark it as resolved in case the rust formatter automatically formats it, otherwise improve it as you may seem fit.

utkarshg6 (commented):
accountant/mod.rs 3572 red

node/src/accountant/scanners.rs
```diff
@@ -215,18 +168,16 @@ impl Scanner<ReportAccountsPayable, SentPayable> for PayableScanner {
...
         );
-        let (qualified_payables, summary) = qualified_payables_and_summary(
-            timestamp,
-            all_non_pending_payables,
-            self.common.payment_thresholds.as_ref(),
-        );
-        debug!(logger, "{}", summary);
-        match qualified_payables.is_empty() {
+
+        let qualified_payable = self.sniff_out_alarming_payables(all_non_pending_payables, logger);
```
utkarshg6: I was expecting the summary to be printed here, but the name suggests that it isn't doing any logging.

node/src/accountant/scanners.rs
```diff
@@ -314,8 +265,79 @@ impl PayableScanner {
         }
     }
 
-    fn handle_blockchain_errors(&self, blockchain_errors: Vec<BlockchainError>, logger: &Logger) {
-        for blockchain_error in blockchain_errors {
+    fn sniff_out_alarming_payables(
```
utkarshg6: A developer might infer from the params that this function is logging something. I understand, but to add more clarity, I'd like if the name of this function was `sniff_out_alarming_payables_and_log_them`.

node/src/accountant/dao_utils.rs
bertllll: Oh... the casting is going on after a different argument actually... it's just   `payment_thresholds.threshold_interval_sec as i128`. Which is safe because `threshold_interval_sec` is of type u64.

node/src/accountant/dao_utils.rs
bertllll: I'll check it out but it might look bad only in the GitHub editor. I dunno. I'll see what I can do.

node/src/accountant/scanners.rs
```diff
@@ -314,8 +265,79 @@ impl PayableScanner {
...
+        &self,
+        non_pending_payables: Vec<PayableAccount>,
+        logger: &Logger,
+    ) -> Vec<PayableAccount> {
```
utkarshg6, line 274: This function can be made simpler. We don't need so many functions inside of it:


Here's a simple implementation (you'll require to derive `Hash` at multiple places):

```rust
    fn sniff_out_alarming_payables(
        &self,
        non_pending_payables: Vec<PayableAccount>,
        logger: &Logger,
    ) -> Vec<PayableAccount> {
        let mut thresholds_by_account = HashMap::new();
        for non_pending_payable in non_pending_payables {
            if let Some(threshold_point) =
                self.payable_exceeded_threshold(&non_pending_payable, SystemTime::now())
            {
                thresholds_by_account.insert(non_pending_payable, threshold_point);
            }
        }

        if logger.debug_enabled() {
            let thresholds_by_account_vec: Vec<(PayableAccount, u128)> = thresholds_by_account
                .iter()
                .map(|(account, threshold)| ((account.to_owned(), threshold.to_owned())))
                .collect();

            payables_debug_summary(&thresholds_by_account_vec, logger);
        }

        thresholds_by_account
            .into_keys()
            .collect::<Vec<PayableAccount>>()
    }
```

If you can use the Hashmap inside the `payables_debug_summary`, then it'll be even better. Up to you.

node/src/accountant/scanners_utils.rs
```diff
@@ -0,0 +1,540 @@
...
+                        p_age.as_secs(),
+                        threshold_point.separate_with_commas(),
+                        payable.wallet
+                    )
+                })
+                .join("\n")
+        })
+    }
+
+    pub trait PayableThresholdsGauge {
```
utkarshg6, line 130: I like traits whenever there are supposed to be multiple implementations. Over here we don't require multiple implementations. I'm not asking you to change it in any way. I prefer to reduce the use of traits and use direct functions or a struct implementation.


It is a personal preference, so I won't force you to think my way, just understanding my perspective is enough.

node/src/accountant/dao_utils.rs
bertllll: Okay, I improved it. :)

node/src/accountant/test_utils.rs
```diff
@@ -92,53 +95,141 @@ impl Default for AccountantBuilder {
...
     }
+    populated_queue
+}
 
-    pub fn pending_payable_dao(mut self, pending_payable_dao: PendingPayableDaoMock) -> Self {
-        match self.pending_payable_dao_factory {
-            None => {
-                self.pending_payable_dao_factory =
-                    Some(PendingPayableDaoFactoryMock::new().make_result(pending_payable_dao))
+macro_rules! update_factory {
```
utkarshg6: This is my personal opinion, I think we're making it complex to understand how the utility was built by using the macro, especially by using `destination_matrix`.

I'm not asking you to modify anything, a good thing to add will be if we explain by comments why those particular numbers inside the matrix.

node/src/accountant/test_utils.rs
```diff
@@ -92,53 +95,141 @@ impl Default for AccountantBuilder {
...
+}
+
+impl AccountantBuilder {
+    pub fn bootstrapper_config(mut self, config: BootstrapperConfig) -> Self {
+        self.config = Some(config);
         self
     }
 
+    pub fn payable_daos(mut self, dao_set: Vec<DaoDestination<PayableDaoMock>>) -> Self {
+        let destination_matrix: [u8; 3] = [0, 1, 3];
```
utkarshg6: A comment over here.

node/src/accountant/test_utils.rs
```diff
@@ -92,53 +95,141 @@ impl Default for AccountantBuilder {
...
+            destination_matrix,
+            PayableDaoFactoryMock,
+            payable_dao_factory,
+            PayableDao,
+            self
+        )
+    }
+
+    pub fn receivable_daos(mut self, dao_set: Vec<DaoDestination<ReceivableDaoMock>>) -> Self {
+        let destination_matrix: [u8; 2] = [0, 2];
```
utkarshg6: A comment over here.

node/src/accountant/test_utils.rs
```diff
@@ -92,53 +95,141 @@ impl Default for AccountantBuilder {
...
+            ReceivableDao,
+            self
+        )
+    }
+
+    pub fn pending_payable_daos(
+        mut self,
+        dao_set: Vec<DaoDestination<PendingPayableDaoMock>>,
+    ) -> Self {
+        let destination_matrix: [u8; 3] = [0, 1, 3];
```
utkarshg6: A comment over here.

utkarshg6 (approved):
Well Done!

node/src/accountant/dao_utils.rs
utkarshg6: Ok, I just didn't liked the use of `_`, it's use devalues passing another parameter.

You can resolve this comment.

bertllll: not true...it works clearly...you can easily check what type `debt_threshold_gwei` is... for me the number would be just noise :man_shrugging:

node/src/accountant/scanners_utils.rs
bertllll: This is a mockable thing. Real and mock.

node/src/accountant/scanners.rs
bertllll: Okay. I like it. I think I will accept :)

bertllll: Well... 
What I did was mostly avoiding clones. You do some but I might swallow it.

utkarshg6 (approved):
Thanks for your efforts, Bert! 🙏

node/src/accountant/test_utils.rs
utkarshg6: I really liked your new implementation. You may resolve this comment. 😊

node/src/accountant/scanners_utils.rs
utkarshg6: I was aware that there were implementations in both Production and Test Code. The point I was referring to was that I only feel that traits are used in the correct scenarios if:


Scenario 1: It requires multiple implementations in Production code.
Scenario 2: We can't directly test the production code, so we need to create a Mock, hence we'll use traits.

In our case, I felt that it is plausible to say that we can test these functions even if they're direct functions, which can be tested directly. So, just for this case, I felt that it doesn't fall in either of the scenarios.

Now, don't take the other way. I highly respect the efforts that you did over there. I just want to let you know about my way of doing things. In the end, it's up to you to reject some or all of my suggestions. Since, we're both right and the differences lie in the style of coding. In the end, I have huge respect for paying extra efforts for making the codebase stronger.

Feel free to resolve the comment, once you've read it. 😉

kauri-hero:
We are doing QA on this starting today @utkarshg6 @bertllll

PR #223, opened 2023-01-17
GH-611 (additional fix): End scan when accountant receives blockchain scan error

utkarshg6:


node/src/accountant/mod.rs
```diff
@@ -4089,4 +4132,87 @@ mod tests {
     fn wei_to_gwei_blows_up_on_overflow() {
         let _: u64 = wei_to_gwei(u128::MAX);
     }
+
+    fn assert_scan_error_is_handled_properly(test_name: &str, message: ScanError) {
+        init_test_logging();
+        let (ui_gateway, _, ui_gateway_recording_arc) = make_recorder();
+        let mut subject = AccountantBuilder::default().build();
+        subject.logger = Logger::new(test_name);
```
dnwiebe: Check the AccountantBuilder to see if you can supply a logger during the build. Part of the purpose of AccountBuilder, I think, is to give you a subject that doesn't have to be mutable.

node/src/accountant/mod.rs
```diff
@@ -4089,4 +4132,87 @@ mod tests {
...
+        let system = System::new("test");
+        let peer_actors = peer_actors_builder().ui_gateway(ui_gateway).build();
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
+
+        subject_addr.try_send(message.clone()).unwrap();
+
+        subject_addr
+            .try_send(AssertionsMessage {
+                assertions: Box::new(move |actor: &mut Accountant| {
+                    let is_scan_running = match message.scan_type {
```
dnwiebe: Maybe `scan_started_at_opt`?

node/src/accountant/mod.rs
```diff
@@ -4089,4 +4132,87 @@ mod tests {
...
+                TestLogHandler::new().assert_logs_contain_in_order(vec![
+                    &format!("ERROR: {}: Received ScanError: {:?}", test_name, message),
+                    &format!(
+                        "ERROR: {}: Sending UiScanResponse: {:?}",
+                        test_name, expected_message_sent_to_ui
+                    ),
+                ]);
+            }
+            None => {
+                let message_sent_to_ui_opt =
```
dnwiebe: I think you can leave out these three lines and just assert that the recording length is zero.

dnwiebe (approved):

PR #224, opened 2023-01-17
GH-664: HOT FIX ... updating at overflow SQL wrongly written 

bertllll:


node/src/accountant/big_int_processing/big_int_db_processor.rs
```diff
@@ -407,6 +407,27 @@ impl StdNumParamFormNamed {
...
+}
+
+impl<'a> Param<'a> {
+    pub fn new(
+        value_pair: (&'a str, &'a (dyn ExtendedParamsMarker + 'a)),
+        participates_in_update_clauses: bool,
+    ) -> Self {
+        Self {
+            value_pair,
+            participates_in_overflow_update_clauses: participates_in_update_clauses,
```
dnwiebe: A name change would be cool, but not necessary.

node/src/accountant/big_int_processing/big_int_db_processor.rs
```diff
@@ -407,6 +407,27 @@ impl StdNumParamFormNamed {
...
+        }
+    }
+
+    fn value_pair(&'a self) -> (&'a str, &'a dyn ToSql) {
```
dnwiebe: Maybe this method could have a name that signals the type change in the second return value so that people aren't confused.

node/src/accountant/big_int_processing/big_int_db_processor.rs
```diff
@@ -417,22 +438,23 @@ impl<'a> From<&'a WeiChangeAsHighAndLowBytes> for [(&'a str, &'a dyn ToSql); 2]
...
-    fn merge_pure_rusqlite_and_wei_params(
+    fn filter_and_merge_pure_and_wei_params(
         &'a self,
+        for_updates_only: bool,
         wei_change_params: [(&'a str, &'a dyn ToSql); 2],
     ) -> Vec<(&'a str, &'a dyn ToSql)> {
-        self.pure_rusqlite_params(wei_change_params.into_iter())
-            .collect()
-    }
+        let param_preselection_key = if !for_updates_only {
```
dnwiebe: Might be easier to understand if the sense of the test was reversed. Also, `param_preselection_key` sounds like a data item rather than a closure. "Predicate" is a word we commonly use to describe a closure that returns either true or false.

node/src/accountant/big_int_processing/big_int_db_processor.rs
```diff
@@ -407,6 +407,27 @@ impl StdNumParamFormNamed {
     }
 }
 
+pub struct Param<'a> {
+    value_pair: (&'a str, &'a dyn ExtendedParamsMarker),
+    participates_in_overflow_update_clauses: bool,
```
dnwiebe: It might be good to put the flag either at the beginning in both places or at the end in both places...although if you split the `filter_and_merge_pure_and_wei_params()` function into two, this is moot.

node/src/accountant/big_int_processing/big_int_db_processor.rs
```diff
@@ -667,6 +698,49 @@ mod tests {
...
+            params_except_wei_change: vec![
+                Param::new(update_positive_2, true),
+                Param::new(update_negative_1, false),
+                Param::new(update_positive_1, true),
+                Param::new(update_negative_2, false),
+            ],
+        };
+
+        let result = subject.filter_and_merge_pure_and_wei_params(
+            true,
```
dnwiebe: Do we have a test where this flag is false?

node/src/accountant/mod.rs
```diff
@@ -113,6 +113,13 @@ pub struct ResponseSkeleton {
 
 #[derive(Debug, Message, PartialEq, Eq)]
 pub struct ReceivedPayments {
+    // TODO discussion: it's almost never true that all received transactions will
+    // be in the same block and we also disregard how far in the past the block was added
+    // to the blockchain since the scan. Time passed between scans can be standardly 10 or more minutes long.
+    // It's technically possible to avoid udjust bans by also quering the blocks timestamps which could be
```
dnwiebe: Unjust, not udjust; querying, not quering; block's, not blocks; supplied, not supllied

dnwiebe: Okay, how about:

When we decide whether to delinquency-ban a debtor, we do so based on the age of his debt. That age is calculated from the last time he made a payment. It would be most accurate to draw that timestamp from the time the block containing the payment was placed on the blockchain; however, we're actually drawing the timestamp from the moment we discovered and accepted the payment, which is less accurate and detects any upcoming delinquency later than the more accurate version would. Is this a problem? Do we want to correct the timestamp? Discuss.

dnwiebe (commented):
Resume at account_payable_dao.rs:470 green.

dnwiebe (approved):

PR #226, opened 2023-01-26
Rename stream_handler_pool to neighbor_stream_handler_pool inside node/src and node/src/sub_lib

bertllll:
We renamed a number of obscure or misleading elements to make their functions clearer.

Also we finally broke down that huge method in (now) `NeighborStreamHandlerPool`.


utkarshg6 (approved):
I'm pointing out to the renames that happened during this PR in case we want to revisit:

Previous Name | New Name
-- | --
`StreamHandlerPool` | `NeighborStreamHandlerPool`
`StreamHandlerPoolSubs` | `NeighborStreamHandlerPoolSubs`
`stream_handler_pool.rs` | `neighbor_stream_handler_pool.rs`
`StreamHandlerPoolCluster` | `NeighborStreamHandlerPoolCluster`
`Endpoint::Key` | `Endpoint::NeighborPublicKey`

Everything looks good to me.

PR #228, opened 2023-02-21
GH-633(for bert): Review 1 Changes

utkarshg6:


multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -162,14 +163,15 @@ fn dns_resolution_failure_no_longer_blacklists_exit_node_for_all_hosts() {
     let relay1_cryptde =
         CryptDENull::from(&relay1_mock.main_public_key(), TEST_DEFAULT_MULTINODE_CHAIN);
     let cheap_exit_cryptde = CryptDENull::from(&cheap_exit_key, TEST_DEFAULT_MULTINODE_CHAIN);
+    let key_length = normal_exit_key.len();
```
bertllll, line 166: I don't know if my concern is grounded but to me it seems like there are odds that two public keys could have different lengths even though they probably shouldn't correctly. However, structurally, it is just a vector of bytes of unbounded size.

To be honest, I cannot form an opinion on this...

multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -191,23 +193,48 @@ fn dns_resolution_failure_no_longer_blacklists_exit_node_for_all_hosts() {
...
+    // TODO GH-674: Once the new exit node is picked, the Node will continue using that exit node
+    // for all its route unless the current exit node faces an extreme penalty. Maybe it's not the
+    // most economical solution. For example, in the assertion below, we may prefer to use
+    // cheaper_exit_node instead.
+    client.send_chunk("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n".as_bytes());
+    let (_, _, live_cores_package) = relay1_mock
+        .wait_for_package(&masquerader, Duration::from_secs(2))
+        .unwrap();
+    let (_, intended_exit_public_key) =
+        CryptDENull::extract_key_pair(key_length, &live_cores_package.payload);
```
bertllll, line 236: Maybe I'm dumb but why is that we catch a CORES package right behind the originating Node, as it goes to us, the first relay and there are no layers of encryption for the second relay and the exit Node. My train of thoughts: 
If you call simply extract_key_pair(), which is one function with just ordinary arguments, how do you specify that you want to decrypt multiple layers and derive the key of deeper layer - which should be intended for a Node to be instructed to make the next hop towards the exit Node.

This is relay 1 and there is also a relay two, a middleman between the relay 1 and the exit Node. So by decrypting just one layer, you should not get close to the information of who was chosen to be the exit Node and so what public key it points to.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1104,7 +1108,7 @@ impl StandardGossipHandler {
         })
     }
 
-    fn handle_root_node(
+    fn add_gossip_node_as_half_neighbor(
```
bertllll: I suggest to use the name "source node" as it is better known to me and I think it's because it's quite common around here and better understandable.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3588,6 +3591,25 @@ mod tests {
...
+        Source Database (A) ==>
+        - Source has updated version of A and B.
+        - C will not reveal IP, but B, E and F will do.
+
+           Root
+         /  |
+        B---A---E
+        |   |
+        C   F
+         */
```
bertllll, line 3615: I'd add another line before the ending of the commented area; this might give an impression that it's a part of the diagram :-)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3588,6 +3591,25 @@ mod tests {
...
+
+        Source Database (A) ==>
+        - Source has updated version of A and B.
+        - C will not reveal IP, but B, E and F will do.
+
+           Root
+         /  |
+        B---A---E
+        |   |
+        C   F
```
bertllll, line 3612: I'd appreciate if there was a mention of what this second figure depicts. I know it's actually both the networking described by the Gossip and also the resulting state after the Gossip is accepted. Maybe it would be cool to show that picture twice with two different labels. One would say this is the Gossip and the second this is the resulting state of Root database after the processing. I know they are identical but the new way would be more illustrative.

node/src/neighborhood/mod.rs
```diff
@@ -533,7 +533,11 @@ impl Neighborhood {
 
     fn handle_route_query_message(&mut self, msg: RouteQueryMessage) -> Option<RouteQueryResponse> {
         // TODO: Don't do this work unless log level is Debug or Trace
-        let msg_str = format!("{:?}", msg);
+        let msg_str = if self.logger.debug_enabled() {
```
bertllll: Maybe I'd prefer an option here. It's more idiomatic, don't you think? 
Like `let debugged_msg_opt = self.logger.debug_enabled().then(||format!("{:?}", msg));`
You'd unwrap it with expect later in the debug!() call... if it blows up, then we wrote the code badly...it shouldn't, never...

node/src/neighborhood/mod.rs
```diff
@@ -1230,8 +1235,10 @@ impl Neighborhood {
                 hostname_opt,
             )
             .into_iter()
-            .filter(|cr| cr.undesirability <= minimum_undesirability)
-            .map(|cr| cr.nodes)
+            .filter_map(|cr| match cr.undesirability <= minimum_undesirability {
```
bertllll, line 1234: You can also write it in a functional way... `(cr.undesirability <= minimum_undesirability).then_some(cr.nodes)`
I think yours is almost prettier :-))

node/src/neighborhood/mod.rs
```diff
@@ -1374,9 +1381,7 @@ impl Neighborhood {
                 UndesirabilityType::ExitRequest(hostname_opt)
             }
             (RouteDirection::Over, _) => UndesirabilityType::Relay,
-            (RouteDirection::Back, _) if undesirability == 0 => {
-                UndesirabilityType::ExitAndRouteResponse
-            }
+            // The exit-and-relay undesirability is initial_undesirability
```
bertllll, line 1380: `exit-and-relay` seems misleading, it's as if you were talking about what's going on across two components, the exit and also a relay... but you probably meant two responsibilities of the same guy: exit the response and relay it to the first relay...both refer to the Exit Node.

node/src/neighborhood/mod.rs
```diff
@@ -3440,6 +3446,11 @@ mod tests {
                     + rate_pack.exit_charge (1_000) as i64 // charge to exit request
                     + UNREACHABLE_HOST_PENALTY // because host is blacklisted
         );
+        TestLogHandler::new().exists_log_containing(
```
bertllll, line 3445: When you make assertions on logs it's always important to include the beginning of the message which states the log level of this log and also at what Node's component this log was emitted. Such as "INFO: Neighborhood: ..."

node/src/sub_lib/cryptde_null.rs
```diff
@@ -103,12 +103,12 @@ impl CryptDE for CryptDENull {
         signature: &CryptData,
         public_key: &PublicKey,
     ) -> bool {
-        let key_data = public_key.as_slice();
-        let (k, _) = Self::key_and_data(key_data.len(), signature);
-        if k != key_data {
+        let public_key_data = public_key.as_slice();
```
bertllll: I suggest using the word `bytes` instead of `data` which is too vague.

node/src/sub_lib/neighborhood.rs
```diff
@@ -59,12 +59,12 @@ pub struct RatePack {
...
 impl RatePack {
-    pub fn routing_charge(&self, payload_len: usize) -> u64 {
-        self.routing_service_rate + (self.routing_byte_rate * payload_len as u64)
+    pub fn routing_charge(&self, payload_size: u64) -> u64 {
+        self.routing_service_rate + (self.routing_byte_rate * payload_size)
     }
 
-    pub fn exit_charge(&self, payload_len: usize) -> u64 {
-        self.exit_service_rate + (self.exit_byte_rate * payload_len as u64)
+    pub fn exit_charge(&self, payload_size: u64) -> u64 {
```
bertllll, line 66: Well... using a numeric type different from `usize` for lengths looks unusual but I guess I brought you here by asking you to change some numeric data type at other places. 
If there is any easy way to retain usize here then it might be a good decision.

node/src/sub_lib/route.rs
```diff
@@ -373,6 +375,19 @@ mod tests {
         );
     }
 
+    #[test]
+    #[should_panic(expected = "Could not decrypt with ebe5f9a0e2 data beginning with ebe5f9a0e1")]
+    fn id_returns_error_when_the_id_fails_to_decrypt() {
+        let cryptde1 = CryptDENull::from(&PublicKey::new(b"key a"), TEST_DEFAULT_CHAIN);
+        let cryptde2 = CryptDENull::from(&PublicKey::new(b"key b"), TEST_DEFAULT_CHAIN);
+
```
bertllll: I wouldn't expect to see this blank line here. The act is the last line only.

node/src/test_utils/mod.rs
```diff
@@ -182,7 +182,7 @@ pub fn make_meaningless_message_type() -> MessageType {
     DnsResolveFailure_0v1::new(make_meaningless_stream_key()).into()
 }
 
-pub fn make_route(public_keys: Vec<&PublicKey>) -> Route {
+pub fn make_route_to_proxy_client(public_keys: Vec<&PublicKey>) -> Route {
```
bertllll: You could include `one_way_route` in the name. It's not a requirement though.

multinode_integration_tests/tests/communication_failure_test.rs
utkarshg6: At least, in this test it appears that it's safe to use. Since its scope is limited I think it's safe to use.

node/src/neighborhood/gossip_acceptor.rs
utkarshg6: These diagrams are trying to depict how the databases of the Source and destination nodes look before the gossip and are not the accurate depictions of how they really are.

In tests we are establishing half neighborship in a certain direction and not the other. These diagrams fail to represent that part. I expect the viewer to take only the rough idea and not the whole picture. For the whole picture, they must read the test.

In the end, I did modified some changes to make these diagrams clearer, as per your request.

multinode_integration_tests/tests/communication_failure_test.rs
utkarshg6: The `relay1` is the mock node and the Originating Node is a real one. All the other nodes (`relay2`, `cheap_exit` and `normal_exit`) are fictional.

So, my understanding is that the `wait_for_package` is waiting for the response. Since, the other nodes are neither real or mocked, the expected behavior is different from the real world scenario.

node/src/neighborhood/mod.rs
utkarshg6: Yeah, got it. It's done.

node/src/neighborhood/mod.rs
utkarshg6: I'm not the biggest fan of using `then`. I prefer the `match` expression, it feels more rusty to me.

node/src/neighborhood/mod.rs
utkarshg6: I see it as a variant. Makes more sense to me this way.

node/src/neighborhood/mod.rs
utkarshg6: Done! :+1:

node/src/sub_lib/cryptde_null.rs
utkarshg6: Done :+1:

node/src/sub_lib/neighborhood.rs
utkarshg6: Using `u64` adds a convenience over here. Since, any value bigger than `u64` would never be used, it makes more sense to keep the fn signature accurate instead of using vague data type like `usize`.

node/src/sub_lib/route.rs
utkarshg6: Done :+1:

node/src/test_utils/mod.rs
utkarshg6: Done :+1:

multinode_integration_tests/tests/communication_failure_test.rs
bertllll: I'd rather talk this over again with Dan, okay?

utkarshg6: Sure.

PR #229, opened 2023-02-22
GH-673: Fixing some weak points in tests

bertllll:


node/src/accountant/mod.rs
```diff
@@ -2166,10 +2168,9 @@ mod tests {
         ];
         let payable_dao = PayableDaoMock::default().non_pending_payables_result(accounts.clone());
         let (blockchain_bridge, _, blockchain_bridge_recordings_arc) = make_recorder();
-        let mut expected_messages_by_type = HashMap::new();
-        expected_messages_by_type.insert(TypeId::of::<ReportAccountsPayable>(), 1);
-        let blockchain_bridge = blockchain_bridge
-            .stop_after_messages_and_start_system_killer(expected_messages_by_type);
+        let blockchain_bridge = blockchain_bridge.stop_conditions(StopConditions::Single(
```
dnwiebe: Macro here

node/src/accountant/mod.rs
```diff
@@ -2206,7 +2207,13 @@ mod tests {
         let test_name = "accountant_does_not_initiate_another_scan_in_case_it_receives_the_message_and_the_scanner_is_running";
         let payable_dao = PayableDaoMock::default();
         let (blockchain_bridge, _, blockchain_bridge_recording) = make_recorder();
-        let report_accounts_payable_sub = blockchain_bridge.start().recipient();
+        let blockchain_bridge_addr = blockchain_bridge
+            .stop_conditions(multiple_type_ids!(
```
dnwiebe: Better name: `match_every_type_id`

node/src/accountant/mod.rs
```diff
@@ -2216,31 +2223,62 @@ mod tests {
...
+        addr.try_send(message_before.clone()).unwrap();
+
         addr.try_send(ScanForPayables {
             response_skeleton_opt: None,
         })
         .unwrap();
 
-        addr.try_send(ScanForPayables {
-            response_skeleton_opt: None,
+        //let's use this assertion message for changing the Accountant's state
```
dnwiebe: Can you comment more extensively? Maybe something like:

We ignored the second ScanForPayables message because the first message meant a scan was already in progress; now let's make it look like that scan has ended so that we can prove the next message will start another one.

node/src/stream_messages.rs
```diff
@@ -78,8 +78,20 @@ impl Debug for PoolBindMessage {
...
+    use crate::sub_lib::tokio_wrappers::{ReadHalfWrapperReal, WriteHalfWrapperReal};
     use crate::test_utils::recorder::peer_actors_builder;
+    use crate::test_utils::tokio_wrapper_mocks::{ReadHalfWrapperMock, WriteHalfWrapperMock};
     use actix::System;
+    use masq_lib::utils::find_free_port;
+    use std::net::{IpAddr, Ipv4Addr, SocketAddrV4, TcpStream};
+
+    impl PartialEq for AddStreamMsg {
+        fn eq(&self, _other: &Self) -> bool {
+            // this trait cannot be implemented really, it's just a requirement from code
```
dnwiebe: Suggestion:

We need to implement PartialEq so that AddStreamMsg can be received by the Recorder; but AddStreamMsg breaks the rules for an actor message by containing references to outside resources (namely, an I/O stream) and therefore cannot have a real implementation of PartialEq. So here we break the rules again to patch up the problems created by the first breach of the rules. Don't move this into the production tree; it only needs to be here for the Recorder, and the Recorder is only in the test tree.

node/src/sub_lib/tokio_wrappers.rs
```diff
@@ -17,9 +17,13 @@ pub trait TokioListenerWrapper: Send {
     fn poll_accept(&mut self) -> Result<Async<(TcpStream, SocketAddr)>, io::Error>;
 }
 
-pub trait ReadHalfWrapper: Send + AsyncRead {}
+pub trait ReadHalfWrapper: Send + AsyncRead {
+    as_any_dcl!();
```
dnwiebe: You didn't like these.

node/src/sub_lib/tokio_wrappers.rs
```diff
@@ -17,9 +17,13 @@ pub trait TokioListenerWrapper: Send {
...
+}
 
-pub trait WriteHalfWrapper: Send + AsyncWrite {}
+pub trait WriteHalfWrapper: Send + AsyncWrite {
+    as_any_dcl!();
```
dnwiebe: Suggestion: I think these macros would be easier to understand if they were named `declare_as_any!` and `implement_as_any!`.

node/src/test_utils/mod.rs
```diff
@@ -624,7 +627,9 @@ pub mod unshared_test_utils {
     {
         let (recorder, _, recording_arc) = make_recorder();
         let recorder = match stopping_message {
-            Some(type_id) => recorder.stop_condition(type_id), // No need to write stop message after this
+            Some(type_id) => {
+                recorder.stop_conditions(StopConditions::Single(StopCondition::StopOnType(type_id)))
```
dnwiebe: Needs macro

node/src/test_utils/recorder.rs
```diff
@@ -235,20 +217,33 @@ impl Recorder {
...
-    ) -> Recorder {
+    fn start_system_killer(&mut self) {
         let system_killer = SystemKillerActor::new(Duration::from_secs(15));
         system_killer.start();
-        self.expected_count_by_msg_type_opt = Some(expected_count_by_messages);
-        self
+    }
+
+    fn handle_msg<T: 'static + PartialEq + Send>(&mut self, msg: T) {
+        let kill_system = if let Some(stop_condition) = &mut self.stop_conditions_opt {
```
dnwiebe: `stop_condition` should be plural

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    fn inspect_mutable<T: PartialEq + Send + 'static>(&mut self, msg: &T) -> bool {
+        match self {
+            StopConditions::All(conditions) => {
+                let indexes_to_remove = Self::indexes_of_matched_conditions(conditions, msg);
+                Self::remove_matched_conditions(conditions, indexes_to_remove);
+                conditions.is_empty()
+            }
+            _ => unreachable!("something is wrong"),
```
dnwiebe: This message can probably be improved. Imagine that somebody else quoted it to you and you had to track it down.

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    fn remove_matched_conditions(
+        conditions: &mut Vec<StopCondition>,
+        indexes_to_remove: Vec<usize>,
+    ) {
+        if !indexes_to_remove.is_empty() {
+            let _ = indexes_to_remove
+                .into_iter()
+                .fold(0, |removed_counter: usize, idx| {
```
dnwiebe: You could do this without an accumulator if you reversed `indexes_to_remove`. After that, you could use `for_each`.

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -0,0 +1,325 @@
...
+    fn matches_stop_on_predicate<T: Send + 'static>(
+        predicate: &dyn Fn(RefMsgExpected) -> bool,
+        msg: &T,
+    ) -> bool {
+        predicate(msg as RefMsgExpected)
+    }
+}
+
+#[macro_export]
+macro_rules! single_type_id {
```
dnwiebe: For consistency, this macro name should change as well, perhaps to `match_type_id!`.

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -0,0 +1,325 @@
...
+            false
+        );
+        assert_eq!(
+            cond_set.resolve_stop_conditions::<ScanError>(&good_msg),
+            true
+        )
+    }
+
+    #[test]
+    fn match_any_works() {
```
dnwiebe: Maybe mention something about the number of conditions not decreasing as the conditions are resolved, since this test does assert that. It'd help explain what `inspect_len_of_any` is all about.

dnwiebe (approved):

PR #230, opened 2023-02-28
GH-673: review 2

bertllll:


node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -34,9 +34,6 @@ impl StopConditions {
 
     fn inspect_immutable<T: PartialEq + Send + 'static>(&self, msg: &T) -> Option<bool> {
```
dnwiebe: Now that `StopConditions::Single` is gone, these `inspect_xxx` functions can probably be reworked to be prettier and more symmetrical.

node/src/accountant/mod.rs
```diff
@@ -1125,19 +1111,22 @@ mod tests {
         );
 
         let financial_statistics = result.financial_statistics().clone();
-        let notify_later = result.notify_later;
+        let notify_later = result.scan_timings;
```
dnwiebe: Variable should be renamed

node/src/accountant/mod.rs
```diff
@@ -1252,7 +1241,7 @@ mod tests {
             PayableDaoMock::new().non_pending_payables_result(vec![payable_account.clone()]);
         let subject = AccountantBuilder::default()
             .bootstrapper_config(config)
-            .payable_daos(vec![PayableScannerDest(payable_dao)])
+            .payable_daos(vec![ToPayableScanner(payable_dao)])
```
dnwiebe: I think the ToXxxScanner (or whatever) name is fine; if you're still not satisfied with it, you might try ForXxxScanner; it kind of fits with the rest of the code when it's called.

dnwiebe (approved):

PR #231, opened 2023-03-01
GH-633 (for bert): Review 2 Changes

utkarshg6:


node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3593,21 +3593,30 @@ mod tests {
...
          /  |
         B---A---E
         |   |
         C   F
+
+        <------------------------------------->
+
+        Source Node will send a gossip to the destination Node with
+        an updated version of Node A and B. The IP of Node C won't
+        be revealed but for Nodes B, E and F, it will be revealed.
```
bertllll: This comment might be too much compared to what it says about the test itself, which is low. Maybe we don't talk about this things because as per my understanding this test is actually about the Node D.

Also, if you insist on the information about B, E, F with address revealed...check it with the GossipBuilder, it was sadly set up differently. 
I caught myself finding out that I don't really know what the flag "reveal address" really means, because Root should definitely not know the addresses of E and F. Right?

bertllll (commented):
Please have hold some thinking about the test with the figures and how to make it even a bit clearer, including the discrepancy in the revealed IPs

node/src/neighborhood/gossip_acceptor.rs
utkarshg6: You are right, the root must not know the address of E and F. That's exactly what we're testing. In the right scenario, the ip addrs of these nodes should never get revealed. In this test we're showing what'll happen if they get revealed accidentally.

I think it's better to remove any written text. Since, in the end the dev has to rely on the test.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -3593,21 +3593,25 @@ mod tests {
...
-        - C will not reveal IP, but B, E and F will do.
 
-           Root
+          Root
          /  |
         B---A---E
         |   |
         C   F
+
+        <------------------------------------->
```
bertllll, line 3614: I still think this test has been long-time wrong, not in a critical way, but those flags for reveal-node-addr should be right only as: 
```
let gossip = GossipBuilder::new(&src_db)
            .node(node_a.public_key(), true)
            .node(node_b.public_key(), true)
            .node(node_c.public_key(), false)
            .node(node_e.public_key(), false)
            .node(node_f.public_key(), false)
            .build();
```

utkarshg6: No, the test is right the way it is. The IP Addresses of Node E and F were revealed on purpose. The test is about how will destination Node respond if the source Node accidentally reveals the IP Addresses of the Nodes it wasn't suppose to reveal.

PR #233, opened 2023-03-02
GH-671: base for PaymentAdjuster-checked payables

bertllll:


node/src/blockchain/blockchain_interface.rs
```diff
@@ -1076,6 +1070,16 @@ mod tests {
     #[test]
     fn blockchain_interface_non_clandestine_returns_an_error_for_unintelligible_response_when_requesting_token_balance(
     ) {
+        let act = |subject: &BlockchainInterfaceNonClandestine<Http>, wallet: &Wallet| {
+            subject.get_token_balance(wallet)
+        };
+        error_on_requesting_balance(act);
+    }
+
+    fn error_on_requesting_balance<F>(act: F)
```
bertllll: Should start with "assert_..."

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -323,19 +323,27 @@ fn assert_balances(
     expected_eth_balance: &str,
     expected_token_balance: &str,
 ) {
-    if let (Ok(eth_balance), Ok(token_balance)) = blockchain_interface.get_balances(&wallet) {
+    if let Ok(eth_balance) = blockchain_interface.get_gas_balance(&wallet) {
```
utkarshg6: I'd rather prefer if you use `expect()` over here.

You can pull out the `etc_balance` and `token_balance` on the consecutive lines and then assert on them later.

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -323,19 +323,27 @@ fn assert_balances(
...
             String::from(expected_eth_balance),
-            "EthBalance"
+            "Actual EthBalance {} doesn't much with expected {}",
+            eth_balance,
+            expected_eth_balance
         );
+    } else {
+        panic!("Failed to retrieve gas balance for {}", wallet);
+    }
+    if let Ok(token_balance) = blockchain_interface.get_token_balance(&wallet) {
```
utkarshg6: This should be pulled after the next line the `eth_balance` is taken out.

node/src/accountant/mod.rs
```diff
@@ -129,6 +133,13 @@ pub struct SentPayables {
     pub response_skeleton_opt: Option<ResponseSkeleton>,
 }
 
+#[derive(Debug, Message, PartialEq, Eq)]
+pub struct AvailableBalancesAndQualifiedPayables {
+    pub accounts: Vec<PayableAccount>,
```
utkarshg6: I'd rather prefer if this field is called `qualified_payables`. It'll confirm that this field is pointing to the same entity the struct is named after.

node/src/accountant/mod.rs
```diff
@@ -201,12 +212,32 @@ impl Handler<ReceivedPayments> for Accountant {
...
 
+impl Handler<AvailableBalancesAndQualifiedPayables> for Accountant {
+    type Result = ();
+
+    fn handle(
+        &mut self,
+        msg: AvailableBalancesAndQualifiedPayables,
+        _ctx: &mut Self::Context,
+    ) -> Self::Result {
+        //TODO GH-672 is gonna take place here
```
utkarshg6: Can you please add a comment saying that:

```
TODO: GH-672 - PaymentAdjuster hasn't been implemented yet
```

node/src/accountant/mod.rs
```diff
@@ -1309,6 +1351,51 @@ mod tests {
...
+    ) {
+        let mut subject = AccountantBuilder::default().build();
+        let (blockchain_bridge, _, blockchain_bridge_recording_arc) = make_recorder();
+        let report_recipient = blockchain_bridge
+            .system_stop_conditions(match_every_type_id!(ReportAccountsPayable))
+            .start()
+            .recipient();
+        subject.report_accounts_payable_sub_opt = Some(report_recipient);
+        let subject_addr = subject.start();
+        let number_closely_fitting_under = ((u32::MAX / 2) - 1) as u64 / WEIS_OF_GWEI as u64;
```
utkarshg6: It's hard to understand why exactly this number was picked.

I anticipate that it's either the corner case, or a value close to some threshold value.

Please rename the constant `WEIS_OF_GWEI` to `WEIS_IN_GWEI`.

If you create something like this, it'll be easier to understand:

```rust
let number_closely_fitting_under =  {
   let numerator = ((u32::MAX / 2) - 1);
   let value_in_gwei = numerator / WEIS_IN_GWEI;
   value_in_gwei as u64
};
```

I used `numerator` as an example, please replace it with something more meaningful. In case, you want to break the numerator further, then I won't stop you. 😉

node/src/accountant/scanners.rs
```diff
@@ -43,7 +44,7 @@ use time::OffsetDateTime;
 use web3::types::TransactionReceipt;
 
 pub struct Scanners {
-    pub payable: Box<dyn Scanner<ReportAccountsPayable, SentPayables>>,
+    pub payable: Box<dyn Scanner<RequestAvailableBalancesForPayables, SentPayables>>,
```
utkarshg6: Correct me if I'm wrong.

I do think, these two things are misleading. 

`Balances`: I think it should be `Balance` instead, since it's the wallet balance of the wallet.
`For`: When we use the word for, it can be interpreted as that there will be a balance dedicated to each payable.

My suggestion is to rename it to `RequestWalletBalanceAndPayables`.

node/src/accountant/mod.rs
```diff
@@ -129,6 +133,13 @@ pub struct SentPayables {
     pub response_skeleton_opt: Option<ResponseSkeleton>,
 }
 
+#[derive(Debug, Message, PartialEq, Eq)]
+pub struct AvailableBalancesAndQualifiedPayables {
+    pub accounts: Vec<PayableAccount>,
+    pub consuming_wallet_balances: WalletBalances,
```
utkarshg6: Although, there are two fields in `WalletBalances`, the user will only pay with MASQ tokens, which in turn represents their balance, gas is more like a fuel to transfer their balance from one place to another.

So, a more easy to understand name would be `WalletBalance`.

The benefit of using such a name is that it will prevent us from some confusions.

Step 1 -> We can make the name of the struct singular to `AvailableBalanceAndQualifiedPayables`.

Step 2 -> Change the name inside scanners to a singular name `RequestAvailableBalanceForPayables` or some other appropriate name. There's a comment over there too.

The benefit of this name change will prevent the confusion that  _there is a balance associated with each payable_. By making it singular it'll be easier to understand that the balance is a unique quantity and is disassociated from payables.

PS: If you do make this change, don' forget to change the field name to `consuming_wallet_balance`.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -46,6 +49,7 @@ pub struct BlockchainBridge {
     persistent_config: Box<dyn PersistentConfiguration>,
     set_consuming_wallet_subs_opt: Option<Vec<Recipient<SetConsumingWalletMessage>>>,
     sent_payable_subs_opt: Option<Recipient<SentPayables>>,
+    balances_and_payables_sub_opt: Option<Recipient<AvailableBalancesAndQualifiedPayables>>,
```
utkarshg6: This field should be renamed to `walltet_balance_and_payables_sub_opt`.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -227,18 +249,77 @@ impl BlockchainBridge {
...
+            Some(wallet) => wallet,
+            None => {
+                return Err(
+                    "Cannot inspect available balances for payables while consuming wallet \
+                    is missing"
+                        .to_string(),
+                )
+            }
+        };
+        //TODO rewrite this into a batch call as soon as a card also involving this feature gets into master
```
utkarshg6: Can you also specify the card number? It'll make it very easy to find the comment.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -227,18 +249,77 @@ impl BlockchainBridge {
...
+            })
+            .expect("Accountant is dead");
+
+        Ok(())
+    }
+
     fn handle_report_accounts_payable(
         &mut self,
-        creditors_msg: &ReportAccountsPayable,
+        creditors_msg: ReportAccountsPayable,
```
utkarshg6, line 314: I don't like that we're preferring ownership over here.

I would rather prefer if we use the reference over here. Since, we're only reading the information ultimately, It would be appropriate if we keep ownership at a higher level.

I understand that there is a reason for doing so, but I don't think it's worth it of removing the flexibility of keeping the ownership at a higher level.

It's your choice if you want to use a reference or ownership. Just please consider the trade-off and decide wisely. 😎

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -350,9 +431,9 @@ impl BlockchainBridge {
         Ok(())
     }
 
-    fn handle_scan<M, F>(&mut self, handler: F, scan_type: ScanType, msg: &M)
+    fn handle_scan<M, F>(&mut self, handler: F, scan_type: ScanType, msg: M)
     where
-        F: FnOnce(&mut BlockchainBridge, &M) -> Result<(), String>,
+        F: FnOnce(&mut BlockchainBridge, M) -> Result<(), String>,
```
utkarshg6, line 431: This makes me a bit nervous. I'm unsure whether this big change is worth it.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -269,7 +350,7 @@ impl BlockchainBridge {
         }
     }
 
-    fn handle_retrieve_transactions(&mut self, msg: &RetrieveTransactions) -> Result<(), String> {
+    fn handle_retrieve_transactions(&mut self, msg: RetrieveTransactions) -> Result<(), String> {
```
utkarshg6, line 348: Same as above comment.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -604,6 +687,188 @@ mod tests {
...
+                    client_id: 11122,
+                    context_id: 444
+                })
+            }
+        );
+    }
+
+    fn assert_failure_on_balance_inspection(
+        blockchain_interface: BlockchainInterfaceMock,
+        error_msg: &str,
```
utkarshg6, line 760: I'd like if we add one more parameter, `test_name`. It can be used at 3 places:

1. Creating System
2. Logger
3. `wallet` field in `RequestAvailableBalancesForPayables`

node/src/blockchain/blockchain_interface.rs
utkarshg6: I'd appreciate if you change it to that.

Well, great work defining this function. 🤙 

Although, I was hoping for more than one occurrence.

I think we can use this design in other places wherever possible.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -1102,43 +1107,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn blockchain_interface_non_clandestine_can_request_both_eth_and_token_balances_happy_path() {
```
utkarshg6, line 1106: I don't see this test migrated anywhere, is there any other test testing something similar?

node/src/sub_lib/accountant.rs
```diff
@@ -94,6 +94,8 @@ pub struct AccountantSubs {
     pub report_routing_service_provided: Recipient<ReportRoutingServiceProvidedMessage>,
     pub report_exit_service_provided: Recipient<ReportExitServiceProvidedMessage>,
     pub report_services_consumed: Recipient<ReportServicesConsumedMessage>,
+    pub report_our_balances_and_qualified_payables:
```
utkarshg6: It would be better to rename it to `report_balance_and_qualified_payables`

node/src/sub_lib/blockchain_bridge.rs
```diff
@@ -22,6 +23,7 @@ pub struct BlockchainBridgeConfig {
 pub struct BlockchainBridgeSubs {
     pub bind: Recipient<BindMessage>,
     pub report_accounts_payable: Recipient<ReportAccountsPayable>,
+    pub request_available_balances_for_payables: Recipient<RequestAvailableBalancesForPayables>,
```
utkarshg6: In case you change the name of this message, please change the name of this field too.

node/src/sub_lib/blockchain_bridge.rs
```diff
@@ -34,27 +36,33 @@ impl Debug for BlockchainBridgeSubs {
...
 
-impl SkeletonOptHolder for ReportAccountsPayable {
+impl SkeletonOptHolder for RequestAvailableBalancesForPayables {
     fn skeleton_opt(&self) -> Option<ResponseSkeleton> {
         self.response_skeleton_opt
     }
 }
 
 #[derive(Clone, PartialEq, Eq, Debug, Message)]
-pub struct SetDbPasswordMsg {
```
utkarshg6, line 49: I don't understand why this message was removed.

node/src/sub_lib/blockchain_bridge.rs
```diff
@@ -34,27 +36,33 @@ impl Debug for BlockchainBridgeSubs {
...
-    pub client_id: u64,
-    pub password: String,
+pub struct ReportAccountsPayable {
+    pub accounts: Vec<PayableAccount>,
+    pub response_skeleton_opt: Option<ResponseSkeleton>,
 }
 
-#[derive(Clone, PartialEq, Eq, Debug, Message)]
-pub struct SetGasPriceMsg {
```
utkarshg6, line 55: Same with this, I don't understand why was it removed?

node/src/test_utils/recorder.rs
```diff
@@ -408,6 +413,10 @@ pub fn make_accountant_subs_from_recorder(addr: &Addr<Recorder>) -> AccountantSu
         report_routing_service_provided: recipient!(addr, ReportRoutingServiceProvidedMessage),
         report_exit_service_provided: recipient!(addr, ReportExitServiceProvidedMessage),
         report_services_consumed: recipient!(addr, ReportServicesConsumedMessage),
+        report_our_balances_and_qualified_payables: recipient!(
```
utkarshg6: This looks like a formatting error. It could be possible that GitHub is not showing it properly. Anyways, don't forget to run the script :)

utkarshg6 (commented):
Thanks for the opportunity! 🙏

node/src/sub_lib/blockchain_bridge.rs
bertllll: Never used in the whole codebase. Technically dead code but masqueraded with the derive impls.

node/src/sub_lib/blockchain_bridge.rs
bertllll: Same as my comment above...

node/src/accountant/scanners.rs
bertllll: Well, there are balanceS. The balance of MATIC or ETH and balance of MASQ (also MASQ on Polygon).

So I think I wanted to say that in the msg name. 

What do you make of these: `RequestBalancesToPayPayables` or rather `InspectConsumingWalletForPayables`,
`InspectConsumingWalletToPayPayables`?

node/src/blockchain/blockchain_interface.rs
bertllll: I deleted that method of the trait therefore this test has no reason to exist. It used to be just a coupled caller of `request_eth_balance()` and `request_token_balance()` which are also methods of the same trait. Using the coupled function with its nested structure made it impossible to mock these methods exactly as wanted to. 

So that's it...

node/src/test_utils/recorder.rs
bertllll: It's a valid formatting. :man_shrugging: 

The formater forces this layout. I couldn't think of a better one which wouldn't be knocked out by the formatter so I left it untouched.

node/src/blockchain/blockchain_bridge.rs
bertllll: To be honest, the ownership doesn't make much sense if we keep it at the tier up by one. Think with me:

Firstly, think about the Actor message Handler and what we could say about it? The best option is to have very clear content in there, ideally a single function and nothing else.
Next, what the handle is suppose to do in general?
To take a message and process it (it usually brings data along). What data is easiest to process? Well, the owned data. In other words _I would find it bad code if somebody widened the competences of the handler, trying to use the message for anything outside the actual msg processing machinery_. 

There is just one thing to my eyes which could make devs nervous. It is the fact that the arguments above don't match with the call of  `self.handle_scan(XXX_the_fn_to_process_the_msgXXX, other param, 
msg)` because technically it is another layer around the msg processing function per se. Yeah, I was thinking about it now as well as then and I do think it's worth a big concern. Look at the guts of it. 
One, it is very simple inside. Just two usages of the msg. Two, more importantly, it has to take it as generics. Think how inconvenient is to work with generics. Nobody sane would deliberately add anything here that would interact with the variable `msg`. He would have to implement a new trait as a complement to the other trait, which requires also defining a supertrait to be able to set bounds for the generic type. It's hassle and if somebody had done that there would have probably been another way to do that, being much easier.    

For these reasons I'm fine to require an owned version of it because it is completely safe to me. Also moving the ownership closer to the guts of the data processing functions is another and huge positive trade-off and somebody should have done it before me.

node/src/blockchain/blockchain_bridge.rs
bertllll: I'm opposed. Explained below...

node/src/blockchain/blockchain_bridge.rs
bertllll: Explained above...

node/src/accountant/mod.rs
bertllll: I agree that my name isn't pretty but that doesn't change the fact that it does refer to two balances of our consuming wallet. 

The gas is as important as the tokens. If you don't have enough gas, the payment will fail. No excuse...
In that sense I put both balances at the same level. 

Suggested renaming: 

Instead of `consuming_wallet_balances` `consuming_balances` should work as we never can know other than our consuming wallet. 

`WalletBalance` should be renamed to `ConsumingWalletBalances`.

`AvailableBalanceAndQualifiedPayables` would work better as `ConsumingBalancesAndQualifiedPayables`.

From here on everything would be abide by this law :point_up: .

node/src/blockchain/blockchain_bridge.rs
bertllll: See the first answer to your comment touching this.

node/src/sub_lib/accountant.rs
bertllll: I will change this one...

But accordingly to the final decision about the greater naming.

node/src/sub_lib/blockchain_bridge.rs
bertllll: See the first answer to your comment touching this.

node/src/blockchain/blockchain_interface.rs
bertllll: Thanks for pointing me out to the missing usage with another function. :pray:

node/src/accountant/scanners.rs
utkarshg6: I really like this name: `RequestBalancesToPayPayables`. I hope you'll use this one. 🙏

node/src/accountant/mod.rs
utkarshg6: I'd appreciate, that you're doing the effort of renames.

Yes, the names that you provided are easier to understand.

node/src/blockchain/blockchain_bridge.rs
utkarshg6: The way you pointed out, that you don't want any mutation to happen inside msg, it totally makes sense.

I totally agree with your point of view. This is an exception, if we would be dealing with some structure which is supposed to be processed differently than a `msg`, I'd like to keep the powers at the highest level possible, so the functions should only be granted the permissions as per their actions. If a function is able to perform all it's instructions just by reading the structure, then I'd prefer references over mutation or ownership.

Since, this is an exception, I just want to appreciate your efforts for making the right choice.

node/src/blockchain/blockchain_interface.rs
utkarshg6: That's what I guessed. I just wanted to confirm that it is intended and not something that your eyes missed.

node/src/sub_lib/accountant.rs
utkarshg6: Ok, let's do that.

node/src/sub_lib/blockchain_bridge.rs
utkarshg6: Ok, let's do that.

node/src/sub_lib/blockchain_bridge.rs
utkarshg6: That's a fault how `rustc` works. Thanks for cleaning it up.

utkarshg6 (approved):

utkarshg6 (approved):

utkarshg6 (approved):

PR #235, opened 2023-03-08
GH-633: review five

bertllll:


multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -146,59 +145,69 @@ impl MASQNode for MASQMockNode {
     }
 }
 
-pub struct MASQMockNodeWithMutableHandle {
+pub struct ModifiableMasqMockNode {
```
dnwiebe: Would like to see consistency in capitalization

multinode_integration_tests/src/masq_node_cluster.rs
```diff
@@ -91,63 +88,62 @@ impl MASQNodeCluster {
...
             index,
             self.host_node_parent_dir.clone(),
-            mock_node_starter_mode,
             public_key_opt,
             self.chain,
         )
     }
 
-    pub fn finalize_mock_node_with_mutable_handle(
+    fn start_mock_node_with_mutable_handle(
```
dnwiebe: There's some duplicate code here. It's not a lot of duplicate code, but getting rid of it would be nice, whether with a closure or with a trait. On the other hand, if that's less comprehensible than leaving it as it is, leaving it is fine.

multinode_integration_tests/src/masq_node_cluster.rs
```diff
@@ -91,63 +88,62 @@ impl MASQNodeCluster {
...
+        MASQMockNodeStarter::start_with_mutable_handle(
+            ports,
+            index,
+            self.host_node_parent_dir.clone(),
+            public_key_opt,
+            self.chain,
+        )
+    }
+
+    pub fn finalize_mutable_mock_node_and_add_it_to_cluster(
```
dnwiebe: If the parameter has a good name, you don't need to mention what is being added in the method name, because the parameter name will take care of that. Also, it should be evident from the call what is being added to (a cluster). Sooo...how about `finalize_and_add`?

multinode_integration_tests/src/neighborhood_constructor.rs
```diff
@@ -188,9 +188,9 @@ fn form_mock_node_skeleton(
         .into_iter()
         .map(|model_node_key| {
             let mut configurable_node =
-                cluster.start_configurable_mock_node_with_public_key(vec![10000], model_node_key);
+                cluster.start_modifiable_mock_node_with_public_key(vec![10000], model_node_key);
```
dnwiebe: Either "modifiable" or "mutable" works, but consistency is good.

dnwiebe (approved):

multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -470,6 +451,58 @@ impl MASQMockNode {
         stream.shutdown(Shutdown::Both).unwrap();
     }
 
+    fn start_masq_mock_nodes_with_bare_guts(
```
dnwiebe: "nodes" should be singular

dnwiebe (approved):

PR #236, opened 2023-03-09
GH-677:  Refactor the extremely long function `handle_dispatcher_node_query_response()`

utkarshg6:


node/src/stream_handler_pool.rs
```diff
@@ -367,218 +370,352 @@ impl StreamHandlerPool {
...
                 }
+            }
+        } else {
+            SequencedPacket::from(&msg.context)
+        };
 
+        let packet_len = packet.data.len();
+        match tx_box.unbounded_send(packet) {
+            Err(e) => {
                 debug!(
```
dnwiebe: This should probably be an `error!` log, unless it turns out to happen very often, in which case it should be a `warning!` log.

node/src/stream_handler_pool.rs
```diff
@@ -367,218 +370,352 @@ impl StreamHandlerPool {
...
+            peer_addr,
+            msg.context.data.len()
+        );
+        let recipient = self
+            .self_subs_opt
+            .as_ref()
+            .expect("StreamHandlerPool is unbound.")
+            .node_query_response
+            .clone();
+        // TODO FIXME revisit once SC-358/GH-96 is done (idea: use notify_later() to delay messages)
```
dnwiebe, line 542: Maybe explore `notify_later()` here; if it's easy, do it in this card, and if it's hard, make another card for it.

node/src/stream_handler_pool.rs
```diff
@@ -367,218 +370,352 @@ impl StreamHandlerPool {
...
-                            peer_addr: peer_addr.ip(),
-                            event: ConnectionProgressEvent::TcpConnectionFailed
-                        };
-                        connection_progress_sub_err.try_send(connection_progress_message).expect("Neighborhood is dead");
-                    });
-
-                debug!(self.logger, "Beginning connection attempt to {}", peer_addr);
-                tokio::spawn(connect_future);
-            }
+    // TODO: This method is wayyyy too big
```
dnwiebe: Remove comment

dnwiebe (approved):

node/src/stream_handler_pool.rs
utkarshg6: Apparently, it'll require us to invent a new message, which can schedule a `notify_later()`. It would be better if this new message stays generic and can be used in other parts of the codebase.

I believe, it deserves a new card. So, I created this card - https://github.com/MASQ-Project/Node-issues/issues/678

dnwiebe (approved):
Approved with the caveat that `stream_handler_pool.rs:1760` is changed to use `find_free_port()`.

node/src/stream_handler_pool.rs
```diff
@@ -300,7 +300,7 @@ impl StreamHandlerPool {
                 };
                 debug!(
                     self.logger,
-                    "Sending node query about {:?} to Neighborhood", key
+                    "Sending node query about {} to Neighborhood", key
```
dnwiebe, line 303: Investigate to see if this will result in three dots replacing most of the key value. If it will, switch back to the Debug implementation.

dnwiebe (approved):

node/src/stream_handler_pool.rs
utkarshg6: It turns out the key is not truncated while printing it either with `Display` or `Debug`.

This is how Key gets printed using:

`Display`: `CAQIBA`
`Debug`: `0x08040804`

Also, there wasn't any assertion for this debug log inside the tests that were testing this section of code. So, I introduced one.

dnwiebe (approved):

dnwiebe (approved):

PR #239, opened 2023-03-10
GH-677: Review 1 Changes

utkarshg6:


multinode_integration_tests/tests/data_routing_test.rs
```diff
@@ -316,7 +316,7 @@ fn multiple_stream_zero_hop_test() {
     let mut another_client = zero_hop_node.make_client(8080);
 
     one_client.send_chunk(b"GET / HTTP/1.1\r\nHost: www.example.com\r\n\r\n");
-    another_client.send_chunk(b"GET / HTTP/1.1\r\nHost: www.fallingfalling.com\r\n\r\n");
+    another_client.send_chunk(b"GET / HTTP/1.1\r\nHost: www.neverssl.com\r\n\r\n");
```
dnwiebe: This particular URL will cause a redirect, which is not really what you want. There's a trick you can use with neverssl.com to avoid that redirect. Pick a string that's not `www` and use it as the first element of the domain name, and end the URL with `/online/`. Like this:
```
b"GET /online/ HTTP/1.1\r\nHost: booga.neverssl.com\r\n\r\n"
```
That should get you in without a redirect, whereupon you can check the body for something reliable, like "neverssl.com will never use SSL".

node/src/stream_handler_pool.rs
```diff
@@ -488,7 +488,7 @@ impl StreamHandlerPool {
         let packet_len = packet.data.len();
         match tx_box.unbounded_send(packet) {
             Err(e) => {
-                debug!(
+                error!(
```
dnwiebe, line 491: Check and make sure there's a test that exercises this. If there is, it should assert on this `error!()` log. If there isn't, one should be written.

PR #240, opened 2023-03-12
fix: replaced old copyright in scripts with masq copyright

FinsaasGH:


.gitignore
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2019, Substratum LLC (https://substratum.net) and/or its affiliates. All rights reserved.
+# Copyright (c) 2019-2021, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
kauri-hero: Thanks @FinsaasGH 

could you change the year to be 2023 to bring it current?

Unsure if the year 2021 was intended for another reason?

kauri-hero (changes requested):

.gitignore
bertllll: Hi @FinsaasGH and @kauri-hero, you might want to take inspiration from the arrangement which we use practically at the first line of every Rust file in our codebase. It states this: `Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.` As you can see, we've back then decided for some reason to simplify the year stamp just to the date of our inception. What do you think?

Regardless what you prefer, you probably will want to synchronize all those other copyright statements with these newly updated in the bash files so that they don't look different.

kauri-hero: agreed that would be appropriate.

Please follow this suggestion @FinsaasGH

kauri-hero (approved):

kauri-hero:
looks good to me now @FinsaasGH 
If you want to check once more @bertllll to ensure we got all the files including tests - looks like they are all there

kauri-hero (approved):

PR #242, opened 2023-03-15
GH-679: Create a script for bumping version

utkarshg6:


ci/bump_version.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -xv
+
+if [ $# != 1 ]
+then
+     echo "Please provide version as the first argument"
+     exit 1
+fi
+
+version=$1
```
dnwiebe: This $1 should have double quotes: "$1"

ci/bump_version.sh
```diff
@@ -0,0 +1,51 @@
...
+CI_DIR="$( cd "$( dirname "$0" )" && pwd )"
+file=Cargo.toml
+final_exit_code=0
+
+declare -a crates=(
```
dnwiebe: Think about getting this from an `ls` command.

ci/bump_version.sh
```diff
@@ -0,0 +1,51 @@
...
+  "multinode_integration_tests"
+  "node"
+  "port_exposer"
+)
+declare -a failed_crates
+
+bump_version() {
+  pushd "$CI_DIR/../$1"
+
+  sed -i '3s/version = .*/version = "'"$version"'"/' $file
```
dnwiebe: Your regular expression matches only `version = ` on line three, with its `=` surrounded by one space on each side. This could miss `Cargo.toml` files that people change after you put this script in place.

For example, if they put a line of comment at the top, pushing the `version` line down to 4, or if they remove one or both of the spaces around the `=`, or if they change one or both of those spaces to tabs, you'll miss the file.

Since most of the `version` lines that you don't want to change won't have `version` at the beginning of the line, you can use that.

Try something like 
```
s/^\s*version\s*=.*/version = "'"$version"'"/
```
This will catch every `version` tag that begins a line. Currently at least in `node/Cargo.toml`, that's the one you want. Somebody could modify a `Cargo.toml` file to look like this
```
time = {
    version = "0.3.11", 
    features = [ "macros" ]
}
```
which would cause problems; but there are ways we might deal with that.

One way might be using the leading whitespace to disqualify it. Another might be using the comma at the end of the line to disqualify it. A third might be accepting `version` only in the first five or ten lines or so by putting `1,5` at the beginning.
```
s/^version\s*=.*/version = "'"$version"'"/ --no whitespace at beginning
s/^\s*version\s*=.*[^,]\s*$/version = "'"$version"'"/ -- no comma at end
1,5 s/^\s*version\s*=.*/version = "'"$version"'"/ -- only applies somewhere in the first five lines
```

ci/bump_version.sh
```diff
@@ -0,0 +1,51 @@
...
+      final_exit_code=1
+      failed_crates+=($1)
+  fi
+
+  popd
+}
+
+for crate in "${crates[@]}"
+do
+   :
```
dnwiebe: I don't think this is necessary.

ci/bump_version.sh
```diff
@@ -0,0 +1,51 @@
...
+  "masq_lib"
+  "multinode_integration_tests"
+  "node"
+  "port_exposer"
+)
+declare -a failed_crates
+
+bump_version() {
+  pushd "$CI_DIR/../$1"
+
```
dnwiebe, line 33: I notice that if `sed` does not find a `version` line to replace, the script will not know, and the user will be unaware that a `Cargo.toml` file remains unchanged.

You could first try `grep`ping for the same expression you're trying to change with `sed` (pipe the result to `/dev/null`, because you don't care about anything but whether there was output or there wasn't), and looking at the exit code. If `grep` found something, it'll be 0; if not, it'll be 1. If you get a 1 from `grep`, you can put the file straight into the failed list and not bother `sed`ding it.

dnwiebe (commented):
Nice job!

dnwiebe (approved):

PR #243, opened 2023-03-16
GH-679: Review 1

utkarshg6:


ci/bump_version.sh
```diff
@@ -1,51 +1,63 @@
...
 if [ $# != 1 ]
 then
      echo "Please provide version as the first argument"
      exit 1
 fi
 
-version=$1
+echo "Searching for crates..."
+
+crates=($(find . -type d -exec sh -c '[ -f "$0"/Cargo.toml ]' '{}' \; -print))
```
dnwiebe: Would like to see you try `bash` here instead of `sh` and try to stay in the same command shell.

ci/bump_version.sh
```diff
@@ -1,51 +1,63 @@
...
+
+if [[ "${#crates[@]}" == "0" ]]; then
+  echo "No crates found."
+  exit 1
+else
+  echo "Found ${#crates[@]} crate(s): ${crates[*]}"
+fi
+
+version="$1"
 CI_DIR="$( cd "$( dirname "$0" )" && pwd )"
```
dnwiebe: If you do this at the beginning, then you can use it as a reference to `cd` to where the crates are, rather than depending on the user to run this script from where the crates are.

dnwiebe (approved):
Again: Nice job!

ci/bump_version.sh
utkarshg6: Bash works too.

There is something that I tried on my terminal. If I use `sh` along with an argument, it keeps me inside the `bash` shell. Similarly if I do something similar in sh shell, that is I use `bash` along with an argument it'll also keep me in the same shell.

However, if I use `bash` or `sh` without an arguments, that switches me to a new shell.

PR #248, opened 2023-03-22
GH-580: test for data_directory_from_context with opt dir and implementation …

bertllll:
…in fn


node/src/node_configurator/mod.rs
```diff
@@ -131,13 +134,20 @@ pub fn data_directory_from_context(
...
                 wrong_local_data_dir.replace(&wrong_home_dir, &right_home_dir);
-            PathBuf::from(right_local_data_dir)
-                .join("MASQ")
-                .join(chain.rec().literal_identifier)
+            create_path_buf(chain, right_local_data_dir.to_string())
         }
     }
 }
 
+pub fn create_path_buf(
```
bertllll: The name is too vague. Have it say what actually is happening inside. The fact you need to use PathBuf isn't relevant; what would be helpful though is something such as: "add_masq_directories" or "add_chain_specific_directories" etc.

node/src/node_configurator/mod.rs
```diff
@@ -131,13 +134,20 @@ pub fn data_directory_from_context(
                 .to_string();
             let right_local_data_dir =
                 wrong_local_data_dir.replace(&wrong_home_dir, &right_home_dir);
-            PathBuf::from(right_local_data_dir)
-                .join("MASQ")
-                .join(chain.rec().literal_identifier)
+            create_path_buf(chain, right_local_data_dir.to_string())
```
bertllll: Could have been written with less code. Look at the match as it has two arms only and both a lot of similar operations; while you call up your "create_path_buf(...)" in each arm, you could call it just once outside the match, because the match would always produce just the root path.

node/src/node_configurator/mod.rs
```diff
@@ -207,6 +217,35 @@ mod tests {
         )
     }
 
+    #[test]
+    fn data_directory_from_dir_opt_creates_new_folder_for_every_blockchain_platform() {
```
bertllll: The name should be improved. Notice that there is already a test testing these things but your name suggests that this test covers all cases. That's not right, this test is only about "creating_the_right_directory_hierarchy_for_user_requested_data_directory"

node/src/node_configurator/mod.rs
```diff
@@ -207,6 +217,35 @@ mod tests {
...
+            .home_dir_result(Some(PathBuf::from("/nonexistent_home/root".to_string())))
+            .data_dir_result(Some(PathBuf::from("/nonexistent_home/root/.local/share")));
+        let real_user = RealUser::new(
+            None,
+            None,
+            None,
+        );
+        let custom_dir = Some(PathBuf::from("~/mynode".to_string())); 
+        let data_dir_opt = custom_dir;
+        let chain_name = "eth-ropsten";
```
bertllll: You might not be aware of that, so no stress, but the Ropsten testnet has been shut down. Avoid using that name, you should pick something like "eth-mainnet" or "polygon-mumbai". 

If I'm informed correctly Dan put some changes removing all references to Ropsten in his unfinished card and so you don't have to care too much about tests around the new one.

node/src/node_configurator/mod.rs
```diff
@@ -111,7 +111,10 @@ pub fn data_directory_from_context(
     chain: Chain,
 ) -> PathBuf {
     match data_directory_opt {
-        Some(data_directory) => data_directory.clone(),
+        Some(data_directory) => {
+            let data_dir = data_directory.clone().to_string_lossy().to_string();
```
bertllll: I've got a feeling that you might be able to get rid of this clone() here. My reasoning is that you usually don't need to have it owned if the (displayable) look of it is the only thing you want. Here you just convert it to a string.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
bertllll: add to gitignore pls

node/src/bootstrapper.rs
```diff
@@ -1140,14 +1141,17 @@ mod tests {
...
             ]))
             .unwrap();
 
         let init_params = init_params_arc.lock().unwrap();
+        let expected_dir = add_chain_specific_directories(Chain::PolyMumbai, &data_dir);
         assert_eq!(
             *init_params,
             vec![(
-                data_dir,
+                expected_dir,
```
bertllll: "expected_data_dir" would be slightly better :wink:

node/src/daemon/setup_reporter.rs
```diff
@@ -119,7 +119,7 @@ impl SetupReporter for SetupReporterReal {
                 .populate(self.dirs_wrapper.as_ref())
         });
         let data_directory = match all_but_configured.get("data-directory") {
-            Some(uisrv) if uisrv.status == Set => PathBuf::from(&uisrv.value),
+            Some(uisrv) if uisrv.status == Set => PathBuf::from(&uisrv.value), // TODO drive in the specific chain dir here
```
bertllll: First run all tests together to see them all are passing. Use todo!() after the arrow (instead of PathBuf::from(&uisrv.value)). If you run all tests again and no test is failing that means this arm is not properly tested. Then you would have to write a test. 

If it is tested just remove the TODO comment.

node/src/daemon/setup_reporter.rs
```diff
@@ -2413,9 +2413,18 @@ mod tests {
             "config_file_not_specified_but_exists",
         );
         {
-            let config_file_path = data_directory.join("config.toml");
-            let mut config_file = File::create(config_file_path).unwrap();
-            config_file.write_all(b"gas-price = \"10\"\n").unwrap();
+            let config_file_path = data_directory
```
bertllll: You could create a variable `data_dir_path` and then you could use it twice, in defining the config-file path and also for supply to the `create_dir_all()` fn.   

`config_file_path` would look like `config_file_path = data_dir_path.join("config.toml");

node/src/daemon/setup_reporter.rs
```diff
@@ -2413,9 +2413,18 @@ mod tests {
...
+                .join("MASQ")
+                .join("polygon-mainnet")
+                .join("config.toml");
+            let created_dir = create_dir_all(data_directory.join("MASQ").join("polygon-mainnet"));
```
bertllll: With regards to the fact that the only error we can have here are wrongly set rights for creating folders, you can just use a function to unwrap the result but maybe .expect() because you can write a special message there like something about why it failed. 

So it will replace the if statement below.

node/src/database/config_dumper.rs
```diff
@@ -217,15 +228,25 @@ mod tests {
             "config_dumper",
             "dump_config_does_not_migrate_obsolete_database",
         );
-        let conn = bring_db_0_back_to_life_and_return_connection(&data_dir.join(DATABASE_FILE));
+        let chain_specific_dir = add_chain_specific_directories(Chain::PolyMainnet, &data_dir);
+        let created_dir = create_dir_all(&chain_specific_dir); //
```
bertllll: There are two backslashes left.

node/src/database/config_dumper.rs
```diff
@@ -217,15 +228,25 @@ mod tests {
...
+        if created_dir.unwrap() != () {
```
bertllll: Repeat what was advised for a similar situation in the different file.

node/src/database/config_dumper.rs
```diff
@@ -217,15 +228,25 @@ mod tests {
...
         let schema_version_before = dao.get("schema_version").unwrap().value_opt.unwrap();
         assert_eq!(schema_version_before, "0");
         let mut holder = FakeStreamHolder::new();
         let args_vec: Vec<String> = ArgsBuilder::new()
             .param("--data-directory", data_dir.to_str().unwrap())
             .param("--real-user", "123::")
-            .param("--chain", TEST_DEFAULT_CHAIN.rec().literal_identifier)
+            .param(
+                "--chain",
+                masq_lib::constants::POLYGON_MAINNET_FULL_IDENTIFIER,
```
bertllll: `Chain::PolyMainnet.rec().literal_identifier` is more idiomatic for this project.

node/src/database/config_dumper.rs
```diff
@@ -244,16 +265,15 @@ mod tests {
...
-        .join("MASQ")
-        .join(TEST_DEFAULT_CHAIN.rec().literal_identifier);
+        );
+        let chain_specific_dir = add_chain_specific_directories(Chain::PolyMainnet, &data_dir);
         let mut holder = FakeStreamHolder::new();
         {
             let conn = DbInitializerReal::default()
                 .initialize(
-                    &data_dir,
+                    &chain_specific_dir,
```
bertllll: `chain_specific_data_dir` tells more

node/src/database/config_dumper.rs
```diff
@@ -362,16 +385,15 @@ mod tests {
         let data_dir = ensure_node_home_directory_exists(
             "config_dumper",
             "dump_config_dumps_existing_database_with_correct_password",
-        )
-        .join("MASQ")
-        .join(TEST_DEFAULT_CHAIN.rec().literal_identifier);
+        );
+        let chain_specific_dir = add_chain_specific_directories(Chain::PolyMainnet, &data_dir);
```
bertllll: Maybe `data_dir` is enough but `chain_specific_data_dir` would be better at minimum.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -434,11 +435,18 @@ mod tests {
             "node_configurator",
             "server_initializer_collected_params_can_read_parameters_from_config_file",
         );
+        let chain_specific_dir = add_chain_specific_directories(Chain::PolyMainnet, &home_dir);
         {
-            let mut config_file = File::create(home_dir.join("config.toml")).unwrap();
-            config_file
-                .write_all(b"dns-servers = \"111.111.111.111,222.222.222.222\"\n")
-                .unwrap();
+            let created_dir = create_dir_all(&chain_specific_dir);
```
bertllll: Please use the `.expect()` here too, as advised previously. It will shrink the if into a single-line call.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -707,16 +717,15 @@ mod tests {
...
         running_test();
-        let data_dir = ensure_node_home_directory_exists(
-            "node_configurator_standard",
-            "server_initializer_collected_params_senses_when_user_specifies_config_file",
-        );
-        let args = ArgsBuilder::new().param("--config-file", "booga.toml"); // nonexistent config file: should stimulate panic because user-specified
+        let home_dir = PathBuf::from("/unexisting_home/unexisting_alice");
+        let chain_specific_data_dir = add_chain_specific_directories(Chain::PolyMainnet, &home_dir);
+        let args = ArgsBuilder::new()
+            .param("--config-file", "booga.toml") // nonexistent config file: should stimulate panic because user-specified
```
bertllll: Not your fault but please change the wording `should stimulate panic` to `should return error`. The reason is that this whole test doesn't experience any panic at all so I don't like the mention in the comment. :wink:

bertllll (commented):
Take your time, no rush, address the comments and I'll be happy to see it again once done.

node/src/node_configurator/mod.rs
czarte: changed

node/src/node_configurator/mod.rs
czarte: renamed to `data_directory_from_context_creates_new_folder_for_every_blockchain_platform`

node/src/node_configurator/mod.rs
czarte: thanks :)

.vscode/launch.json
czarte: added

dnwiebe (approved):

dnwiebe (approved):

dnwiebe (approved):

PR #249, opened 2023-03-24
GH-678: Create a Message Scheduler

utkarshg6:


masq_lib/src/messages.rs
```diff
@@ -843,6 +845,12 @@ pub struct UiWalletAddressesResponse {
 }
 conversation_message!(UiWalletAddressesResponse, "walletAddresses");
 
+#[derive(Message, Clone, PartialEq, Eq)]
```
bertllll: Look at the other message in this file and notice that you don't find any Actor messages here. This file is exclusively intended for Websocket messages between UI and Node or Daemon. 

Usually the place you might have been looking for in masq_lib is considered the file `utils.rs`. However, I don't think we want to go there either. Think about what this whole library serves for, it is supposed to offer code that can occur in both, the frontend side (UI) and the backend side. 

In this case it doesn't make sense to put it there because the UI doesn't use Actors at all... and it is of no use for it then. 

Please find a nice location in the `node` crate. I'm suggesting `/sub_lib/utils.rs` as probably the best one.

masq_lib/src/messages.rs
```diff
@@ -843,6 +845,12 @@ pub struct UiWalletAddressesResponse {
...
+pub struct MessageScheduler<M: Message> {
+    pub schedule_msg: M,
```
bertllll: maybe `scheduled_msg` instead?

masq_lib/src/messages.rs
```diff
@@ -843,6 +845,12 @@ pub struct UiWalletAddressesResponse {
...
+    pub duration: Duration,
```
bertllll: Instead of repeating the word duration, the filed could say something like `msg_delay`

node/src/node_test_utils.rs
```diff
@@ -307,6 +308,7 @@ pub fn make_stream_handler_pool_subs_from_recorder(addr: &Addr<Recorder>) -> Str
         bind: recipient!(addr, PoolBindMessage),
         node_query_response: recipient!(addr, DispatcherNodeQueryResponse),
         node_from_ui_sub: recipient!(addr, NodeFromUiMessage),
+        schedule_message_sub: recipient!(addr, MessageScheduler<DispatcherNodeQueryResponse>),
```
bertllll: This is not actually just `scheduled_msg` it should rather be `scheduled_node_query_response_sub`. Keep in mind that theoretically you could have multiple scheduled messages and the recipients for them in the same actor. So mentioning the message kind is important (by the way, it would be handy to be able to read it even if we were clearly sure it's gonna be the only scheduled message ever in this place).

node/src/stream_handler_pool.rs
```diff
@@ -62,6 +62,7 @@ pub struct StreamHandlerPoolSubs {
     pub bind: Recipient<PoolBindMessage>,
     pub node_query_response: Recipient<DispatcherNodeQueryResponse>,
     pub node_from_ui_sub: Recipient<NodeFromUiMessage>,
+    pub schedule_message_sub: Recipient<MessageScheduler<DispatcherNodeQueryResponse>>,
```
bertllll: The same thing :point_up:

node/src/stream_handler_pool.rs
```diff
@@ -155,6 +157,17 @@ impl Handler<DispatcherNodeQueryResponse> for StreamHandlerPool {
     }
 }
```
bertllll, line 158: This thing below is all good. My only request is, if you haven't done yet, please create a card with tiny description about implementing this through a `derive` macro.

Also, I'd prefer to have a simple TODO here referring to the newly existing card.

node/src/stream_handler_pool.rs
```diff
@@ -533,18 +550,19 @@ impl StreamHandlerPool {
             peer_addr,
             msg.context.data.len()
         );
-        let recipient = self
+        let schedule_message_sub = self
```
bertllll: I think we need to call these variables rather `scheduled_message_sub` as you hardly can have a recipient for a wide range of messages (even if as a generic argument in the MessageScheduler).

node/src/test_utils/recorder.rs
```diff
@@ -92,6 +93,17 @@ macro_rules! recorder_message_handler {
     };
 }
 
+impl<M> Handler<MessageScheduler<M>> for Recorder
```
bertllll: I feel like this place is too prominent for this rather marginal thing. I'm asking you to move the implementation under all those `recorder_message_handler` macro calls.

masq_lib/src/messages.rs
utkarshg6: Instead, I just kept it to `delay`.

node/src/stream_handler_pool.rs
utkarshg6: I've created the card, and marked a TODO referencing it - https://github.com/MASQ-Project/Node-issues/issues/686

bertllll (approved):

PR #250, opened 2023-03-27
GH-541 review 3

Syther007:


node/src/database/db_migrations/migrations/migration_3_to_4.rs
```diff
@@ -272,6 +273,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn database_with_password_but_without_secrets_yet_still_accepted() {
+        let mig_declarator =  DBMigDeclaratorMock::default();
+        let example_encrypted_opt = Some("Password".to_string());
```
dnwiebe, line 279: "Password" is misleading. Try "random garbage" or "bit string" or something like that.

PR #251, opened 2023-03-28
GH-675: Find Alternative to route -n in Linux

Syther007:


automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -13,13 +13,16 @@ pub fn linux_find_routers(command: &dyn FindRoutersCommand) -> Result<Vec<IpAddr
...
-            line.split(' ')
-                .filter(|piece| !piece.is_empty())
-                .collect::<Vec<&str>>()
+            let ip_str: String = line
+                .chars()
+                .skip_while(|char| !char.is_numeric())
+                .take_while(|char| !char.is_whitespace())
+                .collect();
+
+            IpAddr::from_str(&ip_str).expect("Bad syntax from ip route")
```
dnwiebe: If we can't extract the default gateway, we don't want to crash the whole system. Instead, drive in some sort of `AutomapError` here.

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -120,55 +103,67 @@ Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
...
-                "Genmask",
-                "Flags",
-                "Metric",
-                "Ref",
-                "Use",
-                "Iface",
-            ]
+        assert!(
+            lines.len() > 1,
+            "Did not find more than two lines in this vector {:?}",
```
dnwiebe: Assertion no longer required

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -120,55 +103,67 @@ Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
...
+        let reg = ip_route_regex();
+        lines.iter().for_each(|line| {
+            assert!(reg.is_match(line), "Lines: {:?} line: {}", lines, line);
+        });
+    }
+
+    fn ip_route_regex() -> Regex {
+        let reg_for_ip = r"((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}";
+        Regex::new(&format!(
+            r#"{}|^{}(/\d+)?\s(dev|via)(\s\.+)\{{3,\}}"#,
```
dnwiebe: This regular expression probably isn't right. Review it.

This one is probably more correct: `r#"^{}(/\d+)?\s(dev|via)(\s.+){{3,}}"`

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -120,55 +103,67 @@ Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
...
+            reg_for_ip, reg_for_ip
+        ))
+        .unwrap()
+    }
+
+    #[test]
+    fn reg_for_ip_route_command_output() {
```
dnwiebe: If you're serious about testing the IP address, you should supply another test that makes tiny changes to these lines and expects the match to fail on each of them.

bertllll (approved):

bertllll (approved):

PR #261, opened 2023-04-05
GH-671: review two

bertllll:


node/src/accountant/mod.rs
```diff
@@ -1362,24 +1361,25 @@ mod tests {
             .recipient();
         subject.report_accounts_payable_sub_opt = Some(report_recipient);
         let subject_addr = subject.start();
-        let number_closely_fitting_under = ((u32::MAX / 2) - 1) as u64 / WEIS_OF_GWEI as u64;
-        let account_1 = make_payable_account(number_closely_fitting_under);
+        let half_of_u32_max_in_wei = u32::MAX as u64 / (2 * WEIS_OF_GWEI as u64);
```
utkarshg6: This rename was missed, please change `WEIS_OF_GWEI` to `WEIS_IN_GWEI`

node/src/daemon/setup_reporter.rs
```diff
@@ -226,18 +229,20 @@ impl SetupReporterReal {
         }
     }
 
-    fn get_err_conflicts_for_blanked_out(
+    fn prevent_err_induced_impairments_to_blanked_out(
```
utkarshg6: It's hard to grasp what the function is doing by simply reading the new name. I noticed that the guts of this function is still the same, in terms of its functionality.

Maybe I'm missing something, I don't understand why you decided to rename this function.

Also, I find this function very hard to understand, so I would appreciate if you decide to make these changes.

First of all, the constant is very confusing, it might only be meaningful to the author of the code but for other devs, it doesn't say much. So, please modify the constant `ERR_SENSITIVE_BLANKED_OUT_VALUE_PAIRS` to the following:

```rust
struct ErrorSensitiveArgsPair<'arg_names> {
    blanked_arg: &'arg_names str,
    linked_arg: &'arg_names str,
}

const ERR_SENSITIVE_ARGS_PAIRS: [ErrorSensitiveArgsPair; 1] = [ErrorSensitiveArgsPair {
    blanked_arg: "chain",
    linked_arg: "data-directory",
}];
```

Later on, it can be used to add more clarity to this function. Please take a look at this approach, I find it easier to understand.

```rust
fn fix_err_sensitive_args_pairs(
    former_setup: &SetupCluster,
    existing_setup: &SetupCluster,
) -> SetupCluster {
    ERR_SENSITIVE_ARGS_PAIRS
        .iter()
        .fold(HashMap::new(), |mut acc, pair| {
            if former_setup.contains_key(&pair.blanked_arg.to_string()) {
                if let Some(existing_value) = existing_setup.get(&pair.linked_arg.to_string()) {
                    acc.insert(pair.linked_arg.to_string(), existing_value.clone());
                }
            };
            acc
        })
}
```

Later you can use it in the code as:


```rust
let fixed_values = Self::fix_err_sensitive_args_pairs(
  &blanked_out_former_values,
  &existing_setup,
);
```

PR #265, opened 2023-04-14
GH-675-review-2

Syther007:


automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -11,20 +11,31 @@ pub fn linux_find_routers(command: &dyn FindRoutersCommand) -> Result<Vec<IpAddr
...
+                    .take_while(|char| !char.is_whitespace())
+                    .collect();
+
+                match IpAddr::from_str(&ip_str) {
+                    Ok(ip_addr) => {
+                        ip_addr_vec.push(ip_addr);
+                        Ok(ip_addr_vec)
+                    }
+                    Err(e) => Err(AutomapError::FindRouterError(format!(
+                        "Failed to parse an IP from \"ip route\": {:?}",
```
dnwiebe: It could be that some future version of `ip route` on some future version of the operating system could have an output format that has digits before the real IP address. When that happens, everybody using those versions together is going to start failing here, but it will be really hard to tell why, because we won't be able to see the text you're parsing.

Make this error message show at least the part of the line we think should be the IP address; even better would be to show the entire line, so that whoever is tasked with figuring out what the problem is will have an easier time.

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -118,18 +133,47 @@ mod tests {
...
+        let result = linux_find_routers(&find_routers_command);
+
+        assert_eq!(
+            result,
+            Err(AutomapError::ProtocolError("Booga!".to_string()))
+        )
+    }
+
+    #[test]
+    fn find_routers_returns_error_if_ip_addreses_can_not_be_parsed() {
```
dnwiebe: Spelling...

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -153,17 +197,35 @@ mod tests {
         168.63.129.16 via 10.1.0.1 dev eth0 proto dhcp src 10.1.0.84 metric 100\n\
         169.254.169.254 via 10.1.0.1 dev eth0 proto dhcp src 10.1.0.84 metric 100\n\
         172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown";
-        let reg_for_any_other_line = ip_route_regex();
+        let regex = ip_route_regex();
```
dnwiebe: At least one of the lines above should probably begin with `default via` to fully exercise the regular expression.

automap/src/comm_layer/pcp_pmp_common/linux_specific.rs
```diff
@@ -153,17 +197,35 @@ mod tests {
...
-                "Lines: {:?} line: {}",
-                lines,
-                line
-            );
+            assert!(regex.is_match(line), "Lines: {:?} line: {}", lines, line);
         });
     }
+
+    #[test]
+    fn reg_for_ip_route_bad_ip() {
```
dnwiebe: A way to make this test easier to read (and extend in the future) might be to make your `route_n_output` consist of a vector of tuples, like this:
```
let route_n_output = vec![
    ("10.1.0.0/16 dev eth0 proto kernel scope link src 10.1.0.84 metric 100", true, ""),
    ("0.1.0 dev eth0 proto dhcp scope link src 10.1.0.84 metric 100", false, "IPv4 address has only three elements"),
    [...]
];
```
Then you can apply the regex to the first item in each pair and expect the result in the second item of each pair, and if the second item was false and the regex succeeded, you'd be able to print the third item in the failure message. It'd be easier to see what you were expecting from each line, and easier to make sure that an additional line was correctly added.

Also, you could combine this test with the one above and have both successes and failures in the same test, possibly arranged to make aspects of your requirements clearer.

PR #275, opened 2023-05-05
GH-690:  Add new unprivileged arg `--min-hops`

### LINKED ISSUE #690: Add new unprivileged arg `--min-hops`
This card is a child of #567.

These are the tasks expected under this card:

1. Add `--min-hops` to the clap schema and process it into the `NeighborhoodConfig` in `NodeConfiguratorStandardUnprivileged`.
2. Add an instance field in the `Neighborhood` to hold the minimum hop count, and make the Routing Engine actually use the value when calculating routes. (See the comment inside the parent card about single-hop routes.)
3. Make sure it can be used in `masq` to establish the Daemon's in-memory setup and passed okay to a starting Node.

utkarshg6:


ci/bump_version.sh
```diff
@@ -31,7 +31,7 @@ bump_version() {
   find_pattern='^version\s*=.*[^,]\s*$'
   replace_pattern='s/'$find_pattern'/version = "'"$version"'"/'
 
-  grep -q "$find_pattern" "$file" && sed -i "$replace_pattern" "$file"
+  grep -q "$find_pattern" "$file" && sed -i "" "$replace_pattern" "$file"
```
dnwiebe, line 34: If you turn out to need different code for different operating systems, there's an example of that in `node/ci/integration_tests.sh`.

masq/src/notifications/connection_change_notification.rs
```diff
@@ -21,7 +21,7 @@ impl ConnectionChangeNotification {
             UiConnectionStage::ConnectedToNeighbor => {
                 "\nConnectedToNeighbor: Established neighborship with an external node.\n"
             }
-            UiConnectionStage::ThreeHopsRouteFound => {
+            UiConnectionStage::RouteFound => {
                 "\nThreeHopsRouteFound: You can now relay data over the network.\n"
```
dnwiebe: The enum name (with "ThreeHops") is still in this message.

masq_lib/src/multi_config.rs
```diff
@@ -88,6 +88,13 @@ impl<'a> MultiConfig<'a> {
             let message = format!("Invalid value: {}", &captures[2]);
             return ConfiguratorError::required(name, &message);
         }
+        let another_invalid_value_regex =
```
dnwiebe: If you removed the capture group you're not using in each of these cases, you might be able to combine them and use the regex as a parameter--either by making two calls to an outboard method or by putting the regexes in a vector and iterating through it.

masq_lib/src/shared_schema.rs
```diff
@@ -105,6 +105,22 @@ pub const MAPPING_PROTOCOL_HELP: &str =
...
     it will remember the protocol it used, and on its next run it will try that protocol first, unless \
     you specify a different protocol on the command line.";
+pub const MIN_HOPS_HELP: &str =
+    "The Node is a system that routes data through multiple Nodes to enhance security and privacy. \
+    However, the level of anonymity and security provided depends on the number of hops specified \
+    by the user. By default, the system allows the user to customize the number of hops within a \
+    range of 1 to 6.\n\n\
+    It's important to note that if the user selects less than 3 hops, the anonymity of their data \
+    cannot be guaranteed. Here's a breakdown of the different hop counts and their implications:\n\n\
+    1. A 1-hop route allows Exit Nodes to see your requests.\n\
```
dnwiebe: A 1-hop route means that the exit Node will know the IP address of the originating Node. Also, someone snooping traffic on the network will be able to see both the originating Node's IP and the exit Node's IP in the same packet. A 1-hop route makes MASQ the equivalent of a VPN.
A 2-hop route removes the ability to see both the originating and exit IP addresses on the same packet, but it means that the relay Node in the middle (which could be subverted by an attacker) knows both IP addresses.
A 3-hop route is the shortest route that prevents any Node in the network (even the originating Node) from knowing the IP addresses of all the Nodes in the route.

masq_lib/src/shared_schema.rs
```diff
@@ -105,6 +105,22 @@ pub const MAPPING_PROTOCOL_HELP: &str =
...
+    2. A 2-hop route makes it harder to associate your requests with your IP address, but it's \
+    not a foolproof guarantee.\n\
+    3. The minimum number of hops required to guarantee anonymity is 3.\n\
+    4. Increasing the number of hops to 4, 5, or 6 can enhance security, but it will also \
+    increase the cost and latency of the route.\n\
+    If you want to specify a minimum hops count, you can do so by entering a number after the \
+    '--min-hops' command. For example, '--min-hops 4' would require at least 4 hops. If you fail \
```
dnwiebe: `--min-hops` is a parameter, not a command. The command is `MASQNode` (or `setup`, if you're in `masq`). Maybe "by specifying the '--min-hops <n>' parameter."?

multinode_integration_tests/src/masq_cores_client.rs
```diff
@@ -18,7 +18,7 @@ impl<'a> MASQCoresClient<'a> {
     pub fn new(socket_addr: SocketAddr, cryptde: &'a dyn CryptDE) -> MASQCoresClient<'a> {
         MASQCoresClient {
             cryptde,
-            delegate: MASQNodeClient::new(socket_addr),
+            delegate: MASQNodeClient::new(socket_addr, 1000),
         }
     }
```
dnwiebe, line 24: Not this line, but down below, there are two methods that accept parameters of type `&JsonMasquerader`. That's not right.

First, `JsonMasquerader` has no internal state to speak of, which means that every `JsonMasquerader` is essentially equivalent to every other `JsonMasquerader`, and there's no reason to pass one around; you could just create a new one wherever you needed it.

But second, even though `JsonMasquerader` is the only `Masquerader` we have at the moment, we anticipate a future with dozens of implementors of the `Masquerader` trait, so that we can masquerade our data as `ftp` transactions or `Call of Duty` streams or whatever. This parameter should be typed with `&dyn Masquerader`, not `&JsonMasquerader`.

multinode_integration_tests/src/masq_cores_client.rs
```diff
@@ -18,7 +18,7 @@ impl<'a> MASQCoresClient<'a> {
     pub fn new(socket_addr: SocketAddr, cryptde: &'a dyn CryptDE) -> MASQCoresClient<'a> {
         MASQCoresClient {
             cryptde,
-            delegate: MASQNodeClient::new(socket_addr),
+            delegate: MASQNodeClient::new(socket_addr, 1000),
```
dnwiebe: There's a constant (`STANDARD_CLIENT_TIMEOUT_MILLIS`) that could be used here instead of an immediate value.

multinode_integration_tests/src/masq_real_node.rs
```diff
@@ -1289,6 +1305,7 @@ mod tests {
 
     #[test]
     fn node_startup_config_builder_settings() {
+        let min_hops_count = Hops::SixHops;
```
dnwiebe, line 1307: If you `use`d `SixHops` the way you did `TwoHops` above, you'd be able to just say `SixHops` here instead of `Hops::SixHops`.

Or you could go the other way, remove the `use` for `TwoHops`, and qualify both of them with `Hops::` if you think it's clearer.

multinode_integration_tests/src/masq_real_node.rs
```diff
@@ -1415,6 +1435,7 @@ mod tests {
             .build();
 
         assert_eq!(result.neighborhood_mode, neighborhood_mode);
+        assert_eq!(result.min_hops_count, Hops::TwoHops);
```
dnwiebe, line 1437: With the `use` pattern at the time of this writing, you shouldn't need `Hops::TwoHops` here; just `TwoHops` should be enough.

dnwiebe (commented):
Resume at gossip_acceptor.rs:1281 green

node/src/neighborhood/mod.rs
```diff
@@ -435,23 +433,23 @@ enum RouteDirection {
 impl Neighborhood {
     pub fn new(cryptde: &'static dyn CryptDE, config: &BootstrapperConfig) -> Self {
         let neighborhood_config = &config.neighborhood_config;
-        if neighborhood_config.mode.is_zero_hop()
-            && !neighborhood_config.mode.neighbor_configs().is_empty()
-        {
+        let min_hops_count = neighborhood_config.min_hops_count;
+        let neighborhood_mode = &neighborhood_config.mode;
+        let mode: NeighborhoodModeLight = neighborhood_mode.into();
+        let neighborhood_configs = neighborhood_mode.neighbor_configs();
```
dnwiebe: Maybe `neighbor_configs` instead of `neighborhood_configs`

node/src/neighborhood/mod.rs
```diff
@@ -2755,6 +2785,7 @@ mod tests {
             "route_query_succeeds_when_asked_for_one_hop_round_trip_route_without_consuming_wallet",
         );
         let mut subject = make_standard_subject();
+        subject.min_hops_count = Hops::OneHop;
```
dnwiebe, line 2786: This test could be named better. Let's see:

`route_query_works_when_node_is_set_for_one_hop_and_no_consuming_wallet`

Feel free to use a better one if you come up with it.

node/src/neighborhood/mod.rs
```diff
@@ -2836,6 +2867,7 @@ mod tests {
     ) {
         let system = System::new("route_query_responds_with_none_when_asked_for_one_hop_round_trip_route_without_consuming_wallet_when_back_route_needs_two_hops");
         let mut subject = make_standard_subject();
+        subject.min_hops_count = Hops::OneHop;
```
dnwiebe, line 2868: This is a bad test. Not your fault, but it claims to be stimulating a failed query because of a hop-count problem when the query is actually failing because there are no full neighborships.

Recommendation: briefly change `.add_arbitrary_half_neighbor` below to `.add_arbitrary_full_neighbor`. Run the test. If it passes, leave it that way. If it fails, put it back the way it was and make a card about cleaning it up.

ci/bump_version.sh
utkarshg6: The code modifications are inside #281

masq_lib/src/multi_config.rs
utkarshg6: The first regex has `name` and `value` at positions `1` and `2` respectively. The second regex has them at places `2` and `1` respectively.

So, I need to introduce `index_of_name` and `index_of_value` inside the outboard function and it looks something like this:

```rust
    fn check_for_inavlid_value_err(
        error_msg: &str,
        pattern: &str,
        index_of_name: usize,
        index_of_value: usize,
    ) -> Option<ConfiguratorError> {
        let regex = Regex::new(pattern).expect("Bad regex");

        match regex.captures(error_msg) {
            Some(captures) => {
                let name = &captures[index_of_name];
                let message = format!("Invalid value: {}", &captures[index_of_value]);

                Some(ConfiguratorError::required(name, &message))
            }
            None => None,
        }
    }

    pub fn make_configurator_error(e: clap::Error) -> ConfiguratorError {
        let invalid_value_patterns = vec![
            ("Invalid value for.*'--(.*?) <.*? (.*)$", 1, 2),
            ("error: (.*) isn't a valid value for '--(.*?) <.*>'", 2, 1),
        ];
        for (pattern, index_of_name, index_of_value) in invalid_value_patterns {
            if let Some(invalid_value_err) = MultiConfig::check_for_inavlid_value_err(
                &e.message,
                pattern,
                index_of_name,
                index_of_value,
            ) {
                return invalid_value_err;
            };
        }
        
        // Remaining code of this function
    }
```

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -213,8 +213,15 @@ pub fn make_neighborhood_config<T: UnprivilegedParseArgsConfiguration + ?Sized>(
                 .get_past_neighbors(persistent_config, unprivileged_config)?,
         }
     };
+
+    let min_hops_count = value_m!(multi_config, "min-hops", Hops)
+        .expect("Clap schema didn't specified the default value.");
```
dnwiebe: Specify, not specified

node/src/sub_lib/neighborhood.rs
```diff
@@ -1218,4 +1252,21 @@ mod tests {
             .unwrap();
         assert_eq!(subject.ask_about_gossip_interval, Duration::from_secs(10));
     }
+
+    #[test]
+    fn min_hops_count_can_be_converted_from_str() {
+        let result = Hops::from_str("6").unwrap();
```
dnwiebe: We need all six options here, not just one.

multinode_integration_tests/src/masq_real_node.rs
utkarshg6: I preferred to remove the `use` for `TwoHops` and use `Hops::` instead.

node/src/neighborhood/mod.rs
utkarshg6: I've created this card - https://github.com/MASQ-Project/MASQ-Node-issues/issues/704

Although, I didn't mentioned what should be a more accurate routing behaviour. I need your help in postulating some ideas for that.

node/src/neighborhood/mod.rs
utkarshg6: I've created this card - https://github.com/MASQ-Project/MASQ-Node-issues/issues/704

node/src/neighborhood/mod.rs
```diff
@@ -3978,29 +4033,39 @@ mod tests {
...
-                    stage: UiConnectionStage::ThreeHopsRouteFound
+                    stage: UiConnectionStage::RouteFound
                 }
                 .tmb(0),
             }
         );
         TestLogHandler::new().exists_log_containing(&format!(
-            "DEBUG: {}: The connectivity check has found a 3 hops route.",
-            test_name
+            "DEBUG: {}: The connectivity check has found a {}-hop(s) route.",
```
dnwiebe: Regardless of the value of the number, "-hop" is perfectly valid English. 0-hop, 1-hop, 100000-hop, all of that is fine. So you could get rid of the `(s)` here, if you wanted.

node/src/neighborhood/mod.rs
```diff
@@ -5214,6 +5310,54 @@ mod tests {
...
+            hostname_opt: None,
+        });
+
+        let assert_hops = |cryptdes: Vec<CryptDENull>, route: &[CryptData]| {
+            assert_eq!(cryptdes.len(), route.len());
+            for (cryptde, data) in cryptdes.into_iter().zip(route) {
+                decodex::<LiveHop>(&cryptde, data).unwrap();
+            }
+        };
+        let mut route = result.clone().unwrap().route.hops;
```
dnwiebe: "Route" in this codebase has a meaning already; there's a widely-used type for it. Perhaps a better name for this variable would be "hops," since it's only a series of hops, not a complete Route.

Also, Bert pointed out that the `.clone()` probably isn't necessary.

node/src/neighborhood/mod.rs
```diff
@@ -5214,6 +5310,54 @@ mod tests {
...
+        let route_length = route.len();
+        let _accounting = route.pop();
```
dnwiebe: Bert would like a comment telling why you're throwing away the last value. He might be satisfied with a better name for `_accounting`, if the name is really, really good.

dnwiebe (commented):
Pass 1 complete

node/src/neighborhood/mod.rs
utkarshg6: Since `hops` was already used by another variable. I renamed this variable to `route_hops`. Bert was right, the `.clone()` was unnecessary.

node/src/neighborhood/mod.rs
utkarshg6: There is an old test named `make_round_trip_succeeds_when_it_finds_non_next_door_neighbor_exit_node`. It was written during the substratum times. It uses the variable name `_accounting` to refer to the last element of the vector which is intended for the Originating Node.

So, I used this name to keep it consistent with this old test.

I added new comments to explain everything a little better.

dnwiebe (approved):

bertllll (approved):

PR #281, opened 2023-05-10
GH-700:  Improve Version Bump Script

### LINKED ISSUE #700: Improve Version Bump Script
The Version Bump Script has some issues that need to be addressed. 

1. The script uses a command called `cargo generate-lockfile` that updates dependencies to their latest version while generating lock files. However, during version bumps, we only want to update the version number and not the dependencies. A better solution to this is to use the `cargo update --workspace` command, which only updates versions within the workspace without connecting to the internet.

2. The `sed` command behaves differently on Linux and MacOS. On Linux, the command works fine, but on MacOS, an empty string prefix is needed to prevent the creation of backup files while editing the file. However, this prefix causes the command to fail on Linux. To solve this problem, we can either conditionally run the command based on the operating system or use a generalised sed command that works on both systems.

3. The script misses changing the version number in certain places due to interdependencies among crates in the workspace. Specifically, `masq_lib` is a dependency of both `automap` and `dns_utility`, causing their lock files to also require updating after the version number is changed. This 2-level interdependency issue needs a solution, either by addressing the specific case or by implementing a more robust n-level interdependency solution. To solve it for n-levels, we can update the lock files in a loop, create backups during each iteration, and break the loop once the new and backup lock files are identical.

Also, add the usage of this script inside the contributing guidelines of the Markdown.

utkarshg6:


CONTRIBUTING.md
```diff
@@ -261,3 +261,36 @@ discrepancies they find.
...
+
+```bash
+./bump_version.sh 6.9.1
+```
+
+Note: The script only modifies the version numbers inside the cargo files, and does not connect to the internet or 
+modify the project's dependencies in any way.
+
+The script is easy to use and validates the command-line argument to ensure that it is a valid version number. It also 
+reports any errors that occur during the modification process.
```
bertllll, line 296: Not sure if these two last lines need to be here. Technically, they just repeat what was already said before. 
I also cannot say whether it is still meant as continuation of the `Note:` or stands separately. 

If I should chose a piece of information that might bring some value in I'd call that the second sentence, maybe. Still partially debated in the preceding text.

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
 #!/bin/bash
 
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+BLUE='\033[0;34m'
+MAGENTA='\033[0;35m'
+CYAN='\033[0;36m'
+GRAY='\033[0;37m'
```
bertllll: Didn't see any of these used: 
YELLOW, BLUE, MAGENTA, GRAY.

Probably just look distracting then.

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
...
+
+# Reset
```
bertllll: I'd prefer you to change the comment to be more comprehensive, e.g. `Reset output color to the default`.

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
...
+NC='\033[0m'
+
 if [ $# != 1 ]
 then
      echo "Please provide version as the first argument"
```
bertllll, line 16: Maybe you want to use your colors also for this message?

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
...
      exit 1
 fi
 
 version="$1"
+regex="^[0-9]+\.[0-9]+\.[0-9]+$"
```
bertllll: What if somebody supplies numbers with leading zeros or even multiple zeros? For example `000.004.4`. Do we want to have this the new version? Can you find a way how to treat it with an error or how to delete those zeros?

For a single number (so you'd have to repeat the pattern three times) maybe: `0|[1-9][0-9]+?`

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
...
+
+if [[ $version =~ $regex ]]; then
+  echo -e "${CYAN}Changing to the version number $version${NC}"
+else
+  echo -e "${RED}Error: Invalid version number"
```
bertllll: Could that error message contain an example of the proper format? For the poor guy's inspiration :)

bertllll: I also noticed that this line does not end with ${NC}. Will it leave the terminal with the red coloring when this program terminates?

ci/bump_version.sh
```diff
@@ -1,12 +1,32 @@
...
 then
      echo "Please provide version as the first argument"
      exit 1
 fi
 
 version="$1"
+regex="^[0-9]+\.[0-9]+\.[0-9]+$"
+
+if [[ $version =~ $regex ]]; then
+  echo -e "${CYAN}Changing to the version number $version${NC}"
```
bertllll, line 24: What does the `-e` mean for this command? I couldn't find it. I have no idea what it really is but I thought maybe something with an error? Maybe not... But you should know when you use it. 

If it is something about errors this message doesn't seem to be intended as an error...

bertllll (commented):
Continue at line 29 in `bump_version.sh`

ci/bump_version.sh
```diff
@@ -15,50 +35,94 @@ echo "Searching for crates..."
...
+  local crate=$1
+  local file="Cargo.lock"
 
-  cargo generate-lockfile
-  exit_code="$?"
-  if [[ "$exit_code" != "0" ]]; then
+  if cargo update --workspace; then
+    echo -e "${CYAN} Successfully updated $file for ${crate#./}${NC}"
+  else
+    echo -e "${RED} Error: Failed to update $file for ${crate#./}${NC}"
```
bertllll, line 102: I couldn't interpret this part ${crate#./}, particularly `#./`. Could you explain me what those three characters are supposed to do?

ci/bump_version.sh
```diff
@@ -15,50 +35,94 @@ echo "Searching for crates..."
...
-  [[ "${#grep_failures[@]}" != "0" ]] && echo "Failed to find 'version' for ${#grep_failures[@]} crate(s): ${grep_failures[*]}"
-  [[ "${#lockfile_failures[@]}" != "0" ]] && echo "Failed to generate lockfile for ${#lockfile_failures[@]} crate(s): ${lockfile_failures[*]}"
+  [[ "${#grep_failures[@]}" != "0" ]] && echo -e "${RED} Error: Failed to find 'version' for ${#grep_failures[@]} crate(s): ${grep_failures[*]}"
+  [[ "${#lockfile_failures[@]}" != "0" ]] && echo -e "${RED} Error: Failed to generate lockfile for ${#lockfile_failures[@]} crate(s): ${lockfile_failures[*]}"
 else
-  echo "The version number has been changed to $version."
+  echo -e "${GREEN} The version number has been updated to $version."
 fi
 
 exit $final_exit_code
```
bertllll, line 131: This is a very nice script. :clap:

CONTRIBUTING.md
utkarshg6: Yes, there is repetition. I added it to end the docs in conclusive manner along with some motivation. I'm using a line break to separate it from the Note. I hope it's acceptable. :pray:

ci/bump_version.sh
utkarshg6: I'm keeping it so that someone else can use this in their script in case they want to use these colors.

ci/bump_version.sh
utkarshg6: Done 👍

ci/bump_version.sh
utkarshg6: I think this one is better off without any colours.

ci/bump_version.sh
utkarshg6: The program ends here, therefore resetting won't make any difference.

ci/bump_version.sh
utkarshg6: This `-e` is required to print the text using different colours.

ci/bump_version.sh
utkarshg6: It removes the prefix `./` from a string. For example, `./automap` will be printed as `automap` by using this formatting.

ci/bump_version.sh
utkarshg6: Thank you so much! 🙏

ci/bump_version.sh
utkarshg6: Well, we can also use this regex

```
^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```

This is a widely accepted regex for Semantic Versioning followed by the Rust community.

I used this simple regex because it's easy to understand, and this is good enough for the version numbers we use at Masq, but it is vulnerable to the cases you mentioned with the trailing zeros.

Let me know which regex you like more? :)

ci/bump_version.sh
utkarshg6: Done 👍

bertllll: Wouldn't this argument apply to the other case too then? 

Also, are you sure that the print colors are valid only until the end of the program?

ci/bump_version.sh
bertllll: We agreed that those unused colors could stay but be commented out.

ci/bump_version.sh
bertllll: Please use this regex as requires more precise syntax.

^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)$

ci/bump_version.sh
bertllll: True again

bertllll (approved):

PR #289, opened 2023-05-19
GH-687-review2-part2

Syther007:


node/tests/utils.rs
```diff
@@ -171,66 +171,75 @@ impl MASQNode {
...
-                                capture.get(idx).map(|particular_group_match| {
-                                    particular_group_match.as_str().to_string()
-                                })
-                            })
-                            .collect::<Vec<String>>()
-                    })
-                })
-                .collect::<Vec<Vec<String>>>();
-            Some(structured_captures)
+        let log_output_processor = |log_output: &str, regex: &Regex| {
```
dnwiebe, line 197: Bert wasn't happy with this name. He thought maybe "regex" should go in it somewhere.

node/tests/utils.rs
```diff
@@ -171,66 +171,75 @@ impl MASQNode {
...
+        regex: &Regex,
+        required_repetition: usize,
+    ) -> Option<Vec<Vec<String>>> {
+        let captures = regex.captures_iter(log_output).collect::<Vec<Captures>>();
+        if captures.len() < required_repetition {
+            return None;
+        }
+        let structured_captures = (0..captures.len())
+            .flat_map(|idx| {
+                captures.get(idx).map(|capture| {
```
dnwiebe: Suggestion: change `flat_map` to `map` and `captures.get(idx).map()` to `captures.get(idx).unwrap()`

Also, `capture` would be clearer as `groups`.

node/tests/utils.rs
```diff
@@ -171,66 +171,75 @@ impl MASQNode {
...
+                    (0..capture.len())
+                        .flat_map(|idx| {
+                            capture.get(idx).map(|particular_group_match| {
```
dnwiebe: Play the same trick here. If Regex tells you there are N captures and then doesn't want to give you all of them, that should be a panic.

node/tests/utils.rs
```diff
@@ -171,66 +171,75 @@ impl MASQNode {
...
+    }
+
+    fn wait_for_log_at_directory<T, F>(
         pattern: &str,
         path_to_logfile: &Path,
-        do_with_log_output: &dyn Fn(&String, &Regex) -> Option<T>,
+        log_output_processor: F,
         limit_ms: Option<u64>,
-    ) -> Option<T> {
+    ) -> Option<T>
```
dnwiebe: Could be `T` instead of `Option<T>`

node/tests/utils.rs
```diff
@@ -415,7 +424,17 @@ impl MASQNode {
         config: Option<CommandConfig>,
         remove_database: bool,
     ) -> process::Command {
-        let mut args = Self::minimum_args();
+        let mut args = vec![];
```
dnwiebe: Since `args` now starts out empty, we don't have to make it mutable and then extend it.

dnwiebe (approved):

PR #292, opened 2023-05-24
GH-706: cleanup at existing response producing messages

bertllll:


node/src/sub_lib/neighborhood.rs
```diff
@@ -381,6 +381,7 @@ pub struct NeighborhoodSubs {
     pub bind: Recipient<BindMessage>,
     pub start: Recipient<StartMessage>,
     pub new_public_ip: Recipient<NewPublicIp>,
+    #[cfg(test)]
```
dnwiebe: If we could use `AssertionMessage` (or whatever it's called) to perform the functions now handled by `NodeQueryMessage`, we could get rid of `NodeQueryMessage` and all its special-casing entirely.

dnwiebe (approved):

PR #294, opened 2023-05-26
GH-691: Integrate `--min-hops` with `PersistentConfiguration`

### LINKED ISSUE #691: Integrate `--min-hops` with `PersistentConfiguration`
This card is a child of #567 and is blocked by #690.

Make a place for it in `PersistentConfiguration` and modify `SetupReporter` to load it from the database if it isn't specified anywhere with higher priority. Remember to add a database migration for the new configuration property and bump the `CURRENT_SCHEMA_VERSION` constant.

utkarshg6:


node/src/daemon/setup_reporter.rs
```diff
@@ -1283,7 +1283,7 @@ mod tests {
             ("ip", "4.3.2.1", Set),
             ("log-level", "warn", Default),
             ("mapping-protocol", "", Blank),
-            ("min-hops", "3", Configured), // TODO: GH-698: This should be changed to Default after this card is played
+            ("min-hops", "", Blank),
```
bertllll: Really? I 'd expect `("min-hops","3", Default)`, but you might have talked this over with Dan and this is it. 

Is the reason, that we weren't satisfied by the current state of the setup_reporter algorithm? I remember we wondered a lot about how it works and how it should work ideally. I don't know why this shouldn't become a default. I think it should, because it sounds silly that the min hop count would stay blank, this value must be always known...

node/src/database/db_initializer.rs
```diff
@@ -1032,6 +1039,7 @@ mod tests {
             false,
         );
         verify(&mut config_vec, "mapping_protocol", None, false);
+        verify(&mut config_vec, "min_hops_count", Some("3"), false);
```
bertllll: This is okay. 

If you want to please me, I'd like to see this test (`existing_database_with_correct_version_is_accepted_without_changes`) be separated into two different tests. You might remember it was said that it combines two different concerns. One, it makes sure that when you create a new database and modify and close it and open again, the values put in the last time will still be there.
Dan said something like it would be even nicer to test that if you put records in all those financials tables (payable, receivable, pending_payable, banned) the test will be much better. 

You would then have also the second test testing just the config table alone for right values after it was created the first time. Don't forget to give a well illustrative name because we don't want to repeat that we had to look hard for this test when we wanted to add a new parameter into the config table. 

The new name also can reflect that this tests the creation not a migration.

node/src/db_config/persistent_configuration.rs
```diff
@@ -335,6 +337,20 @@ impl PersistentConfiguration for PersistentConfigurationReal {
...
+        let result = self.get("min_hops_count")?.map(|val| Hops::from_str(&val));
+        match result {
+            None => Err(PersistentConfigError::NotPresent),
+            Some(Ok(hops)) => Ok(hops),
+            Some(Err(msg)) => Err(PersistentConfigError::DatabaseError(msg)),
+        }
+    }
+
+    fn set_min_hops_count(&mut self, value: Hops) -> Result<(), PersistentConfigError> {
+        let value = format!("{}", value as usize);
```
bertllll: Could be `(value as usize).to_string()`. But I don't care so much.

node/src/db_config/persistent_configuration.rs
```diff
@@ -1685,6 +1701,52 @@ mod tests {
...
+    fn set_min_hops_count_to_some() {
+        let set_params_arc = Arc::new(Mutex::new(vec![]));
+        let config_dao = ConfigDaoMock::new()
+            .set_params(&set_params_arc)
+            .set_result(Ok(()));
+        let mut subject = PersistentConfigurationReal::new(Box::new(config_dao));
+
+        let result = subject.set_min_hops_count(Hops::TwoHops);
+
+        assert!(result.is_ok());
```
bertllll: I'd prefer `assert_eq!(result, Ok(()))` because yours could potentially contain something, not shown to us.

node/src/db_config/persistent_configuration.rs
```diff
@@ -335,6 +337,20 @@ impl PersistentConfiguration for PersistentConfigurationReal {
             .set("mapping_protocol", value.map(|v| v.to_string()))?)
     }
 
+    fn min_hops_count(&self) -> Result<Hops, PersistentConfigError> {
+        let result = self.get("min_hops_count")?.map(|val| Hops::from_str(&val));
```
bertllll: I think this is misunderstanding from Dan. 

Methods like this one, where the value should always be found, are treated in the standard way by using 
Self::missing_value_panic("min_hops_count"). If you don't do that you break the consistency.

I believe what Dan meant and was mistaken by is what should happen if this exact row does not exist in the config table at all. However, my answer is that the question mark at this line 341 does this job for us. It will raise an error if it happens. 

So that situation might not even be costumed to be tested, looking around here, even though it's debatable if it should be. 

What your third test, affected by what I'm getting at here, should be like is a test that tests a panic with the standard message from inside the function I mentioned in this comment a bit higher.

node/src/db_config/persistent_configuration.rs
```diff
@@ -1685,6 +1701,52 @@ mod tests {
...
+        assert!(result.is_ok());
+        let set_params = set_params_arc.lock().unwrap();
+        assert_eq!(
+            *set_params,
+            vec![("min_hops_count".to_string(), Some("2".to_string()))]
+        );
+    }
+
+    #[test]
+    fn throws_err_when_min_hops_count_value_is_missing() {
```
bertllll: I hope you read my comment at the implementation.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -251,11 +251,16 @@ fn configure_database(
         if let Err(pce) = persistent_config.set_clandestine_port(port) {
             return Err(pce.into_configurator_error("clandestine-port"));
         }
-    }
+    } // TODO: Shouldn't we be testing for the else case when no clandestine port is returned
```
bertllll: Of course we should. 

However, putting an expect up here might be the smoothest solution. You can think about it. My guess is that the missing port would mean some of our procedures that are supposed to run before this place don't work right.  

At any rate, I'd like to encourage you to fix it instead of adding more TODOs in. I wouldn't if the fix would require a lot of work.

node/src/sub_lib/neighborhood.rs
```diff
@@ -396,6 +396,19 @@ impl FromStr for Hops {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match *self {
+            Hops::OneHop => write!(f, "1"),
+            Hops::TwoHops => write!(f, "2"),
+            Hops::ThreeHops => write!(f, "3"),
+            Hops::FourHops => write!(f, "4"),
+            Hops::FiveHops => write!(f, "5"),
+            Hops::SixHops => write!(f, "6"),
+        }
+    }
```
bertllll, line 402: Please rewrite it so that the whole match returns this type `&str` (desugared as &static' str) and this value is assigned to a variable which you put in a single call like `write!(f, min_hops)` or if unsupported then `write!(f, "{}", min_hops)` but the former should work too.  

I find such a solution more decent.

node/src/test_utils/mod.rs
```diff
@@ -555,6 +555,7 @@ pub mod unshared_test_utils {
         NLSpawnHandleHolder, NLSpawnHandleHolderReal, NotifyHandle, NotifyLaterHandle,
     };
     use crate::test_utils::database_utils::bring_db_0_back_to_life_and_return_connection;
+    use crate::test_utils::neighborhood_test_utils::MIN_HOPS_COUNT_FOR_TEST;
```
bertllll: Is there a good reason why there must be two constants, one for each, either the production or test tree?

I think nothing would backfire us if you were just using the official default.

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -44,6 +44,10 @@ pub struct PersistentConfigurationMock {
     mapping_protocol_results: RefCell<Vec<Result<Option<AutomapProtocol>, PersistentConfigError>>>,
     set_mapping_protocol_params: Arc<Mutex<Vec<Option<AutomapProtocol>>>>,
     set_mapping_protocol_results: RefCell<Vec<Result<(), PersistentConfigError>>>,
+    min_hops_count_params: Arc<Mutex<Vec<()>>>,
```
bertllll: We reach the option with asserting on empty params only really rarely. In special situations. I don't think you need it here. Just don't create it and leave it out. 

You can easily prove that the function was called. You just need to provide such a result which is different from the default and if you can see the effects of that at the end you are warm and dry.

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -172,6 +176,16 @@ impl PersistentConfiguration for PersistentConfigurationMock {
         self.set_mapping_protocol_results.borrow_mut().remove(0)
     }
 
+    fn min_hops_count(&self) -> Result<Hops, PersistentConfigError> {
+        self.min_hops_count_params.lock().unwrap().push(());
```
bertllll: unimportant line

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -172,6 +176,16 @@ impl PersistentConfiguration for PersistentConfigurationMock {
...
+        self.min_hops_count_results.borrow_mut().remove(0)
+    }
+
+    fn set_min_hops_count(&mut self, value: Hops) -> Result<(), PersistentConfigError> {
+        self.set_min_hops_count_params.lock().unwrap().push(value);
```
bertllll: unimportant line

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -374,6 +388,32 @@ impl PersistentConfigurationMock {
         self
     }
 
+    pub fn min_hops_count_params(mut self, params: &Arc<Mutex<Vec<()>>>) -> Self {
```
bertllll: Shouldn't be here.

node/src/actor_system_factory.rs
```diff
@@ -1089,8 +1089,9 @@ mod tests {
             payment_thresholds_opt: Some(PaymentThresholds::default()),
             when_pending_too_long_sec: DEFAULT_PENDING_TOO_LONG_SEC,
         };
-        let persistent_config =
-            PersistentConfigurationMock::default().chain_name_result("eth-ropsten".to_string());
+        let persistent_config = PersistentConfigurationMock::default()
+            .chain_name_result("eth-ropsten".to_string())
```
bertllll, line 1093: You'll do better if you chose a different chain. Otherwise you can add work for other people who will finally decide to tear all ropsten code out.

node/src/actor_system_factory.rs
```diff
@@ -1320,7 +1322,7 @@ mod tests {
         let _ = subject.prepare_initial_messages(
             make_cryptde_pair(),
             config.clone(),
-            Box::new(PersistentConfigurationMock::new()),
+            Box::new(PersistentConfigurationMock::new().set_min_hops_count_result(Ok(()))),
```
bertllll: It really surprises me that tests at this level (We are in actor_system_factory.rs!!) ask about results for setting the min hops count. This seems wrong. Investigate whether, why and where...

My understanding is that the `set_min_hops_count()` method ought to be used only in the function where various values are set from the BootstrapperConfig into the database. That should happen when the config is complete and I'm pretty sure entering the actor system factory domain means we've done with the setup and the config is complete. 

In other words, that method should definitely be called outside of this space. When you are already here you should be able to read everything you need to know from the bootstrapper_config because that's exactly its purpose.

Therefore I hope this is just a mistake and you can remove this results (in other tests too) and they will keep running fine. If that's not true, you will have to reconsider whether the code is okay.

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct Migrate_7_to_8;
+
+impl DatabaseMigration for Migrate_7_to_8 {
+    fn migrate<'a>(
+        &self,
+        mig_declaration_utilities: Box<dyn DBMigDeclarator + 'a>,
+    ) -> rusqlite::Result<()> {
+        let statement =
+            "INSERT INTO config (name, value, encrypted) VALUES ('min_hops_count', '3', 0)";
+        mig_declaration_utilities.execute_upon_transaction(&[&statement])
```
bertllll, line 18: Maybe the second `&` in the nested slices can be omitted? Thinking that statement is also a slice `&str`.

Later when I've seen the test I had to go back here and see it again. My late impression is that the number 3 should actually not be exactly hard-coded but it should be the DEFAULT_MIN_HOPS_COUNT. So my comment from the beginning might become irrelevant because you will need `format!()`.

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+        );
+        let db_path = dir_path.join(DATABASE_FILE);
+        let _ = bring_db_0_back_to_life_and_return_connection(&db_path);
+        let subject = DbInitializerReal::default();
+        {
+            subject
+                .initialize_to_version(
+                    &dir_path,
+                    start_at,
+                    DbInitializationConfig::create_or_migrate(make_external_data()),
```
bertllll: Nope, we'd normally use `DbInitializationConfig::test_default()`

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+                )
+                .unwrap();
+        }
+
+        let result = subject.initialize_to_version(
+            &dir_path,
+            start_at + 1,
+            DbInitializationConfig::create_or_migrate(ExternalData::new(
```
bertllll: Using just `make_external_data()` would be justifiable because there is no link between your new parameter and the data that has to come from outside. Your implementation draws too much attention to these values as though they were important. I'd prefer to see the change.

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+                DEFAULT_CHAIN,
+                NeighborhoodModeLight::ConsumeOnly,
+                None,
+            )),
+        );
+
+        let connection = result.unwrap();
+        let (bchs_value, bchs_encrypted) =
```
bertllll: These variable names seems like something what you copied from a different test but here they lack the sense. Please chose some better ones. 

A key could be that those are probably shortcuts like for blockchain_service_value and blockchain_service_encrypted. :man_shrugging:

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+            retrieve_config_row(connection.as_ref(), "min_hops_count");
+        assert_eq!(bchs_value, Some("3".to_string()));
+        assert_eq!(bchs_encrypted, false);
+        let (cs_value, cs_encrypted) = retrieve_config_row(connection.as_ref(), "schema_version");
```
bertllll: Also here, I'd expect something like `schv_value`, maybe, or something even better.

node/src/database/db_migrations/migrations/migration_7_to_8.rs
```diff
@@ -0,0 +1,86 @@
...
+                DEFAULT_CHAIN,
+                NeighborhoodModeLight::ConsumeOnly,
+                None,
+            )),
+        );
+
+        let connection = result.unwrap();
+        let (bchs_value, bchs_encrypted) =
+            retrieve_config_row(connection.as_ref(), "min_hops_count");
+        assert_eq!(bchs_value, Some("3".to_string()));
```
bertllll: After you change the insert statement at the top of this file you will want to use the same constant here.

node/src/neighborhood/mod.rs
```diff
@@ -546,6 +547,21 @@ impl Neighborhood {
         }
     }
 
+    fn validate_database_min_hops_count(&self) {
```
bertllll: Hmmm. Is this really necessary? 

Think with me. How was the value stored in the actor's body set up? It happened in the actor system factory. Do you know what those actors use during their setup? The vast majority (if not all) of parameters that they need comes from the BootstrapperConfig whose existence is conditioned by this very moment. Ideally, the actors should just be fed by the relevant fields (some have their own whole configs there) in that aggregated config.

That means the Neighborhood when we're preparing it to start it up later should get the min_hops_count out of the BootstrapperConfig. (If it doesn't you need to go and change things so that it will. This pattern is pretty important.) That implies that this value here at the line 555 should be exactly the value that was finding itself in the BootstrapperConfig when it was finalized. 

Then ask yourself, what was that what assisted when the desired mins_hop_count was newly set to the database. Yes, again, the BootstrapperConfig on its own. Now you might see that both of those places that you test in the function here for equality got their values from exactly the same source. Then tell me why do you assert whether they are the same.

I don't see why 
a) it might have changed
b) why it is such a dangerous parameter that it stimulated you to write this code

I think we can stay calm even without this function so please delete it.

bertllll (commented):
Please write a comment if you want to discuss the issue or disagree with me. Thanks

node/src/daemon/setup_reporter.rs
utkarshg6: It will become `Default` if it is provided by the Clap. Now, we're not using Clap to provide the default value, instead we're using Multi Config and Database for identifying the value.

Priority Order is `Multi Config > Database`.

For a newly created Databases, the default count of 3 will be picked.

Due to that reason, it should stay `Blank`.

node/src/database/db_initializer.rs
utkarshg6: Last week I created this Card for the same. I did that because it could be a good first issue.

https://github.com/MASQ-Project/MASQ-Node-issues/issues/715

Feel free to edit it.

node/src/db_config/persistent_configuration.rs
utkarshg6: Done 👍

node/src/db_config/persistent_configuration.rs
utkarshg6: Done 👍

node/src/sub_lib/neighborhood.rs
utkarshg6: Done 👍

node/src/test_utils/mod.rs
utkarshg6: It serves a purpose. If you'll find usages for `DEFAULT_MIN_HOPS_COUNT`, you'll see it's usages inside production code, and if you search for usages for this constant, you'll see it's usages only inside the test code.

So, you can enjoy it's benefits inside the IDE :)

node/src/test_utils/persistent_configuration_mock.rs
utkarshg6: I prefer it the other way, as it makes the tests easier to read. When a reader says two values in the params, and we're asserting on it, a user can be confident that the function ran exactly 2 times and we expect this behaviour.

node/src/test_utils/persistent_configuration_mock.rs
utkarshg6: Same as the above comment

node/src/test_utils/persistent_configuration_mock.rs
utkarshg6: Same as the above comment

node/src/test_utils/persistent_configuration_mock.rs
utkarshg6: Same as the above comment

node/src/actor_system_factory.rs
utkarshg6: Doing that will cause a panic, also it was used like that before I modified.

node/src/db_config/persistent_configuration.rs
utkarshg6: Done 👍

node/src/db_config/persistent_configuration.rs
utkarshg6: Done 👍

node/src/node_configurator/node_configurator_standard.rs
utkarshg6: Done 👍 

Instead of doing what you mentioned in the comment, I discussed that in the dev meeting, and the current code represents the desired behaviour. So, all I did was to change the TODO into a comment to eliminate confusion.

node/src/actor_system_factory.rs
utkarshg6: Done 👍 

This is just a residue of old code for this card.

Previously, I was setting up the value while starting the actors. This has now changed and we're setting up the value in `configure_database`.

I was able to remove this code as if it was a dead leaf of tree.

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/database/db_migrations/migrations/migration_7_to_8.rs
utkarshg6: Done 👍

node/src/neighborhood/mod.rs
utkarshg6: I added this to make it safe for any changes that we may or may not do in the future.

If by any chance there are any side effects due to such changes, then the panic will remind us that we've broken the pattern somewhere.

I believe it makes our code more secure.

node/src/daemon/setup_reporter.rs
bertllll: Nope. Look at the test again, we do create a database with values there. If you want to be closer to the reality you should add 3 in the database. And thus I'd expect to see `("min-hops", "3", Default)` as the result.

node/src/database/db_initializer.rs
bertllll: Okay. I did.

node/src/node_configurator/node_configurator_standard.rs
bertllll: Okay...

node/src/test_utils/mod.rs
bertllll: I don't get. What is the benefit again? 

I cannot come to an idea why they should ever differ. Until then, if it by chance happen, it is just one more piece of confusion in our codebase.

node/src/test_utils/persistent_configuration_mock.rs
bertllll: You may prefer it but we've historically fought against that... You'd find many and many mocks where the param assertions for certain methods simply miss. 

Your solution is unusual and it is so called non-standard mock. There were times, when I was early here, when Dan insisted on avoiding that by all means. I remember I simply learned to obey him in this matter as it made quite a lot of sense to me. Asserting this way looks awkward and you could hardly see Dan to decide for it. As I said in the previous comment, it would've had to be a special case.

Sometimes when I run into alike assertions I tend to want to fix it. 

I'm asking you to bring this up in the dev meeting with Dan. :pray:

node/src/test_utils/persistent_configuration_mock.rs
bertllll: I still thing this is just bloating code.

node/src/test_utils/persistent_configuration_mock.rs
bertllll: Same as my response before.

node/src/test_utils/persistent_configuration_mock.rs
bertllll: I don't think you were reasoning right about it.

node/src/actor_system_factory.rs
bertllll: Ok

node/src/actor_system_factory.rs
bertllll: The answer is poetic. Nice :)

node/src/neighborhood/mod.rs
bertllll: Because we don't act like that with other parameters it still seems out of place here. I believe Dan and I wouldn't have done this or at least insisted to keep it.

I'd like you to discuss this with Dan because it is potentially needless code.

node/src/test_utils/mod.rs
utkarshg6: It makes it easier to find the usages of default value in the production code. Uncle Dan liked this a lot.

node/src/daemon/setup_reporter.rs
utkarshg6: Done 👍

node/src/neighborhood/mod.rs

We're picking the value from db now.

bertllll (approved):

bertllll (approved):

PR #298, opened 2023-06-02
GH 499 Review 2

dnwiebe:


automap/src/comm_layer/igdp.rs
```diff
@@ -440,33 +441,31 @@ impl IgdpTransactor {
             let _ = inner.housekeeping_commander_opt.take();
             error!(inner.logger, "Can't find router");
             change_handler(AutomapChange::Error(AutomapError::CantFindDefaultGateway));
-            return false;
+            return Finished::Yes;
         };
-        let (old_public_ip, current_public_ip) = match Self::retrieve_old_and_new_public_ips(
+        if let Ok((old_public_ip, current_public_ip)) = Self::retrieve_old_and_new_public_ips(
             inner.gateway_opt.as_ref().expectv("gateway_opt").as_ref(),
```
bertllll: Not needed for a repair, just showing you my point.

Philosophically though, ".expectv()" is meant to be used to reduce the message that something that by the logic should be there but is missing because I noticed we had to write such messages again and again anew and so in different words. Therefore, precisely, it should be `expectv("gateway")` not `expectv("gateway_opt")`.

automap/src/comm_layer/igdp.rs
```diff
@@ -1265,15 +1270,15 @@ mod tests {
...
-        assert_eq!(params_triple.2, 600);
+        let (_, hole_port, lifetime) = add_mapping_params.remove(0);
+        assert_eq!(hole_port, 6666);
+        assert_eq!(lifetime, 600);
+        let (_, hole_port, lifetime) = add_mapping_params.remove(0);
+        assert_eq!(hole_port, 6666);
+        assert_eq!(lifetime, 600);
+        let (_, hole_port, lifetime) = add_mapping_params.remove(0);
+        assert_eq!(hole_port, 6666);
+        assert_eq!(lifetime, 600);
```
bertllll: Hmm, these three times repeated three lines could easily be refactored into and iterator with `for_each()`over the range `(0..3)`. I think it'll be clearer and shorter.

automap/src/comm_layer/igdp.rs
```diff
@@ -1458,6 +1463,43 @@ mod tests {
...
+        IgdpTransactor::thread_guts_iteration(
+            &change_handler,
+            &inner_arc,
+            &mut Instant::now().sub(Duration::from_millis(1000)),
+            &Some(mapping_config),
+        );
+
+        let (_, hole_port, lifetime) = add_mapping_params_arc.lock().unwrap().remove(0);
+        assert_eq!(hole_port, 6689);
+        assert_eq!(lifetime, 1);
```
bertllll: Is this "one" related to the line 1494 in millis? If my interpretation is right, could you make the test clearer by using `Duration::from_secs(1)` instead. If no, could you try something out also (to get the same effect)?

automap/src/comm_layer/mod.rs
```diff
@@ -94,6 +94,33 @@ impl AutomapError {
             }
         }
     }
+
+    pub fn should_crash(&self) -> bool {
```
bertllll, line 96: I like this new method.

Just a detail though. I don't think I need to push you to alphabetical ordering. I don't think personally it'd be so important here. I was thinking, anyway, that sorting these errors as two compact groups of all trues followed by all falses could give it a nice facelift.

automap/src/comm_layer/mod.rs
```diff
@@ -94,6 +94,33 @@ impl AutomapError {
...
+            AutomapError::TemporaryMappingError(_) => false,
+            AutomapError::PermanentMappingError(_) => true,
+            AutomapError::ProbeServerConnectError(_) => true,
+            AutomapError::ProbeRequestError(_, _) => true,
+            AutomapError::ProbeReceiveError(_) => true,
+            AutomapError::DeleteMappingError(_) => false,
+            AutomapError::TransactionFailure(_) => true,
+            AutomapError::AllProtocolsFailed(_) => true,
+            AutomapError::HousekeeperAlreadyRunning => false,
+            AutomapError::HousekeeperCrashed => false,
```
bertllll: This error sounds serious, isn't it? 

I might see your reasoning: let's not kill the Node if we lose the ability to catch ip changes - it will eventually become useless, if the change comes to here, but we maybe can gain a couple of minutes or hours of the Node's run before so. Without necessary human assistance.

automap/src/comm_layer/mod.rs
```diff
@@ -264,4 +291,61 @@ mod tests {
...
+        ]
+        .into_iter()
+        .for_each(|(error, should_crash)| {
+            assert_eq!(
+                error.should_crash(),
+                should_crash,
+                "{:?}.should_crash should be {}, but was {}",
+                error,
+                should_crash,
+                !should_crash
```
bertllll, line 345: This is a cunning `!`. Consider resolved.

automap/src/comm_layer/pcp_pmp_common/mod.rs
```diff
@@ -379,8 +381,9 @@ pub mod tests {
         let result = subject.execute_command("ls booga");
 
         match result {
-            Err(stderr) if stderr.contains("No such file or directory") => (),
-            Err(stderr) => panic!("Unexpected content in stderr: '{}'", stderr),
+            Err(Either::Right(e)) => panic!("Unexpected error: '{:?}'", e),
+            Err(Either::Left(stderr)) if stderr.ends_with("No such file or directory\n") => (),
```
bertllll: Using this clause as the first one to show out that it's the test was written for seems worth a change to me. 

However this is even better: 
```
match result {
            Err(Either::Right(e)) => panic!("Unexpected error: '{:?}'", e),
            Err(Either::Left(stderr)) => assert_eq!(stderr.ends_with("No such file or directory\n"), true, "Unexpected content in stderr: '{}'", stderr)
            x => panic!("Expected error message in stderr; got {:?}", x),
        }
```

automap/src/comm_layer/pcp_pmp_common/mod.rs
```diff
@@ -393,13 +396,25 @@ pub mod tests {
         let result = subject.execute_command("dir booga");
 
         match result {
-            Err(stderr)
+            Err(Either::Left(stderr))
                 if stderr.contains("The system cannot find the file specified")
                     || stderr.contains("No such file or directory") =>
             {
                 ()
```
bertllll: Please use the standard assert_eq! or assert! macro here too.

automap/src/comm_layer/pcp_pmp_common/mod.rs
```diff
@@ -393,13 +396,25 @@ pub mod tests {
...
             }
-            Err(stderr) => panic!("Unexpected content in stderr: '{}'", stderr),
+            x => panic!("Expected error message in stderr; got {:?}", x),
+        }
+    }
+
+    #[test]
+    fn find_routers_command_works_when_error_is_returned() {
```
bertllll, line 412: Maybe `when_os_error_is_returned`?

masq_lib/src/logger.rs
```diff
@@ -389,12 +386,9 @@ mod tests {
...
             .expect("Unable to unwrap the duration_since for template after");
-        let received_message_count = Arc::new(AtomicU32::new(0));
-        let fake_ui_gateway = TestUiGateway::new(
-            (thread_count * msgs_per_thread) as u32,
-            received_message_count.clone(),
-            10,
-        );
+        let received_message_count_arc = Arc::new(AtomicU32::new(0));
+        let fake_ui_gateway =
+            TestUiGateway::new(total_msg_count as u32, received_message_count_arc.clone());
```
bertllll, line 391: The `.clone()` could be used inside, here a reference only, but I don't insist.

multinode_integration_tests/src/masq_cores_client.rs
```diff
@@ -9,6 +9,7 @@ use node_lib::sub_lib::cryptde::PublicKey;
 use node_lib::sub_lib::hopper::IncipientCoresPackage;
 use std::net::SocketAddr;
 
+// TODO: This is only used in one place. See if it can be replaced with MASQNodeClient.
```
bertllll: This still sounds like a good idea. Please investigate.

node/src/lib.rs
```diff
@@ -29,6 +29,7 @@ mod listener_handler;
 pub mod masquerader;
 pub mod neighborhood;
 pub mod node_configurator;
+#[cfg(test)]
```
bertllll: I'd just check out if that file itself doesn't have `#![cfg(test)]` in it at the top, then I might pick just one to stay.

node/src/neighborhood/gossip_producer.rs
```diff
@@ -335,7 +335,8 @@ mod tests {
     }
 
     #[test]
-    fn produce_debut_creates_a_gossip_to_a_target_about_ourselves_when_accepting_connections() {
+    fn produce_debut_or_ipchange_creates_a_gossip_to_a_target_about_ourselves_when_accepting_connections(
```
bertllll: Maybe this is my weak English but I cannot say for sure just from the name of the test whether it is the source or the target who is accepting connections. 


I could read it from the code or even deduce it but the name would be best if it was enough to make it clear.

node/src/neighborhood/gossip_producer.rs
```diff
@@ -362,7 +363,8 @@ mod tests {
     }
 
     #[test]
-    fn produce_debut_creates_a_gossip_to_a_target_about_ourselves_when_not_accepting_connections() {
+    fn produce_debut_or_ipchange_creates_a_gossip_to_a_target_about_ourselves_when_not_accepting_connections(
```
bertllll: If you think I was right above you can make changes here too.

node/tests/utils.rs
```diff
@@ -550,6 +550,6 @@ pub fn make_conn(home_dir: &Path) -> Box<dyn ConnectionWrapper> {
 }
 
 pub fn handle_connection_error(stream: TcpStream) {
-    let _ = stream.shutdown(Shutdown::Both).is_ok();
-    thread::sleep(Duration::from_millis(5000));
+    let _ = stream.shutdown(Shutdown::Both);
+    thread::sleep(Duration::from_millis(1000));
```
bertllll: I'm hoping this limit will be ok. I have no qualified guess on this.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -89,13 +89,21 @@ impl NeighborhoodDatabase {
...
-        match self.by_ip_addr.get(ip_addr) {
+    pub fn node_by_ip(&self, ip_addr: IpAddr) -> Option<&NodeRecord> {
+        match self.by_ip_addr.get(&ip_addr) {
             Some(key) => self.node_by_key(key),
             None => None,
         }
     }
 
+    pub fn node_by_ip_mut(&mut self, ip_addr: IpAddr) -> Option<&mut NodeRecord> {
+        let key_opt = self.by_ip_addr.get(&ip_addr).cloned();
```
bertllll, line 100: I don't think the `cloned()` has to be there. For referring to a Node to gets the right NodeRecord for it should a reference of the key be enough. I think it should work similarly like with the function above.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -503,20 +511,63 @@ mod tests {
...
+    #[test]
+    fn node_by_ip_mut_works() {
+        let this_node = make_node_record(1234, true);
+        let this_node_original_version = this_node.inner.version;
+        let one_node = make_node_record(4567, true);
+        let one_node_original_version = one_node.inner.version;
+        let another_node = make_node_record(5678, true);
+        let mut subject = db_from_node(&this_node);
+
+        subject.add_node(one_node.clone()).unwrap();
```
bertllll, line 543: I don't see this guy be acting the prime role. Yeah, it's harder here, but what about this central island: 
```
      //the end of the setup phase
      
        let mut this_node_actual = subject
            .node_by_ip_mut(this_node.node_addr_opt().unwrap().ip_addr())
            .unwrap();
        this_node_actual.inner.version += 1;
         let mut one_node_actual = subject
            .node_by_ip_mut(one_node.node_addr_opt().unwrap().ip_addr())
            .unwrap();
        one_node_actual.inner.version += 1;
        let excluded_node_record_opt = subject.node_by_ip_mut(another_node.node_addr_opt().unwrap().ip_addr();
        
        //the start of the assertions
        
```
...and then use assert_eq!() consecutively here

node/src/node_configurator/mod.rs
```diff
@@ -142,6 +143,42 @@ pub fn port_is_busy(port: u16) -> bool {
     TcpListener::bind(SocketAddr::new(localhost(), port)).is_err()
 }
 
+pub fn compute_mapping_protocol_opt(
```
bertllll: I'm certain that this function doesn't belong here.

a) the usage in production code is not in this file, where it fits fine,
b) no tests migrated over here; while I know this function takes a lot of tests. Those tests _should be_ also in `unprivileged_parse_args_configuration.rs` and I hope they are. Because chances are that the complicated merge confused you enough to lose some of them. Just be careful. I counted up six tests in master with this function as the act.  
c) putting the function here lacks a good reason; so the earned public visibility is awkward 

Edit: the count I said minus one because your version is different from master and has eliminated one case which would've required a separate test. 

Edit2: I think I've seen them live in `unprivileged_parse_args_configuration` which is good. I didn't count them up, yet they looked like lots.

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -481,6 +481,18 @@ impl PersistentConfigurationMock {
         self
     }
 
+    pub fn encrypt_past_neighbors_result(
```
bertllll: I don't like the nature of this method. It breaks the tight pattern we've always wanted to pursue so much.
I understand the benefits but it shouldn't be an excuse. You can take the extra logic use it in a test in its setup phase before you feed the result simply for `past_neighbors_result` which is expected.

What I don't like the most is that this looks like the trait has a member called `encrypt_past_neighbors` but no, we don't have it. 

If its used more widely you can create a test utility that you can call anytime you need to prepare an encrypted examples of the neighbors and then continue normally.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1573,11 +1061,12 @@ mod tests {
...
-        let mut persistent_config = make_default_persistent_configuration()
-            .check_password_result(Ok(true))
-            .check_password_result(Ok(true))
-            .check_password_result(Ok(true))
-            .change_password_result(Err(NotPresent));
+        let mut persistent_config =
+            configure_persistent_config(vec![PastNeighbors, GasPrice, MappingProtocol])
+                .check_password_result(Ok(true))
+                .check_password_result(Ok(true))
+                .check_password_result(Ok(true))
```
bertllll: Isn't that strange to you that there are so many results for `check_password` together?  Worth an investigation, says me.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -1001,9 +989,10 @@ mod tests {
     fn get_past_neighbors_handles_unavailable_password_for_parse_args_configuration_dao_real() {
         //sets the password in the database - we'll have to resolve if the use case is appropriate
         running_test();
-        let mut persistent_config = configure_default_persistent_config(ZERO)
-            .check_password_result(Ok(true))
-            .change_password_result(Ok(()));
+        let mut persistent_config =
+            configure_persistent_config(vec![PastNeighbors, GasPrice, MappingProtocol])
+                .check_password_result(Ok(true))
+                .change_password_result(Ok(())); // TODO: Do we need a change_password_result here?
```
bertllll: Please take it out for a sec and run it. My guts are saying it will be actually required because I think `get_past_neighbors` contains the weird piece of code that is enabled to initialize the database password if it's a brand new database and the password yet was supplied in multi_config. 

At any case, some removal is necessary, either the result or the TODO.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -2478,9 +2514,12 @@ mod tests {
             .unprivileged_parse_args(
                 &make_simplified_multi_config(args),
                 &mut bootstrapper_config,
-                &mut configure_default_persistent_config(
-                    ACCOUNTANT_CONFIG_PARAMS | MAPPING_PROTOCOL | RATE_PACK,
-                ),
+                &mut configure_persistent_config(PCField::base_and(vec![
```
bertllll: Somewhere you use `PCField::just_base()`, elsewhere `PCField::base_and(Vec<...>)`, but elsewhere just a vector of a lot of parameter names mentioned. I'm asking why are you using the second item just here and not in the heavy fed, long vectors? 

Also, I don't know exactly what all `just_base` means but I had a perception that it might have been the a lot repeated four parameters I can see at many tests here. If so what about their replacement with this nice ready function?

node/src/sub_lib/node_addr.rs
```diff
@@ -57,15 +61,21 @@ impl<'a> From<&'a SocketAddr> for NodeAddr {
     }
 }
 
-impl From<NodeAddr> for SocketAddr {
-    fn from(node_addr: NodeAddr) -> Self {
+impl From<&NodeAddr> for SocketAddr {
+    fn from(node_addr: &NodeAddr) -> Self {
         let all: Vec<SocketAddr> = node_addr.into();
         all[0]
```
bertllll, line 67: This is such a hack :man_facepalming:

node/src/sub_lib/node_addr.rs
```diff
@@ -57,15 +61,21 @@ impl<'a> From<&'a SocketAddr> for NodeAddr {
...
 }
 
-impl From<NodeAddr> for Vec<SocketAddr> {
+impl From<NodeAddr> for SocketAddr {
     fn from(node_addr: NodeAddr) -> Self {
+        Self::from(&node_addr)
+    }
+}
+
+impl From<&NodeAddr> for Vec<SocketAddr> {
```
bertllll, line 77: If you remove this code will we get enough broken code to say it's covered?

node/src/test_utils/mod.rs
```diff
@@ -577,63 +579,92 @@ pub mod unshared_test_utils {
...
         app_args.append(&mut array_of_borrows_to_vec(&args));
         let arg_matches = app_node().get_matches_from_safe(app_args).unwrap();
         MultiConfig::new_test_only(arg_matches)
     }
 
-    pub const ZERO: u32 = 0b0;
-    pub const MAPPING_PROTOCOL: u32 = 0b000010;
-    pub const ACCOUNTANT_CONFIG_PARAMS: u32 = 0b000100;
-    pub const RATE_PACK: u32 = 0b001000;
+    pub enum PCField {
```
bertllll, line 589: I appreciate the translation to this more friendly design. Well done.

node/src/actor_system_factory.rs
```diff
@@ -334,7 +334,12 @@ impl ActorSystemFactoryToolsReal {
     }
 
     fn handle_automap_error(prefix: &str, error: AutomapError) {
-        exit_process(1, &format!("Automap failure: {}{:?}", prefix, error));
+        let msg = format!("Automap failure: {}{:?}", prefix, error);
+        if error.should_crash() {
+            exit_process(1, &format!("Automap failure: {}{:?}", prefix, error));
+        } else {
+            todo!("What do we do if we don't crash? ({})", msg)
```
bertllll: Well, I don't know the answer, honestly, but the clear thing is that we don't want to panic on this `todo!()`. So please return to it and give it more thoughts. 

My suggestion: I'd act accordingly to the place of use of this function. In overall, though, it would be attempts to find a way how to log and let it run on or evaporate. I don't know if there are chances to find a similar solution.

node/src/actor_system_factory.rs
```diff
@@ -1787,8 +1775,8 @@ mod tests {
         };
     }
 
-    fn check_pool_bind_message(recording: &Arc<Mutex<Recording>>) {
-        let _pool_bind_message = Recording::get::<PoolBindMessage>(recording, 1);
+    fn check_pool_bind_message(recording: &Arc<Mutex<Recording>>, idx: usize) {
```
bertllll: When I saw this the first time from outside the function it was confusing because I falsely thought the number meant a number of messages expected. I know, it's stupid, if I think of it a bit deeper but yet the immediate impression makes me say that the numbers might be confusing in some way.

I don't think you need to necessarily find a better way, and so this can be accepted, because I would feel myself uneasy about such a request if you were me and I received it. I just let my thoughts be here.   
(making these assert helpers clearer could be done by well named enums as the argument(s) but risking being cumbersome after all again)

multinode_integration_tests/tests/ip_change_test.rs
```diff
@@ -54,5 +53,5 @@ fn receiving_announce_from_router_produces_ipchange_gossip() {
     // Have the mock router announce a change in public IP.
     // Verify that the mock Node receives IpChange Gossip.
     // Connect a client and send a request. Verify that the request shows up at the mock Node.
-    todo! ("Finish me")
+    todo!("Finish me")
```
bertllll: This `todo!()` seems serious. Do you have a plan for it? Shouldn't it be driven away yet in this card?

node/src/neighborhood/mod.rs
```diff
@@ -118,29 +118,30 @@ impl Handler<BindMessage> for Neighborhood {
...
 
     fn handle(&mut self, msg: PoolBindMessage, _: &mut Self::Context) -> Self::Result {
-        let gossip_acceptor_subs = self.gossip_acceptor_info_opt
+        let gossip_acceptor_subs = self
+            .gossip_acceptor_info_opt
             .take()
             .expect("Neighborhood never got its BindMessage")
-            .right_or_else (|_| panic! ("Neighborhood got multiple PoolBindMessages"));
-        self.gossip_acceptor_info_opt = Some(Either::Left (Box::new(GossipAcceptorReal::new(
+            .right_or_else(|_| panic!("Neighborhood got multiple PoolBindMessages"));
```
bertllll, line 140: I remember that you once exclaimed victoriously that the need for this complexity with different stages in the long initialization of the `gossip_acceptor` was gone. 
I was probably mistaken then but yes I'm surprised to see it still so bright here in this place.   

Speaking of the Either referred about both on the line above and below.

node/src/neighborhood/mod.rs
```diff
@@ -4057,13 +4061,9 @@ mod tests {
...
 
-        subject.process_acceptance_result(
-            GossipAcceptanceResult::Absorbed,
-            "new-IP Node",
-            27,
-        );
+        subject.process_acceptance_result(GossipAcceptanceResult::Absorbed, "new-IP Node", 1);
 
-        TestLogHandler::new().exists_log_containing("27-Node Gossip from new-IP Node absorbed");
+        TestLogHandler::new().exists_log_containing("1-Node Gossip from new-IP Node absorbed");
```
bertllll: This log isn't strong enough. It lacks the log level and also the name of the logger which should correspond to the name of this test.

automap/src/comm_layer/pmp.rs
```diff
@@ -1539,58 +1552,159 @@ mod tests {
...
+            Box::new(announcement_socket),
+            Box::new(|_| ()),
+            rx,
+        );
+
+        let result = subject.single_iteration(&mut None, &mut Instant::now());
+
+        assert_eq!(result, Finished::Yes);
+        TestLogHandler::default().exists_log_containing(
+            "ERROR: PmpTransactor: Node died; housekeeping thread is terminating",
```
bertllll: I'd appreciate if the logger had its own name for this test. To be 100% sure we're getting it from here.

automap/src/comm_layer/pmp.rs
```diff
@@ -1539,58 +1552,159 @@ mod tests {
...
-                &buffer[0..len_to_send],
-                router_connections.multicast_address,
-            )
-            .unwrap();
-        assert_eq!(sent_len, len_to_send);
-        thread::sleep(Duration::from_millis(1)); // yield timeslice
-        let _ = subject.stop_housekeeping_thread();
-        assert!(subject.housekeeper_commander_opt.is_none());
+        let announcement_socket = UdpSocketWrapperMock::new()
+            .recv_from_result(Ok((len_to_send, router_addr)), buffer.to_vec());
```
bertllll, line 1591: This is a non-standardly supplied staged result. This feeding function takes two args but we are used to just one. 

Could it possibly take a tuple if two different values must be supplied?

automap/src/comm_layer/pmp.rs
```diff
@@ -1539,58 +1552,159 @@ mod tests {
...
+            factories_arc,
+            router_addr,
+            change_handler: Box::new(change_handler),
+            announcement_read_timeout_millis: 100,
+            logger: Logger::new("test"),
+        };
+
+        let result = subject.handle_announcement_if_present(&None);
+
+        assert_eq!(result, ContinueWithIteration::Yes);
```
bertllll: I'm not a native so it's problematic for me to make up my mind about that but I'd like for example `ContinueInIteration` better or maybe `ContinueInThisIteration` even more determinant. You probably can rephrase it into something by one level better...

The idea is that `ContinueWithIteration` doesn't necessarily mean "_continue with the last, still incomplete and already partially executed iteration_".

automap/src/comm_layer/pmp.rs
```diff
@@ -1539,58 +1552,159 @@ mod tests {
...
+            vec![AutomapChange::NewIp(IpAddr::V4(new_ip_addr))]
+        );
+    }
+
+    #[test]
+    fn handle_announcement_if_present_handles_unparseable_data_from_router() {
+        let router_addr =
+            SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 20, 30, 40)), ANNOUNCEMENT_PORT);
+        let announcement_socket =
+            UdpSocketWrapperMock::new().recv_from_result(Ok((0, router_addr)), vec![]);
```
bertllll: Maybe I'd give a variable to this empty vector to allow less thinking about why is that the data is unparseable. 

Before you decide to answer this, look at the next comment I left under the following test. It might affect this comment in its implementation.

automap/src/comm_layer/pmp.rs
```diff
@@ -1539,58 +1552,159 @@ mod tests {
...
+            announcement_read_timeout_millis: 100,
+            logger: Logger::new("test"),
+        };
+
+        let result = subject.handle_announcement_if_present(&None);
+
+        assert_eq!(result, ContinueWithIteration::Yes);
         let changes = changes_arc.lock().unwrap();
         assert_eq!(*changes, vec![]);
     }
```
bertllll, line 1681: Three of the four previous tests could be easily refactored into tests sharing like 95% of their bodies through a special assertion function. 

See:
handle_announcement_if_present_ignores_data_if_not_from_router
handle_announcement_if_present_handles_unparseable_data_from_router
handle_announcement_if_present_handles_error_reading_from_router

They differ just in the staged result for the  `UdpSocketWrapperMock::new()
            .recv_from_result(...)` method.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -134,40 +146,47 @@ debug! (self.logger, "IpChangeHandler qualifies() found a database Node with nod
...
+                        peer_addr: old_addr,
+                        stream_type: RemovedStreamType::Clandestine,
+                        sub: self.stream_shutdown_sub.clone(),
+                    };
+                    self.remove_stream_sub
+                        .try_send(msg)
+                        .expect("StreamHandlerPool is dead");
+                });
+            } else {
+                panic!("Node {:?} in the database sent Gossip, accepts connections, but has no NodeAddr!", db_node.public_key());
```
bertllll: We should never get here because it doesn't offer how it could have passed the `qualifies` method called beforehand. I hope so, because otherwise this panic seems a vulnerability.

I don't actually like that this condition and the condition in `qualifies` might be tested separately. I'd like you to make sure that if qualifies changes and will accidentally allow the gossip to leak down here it should kill a test (running both `qualifies` and `handle` in a tandem). 
Not sure if it makes really sense to do that though. So it's more like asking you for some thoughts about it.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -985,13 +1008,28 @@ impl GossipHandler for StandardGossipHandler {
...
+                    && agr.inner.accepts_connections
+                    && agr.node_addr_opt.is_none()
+            })
+            .map(|agr| &agr.inner.public_key)
+            .collect();
+        if !violators.is_empty() {
+            return Qualification::Malformed(format!(
+                "Neighboring Node(s) claim to accept connections but present no NodeAddr: {:?}",
+                violators
+            ));
```
bertllll: Interestingly, this message reassembles the situation in the IpGossipChangeHandler where we don't hesitate and panic when this happens. Why the double standard?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1518,57 +1563,89 @@ mod tests {
...
+            IpChangeHandler::new(stream_handler_pool_sub, dispatcher_sub, Logger::new("test"));
+
+        let result = subject.qualifies(&mut dest_db, agrs_vec.as_slice(), gossip_source);
+
+        assert_eq!(
+            result,
+            Qualification::Malformed(
+                "IpChange Gossip: Node AgMEBQ accepts connections but provides no NodeAddr"
+                    .to_string()
+            ),
```
bertllll, line 1603: Is there any logging placed near to this event (not necessarily in the tested function) so that we can later do an analysis if something like this has come about?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -134,40 +146,47 @@ debug! (self.logger, "IpChangeHandler qualifies() found a database Node with nod
...
+                        .try_send(msg)
+                        .expect("StreamHandlerPool is dead");
+                });
+            } else {
+                panic!("Node {:?} in the database sent Gossip, accepts connections, but has no NodeAddr!", db_node.public_key());
+            };
+            return GossipAcceptanceResult::Absorbed;
         }
-        GossipAcceptanceResult::Absorbed
+        GossipAcceptanceResult::Ignored
```
bertllll: A comment here or at the expression of the if condition `accept_connections`, saying this bears on originate only Nodes that _somehow_ (you know better than me why) can get in here. 

I wonder in the first place why originate only Nodes are able to send this IpChangeGossip at all? Wouldn't leaving them out serve with more utility than allowing it and then having to cope with it on this other side with certain difficulties?

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1518,57 +1563,89 @@ mod tests {
...
 
     #[test]
-    fn properly_constructed_originate_only_ipchange_is_identified_and_handled() {
-        let old_addr = NodeAddr::new (&IpAddr::from_str("1.2.3.4").unwrap(), &vec![2000, 2001]);
-        let new_addr = NodeAddr::new (&IpAddr::from_str("2.3.4.5").unwrap(), &vec![2000, 2001]);
-        let gossip_source = SocketAddr::from (new_addr.clone());
-        let (gossip, new_node, mut db) =
-            make_ipchange (2345, Some (old_addr.clone()), Some (new_addr),
-                           Mode::OriginateOnly);
+    fn properly_constructed_originate_only_ipchange_is_identified_and_ignored() {
```
bertllll: Actually. I've got another question, after thinking about originate only Nodes again. 

Why is that we cannot reestablish a new connection with them? I understand that **other** Nodes cannot initiate the process. However, if we look at the standard debut Gossip, I suppose any originate only Node must make initiative and send it and later accept fully, if the other party also decided to accept. 
I also know the IpChange Gossip is extremely similar to a Debut. 

So why is it that we cannot handle this reconstitution of the relationship (perhaps shadowing the procedure happening during handling the Debut etc., in other words, the originate only one is the active one in this so why we cannot absorb it and instead the ignored part has to be played?)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2437,11 +2513,14 @@ mod tests {
...
-        assert_eq!(
-            Some(&expected_introducer),
-            dest_db.node_by_key(&agrs[0].inner.public_key)
-        );
+        let mut expected_introducer = NodeRecord::from(&agrs[0]);
+        let actual_introducer = dest_db.node_by_key(&agrs[0].inner.public_key).unwrap();
+        let expected_last_update = expected_introducer.metadata.last_update;
+        expected_introducer.metadata.last_update = actual_introducer.metadata.last_update;
+        assert_eq!(&expected_introducer, actual_introducer);
+        assert! (((expected_last_update as i32) - (actual_introducer.metadata.last_update as i32)).abs() < 10,
```
bertllll: First, why don't we know the exact number? This is not a multinode test or something, here we don't chaotically send Gossip around. I'd expect it predictable so I'm surprised by the loose assertion. However, I didn't think too seriously about how this value is formed. 

Also, if we need to keep this assertion so complex, removing the parenthesis around the recast last update values would make it leaner and more readable, I think. Rust doesn't require the parenthesis. It knows it has to cast values as early as possible.  

`((expected_last_update as i32) - (actual_introducer.metadata.last_update as i32)).abs()` could be 
`(expected_last_update as i32 - actual_introducer.metadata.last_update as i32).abs()`

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -4587,11 +4702,11 @@ mod tests {
...
-
-    fn node_by_ip_mut(db: &mut NeighborhoodDatabase, ip_addr: IpAddr) -> &mut NodeRecord {
-        let key = db.node_by_ip(&ip_addr).unwrap().public_key().clone();
-        db.node_by_key_mut(&key).unwrap()
-    }
+    //
+    // fn node_by_ip_mut(db: &mut NeighborhoodDatabase, ip_addr: IpAddr) -> &mut NodeRecord {
+    //     let key = db.node_by_ip(ip_addr).unwrap().public_key().clone();
+    //     db.node_by_key_mut(&key).unwrap()
+    // }
```
bertllll: Seems like you can delete these lines now.

automap/src/comm_layer/igdp.rs
dnwiebe: No, the lifetime is denominated in seconds by the IGDP protocol itself, so `Duration` doesn't really make sense. But I renamed the parameter to make its purpose clearer.

automap/src/comm_layer/igdp.rs
dnwiebe: I did it a little differently, but I don't think you'll object.

automap/src/comm_layer/igdp.rs
dnwiebe: ...okay...

automap/src/comm_layer/mod.rs
dnwiebe: Yeah, that was my thinking. Lots of people never get IP changes from their ISPs (me, for instance), so killing their Nodes when the housekeeping thread dies would be silly.

automap/src/comm_layer/mod.rs
dnwiebe: Alphabetical order is a good idea.

automap/src/comm_layer/pcp_pmp_common/mod.rs
dnwiebe: You think so? `ErrorKind::NotFound` is just a regular Rust standard-library error, not an OS error.

If I've misunderstood you, let me know; otherwise, I'll leave it like this.

masq_lib/src/logger.rs
dnwiebe: I couldn't find this bit of code in the repo. I think maybe it's already gone.

automap/src/comm_layer/pcp_pmp_common/mod.rs
dnwiebe: Well, not _standard,_ but more standard than a match.

automap/src/comm_layer/pcp_pmp_common/mod.rs
dnwiebe: I suppose I could live with that.

multinode_integration_tests/src/masq_cores_client.rs
dnwiebe: Okay, I got rid of the whole file.

node/src/lib.rs
dnwiebe: Good eye.

node/src/neighborhood/gossip_producer.rs
dnwiebe: Rearranged

node/src/neighborhood/gossip_producer.rs
dnwiebe: Rearranged

node/src/neighborhood/neighborhood_database.rs
dnwiebe: I tried to clean it up.

node/src/neighborhood/neighborhood_database.rs
dnwiebe: Turns out the Borrow Checker wants it that way.

node/src/node_configurator/mod.rs
dnwiebe: Moved.

node/src/test_utils/persistent_configuration_mock.rs
dnwiebe: Done.

node/src/node_configurator/node_configurator_standard.rs
dnwiebe: I pulled two of them out and it still works fine.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
dnwiebe: I added another utility method; then you got impatient.

node/src/sub_lib/node_addr.rs
dnwiebe: Until we eliminate the concept of clandestine ports, I don't think it's a hack. That sweeping change will remove this, of course, but by that time we won't need it. I'll leave it for now.

node/src/sub_lib/node_addr.rs
dnwiebe: Eeeeyup.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
dnwiebe: You're right!

node/src/test_utils/mod.rs
dnwiebe: Thanks!

node/src/actor_system_factory.rs
dnwiebe: I renamed some functions. Thanks for the heads-up: I think they're much clearer now.

node/src/actor_system_factory.rs
dnwiebe: I decided to log and return.

multinode_integration_tests/tests/ip_change_test.rs
dnwiebe: Yup, I have a plan for it. It's not done  yet, but the plan is there.

node/src/neighborhood/mod.rs
dnwiebe: Fair enough.

node/src/neighborhood/mod.rs
dnwiebe: I think that was a change that Utkarsh made. It's either in `master` or will be there soon; I'll eventually have to merge it in.

automap/src/comm_layer/pmp.rs
dnwiebe: Okay...

automap/src/comm_layer/pmp.rs
dnwiebe: Yeah...the whole `recv_from()` method of the mock is quite nonstandard. I don't completely remember why it is, but part of it is the signature of `recv_from()`, which takes as a parameter an empty buffer that is to be filled by the method. Our standard pattern for mocks doesn't handle output parameters like that very well. I do remember that I tried figuring out how to standardize it for awhile until I gave up.

automap/src/comm_layer/pmp.rs
dnwiebe: Wasn't _quite_ that easy, but close. Good catch.

automap/src/comm_layer/pmp.rs
dnwiebe: Not a bad idea.

automap/src/comm_layer/pmp.rs
dnwiebe: Okay, I changed it...perhaps not in the way you expected, but I think it's clearer.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: They're timestamps. They should be very close, but because they're stamped at different places in the code, we can't guarantee that they'll be exactly the same. We also can't guarantee they'll be at most one millisecond apart, both because Actions VMs are slow and because the cheap low-resolution timers used by most OSes have a granularity greater than 1ms. (The granularity of the old 4.77MHz IBM PC's timer was 55.6ms, or 18 ticks per second; I still remember that.) The assumption here is that the system's timer granularity will be 10ms or less, and that not more than one tick transition will elapse during the test.

And I think the extra parentheses make it easier to read, not harder. However, I'll take 'em out for you.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: I think so. All the malformations are logged, and they'll be right next to the gossip diagram, if TRACE logging is on.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: This turned out to be a stupid idea, and I took it out.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: Yes: if we get there, then `qualify()` is not working. If `qualify()` is modified so it doesn't catch that, it should fail a test too, so I think the `panic!` is justified.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: Good...I think we should keep it.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: I fixed it.

PR #302, opened 2023-06-19
GH-691: Review 2

utkarshg6:


node/src/daemon/setup_reporter.rs
```diff
@@ -835,6 +836,18 @@ impl ValueRetriever for MinHops {
...
+
+    fn computed_default(
+        &self,
+        _bootstrapper_config: &BootstrapperConfig,
+        persistent_config: &dyn PersistentConfiguration,
+        _db_password_opt: &Option<String>,
+    ) -> Option<(String, UiSetupResponseValueStatus)> {
+        match persistent_config.min_hops() {
+            Ok(min_hops) => Some((min_hops.to_string(), Configured)),
+            Err(e) => panic!("Database is corrupt: {:?}", e),
```
bertllll: Not sure if we ever want to make Daemon panic. Remember that it starts up during booting the hosting machine. It's not a process that you could see as one of your open windows. Killing it means the user will not know what's wrong with our software and also how he could make it run again. He'd have to reboot his machine which isn't friendly from our side. So I'd consider (maybe to log first; feel free to create a new short-life logger in  place for that purpose) and pick the default if this happens. If the problems repeat in the Node too, it will be its end but Daemon should probably be steadier.

bertllll: Also, I think we can write this better, more indicative. When the value from the persistent configuration equals the default you should return `UiSetupResponseValueStatus` being `Default` and not `Configured`. Can you please add in this little logic?

node/src/neighborhood/mod.rs
```diff
@@ -547,17 +547,22 @@ impl Neighborhood {
...
                 .expect("Min Hops Count value is not initialized inside Database");
-            let demanded = self.min_hops_count;
-            if demanded != from_db {
-                panic!(
-                    "Database with wrong min hops count value detected; expected: {:?}, was: {:?}",
-                    demanded, from_db
-                )
+            let value_in_neighborhood = self.min_hops;
+            if value_in_neighborhood != value_in_db {
+                debug!(
```
bertllll: I can imagine this should be `info!()`. Think, if this happen it means the user (although a bit unrealistic) managed to change the value in the meantime, then it makes sense to confirm for him that we detected the change and will use the new value, the same one as which was in his instructions.

node/src/neighborhood/mod.rs
```diff
@@ -4640,17 +4642,17 @@ mod tests {
 
         System::current().stop();
         system.run();
-        let min_hops_count_params = min_hops_count_params_arc.lock().unwrap();
-        assert_eq!(*min_hops_count_params, vec![()]);
+        let min_hops_params = min_hops_params_arc.lock().unwrap();
+        assert_eq!(*min_hops_params, vec![()]);
```
bertllll: We said you want to use the assertion message to test the state of the Actor after the validation. It'd be strong enough assertion. Your version does not prove things at all. It just says that somewhere within the code somebody called `min_hops()`, however, what was done with the value or what the action influenced is a big unknown. 

Only if you prove that the state of the actor, with the concern of the min_hops parameter, was the same before and after we can speak of some conclusiveness.    

Bonus: it might also mean that in the end you won't really need your param assertion tool set. If I were in your shoes, I'd probably get rid of those rather unnecessary pieces of code with the unit type as the asserting medium.

node/src/neighborhood/mod.rs
```diff
@@ -4660,72 +4662,41 @@ mod tests {
...
                 make_wallet("earning"),
                 None,
                 test_name,
             ),
         );
+        subject.logger = Logger::new(test_name);
         subject.persistent_config_opt = Some(Box::new(
             PersistentConfigurationMock::new()
-                .min_hops_count_result(Err(PersistentConfigError::NotPresent)),
+                .min_hops_params(&min_hops_params_arc)
```
bertllll: No need for `.min_hops_params(...)`, the assertion is now done by different means, and this one is just bloating code.  The fact that `.min_hops_result(Ok(min_hops_in_db))` sets a certain result which also finds the assertion message in the survey of the actor's guts doesn't actually need any further extensions.

node/src/daemon/setup_reporter.rs
utkarshg6: Done 👍

utkarshg6: Done 👍

node/src/neighborhood/mod.rs

node/src/neighborhood/mod.rs

node/src/neighborhood/mod.rs

PR #304, opened 2023-06-21
GH-691: Review 3

utkarshg6:


node/src/neighborhood/mod.rs
```diff
@@ -4690,10 +4688,8 @@ mod tests {
         addr.try_send(assertions_msg).unwrap();
         System::current().stop();
         system.run();
-        let min_hops_params = min_hops_params_arc.lock().unwrap();
-        assert_eq!(*min_hops_params, vec![()]);
         TestLogHandler::new().exists_log_containing(&format!(
-            "DEBUG: {test_name}: Database with different min hops value detected; \
+            "INFO: {test_name}: Database with different min hops value detected; \
             expected: {:?}, found in db: {:?}; replacing to {:?}",
```
bertllll: Maybe instead of "expected" something like "Currently set" or "Up to now:" 
And I don't know, "replacing to" sounds out of place to me. I'd rather expect a different verb. "Changing to?" or "resetting to".

Maybe like: `Database with different min hops value detected; currently set: {:?}, found in db: {:?}; changing to {:?}.`

PR #305, opened 2023-06-21
GH-719: Dumping BlockchainInterfaceClandestine

bertllll:


dnwiebe (commented):
My opinion is that the Null Object Pattern would be cleaner here, but what you have works, and I don't feel justified mandating it. I'll leave the decision in your hands.

PR #306, opened 2023-06-22
GH-499 Review 3

dnwiebe:


automap/src/comm_layer/pmp.rs
```diff
@@ -1565,43 +1565,16 @@ mod tests {
...
             rx,
         );
+        subject.logger = Logger::new("single_iteration_terminates_if_commander_channel_dies");
 
         let result = subject.single_iteration(&mut None, &mut Instant::now());
 
         assert_eq!(result, Finished::Yes);
         TestLogHandler::default().exists_log_containing(
-            "ERROR: PmpTransactor: Node died; housekeeping thread is terminating",
+            "ERROR: single_iteration_terminates_if_commander_channel_dies: Node died; housekeeping thread is terminating",
```
bertllll, line 1574: This is optional. I think Utharhe's way looks best, which is:

Usually a variable defined somewhere at the top of the test:
```
init_test_logging();
let test_name = "single...dies";
```
later
```
subject.logger = Logger::new(test_name);
```
and the best part 
```
TestLogHandler::default().exists_log_containing(
            &format!("ERROR: {test_name}: Node died; housekeeping thread is terminating")
)
```
Notice how the asserted log message becomes less clumsy with the name, usually long, and the reader can quickly concentrate to the actual log message, its meaning.

automap/src/comm_layer/pmp.rs
```diff
@@ -1637,76 +1610,74 @@ mod tests {
...
         );
     }
 
-    #[test]
-    fn handle_announcement_if_present_handles_unparseable_data_from_router() {
-        let router_addr =
-            SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 20, 30, 40)), ANNOUNCEMENT_PORT);
-        let announcement_socket =
-            UdpSocketWrapperMock::new().recv_from_result(Ok((0, router_addr)), vec![]);
+    fn router_addr() -> SocketAddr {
```
bertllll, line 1621: Maybe a static variable from `lazy_static!`? Seems more idiomatic.

masq_lib/src/utils.rs
```diff
@@ -190,6 +191,10 @@ where
     }
 }
 
+pub fn to_millis(dur: &Duration) -> u64 {
```
bertllll, line 194: Cannot you just delete this function and instead call the standard library function, probably:
```
let dur: Duration = ...;
let millis = dur.as_millis(); 
```
Either like I wrote or sounding similar. I do believe you would find one if you look there. 

Suggesting to get rid of this entire function and replace it accordingly.

node/src/bootstrapper.rs
```diff
@@ -376,7 +375,7 @@ impl BootstrapperConfig {
             },
             blockchain_bridge_config: BlockchainBridgeConfig {
                 blockchain_service_url_opt: None,
-                chain: TEST_DEFAULT_CHAIN,
+                chain: TEST_DEFAULT_CHAIN, // TODO Shouldn't this be just DEFAULT_CHAIN?
```
bertllll, line 378: Yeah. I was also always surprised passing by. One day I came to a conclusion that somebody before had decided to use this chain in the production code because it'd made tests easier to write tests for them. A huge bunch of tests are dependent on the BootstrapperConfig and also a big amount of them expect the chain used in them to be the special test chain. This way, the test chain stayed defaulted and then somebody made sure that in production it would be properly changed to the desired chain. This is what I think was going on in someone's mind when allowing this. 

Please decide if you want to make a change on this (and it should perhaps be a new card) or you don't care so much and remove the TODO. If you thought it through carefully I would appreciate it because it is once over a long time when we get to talk about this and so this is an occasion.

node/src/database/db_initializer.rs
```diff
@@ -6,12 +6,12 @@ use crate::db_config::secure_config_layer::EXAMPLE_ENCRYPTED;
 use crate::sub_lib::accountant::{DEFAULT_PAYMENT_THRESHOLDS, DEFAULT_SCAN_INTERVALS};
 use crate::sub_lib::neighborhood::DEFAULT_RATE_PACK;
 use masq_lib::blockchains::chains::Chain;
+#[cfg(test)]
```
bertllll, line 9: I doubt that it is truly `#[cfg(test)]`, it cannot be, because you put it in `constants` in `masq_lib`, and I like it at the end of the day, but how obvious this must be that it cannot be technically done as that would be inter-crate compilation configuration with this mark, which doesn't work.   

Please remove.

node/src/sub_lib/route.rs
```diff
@@ -349,7 +349,7 @@ mod tests {
     use crate::sub_lib::cryptde_null::CryptDENull;
     use crate::test_utils::make_wallet;
     use crate::test_utils::{main_cryptde, make_paying_wallet};
-    use masq_lib::test_utils::utils::TEST_DEFAULT_CHAIN;
```
bertllll, line 352: Do we really have such a silly naming that we `utils` is a file in `test_utils`? Maybe `other_utils` or `miscellaneous_utils` (oh yeah, the last one is not gonna win this).

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -639,3 +627,13 @@ impl PersistentConfigurationMock {
         }
     }
 }
+
+pub fn encrypted_past_neighbors(
+    cryptde: &dyn CryptDE,
+    past_neighbors: &str,
+) -> Vec<NodeDescriptor> {
+    past_neighbors
+        .split(",")
```
bertllll, line 636: We probably accept also syntax with more than one spaces. Although, I don't think somebody would be so absent-minded to do this in a test and if so he would probably soon understand what advert consequences it brought to him.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -4702,11 +4688,6 @@ mod tests {
             }
         }
     }
-    //
-    // fn node_by_ip_mut(db: &mut NeighborhoodDatabase, ip_addr: IpAddr) -> &mut NodeRecord {
```
bertllll, line 4706: Seems unneeded after all. Please go and delete it for sure.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1658,7 +1642,7 @@ mod tests {
         System::current().stop();
         system.run();
         assert_eq!(Qualification::Matched, qualifies_result);
-        assert_eq!(handle_result, GossipAcceptanceResult::Ignored,);
+        assert_eq!(handle_result, GossipAcceptanceResult::Absorbed);
```
bertllll, line 1645: How is that possible that you say Absorbed, which I consider different enough from Ignored, and despite that the database data did not change? Are the assertions below right and complete? Why do we get None for the Node which sent the Gossip and wanted us to know that his IP is different. Strangely enough, we won't find it at all in our database when we run our assertions.

The same with the stream handler pool, I'd expect some activity. 

If the origin only guy just contacted us with this special delivery we've got them waiting at the doorsteps. They've already initiated a connection, different from the previous one, and so we can accept and hold it.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -4318,14 +4302,16 @@ mod tests {
     }
 
     #[test]
-    fn standard_gossip_containing_neighbor_node_that_accepts_connections_but_has_no_node_addr_is_rejected(
+    fn standard_gossip_containing_non_transmitting_neighbor_node_that_accepts_connections_but_has_no_node_addr_is_accepted(
```
bertllll, line 4305: Hmmm, that I'm asking might be because I don't have an enough good nose for this but what do you mean by "non_transmitting_neighbor"...is that "consume only"?

multinode_integration_tests/src/main.rs
```diff
@@ -136,9 +137,14 @@ impl MockNode {
 
     #[allow(clippy::map_entry)]
```
bertllll, line 138: Could you please look at this? I suppose that if you remove it and it does still causes a Clippy error it will be one that you can easily fix. I'm forecasting a laughably easy thing to fix out there. It seems worth the hassle, in the name of prettier code.

multinode_integration_tests/src/main.rs
```diff
@@ -148,8 +154,9 @@ impl MockNode {
             writeln!(stderr, "{}", open_err_msgs.join("\n")).unwrap();
             return 1;
         }
+        // TODO: Remember to open the UDP ports too
```
bertllll, line 157: Does this mean we don't actually have any UDP ports that could be open? Right? 

Well, if not, then some code is missing... if yes, then I'd appreciate a better message for this TODO

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
+    port_specs: Vec<PortSpec>,
+}
+
+impl FromStr for ProbeTarget {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut main_pieces = s.split(':').collect_vec();
+        if main_pieces.len() != 2 {
+            return Err("Syntax: <IP address>:<U|T><port>/<U|T><port>/...".to_string());
```
bertllll, line 364: The repeatedly used message for incorrect inputs is worth creating a manifest constant e.g. somewhere near above this implementation.

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
+        }
+        let ip_address_string = main_pieces.remove(0);
+        let port_specs_string = main_pieces.remove(0);
```
bertllll, line 367: In theory, you did not have to make the vector mutable and remove the items out.  Because all you need are referenced pieces of &str you might have been able to use just the slice indexing like `main_pieces[0]`. But let's perhaps not make this test code as dry as spots in Sahara.

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
+        let ip_address = match IpAddr::from_str(ip_address_string) {
+            Ok(ip_addr) => ip_addr,
+            Err(_) => return Err("Syntax: <IP address>:<U|T><port>/<U|T><port>/...".to_string()),
+        };
+        let port_spec_strings = port_specs_string.split('/').collect_vec();
+        let port_spec_results = port_spec_strings
+            .into_iter()
+            .map(PortSpec::from_str)
+            .collect_vec();
```
bertllll, line 376: This chain of iterator methods above could have been just a smooth continuation of this part `port_specs_string.split('/')`, with omitting the first `collect_vec()` methods and just one use of it at all. Not sure if you could be interested in such thinking.

bertllll, line 376: Also I understand why using `.collect_vec()`  feels cool but has an disadvantage. It does not show you what type the collected items have. You have to read it out from elsewhere or simply guess. Here I caught myself wondering because of the next chain of iterators for finding the errors. Knowing the type would've made my happier.

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
+        let first_error_opt = port_spec_results
+            .iter()
+            .find(|result| result.is_err())
+            .map(|err| err.clone().err().unwrap());
```
bertllll, line 380: Because of the previously depicted quastionability about the types, I cannot make up my mind about if the `.clone()` belongs there and why. It might be possible to factor it out but it depends on the type...yeah.

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
 }
 
+#[derive(Debug, Eq, PartialEq, Clone)]
+struct ProbeTarget {
+    ip_address: IpAddr,
+    port_specs: Vec<PortSpec>,
+}
+
+impl FromStr for ProbeTarget {
+    type Err = String;
```
bertllll, line 359: This is actually rather a supplement to the following comment, but if I were you, I'd make the Err type a static str `&'static str`, because the always the same error messages allow it, by which you wouldn't need to write `.to_string()` after each of those (repeated) error messages.

multinode_integration_tests/src/main.rs
bertllll, line 380: Interestingly, I realized that you are preparing the error for yourself here but you're not going to use it anywhere in the rest of this function. You need to decide  if you want to include it into the returned error message. 

You might not want to... then a boolean would be better than an option. I'd recommend to use `.any(|result|result.is_err())`, an iterator nicely fitting for this situation.

multinode_integration_tests/src/main.rs
```diff
@@ -342,6 +349,104 @@ impl MockNode {
...
+}
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+struct PortSpec {
+    protocol: NetworkProtocol,
+    port: u16,
+}
+
+impl FromStr for PortSpec {
+    type Err = String;
```
bertllll, line 414: Again, `&'static str` is an attractive option here. Shaking off the `.to_string()`s.

multinode_integration_tests/src/main.rs
```diff
@@ -489,6 +592,125 @@ mod tests {
         assert_eq!(data_hunk.data, vec!(1, 2, 3, 4));
     }
 
+    #[test]
+    fn can_read_probe_target_from_str() {
+        probe_target_pairs()
+            .into_iter()
+            .for_each(|(string, expected)| {
```
bertllll, line 599: You probably should name the local variable `string` better. Perhaps `literal_target` or `str_target_address`.

Even the word `expected` is a bit vague, lacking a good context. I know you can argue that the test name can lead you around down to the idea but there might be a way to write it more expressively.

multinode_integration_tests/src/main.rs
```diff
@@ -489,6 +592,125 @@ mod tests {
...
+
+                assert_eq!(result, expected);
+            })
+    }
+
+    #[test]
+    fn display_implementation_works() {
+        probe_target_pairs()
+            .into_iter()
+            .for_each(|(expected, target)| {
```
bertllll, line 610: Similar issue with expected. 

The problem is that this kind of simplicity forces the reader to jump away to see the `probe target pairs` and learn about their data types.

multinode_integration_tests/src/main.rs
```diff
@@ -489,6 +592,125 @@ mod tests {
...
+                "12.23.34.45:U4321/5432",
+                "Syntax: <IP address>:<U|T><port>/<U|T><port>/...",
+            ),
+            (
+                "12.23.34.45:U4321/T",
+                "Syntax: <IP address>:<U|T><port>/<U|T><port>/...",
+            ),
+            (
+                "12.23.34.45:X4321",
+                "Syntax: <IP address>:<U|T><port>/<U|T><port>/...",
```
bertllll, line 653: Since the error response is always the same it is teasing to rewrite this so that the initial vector contains only the strings with the bad syntax, you can either hard-code the expected error message into the assertion in the `for_each()` or if you have the manifest constant for it by now, which I've already advised for its potentially handiness, you can simply put it in anyaway.

multinode_integration_tests/src/mock_router.rs
```diff
@@ -8,13 +7,19 @@ pub trait MockRouter {
 pub struct MockPcpRouter {}
 
 impl MockRouter for MockPcpRouter {
-    fn announce_ip_change(&self, target_ip: IpAddr, new_ip_address: IpAddr) {
+    fn announce_ip_change(&self, _target_ip: IpAddr, _new_ip_address: IpAddr) {
         todo!()
```
bertllll, line 11: Strange that you don't have a usage for this yet. Maybe in the second test you're still working   on...?

node/src/actor_system_factory.rs
```diff
@@ -1172,27 +1183,27 @@ mod tests {
...
-        check_pool_bind_message(&recordings.neighborhood, 1);
-        check_new_ip_message(
+        check_for_initial_bind_message(&recordings.dispatcher, false);
+        check_for_initial_bind_message(&recordings.hopper, false);
+        check_for_initial_bind_message(&recordings.proxy_client, false);
+        check_for_initial_bind_message(&recordings.proxy_server, false);
+        check_for_initial_bind_message(&recordings.neighborhood, false);
+        check_for_initial_bind_message(&recordings.ui_gateway, false);
+        check_for_initial_bind_message(&recordings.accountant, false);
+        check_for_pool_bind_message_at(&recordings.stream_handler_pool, 0); // what _should_ we be doing here?
```
bertllll, line 1193: Do you happen to remember if the comment at the end of the line is from you? Could you write it while you were working on this card? It sounds possible to me. Please decide the destiny of it yourself, but I think it won't do anybody any good in the future if they are missing the background story you have.

node/src/actor_system_factory.rs
```diff
@@ -1605,6 +1616,39 @@ mod tests {
         system.run();
     }
 
+    #[should_panic(expected = "1: Automap failure: prefixProtocolError(\"Booga\")")]
```
bertllll, line 1619: I don't understand why one couldn't insert a space in between the prefix and the actual error, but if you insist on this I believe you do based on an actuality which I'm not getting at the moment.

node/src/actor_system_factory.rs
```diff
@@ -1605,6 +1616,39 @@ mod tests {
...
+        let automap_control_factory = AutomapControlFactoryMock::new()
+            .make_params(&make_params_arc)
+            .make_result(Box::new(AutomapControlMock::new()));
+        let mut subject = ActorSystemFactoryToolsReal::new();
+        subject.automap_control_factory = Box::new(automap_control_factory);
+
+        let _ = subject.make_automap_control(None, vec![]);
+
+        let mut make_params = make_params_arc.lock().unwrap();
+        let change_handler = make_params.remove(0).1;
```
bertllll, line 1645: Sometimes mentioning even the other items in the tuple is valuable, giving some more context to the test. Not sure this is the case, easily not.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -532,44 +532,43 @@ mod tests {
...
+        let another_node_ip_addr = another_node.node_addr_opt().unwrap().ip_addr();
         let mut subject = db_from_node(&this_node);
-
         subject.add_node(one_node.clone()).unwrap();
+        fn mutable_node(db: &mut NeighborhoodDatabase, ip_addr: IpAddr) -> &mut NodeRecord {
+            db.node_by_ip_mut(ip_addr).unwrap()
+        }
+        fn node_version(db: &NeighborhoodDatabase, ip_addr: IpAddr) -> u32 {
+            db.node_by_ip(ip_addr).as_ref().unwrap().inner.version
+        }
```
bertllll, line 549: I developed a certain habit and I don't mean to impose it onto you but I like when any inner functions stand at the top of the outer function, as the first lines, and after that I leave a blank line there to end that area after which I can start writing the actual code of the outer function (or of the test) .

Thinking more about it, I think here this is a special situation where it makes some sense if them stay as central as possible because they represent something like an Act.

node/src/neighborhood/gossip_producer.rs
```diff
@@ -335,46 +336,37 @@ mod tests {
     }
 
     #[test]
-    fn produce_debut_or_ipchange_creates_a_gossip_to_a_target_about_ourselves_when_accepting_connections(
+    fn produce_debut_or_ipchange_creates_properly_formed_gossip_with_node_addr_opt_if_we_accept_connections_through_the_firewall(
```
bertllll, line 339: I don't like the phrase `...with_node_addr_opt_` because, after all, the true meaning is that the node_addr_opt is Some(...) and on the contrary , `...without_node_addr_opt_` is simply None. 

Partially I feel the problem lies in the word `opt` and should be dropped. `...with_node_addr_present...` is is cleaner to me in overall.

multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -579,7 +579,19 @@ impl MASQMockNode {
     }
 
     fn make_node_args(node_addr: &NodeAddr) -> Vec<String> {
-        vec![format!("{}", node_addr)]
+        vec![Self::add_tcp_specifications(format!("{}", node_addr))]
```
bertllll, line 582: `format!("{}", node_addr)` is an unidiomatic way to say `node_addr.to_string()`.

multinode_integration_tests/src/masq_mock_node.rs
```diff
@@ -579,7 +579,19 @@ impl MASQMockNode {
     }
 
     fn make_node_args(node_addr: &NodeAddr) -> Vec<String> {
```
bertllll, line 581: These are definitely not "node args", so the name should be improved. This name speaks with strong self-esteem so to speak.

Maybe something about addresses or the place where it will be eventually used or the purpose?

multinode_integration_tests/tests/ip_change_test.rs
```diff
@@ -62,29 +60,34 @@ fn receiving_announce_from_router_produces_ipchange_gossip() {
...
+            octets[2],
+            octets[3] + 1,
+        ))
     };
     let mock_router = MockPcpRouter::new();
-    let mut mock_node = node_map.remove(&neighbor_key).unwrap();
+    let mock_node = node_map.remove(&neighbor_key).unwrap();
     thread::sleep(Duration::from_secs(1)); // Wait for the new Node to get situated
-    // Have the mock router announce a change in public IP.
+                                           // Have the mock router announce a change in public IP.
```
bertllll, line 80: Why this long indentation? Can it be made prettier? Maybe the comments should simply by above or below the code lines, not after it.

multinode_integration_tests/tests/ip_change_test.rs
```diff
@@ -62,29 +60,34 @@ fn receiving_announce_from_router_produces_ipchange_gossip() {
...
+    let (_, _, live_cores_package) = mock_node
         .wait_for_package(&JsonMasquerader::new(), Duration::from_secs(2))
         .unwrap();
     let mut gossip = decodex::<Gossip_0v1>(
         mock_node.main_cryptde_null().unwrap(),
         &live_cores_package.payload,
-    ).unwrap();
+    )
+    .unwrap();
     let node_record = gossip.node_records.remove(0); // now verify the gossip
```
bertllll, line 91: The comment would better be above the line, introducing the final stage of the test.

automap/src/comm_layer/mod.rs
```diff
@@ -22,103 +22,101 @@ pub const DEFAULT_MAPPING_LIFETIME_SECONDS: u32 = 600; // ten minutes
...
+            }
+            AutomapError::HousekeeperCrashed => {
+                AutomapErrorCause::Unknown("Thread crash".to_string())
+            }
+            AutomapError::IPv6Unsupported(_) => AutomapErrorCause::NetworkConfiguration,
+            AutomapError::NoLocalIpAddress => AutomapErrorCause::NetworkConfiguration,
             AutomapError::SocketBindingError(_, _) => AutomapErrorCause::UserError,
-            AutomapError::SocketSendError(aec) => aec.clone(),
             AutomapError::SocketReceiveError(aec) => aec.clone(),
+            AutomapError::SocketSendError(aec) => aec.clone(),
```
bertllll, line 80: Maybe consider if AutomapErrorCause would be a good candidate to implement Copy. I'm inclining to say yes.

automap/src/comm_layer/mod.rs
```diff
@@ -22,103 +22,101 @@ pub const DEFAULT_MAPPING_LIFETIME_SECONDS: u32 = 600; // ten minutes
...
             AutomapError::CantFindDefaultGateway => true,
-            AutomapError::IPv6Unsupported(_) => true,
+            AutomapError::DeleteMappingError(_) => false,
             AutomapError::FindRouterError(_) => true,
             AutomapError::GetPublicIpError(_) => true,
-            AutomapError::SocketBindingError(_, _) => true,
-            AutomapError::SocketSendError(_) => true,
-            AutomapError::SocketReceiveError(_) => true,
+            AutomapError::HousekeeperAlreadyRunning => false,
+            AutomapError::HousekeeperCrashed => false,
```
bertllll, line 104: Looks a bit controversial to me... but your reasoning is that if we the Node can make use of at least some time before the public IP address changes then we should allow even such a mystic hardness happening behind the wall as which goes to the housekeeping thread meeting the death... I know your decision has been probably right but it takes time before a person forgets the initial bad taste of it.

automap/src/comm_layer/mod.rs
```diff
@@ -251,36 +238,47 @@ mod tests {
...
             ),
             (
-                AutomapError::TemporaryMappingError(String::new()),
-                AutomapErrorCause::RouterFailure,
+                AutomapError::ProbeReceiveError(String::new()),
+                AutomapErrorCause::ProbeFailed,
+            ),
+            (
+                AutomapError::ProbeRequestError(AutomapErrorCause::ProbeFailed, String::new()),
+                AutomapErrorCause::ProbeFailed,
```
bertllll, line 246: Also a bit strange. Why does the original error already bears the AutomapErrorCause inside. Shouldn't it ask for it if it needs it instead? Seems like mixing roles. I do think the code could be repaired accordingly...

automap/src/comm_layer/mod.rs
```diff
@@ -251,36 +238,47 @@ mod tests {
...
+                AutomapErrorCause::ProtocolNotImplemented,
             ),
             (
-                AutomapError::ProbeReceiveError(String::new()),
-                AutomapErrorCause::ProbeFailed,
+                AutomapError::SocketBindingError(
+                    String::new(),
+                    SocketAddr::from_str("1.2.3.4:1234").unwrap(),
+                ),
+                AutomapErrorCause::UserError,
```
bertllll, line 261: Hmmm. What exactly does "user error" mean? With this name alone and the shown example it's hard for me to see the link between.

(This comment was written late in the night - in fact, it's the morning already - so my senses are generally numb. And so with any comment yet to come.)

automap/src/comm_layer/mod.rs
```diff
@@ -251,36 +238,47 @@ mod tests {
...
             ),
             (
-                AutomapError::DeleteMappingError(String::new()),
-                AutomapErrorCause::ProtocolFailed,
+                AutomapError::SocketReceiveError(AutomapErrorCause::Unknown("Booga".to_string())),
+                AutomapErrorCause::Unknown("Booga".to_string()),
```
bertllll, line 265: I can see now that you use the tight coupling of the Error itself with the ErrorCause quite a lot. Still, if I compare two representatives like   
```
AutomapError::TemporaryMappingError(String::new())
AutomapError::SocketReceiveError(AutomapErrorCause::Unknown("Booga".to_string()))
```
...I must be wondering why either, in the first case, it isn't `AutomapError::TemporaryMappingError(AutomapErrorCause::TemporaryDisfunction(String::new()))` 
or reversely, why the second error isn't `AutomapError::SocketReceiveError("Booga".to_string())`

Consistency is what caused me to write the comment. It brings me to say. anyway, that if this works good, it probably would be too much to require a revamp in the sake of consistency. On the other hand after I had a look cross all the variants it is obvious that the style of the first example has been adopted in the vast majority of them and so maybe it still could be a good idea to find the few of them which stand out of the line and you can make them look similar to the majority. That means, to react in the same direction as my calling up here was in.

automap/src/comm_layer/mod.rs
```diff
@@ -295,46 +293,46 @@ mod tests {
...
             (
-                AutomapError::SocketBindingError(
-                    "".to_string(),
-                    SocketAddr::from_str("0.0.0.0:0").unwrap(),
-                ),
+                AutomapError::PacketParseError(ParseError::WrongVersion(0)),
                 true,
             ),
+            (AutomapError::PermanentLeasesOnly, false),
+            (AutomapError::PermanentMappingError("".to_string()), true),
```
bertllll, line 310: Just a picking up a question, don't look the answer in my words. If you chose to attribute the `HousekeeperCrashed` error with `false`, then philosophically, what's the difference from an error like `PermanentMappingError(...)` or even a more interesting example. I noticed you have here these:
```
(AutomapError::PermanentMappingError("".to_string()), true),
...and...
(AutomapError::TemporaryMappingError("".to_string()), false),
```
I could attempt to interpret it as "if it's a temporary mapping, we might try again and again and we try the permanent one out only a single time" but that doesn't make sense so much. I'd be more open for a criticizing interpretation saying that `TemporaryMappingError` is very likely about to cause a cut-off of our Node since the last mapping has expired and therefore it has attributes for which might qualify as the opposite, with `true`. 
(Again, it's late here, and I'm having hard time to keep eyes open. Consider this more as food for thoughts than well formulated argumentation)

automap/src/comm_layer/pmp.rs
dnwiebe: Good, good...

automap/src/comm_layer/pmp.rs
dnwiebe: I like it...

masq_lib/src/utils.rs
dnwiebe: Sounds good.

node/src/bootstrapper.rs
dnwiebe: I decided to leave the decision for another day. 499 is already too big.

node/src/database/db_initializer.rs
dnwiebe: I didn't understand you objection, but removing the directive didn't seem to hurt.

node/src/sub_lib/route.rs
dnwiebe: I don't think it's there anymore.

node/src/test_utils/persistent_configuration_mock.rs
dnwiebe: I don't feel like tracking it down, but I _hope_ we don't accept spaces in there.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: No, it means...hmm. Okay: you're Node A, and you have neighbors B and C. Node B sends you Gossip including information about both itself and C. B is the transmitting neighbor, and C is a non-transmitting neighbor.

If B accepts connections but does not transmit its IP address, that's a problem. However, if C accepts connections, it's fine for B not to disclose C's IP address.

I can't come up with a better way to communicate that in the test name; I'll put it in a comment.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: Actually, the database _did_ change (we changed the NodeAddr of the transmitting Node), but it's not a change that should trigger a Gossip broadcast.

Remember, this change is because of a comment you made in the last pass.

node/src/neighborhood/gossip_acceptor.rs
dnwiebe: It's red code, right? I think it's already deleted.

multinode_integration_tests/src/main.rs
dnwiebe: Good suggestion.

multinode_integration_tests/src/main.rs
dnwiebe: Fixed

multinode_integration_tests/src/main.rs
dnwiebe: Fixed

multinode_integration_tests/src/main.rs
dnwiebe: Yes: there's a bunch of code missing. Again, this is for that multinode test that I haven't completed yet. I didn't want to delay the GH-499 review until it was done, so I started it early. More code will be coming.

multinode_integration_tests/src/main.rs
dnwiebe: I did it...

multinode_integration_tests/src/main.rs
dnwiebe: Okay, I'll try it. I'm handicapped by the fact that even after years of Rust, I still don't understand exactly what `'static` means. I get the feeling that almost nobody _actually_ understands it, but most people go ahead and use it anyway. Maybe I'm wrong about that.

multinode_integration_tests/src/main.rs
dnwiebe: Okay...

multinode_integration_tests/src/main.rs
dnwiebe: ...okay...

multinode_integration_tests/src/main.rs
dnwiebe: Okay, if you like...

multinode_integration_tests/src/main.rs
dnwiebe: This has become moot.

multinode_integration_tests/src/main.rs
dnwiebe: Okay.

multinode_integration_tests/src/main.rs
dnwiebe: I kind of like consuming the pieces; it seems more Rust-like than letting them lie around and referencing them.

multinode_integration_tests/src/main.rs
dnwiebe: I got rid of it. We'll see what happens.

multinode_integration_tests/src/mock_router.rs
dnwiebe: Yes: it's the test I'm still working on.

node/src/actor_system_factory.rs
dnwiebe: Okay, but it's not really meaningful in this case, as you'll see...

node/src/actor_system_factory.rs
dnwiebe: I think it's because in at least one case there isn't any prefix--by which I mean the prefix is `""`-- and leaving out the space means that the error message won't begin with a space in that case.

node/src/actor_system_factory.rs
dnwiebe: I think it's obsolete--overtaken by events. I got rid of it.

node/src/neighborhood/neighborhood_database.rs
dnwiebe: Yeah, I kind of like these functions to be defined right here near where they're used.

node/src/test_utils/persistent_configuration_mock.rs
bertllll: Yeah. You did right. I talked silly. First of all, it doesn't seem realistic the user has any official channel to set this parameter in the database... and so I probably was thinking of something else while now it seems clear to me. Never mind.

node/src/neighborhood/gossip_acceptor.rs
bertllll: Oh... silly me.

node/src/neighborhood/gossip_acceptor.rs
bertllll: I'm still not so calm about the assertions, and don't know if you thought about them. 

Why does the record contain no NodeAddr? I'm coming to an explanation that according to the real world it should not contain it also before the test fires off. If yes, then I don't understand your point saying that we changed the NodeAddr of the Node which sent this to us. If it should have been None before (because it also was running in this same mode) and is None now still, then we did not change any NodeAddr.

Also, what do you take from my question whether we actually consider this guy an ongoing full neighbor of ours because we just assured ourselves about having a TCP connection with the guy (if not dropped from them - anyway, we can design it so that they won't do it) and so we know all unknowns, eh, which is just the address, they also initiated this connection, so we don't break rules that they want to be OriginateOnly and so we can start using a route through this updated link. 

That's why I was thinking why is that we don't make the stream handler pool involved. Because, not knowing all details, I suppose it participate in some kind of reestablishment of the existing stream keys previously created for the contacting Node.

multinode_integration_tests/src/main.rs
bertllll: That's legit. :-)

node/src/actor_system_factory.rs
bertllll: Yeah, I just hope the true prefixes (in the prod code, saying something real) ends with a space. Would look better definitely.

node/src/neighborhood/gossip_producer.rs
dnwiebe: Okay...

multinode_integration_tests/src/masq_mock_node.rs
dnwiebe: Actually, they _are_ node args...just _mock_ node args. But I can make that clearer...

multinode_integration_tests/src/masq_mock_node.rs
dnwiebe: True, true...

multinode_integration_tests/src/main.rs
bertllll: With `str`s, the interpretation of `'static` is easy. It de facto means the string is firmly determined for the whole life span of the program. It is hard-coded.

In other situations, especially with its concurrences in type boundaries, it translates to me as "the type cannot be referenced. It must be passed in together with the ownership. :shrug:  

The hardest interpretations comes for me with trait objects, which by default has the `'static` quality if you don't specify elsehow. The fun starts with the `&dyn` and `&mut dyn` also implement it, talking about the desuggared syntax looking like `&dyn MyTrait + 'static` - that's the default. However if in your trait object, in the concrete struct hiding behind is a reference (it has a referenced field) then the trait over it cannot implement `'static` but you must introduce a specific limited lifetime say of the name `'a` like here `&dyn MyTrait + 'a`. 

Yeah, sometimes it can be a tough fight but I feel like it always has a solution.    


Otherwise consider this conversation just a side track. No issues. It should be resolved again.

multinode_integration_tests/tests/ip_change_test.rs
dnwiebe: This test isn't complete yet. When it is, these comments will probably be gone.

multinode_integration_tests/tests/ip_change_test.rs
dnwiebe: This test isn't complete yet. When it is, these comments will probably be gone.

automap/src/comm_layer/mod.rs
dnwiebe: With all the `String`s in its values? No, I don't want a bunch of unnecessary `String` copies. I'll leave it the way it is. But thanks for the suggestion!

automap/src/comm_layer/mod.rs
dnwiebe: It's a good catch that a `TemporaryMappingError` should stimulate more-often-than-normal retries, but A) that's not trivial, B) this card is already pretty big, and C) we don't have any idea yet how common `TemporaryMappingError`s are in real life. I made a card about that (GH-723) that we can play if they turn out to be a real problem.

automap/src/comm_layer/mod.rs
dnwiebe: I guess I don't understand your objection. Maybe I can get clarification in person on one of our daily calls.

automap/src/comm_layer/mod.rs
dnwiebe: I looked through the code. At the moment, it seems like the main reason we get `SocketBindingError`s is that there's already a Node running, and we can't open a socket because one is already open.

However, it seems as though that problem should have been solved by bringing in the `socket2` library. I'll look at it.

dnwiebe: It looks like with `socket2` in, the only thing that could reasonably cause a `SocketBindingError` would be some sort of mistake by the developers; so I switched this to `AutomapErrorCause::Unknown` and removed `AutomapErrorCause::UserError`.

automap/src/comm_layer/mod.rs
dnwiebe: No, I think this is appropriate. The `ProbeRequestError` is a higher-level error that just says requesting the probe failed, without saying why. The `cause` field has a lower level of information, and the `String` field has the lowest level.

Eventually, of course, I think all this probe stuff will be removed, once we're confident that Automap works, so I don't want to put a lot of effort in on it anyway.

automap/src/comm_layer/mod.rs
dnwiebe: Not sure I completely understand this, but yes: the Housekeeping thread crashing means nothing at all until the ISP changes the IP address. Lots of our customers will have ISPs that don't change their IP addresses; if some random thing prevents the Housekeeping thread from running for them, we shouldn't crash their Nodes.

PR #307, opened 2023-06-23
GH-719: Replace BlockchainInterfaceClandestine with BlockchainInterfaceNull

bertllll:


node/src/blockchain/blockchain_interface.rs
```diff
@@ -187,7 +179,10 @@ impl BlockchainInterface for BlockchainInterfaceClandestine {
         _start_block: u64,
         _recipient: &Wallet,
     ) -> Result<RetrievedBlockchainTransactions, BlockchainError> {
-        let msg = "Can't retrieve transactions clandestinely yet".to_string();
+        let msg = "\
+        Trying to retrieve transactions with uninitialized blockchain interface. \
+        The blockchain service URL was not supplied."
```
dnwiebe: "set" would be more consistent than "supplied."

node/src/blockchain/blockchain_interface.rs
```diff
@@ -156,28 +155,21 @@ pub trait BlockchainInterface<T: Transport = Http> {
...
 
-impl Default for BlockchainInterfaceClandestine {
-    fn default() -> Self {
-        Self::new(DEFAULT_CHAIN)
-    }
-}
-
-impl BlockchainInterface for BlockchainInterfaceClandestine {
+impl BlockchainInterface for BlockchainInterfaceNull {
     fn contract_address(&self) -> Address {
```
dnwiebe, line 204: Where is this method used? It's possible that `BlockchainInterfaceNull` has no use for a `Chain`, and its constructor could be simplified and this method modified to either panic or return a sentinel `Address`, like 0 or 0xFFFFF... or something.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -200,38 +195,55 @@ impl BlockchainInterface for BlockchainInterfaceClandestine {
...
+        The blockchain service URL was not set."
+        );
         Err(PayableTransactionError::Sending {
-            msg: "invalid attempt to send txs clandestinely".to_string(),
+            msg: "invalid attempt to send transactions".to_string(),
             hashes: vec![],
         })
     }
 
     fn get_gas_balance(&self, _address: &Wallet) -> ResultForBalance {
```
dnwiebe: Would it be expensive to change the name of this method to something like `get_transaction_fee_balance` so that it was already more generic?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -200,38 +195,55 @@ impl BlockchainInterface for BlockchainInterfaceClandestine {
...
-        error!(self.logger, "Can't get eth balance clandestinely yet",);
-        Ok(0.into())
+        error!(
+            self.logger,
+            "Trying to get eth balance with uninitialized blockchain interface. \
```
dnwiebe: Maybe "transaction fee balance" instead of "eth balance?"

node/src/blockchain/blockchain_interface.rs
```diff
@@ -200,38 +195,55 @@ impl BlockchainInterface for BlockchainInterfaceClandestine {
...
+        The blockchain service URL was not set."
+        );
+        Ok(U256::zero())
```
dnwiebe: Philosophically, it is an error to try to get the consuming-wallet balance without a blockchain service URL. If we return zeros from these two methods, will we fool other code into doing stupid things because it thinks there's a real wallet with a real zero balance? Or are those "stupid" things actually not so stupid? Do they turn out to be exactly the things that should be done when you're consuming services without a blockchain service URL?

node/src/blockchain/blockchain_interface.rs
```diff
@@ -868,7 +877,7 @@ mod tests {
     }
 
     #[test]
-    fn blockchain_interface_non_clandestine_retrieves_transactions() {
+    fn blockchain_Interface_web3_retrieves_transactions() {
```
dnwiebe: Several of these test methods have `Interface` beginning with a capital I.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -157,34 +169,55 @@ pub trait BlockchainInterface<T: Transport = Http> {
...
             logger: Logger::new("BlockchainInterfaceNull"),
-            chain,
         }
     }
+
+    fn look_out_am_null<O, E>(&self, operation: &str) -> Result<O, E>
+    where
+        E: BlockchainInterfaceUninitializedError,
+    {
+        self.log_null_translates_as_bi_uninitialized(operation);
```
dnwiebe: "bi" is a prefix that means "two." For example: "biweekly" means "every two weeks." That makes these names a little hard to understand. I know that "blockchain_interface" is quite a bit longer than "bi," but maybe you can find a happy medium.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -90,35 +98,39 @@ pub enum PayableTransactionError {
...
+            )),
+            Self::Signing(msg) => Right(format!("Signing phase: \"{}\"", msg)),
+            Self::Sending { msg, hashes } => Right(format!(
                 "Sending phase: \"{}\". Signed and hashed transactions: {}",
                 msg,
                 comma_joined_stringifiable(hashes, |hash| format!("{:?}", hash))
-            ),
-        }
+            )),
+            Self::UninitializedBlockchainInterface => Left("Uninitialized blockchain interface"),
```
dnwiebe: Maybe this message should mention the `blockchain-service-url` parameter, since that's what the user will understand.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -157,34 +169,55 @@ pub trait BlockchainInterface<T: Transport = Http> {
...
+        E: BlockchainInterfaceUninitializedError,
+    {
+        self.log_null_translates_as_bi_uninitialized(operation);
+        Err(E::bi_uninitialized_error())
+    }
+
+    fn log_null_translates_as_bi_uninitialized(&self, operation: &str) {
+        error!(
+            self.logger,
+            "Trying to {} with uninitialized blockchain interface. The blockchain \
```
dnwiebe: A specific reference to the `blockchain-service-url` parameter might do better here, especially for users who aren't familiar enough with the CLI to immediately recognize the concept of the service URL as something provided on the command line.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -157,34 +169,55 @@ pub trait BlockchainInterface<T: Transport = Http> {
...
+}
+
+impl Default for BlockchainInterfaceNull {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+trait BlockchainInterfaceUninitializedError {
+    fn bi_uninitialized_error() -> Self;
```
dnwiebe: Here's "bi" again.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -157,34 +169,55 @@ pub trait BlockchainInterface<T: Transport = Http> {
...
         _start_block: u64,
         _recipient: &Wallet,
     ) -> Result<RetrievedBlockchainTransactions, BlockchainError> {
-        let msg = "\
-        Trying to retrieve transactions with uninitialized blockchain interface. \
-        The blockchain service URL was not supplied."
-            .to_string();
-        error!(self.logger, "{}", &msg);
-        Err(BlockchainError::QueryFailed(msg))
+        self.look_out_am_null("retrieve transaction")
```
dnwiebe: Should be "transactions" plural.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2331,18 +2337,20 @@ mod tests {
             BlockchainError::QueryFailed(
                 "Don't query so often, it gives me a headache".to_string(),
             ),
+            BlockchainError::UninitializedBlockchainInterface,
         ];
         let pretty_print_closure = |err_to_resolve: &BlockchainError| match err_to_resolve {
             BlockchainError::InvalidUrl => (err_to_resolve.to_string(), 11),
```
dnwiebe: It's not clear to me that these check numbers are necessary. I don't object to additional assertions, but if they're there they should be simple and easy to understand.

You can take a look at the `enum_iterator` crate, but I'm thinking it may be a case where the juice isn't worth the squeeze.

https://docs.rs/enum-iterator/latest/enum_iterator

node/src/blockchain/blockchain_interface.rs
```diff
@@ -72,14 +73,17 @@ impl Display for BlockchainError {
             InvalidAddress => Left("Invalid address"),
             InvalidResponse => Left("Invalid response"),
             QueryFailed(msg) => Right(format!("Query failed: {}", msg)),
-            UninitializedBlockchainInterface => Left("Uninitialized blockchain interface. The parameter blockchain-service-url was not set")
+            UninitializedBlockchainInterface => Left(
+                "The parameter blockchain-service-url must be \
+            set for scanners to function properly",
```
dnwiebe: "We can't contact the blockchain to make or receive payments because the blockchain-service-url was not specified. As a result, this Node may be delinquency-banned."

"The parameter blockchain-service-url was not set, but must be set in order to make or receive payments."

"The value for blockchain-service-url is missing, so we can't make or receive payments. You may be delinquency-banned soon."

"To avoid being delinquency-banned, you should restart the Node with a value for blockchain-service-url."

node/src/blockchain/blockchain_interface.rs
```diff
@@ -122,14 +126,17 @@ impl Display for PayableTransactionError {
                 msg,
                 comma_joined_stringifiable(hashes, |hash| format!("{:?}", hash))
             )),
-            Self::UninitializedBlockchainInterface => Left("Uninitialized blockchain interface"),
+            Self::UninitializedBlockchainInterface => Left(
+                "The parameter blockchain-service-url \
+            must be set for scanners to function properly",
```
dnwiebe: The value of this message should be informed by the value chosen for the message above.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -178,19 +185,19 @@ impl BlockchainInterfaceNull {
...
+        Err(E::interface_uninitialized_error())
     }
 
-    fn log_null_translates_as_bi_uninitialized(&self, operation: &str) {
+    fn log_interface_uninitialized(&self, operation: &str) {
         error!(
             self.logger,
-            "Trying to {} with uninitialized blockchain interface. The blockchain \
-        service URL was not set.",
+            "Trying to {} with uninitialized blockchain interface. Parameter blockchain-service-url \
```
dnwiebe: "Failed to {}" would be better than "Trying to {}", because the latter isn't specific about what happened as a result.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2339,30 +2321,34 @@ mod tests {
...
+            .map(|err| err.to_string())
+            .collect::<Vec<_>>();
 
+        let original_errors_len = original_errors.len();
+        assert_eq!(
+            original_errors_len,
+            BlockchainError::VARIANT_COUNT,
+            "you forgot to add all variants in this test"
+        );
+        let check_hash_set: HashSet<BlockchainError> = HashSet::from_iter(original_errors);
```
dnwiebe: I think this check is unnecessary, and I can't think of any way to force the future developer to do what we want here.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -2385,24 +2371,24 @@ mod tests {
...
+            .map(|err| err.to_string())
+            .collect::<Vec<_>>();
 
+        let original_errors_len = original_errors.len();
+        assert_eq!(
+            original_errors_len,
+            PayableTransactionError::VARIANT_COUNT,
+            "you forgot to add all variants in this test"
+        );
+        let check_hash_set: HashSet<PayableTransactionError> = HashSet::from_iter(original_errors);
```
dnwiebe: Here too

dnwiebe (approved):

PR #311, opened 2023-07-02
GH-499 Review 4

dnwiebe:


automap/src/comm_layer/pmp.rs
```diff
@@ -1916,12 +1918,14 @@ mod tests {
                 remap_interval: Duration::from_secs(300)
             }
         );
-        TestLogHandler::new().exists_log_containing("INFO: timed_remap_test: Remapping port 6689");
+        TestLogHandler::new()
+            .exists_log_containing(&format!("INFO: {}: Remapping port 6689", test_name));
```
bertllll, line 1922: This case and many other ones are good, I don't mind approving it, although I consider the Utkarshe's style in its full purity slightly better because he take advantage of the chance of inlining the variable to be printed inside the placeholder that sets where it should be printed.  

`INFO: {test_name}: Remapping port 6689"` 

It feels like a decently applied symbiosis with the new changes that have been introduced in Rust recently. 

Don't do it if you feel more comfortable this way.

masq_lib/src/utils.rs
```diff
@@ -190,10 +189,10 @@ where
...
     }
 }
-
-pub fn to_millis(dur: &Duration) -> u64 {
-    (dur.as_secs() * 1000) + (u64::from(dur.subsec_nanos()) / 1_000_000)
-}
+//
+// pub fn to_millis(dur: &Duration) -> u64 {
+//     (dur.as_secs() * 1000) + (u64::from(dur.subsec_nanos()) / 1_000_000)
+// }
```
bertllll, line 195: Please get rid of this code totally.

multinode_integration_tests/src/main.rs
```diff
@@ -411,22 +407,22 @@ struct PortSpec {
 }
 
 impl FromStr for PortSpec {
-    type Err = String;
+    type Err = &'static str;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         if s.len() < 2 {
-            return Err("Syntax: <U|T><port>".to_string());
+            return Err("Syntax: <U|T><port>");
```
bertllll, line 414: Given you have deployed the change where a repetitively used msg was converted into a manifest constant (see ln. 351) it would be nice, in the sake of consistency, to have it here too.

node/src/database/db_initializer.rs
```diff
@@ -624,16 +608,35 @@ impl Debug for DbInitializationConfig {
 #[cfg(test)]
```
bertllll, line 608: Hmmm. It causes me pain making such a note but I believe this is breaking a pattern so we should move this whole module `test_utils` from here to a place like `node/src/database/test_utils` if one exists. Otherwise we ought to make one now and do it then.

node/src/sub_lib/node_addr.rs
```diff
@@ -165,6 +132,39 @@ impl FromStr for NodeAddr {
     }
 }
 
+impl NodeAddr {
+    pub const PORTS_SEPARATOR: &'static str = "/";
+
+    pub fn new(ip_addr: &IpAddr, ports: &[u16]) -> NodeAddr {
+        let mut ports = ports.to_owned();
+        ports.sort_unstable();
```
bertllll, line 140: An interesting name...

I bet it's a completely innocent function though. :)

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -4304,6 +4304,12 @@ mod tests {
     #[test]
     fn standard_gossip_containing_non_transmitting_neighbor_node_that_accepts_connections_but_has_no_node_addr_is_accepted(
     ) {
+        // src_root, which accepts connections, is about to tell dest_root about non_malefactor,
+        // which also accepts connections. The NodeRecord for non_malefactor will not contain its
+        // NodeAddr. It's a violation for src_root to claim to accept connections but not to
+        // provide its own NodeAddr, but this test ensures that it's okay for src_root not to
+        // provide the NodeAddr of non_malefactor, since non_malefactor is not the transmitter of
+        // this Gossip.
```
bertllll, line 4312: Oh thanks, this comment makes a lot of sense. It's bulky but I hope we can afford such an aide from time to time.

automap/src/comm_layer/pmp.rs
dnwiebe: ...okay...

multinode_integration_tests/src/main.rs
dnwiebe: Okay.

node/src/sub_lib/node_addr.rs
dnwiebe: A sort is stable if it preserves the order of equal elements. For example, if you have Stanley and Denise, both of whom are 23 years old, and you sort them by age, and you get back Stanley and Denise, then the sort is stable. If you get back (or _might_ get back) Denise and Stanley, then the sort is unstable.

With some sort algorithms--apparently including the one Rust's library collections use--unstable sorts are cheaper than stable sorts; so you choose unstable sorts unless you have a special need for the stable kind.

node/src/database/db_initializer.rs
dnwiebe: Okay, I did it.

masq_lib/src/utils.rs
dnwiebe: Sure.

PR #322, opened 2023-07-25
GH-499 review 5 head

dnwiebe:


masq_lib/src/shared_schema.rs
```diff
@@ -367,6 +367,15 @@ pub fn shared_app(head: App<'static, 'static>) -> App<'static, 'static> {
...
             .hidden(true),
     )
+    .arg(
+        Arg::with_name("fake-router-ip")
+            .long("fake-router-ip")
+            .value_name("FAKE-ROUTER-IP")
+            .required(false)
+            .min_values(0)
+            .max_values(1)
+            .hidden(true),
```
bertllll, line 377: You said something about mentioning this hidden feature in a README file. You'll know better than me where the appropriate file is. 

If you decide to do it I think the other hidden parameters also don't take any space in the docs but would be nice to have a place where we can read about it. Well, you doubted that we need it. So decide yourself.

multinode_integration_tests/src/main.rs
```diff
@@ -229,10 +230,7 @@ impl DataProbe {
     }
 
     fn usage(stderr: &mut dyn Write) -> u8 {
-        writeln! (stderr, "Usage: DataProbe <IP address>:<U|T><port>/<U|T><port>/... where <IP address> is the address DataProbe is running on, U means UDP and T means TCP, and <port> is between {} and {}",
-            LOWEST_USABLE_INSECURE_PORT,
-            HIGHEST_USABLE_PORT,
-        ).unwrap ();
+        writeln! (stderr, "{DATAPROBE_USAGE}").unwrap ();
```
bertllll, line 233: unwrap() in prod code...

multinode_integration_tests/src/utils.rs
```diff
@@ -125,6 +126,64 @@ pub fn open_all_file_permissions(dir: PathBuf) {
     }
 }
 
+pub fn do_docker_run(
+    ip_addr: IpAddr,
+    host_node_parent_dir: Option<String>,
+    name: &str,
+    program_args: Vec<String>,
```
bertllll, line 133: maybe `data_probe_args` rather?

node/src/actor_system_factory.rs
```diff
@@ -568,10 +566,10 @@ pub struct AutomapControlFactoryReal {}
 impl AutomapControlFactory for AutomapControlFactoryReal {
     fn make(
         &self,
-        usual_protocol_opt: Option<AutomapProtocol>,
+        automap_config: AutomapConfig,
         change_handler: ChangeHandler,
     ) -> Box<dyn AutomapControl> {
-        Box::new(AutomapControlReal::new(usual_protocol_opt, change_handler))
+        Box::new(AutomapControlReal::new(automap_config, change_handler))
```
bertllll, line 572: If you feel it's actually untested code, but for a good reason, I'd prefer a comment here.

node/src/actor_system_factory.rs
```diff
@@ -1629,6 +1626,49 @@ mod tests {
         );
     }
 
+    #[test]
+    fn handle_automap_error_delegates_correctly_to_factory() {
```
bertllll, line 1630: I'm a bit confused when I read a function name at the beginning of the test name and then I find the Act to be something else. Could it be made clearer?

node/src/actor_system_factory.rs
```diff
@@ -1629,6 +1626,49 @@ mod tests {
...
+            AutomapConfig::new(Some(mapping_protocol), Some(fake_router_ip)),
+            vec![new_ip_recipient_sub],
+        );
+
+        let mut make_params = make_params_arc.lock().unwrap();
+        // Make sure the AutomapConfig was passed correctly
+        let (automap_config, change_handler) = make_params.remove(0);
+        assert_eq!(
+            automap_config,
+            AutomapConfig::new(Some(AutomapProtocol::Pcp), Some(fake_router_ip))
```
bertllll, line 1653: I like more if you assign the data going in at the Act to a variable and later when you're trying to assert on it you can call it again and use it as the expected data.

node/src/actor_system_factory.rs
```diff
@@ -1629,6 +1626,49 @@ mod tests {
...
+        );
+        // Make sure the generated change_handler handles new IP addresses properly
+        let new_ip = IpAddr::from_str("1.2.3.4").unwrap();
+        let system = System::new("handle_automap_error_delegates_correctly_to_factory");
```
bertllll, line 1657: You'll do better with the test_name variable used multiple times here.

node/src/database/mod.rs
```diff
@@ -4,3 +4,5 @@ pub mod config_dumper;
 pub mod connection_wrapper;
 pub mod db_initializer;
 pub mod db_migrations;
+#[cfg(test)]
```
bertllll, line 7: I'm always fighting with this myself. I don't know exactly what has a better impact on proper division between the prod and test tree and the resulting binary files. The question is should we use the attribute in a place like this one, meaning at the declaration of use of another mod (which is a file in this case) or should we leave this alone and go to the thought file and add in the global attribute (#![cfg(test)]) at the top of it. There is also the third option to use it in both places but I like it probably the least. 

I'm just not sure what happens with a file where only a reference to it (the `mod` manifest) in a different file is disabled. Does it also mean the all the code within the file will be disregarded? Cannot it mean the file still is part of the binaries with the code present but never used?

node/src/database/test_utils.rs
```diff
@@ -0,0 +1,129 @@
...
+        init_config: DbInitializationConfig,
+    ) -> Result<Box<dyn ConnectionWrapper>, InitializationError> {
+        self.initialize_params
+            .lock()
+            .unwrap()
+            .push((path.to_path_buf(), init_config));
+        self.initialize_results.borrow_mut().remove(0)
+    }
+
+    #[allow(unused_variables)]
```
bertllll, line 95: You can get rid of this attribute if you do what we normally do and that is to put underscore before all the args.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -257,6 +258,14 @@ fn handle_fake_public_key_if_supplied(
...
 }
 
+fn handle_fake_router_ip_if_supplied(
+    multi_config: &MultiConfig,
+    privileged_config: &mut BootstrapperConfig,
+) {
+    privileged_config.automap_config.fake_router_ip_opt =
+        value_m!(multi_config, "fake-router-ip", IpAddr);
+}
+
```
bertllll, line 268: Make sure this function is covered with tests. I cannot see a test out here so it raised an imaginary flag at me. :-)

masq_lib/src/shared_schema.rs
```diff
@@ -367,6 +367,15 @@ pub fn shared_app(head: App<'static, 'static>) -> App<'static, 'static> {
             .max_values(1)
             .hidden(true),
     )
+    .arg(
+        Arg::with_name("fake-router-ip")
```
bertllll, line 371: I've once attempted (on your advice) to write tests that would verify some of the attributes set on the parameter in the Clap schema. Here it is marginal above all because it is a test parameter...

node/src/test_utils/automap_mocks.rs
```diff
@@ -57,6 +54,7 @@ impl AutomapControlFactoryMock {
     }
 }
 
+#[derive(Clone)]
```
bertllll, line 57: Cloning mocks tastes fishy. Do you really need that? I always try as hard as possible to avoid this. 

If you're cloning it some of your tests, maybe a scenario test, you could have a closure producing this, that you could call multiple times to produce the same subject any times you need.

masq_lib/src/shared_schema.rs
dnwiebe: I documented the test-only parameters in help strings, even though help is never printed; I figured that was probably appropriate for developers.

multinode_integration_tests/src/main.rs
dnwiebe: Aha! No it's not: it's in `multinode_integration_tests`! Not production code. I missed that when you were reviewing live; sorry.

multinode_integration_tests/src/utils.rs
dnwiebe: Okay.

node/src/actor_system_factory.rs
dnwiebe: Okay.

node/src/actor_system_factory.rs
dnwiebe: Okay.

node/src/actor_system_factory.rs
dnwiebe: Okay.

node/src/database/mod.rs
dnwiebe: I suppose I could move it...

node/src/actor_system_factory.rs
dnwiebe: Way ahead of you.

node/src/database/test_utils.rs
dnwiebe: Huh. I don't remember writing that method. Oh well...

node/src/node_configurator/node_configurator_standard.rs
dnwiebe: Okay.

masq_lib/src/shared_schema.rs
dnwiebe: Yeah...it's not clear to me at this point what value bringing the attributes of the `fake-router-ip` under test would have, since it's test code itself.

node/src/test_utils/automap_mocks.rs
dnwiebe: Nope, don't need it.

PR #325, opened 2023-08-02
Gh 580 review 5 head

czarte:


masq/src/commands/setup_command.rs
```diff
@@ -206,6 +206,7 @@ mod tests {
                 running: false,
                 values: vec![
                     UiSetupResponseValue::new("chain", "eth-ropsten", Configured),
+                    UiSetupResponseValue::new("data-directory", "/home/booga", Default),
```
dnwiebe, line 209: Can you put `/eth-ropsten` on the end of `/home/booga` so that in the assertion it looks as though the data directory really has been modified the way the NOTE claims ti has been?

Also, we're no longer using ropsten for anything, and it will soon be pulled entirely out of the codebase. I realize that doesn't matter to this test, but future developers might have no idea what it is.

masq/src/commands/setup_command.rs
```diff
@@ -279,6 +281,7 @@ scans                         off
                 running: true,
                 values: vec![
                     UiSetupResponseValue::new("chain", ETH_ROPSTEN_FULL_IDENTIFIER, Set),
+                    UiSetupResponseValue::new("data-directory", "/home/booga", Set),
```
dnwiebe, line 284: Here too: if you could put the chain on the end of this data directory, the assertion doesn't look like a lie.

node/src/daemon/setup_reporter.rs
```diff
@@ -3631,17 +3631,19 @@ mod tests {
         let _guard = EnvironmentGuard::new();
         let existing_setup =
             setup_cluster_from(vec![("real-user", "1111:1111:/home/booga", Default)]);
+        let masqhome = Path::new("/home/booga/masqhome");
+        let expected = &masqhome.join("polygon-mainnet");
```
dnwiebe, line 3635: In general, `expected` values are considered part of the Assert portion of the test, not the Arrange portion. Sometimes, `expected` values are used in the Act portion of the test, and when that happens, they need to be created in the Arrange portion; but when they're _not_ used in Act, the test is cleaner if they're confined to the Assert portion.

node/src/daemon/setup_reporter.rs
```diff
@@ -3631,17 +3631,19 @@ mod tests {
...
         let incoming_setup = vec![UiSetupRequestValue::new(
             "data-directory",
-            "/home/booga/masqhome",
+            masqhome.to_str().unwrap(),
         )];
         let dirs_wrapper = Box::new(DirsWrapperReal);
         let subject = SetupReporterReal::new(dirs_wrapper);
 
         let result = subject.get_modified_setup(existing_setup, incoming_setup);
         assert_eq!(
```
dnwiebe, line 3644: Standard form would put a blank line before this assert.

PR #326, opened 2023-08-08
GH-651: Automatic Retries for the URLs in case of DNS Resolve Failure

Syther007:


multinode_integration_tests/src/neighborhood_constructor.rs
```diff
@@ -55,6 +56,7 @@ pub fn construct_neighborhood(
     cluster: &mut MASQNodeCluster,
     model_db: NeighborhoodDatabase,
     additional_keys_to_mock: Vec<&PublicKey>,
+    min_hops: Hops,
```
dnwiebe: It'd be more general to put in here a closure capable of modifying the `NodeStartupConfigBuilder`, and then call it below after creating the `NodeStartupConfigBuilder` off of `.standard()`. If you did so, and then looked at the places the compiler got excited about, you might even find some existing tests that you could severely simplify.

node/src/dispatcher.rs
```diff
@@ -30,8 +30,8 @@ lazy_static! {
 struct DispatcherOutSubs {
     to_proxy_server: Recipient<InboundClientData>,
```
dnwiebe: I think it's probably a good idea to change some names in here, because the existing names are inconsistent, with some of them ending with `_sub` and some not; this is especially annoying after Actix changed the name from `Subscriber` to `Recipient`, so that we have both `_sub` and `_recipient` at various places in the code.

However, since you can send only one kind of message to each `Recipient`, I think it's reasonable to reference that message in the name of the field, especially in contexts like this one, where (as Bert pointed out) there are multiple `Recipient`s that lead to the same Actor.

node/src/proxy_server/server_impersonator_http.rs
```diff
@@ -20,25 +19,18 @@ impl ServerImpersonator for ServerImpersonatorHttp {
...
             "DNS Resolution Problem",
-            &format!("Exit Node couldn't resolve {}", quoted_server_name),
+            &format!("Exit Nodes couldn't resolve {}", quoted_server_name),
             &format!(
-                "We chose the exit Node {} for your request to {}; but when it asked \
-        its DNS server to look up the IP address for {}, it wasn't found. If {} exists, \
-        it will need to be looked up by a different exit Node. We've deprioritized this exit Node. \
-        Reload the page, and we'll try to find another.",
-                exit_key, server_name, server_name, server_name
+                "DNS Failure, We have tried multiple Exit Nodes and all have failed to resolve this address {}",
```
dnwiebe, line 32: Stimulate this error in a browser a couple of times, look at it, and make sure it's friendly enough for a real user to see. If it's not, adjust it until it is.

node/src/stream_handler_pool.rs
```diff
@@ -256,7 +256,7 @@ impl StreamHandlerPool {
             origin_port,
             ibcd_sub,
             remove_sub,
-            stream_shutdown_sub,
+            dispatcher_sub,
```
dnwiebe: I like the documentation here that the message is going to the Dispatcher Actor (presuming that it actually is, and this isn't a mistake), but it would also be nice to retain the documentation that it's supposed to be a Stream Shutdown message that uses this field to go to the Dispatcher.

node/src/stream_handler_pool.rs
```diff
@@ -310,6 +310,7 @@ impl StreamHandlerPool {
             .clone();
```
dnwiebe, line 309: I can't place this comment any higher in the file, but I think the `// TODO` right at the beginning of this method may already have been taken care of. Do you see any reference to `NodeQueryMessage` in here? I think it may actually have been deleted from the codebase by now.

It feels like the right thing to do is not to replace `NodeQueryMessage` with `DispatcherNodeQueryMessage`, because I'll bet the message now comes from more places than just the Dispatcher, but I do think it's the only node query message now. If it seems reasonable, try renaming `DispatcherNodeQueryMessage` to just `NodeQueryMessage`.

node/src/stream_handler_pool.rs
```diff
@@ -310,6 +310,7 @@ impl StreamHandlerPool {
...
         match msg.endpoint.clone() {
```
dnwiebe, line 310: Not your fault, but instinct says we probably don't need this `.clone()` here. It looks like it could result in a performance issue, especially if public keys are being unnecessarily copied a lot.

node/src/stream_handler_pool.rs
```diff
@@ -326,6 +327,7 @@ impl StreamHandlerPool {
                     .expect("Neighborhood is Dead")
             }
             Endpoint::Socket(socket_addr) => {
+                // The socket_addr can either be for the Neighbor or the browser
```
dnwiebe, line 329: This feels uncomfortable. Is it really true? If it is, then documenting it like this is a great thing to do; but it doesn't feel true to me. Of course, this is old code that I haven't been in for years, so actual contemporary investigation of the code beats a vague feeling I might have.

node/src/stream_reader.rs
```diff
@@ -22,7 +22,7 @@ pub struct StreamReaderReal {
     reception_port: Option<u16>,
     ibcd_sub: Recipient<dispatcher::InboundClientData>,
     remove_sub: Recipient<RemoveStreamMsg>,
-    stream_shutdown_sub: Recipient<StreamShutdownMsg>,
+    dispatcher_sub: Recipient<StreamShutdownMsg>,
```
dnwiebe: Again, it'd be nice to have both the "stream shutdown" and "dispatcher" pieces of documentation in this name.

node/src/proxy_server/mod.rs
```diff
@@ -162,12 +164,35 @@ impl AddReturnRouteMessage {
...
+                debug!(
+                    self.logger,
+                    "Found a new route for hostname: {:?} - stream key: {}  retries left: {}",
+                    dns_failure.unsuccessful_request.target_hostname,
+                    msg.stream_key,
+                    dns_failure.retries_left
+                );
+                self.stream_key_routes
+                    .insert(msg.stream_key, route_query_response);
+                // TODO Preserve all the previously know routes and halt the retry process if the new route is already present.
```
dnwiebe: Has this `// TODO` been done? If so, remove the comment...

node/src/proxy_server/mod.rs
```diff
@@ -162,12 +164,35 @@ impl AddReturnRouteMessage {
...
+                // In case we are exhausting retry counts over here Send a message to the browser.
+            }
+            Err(e) => {
+                error!(self.logger, "No route found for hostname: {:?} - stream key {} - retries left: {} - RouteResultMessage Error: {}",dns_failure.unsuccessful_request.target_hostname, msg.stream_key, dns_failure.retries_left, e);
```
dnwiebe: If this is going to be automatically retried, it probably shouldn't be an `error!` log. Save the `error!` log for the final failure after the retries are exhausted.

node/src/proxy_server/mod.rs
```diff
@@ -162,12 +164,35 @@ impl AddReturnRouteMessage {
...
 
-    fn handle(&mut self, msg: AddRouteMessage, _ctx: &mut Self::Context) -> Self::Result {
-        debug!(self.logger, "Establishing stream key {}", msg.stream_key);
-        self.stream_key_routes.insert(msg.stream_key, msg.route);
+    fn handle(&mut self, msg: RouteResultMessage, _ctx: &mut Self::Context) -> Self::Result {
+        let dns_failure = self
+            .dns_failure_retries
+            .get(&msg.stream_key)
+            .unwrap_or_else(|| {
+                panic!("RouteResultMessage Handler: stream key: {} not found within dns_failure_retries", msg.stream_key);
```
dnwiebe: Are we absolutely certain sure that this can't be triggered by anything that comes in from outside over the network? If it can, we need to log a message here and continue, rather than panicking.

dnwiebe (commented):
Resume at proxy_server/mod.rs:281

node/src/proxy_server/mod.rs
```diff
@@ -245,13 +271,72 @@ impl ProxyServer {
...
             add_return_route: recipient!(addr, AddReturnRouteMessage),
-            add_route: recipient!(addr, AddRouteMessage),
             stream_shutdown_sub: recipient!(addr, StreamShutdownMsg),
             set_consuming_wallet_sub: recipient!(addr, SetConsumingWalletMessage),
             node_from_ui: recipient!(addr, NodeFromUiMessage),
+            route_result_sub: recipient!(addr, RouteResultMessage),
+        }
+    }
+
+    fn pop_dns_failure_retry(&mut self, stream_key: &StreamKey) -> Result<DNSFailureRetry, String> {
```
dnwiebe: Since you're looking up (with a key) the element to remove, this isn't really the same operation as popping a stack, which is what the name would make people think of. Can you name this in a way that more immediately reveals its intention?

node/src/proxy_server/mod.rs
```diff
@@ -245,13 +271,72 @@ impl ProxyServer {
...
+        match self.dns_failure_retries.remove(stream_key) {
+            None => Err(format!(
+                "No entry found inside dns_failure_retries hashmap for the stream_key: {:?}",
+                stream_key
+            )),
+            Some(retry) => Ok(retry),
         }
     }
 
+    fn retry_dns_resolution(&mut self, mut retry: DNSFailureRetry, source_addr: SocketAddr) {
```
dnwiebe: Can you give `source_addr` a more descriptive name, so that we instantly know what it means upon seeing it?

node/src/proxy_server/mod.rs
```diff
@@ -245,13 +271,72 @@ impl ProxyServer {
...
+        let route_source = self.out_subs("Neighborhood").route_source.clone();
+        let proxy_server_sub = self.out_subs("ProxyServer").route_result_sub.clone();
+        let inbound_client_data_helper = self
+            .inbound_client_data_helper_opt
+            .as_ref()
+            .expect("IBCDHelper uninitialized");
+
+        inbound_client_data_helper.request_route_and_transmit(args, route_source, proxy_server_sub);
+        retry.retries_left -= 1;
+        self.dns_failure_retries.insert(stream_key, retry);
```
dnwiebe: There's a bit of asymmetry here. It seems as though the entity responsible for retrieving the `DNSFailureRetry` ought to be the entity responsible for putting it back. Maybe there's a good reason for this asymmetry, but it's not clear to me yet at this point in the review.

node/src/proxy_server/mod.rs
```diff
@@ -274,51 +359,49 @@ impl ProxyServer {
...
-                    .expect("Dispatcher is dead");
-                debug!(
-                    self.logger,
-                    "Retiring stream key {}: DnsResolveFailure", &response.stream_key
-                );
-                self.purge_stream_key(&response.stream_key);
+                let retry = match self.pop_dns_failure_retry(&response.stream_key) {
+                    Ok(retry) => retry,
+                    Err(error_msg) => {
+                        panic!("{}", error_msg);
```
dnwiebe: Again, a couple of things. If this condition can be created by an outside actor, we don't want a panic here: panics should only be caused by bad data (corrupted from inside, not outside) or bad developers.

If it _can't_ be created by an outside actor, then it makes sense to put the panic inside `pop_dns_failure_retry` so that we don't have to deal with it here.

node/src/proxy_server/mod.rs
```diff
@@ -274,51 +359,49 @@ impl ProxyServer {
...
+                    }
+                };
+                if retry.retries_left > 0 {
+                    self.retry_dns_resolution(retry, socket_addr);
```
dnwiebe: Okay, yeah. The `retry` object is clearly pulled out of the HashMap here in this method, and from this level it is not at all clear that `retry_dns_resolution` returns it to the HashMap. I think `retry_dns_resolution` ought to return the `retry` object passed into it after modification, and it ought to be put back into the HashMap right here in the caller.

node/src/proxy_server/mod.rs
```diff
@@ -851,10 +979,77 @@ pub trait IBCDHelper {
...
+        proxy_server_sub: Recipient<RouteResultMessage>,
+    );
+}
+
+trait RouteQueryResponseResolver: Send {
+    fn resolve_message(
+        &self,
+        args: TryTransmitToHopperArgs,
+        proxy_server_sub: Recipient<RouteResultMessage>,
+        route_result: Result<Option<RouteQueryResponse>, MailboxError>,
```
dnwiebe: The type of this parameter is more complicated than its name would imply. I'm not yet sure why it's a `Result`, but at least it seems as though the name should end with `_opt`.

node/src/proxy_server/mod.rs
```diff
@@ -851,10 +979,77 @@ pub trait IBCDHelper {
...
+    );
+}
+struct RouteQueryResponseResolverReal {}
+
+impl RouteQueryResponseResolver for RouteQueryResponseResolverReal {
+    fn resolve_message(
+        &self,
+        args: TryTransmitToHopperArgs,
+        proxy_server_sub: Recipient<RouteResultMessage>,
+        route_result: Result<Option<RouteQueryResponse>, MailboxError>,
```
dnwiebe: This `MailboxError` type seems as though it couples this method (and trait) a little too tightly to Actix. The only thing this method does with the `MailboxError` is convert it into a `String`. Could you have the caller of this method (which is already coupled to Actix error handling) do that conversion and pass in a `Result<Option<RouteQueryResponse>, String>`?

node/src/proxy_server/mod.rs
```diff
@@ -1020,6 +1197,35 @@ impl StreamKeyFactory for StreamKeyFactoryReal {
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct Hostname {
+    hostname: String,
+}
+
+impl Hostname {
+    #[allow(dead_code)]
+    fn new(raw_url: &str) -> Self {
```
dnwiebe, line 1217: Is `new` the best name for this constructor function? How about implementing `FromStr` and using `from_str` to make a `Hostname` out of a URL? Or maybe not implementing `FromStr` at all, but having the `impl Hostname` section have something called `from_url(raw_url: &str) -> Self` that does what `new()` does now?

node/src/proxy_server/mod.rs
```diff
@@ -1020,6 +1197,35 @@ impl StreamKeyFactory for StreamKeyFactoryReal {
...
+        let regex = Regex::new(
+            r"^((http[s]?|ftp):/)?/?([^:/\s]+)((/\w+)*/)([\w\-.]+[^#?\s]+)(.*)?(#[\w\-]+)?$",
+        )
+        .expect("Bad Regex");
+        let hostname = match regex.captures(raw_url) {
+            None => raw_url.to_string(),
```
dnwiebe, line 1223: Returning the entire URL if it doesn't match seems dangerous. Could you have another regex that you try if this one doesn't match to make sure it's an IP address? You might need _two_ more regexes to be IPv6 compatible.

dnwiebe (commented):
Resume at proxy_server/mod.rs:1230 green

node/src/proxy_server/mod.rs
```diff
@@ -162,12 +164,35 @@ impl AddReturnRouteMessage {
...
 
-impl Handler<AddRouteMessage> for ProxyServer {
+impl Handler<RouteResultMessage> for ProxyServer {
     type Result = ();
 
-    fn handle(&mut self, msg: AddRouteMessage, _ctx: &mut Self::Context) -> Self::Result {
-        debug!(self.logger, "Establishing stream key {}", msg.stream_key);
-        self.stream_key_routes.insert(msg.stream_key, msg.route);
+    fn handle(&mut self, msg: RouteResultMessage, _ctx: &mut Self::Context) -> Self::Result {
+        let dns_failure = self
```
bertllll, line 157: I might give it a shot and think of a better name. We're not creating the variable in a response to a DNS failure. Maybe `retries_info`?

node/src/proxy_server/mod.rs
```diff
@@ -162,12 +164,35 @@ impl AddReturnRouteMessage {
     }
 }
 
-impl Handler<AddRouteMessage> for ProxyServer {
+impl Handler<RouteResultMessage> for ProxyServer {
```
bertllll: RouteResultMessage?

To me AddRouteMessage had a clearer meaning. If the purpose of the message didn't change, and I suppose it did not, I'd rather see it be more obvious what the words "RouteResult" mean.

I'd like AddRouteResultMessage better.

Also I think the error in the Result is little fishy. It actually isn't a real error which we should fear of too much. When I saw it first time, though, I was like, oh, this must be a serious thing if we need to send it through an Actor message. Nope, it really isn't. Normally, I'd advise to log it right away, maximally, and go away from there but I understand you might haven't wanted to clone the Logger so that you could have moved it into the Future completing the result from that async actix message.

I still think wrapping the type in an error isn't a good decision. There must be a better way. 

Well. What about this. I later noticed that in request_route_and_transmit() there is a place where the logger actually is cloned(), ironically, while you don't need it. So think about using it in the Future, just move it in, and then log the MailboxError inside the future. It should be detectable from tests. 
The sweetest part is that you can get rid of the vexing result type from inside the Actor message. A lot of code will become simpler and easier to read. 

So eventually, this is the right thing to do, I'm sure. :point_up:  Don't forget to rename the msg back.

node/src/proxy_server/mod.rs
```diff
@@ -1049,24 +1255,104 @@ mod tests {
...
+                >,
+            >,
+        ) -> Self {
+            self.resolve_message_params = param.clone();
+            self
+        }
+
+        // fn resolve_message_result(self) -> Self {
+        //
+        // }
```
bertllll: You can delete the three lines.

node/src/proxy_server/mod.rs
```diff
@@ -2241,26 +2552,87 @@ mod tests {
...
+            peer_actors.proxy_server.route_result_sub = dns_retry_result_recipient; //Partial mocking
             subject_addr.try_send(BindMessage { peer_actors }).unwrap();
 
             subject_addr.try_send(msg_from_dispatcher).unwrap();
 
             system.run();
         });
 
-        let expected_add_route_message = AddRouteMessage {
+        let expected_dns_retry_result_message = RouteResultMessage {
```
bertllll: Isn't this naming style potentially misleading? Does every RouteResultMessage automatically mean it is an action of retry? Don't we send the same message initially when yet don't know if we're gonna have to retry? In that case it should probably be called differently.

node/src/proxy_server/mod.rs
```diff
@@ -2241,26 +2552,87 @@ mod tests {
...
+            .proxy_server(proxy_server_mock)
+            .neighborhood(neighborhood_mock)
+            .build();
+        // Get the dns_retry_result recipient so we can partially mock it...
+        let dns_retry_result_recipient = peer_actors.proxy_server.route_result_sub;
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
+        peer_actors.proxy_server.route_result_sub = dns_retry_result_recipient; //Partial mocking
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
+
+        subject_addr.try_send(msg_from_dispatcher).unwrap();
```
bertllll, line 2524: Strictly speaking, you should isolate this line by blank lines, yes, but as the only one in the whole test.

node/src/proxy_server/mod.rs
```diff
@@ -2241,26 +2552,87 @@ mod tests {
...
+    #[test]
+    fn proxy_server_sends_a_message_when_dns_retry_cannot_find_a_route() {
+        let test_name = "proxy_server_sends_a_message_when_dns_retry_cannot_find_a_route";
+        let cryptde = main_cryptde();
+        let http_request = b"GET /index.html HTTP/1.1\r\nHost: nowhere.com\r\n\r\n";
+        let (proxy_server_mock, _, proxy_server_recording_arc) = make_recorder();
+        let route_query_response = None;
+        let (neighborhood_mock, _, _) = make_recorder();
+        let neighborhood_mock = neighborhood_mock
+            .stop_on_message(type_id!(RouteResultMessage))
```
bertllll: You should delete this macro `type_id!`. The idea has always been that all the programmer needs can be done by `match_every_type_id!()`. It's designed to handle whatever, a single or multiple messages specified.

bertllll: Please, also delete the method `.stop_on_message()`, it can be easily substituted by the old `.system_stop_conditions()`. We don't want to have multiple ways of how we can do the same thing.

Your new infrastructure is used just seldomly but the other one has already turned into a standard.

node/src/proxy_server/mod.rs
```diff
@@ -2241,26 +2552,87 @@ mod tests {
...
+
+        let recording = proxy_server_recording_arc.lock().unwrap();
+        let message = recording.get_record::<RouteResultMessage>(0);
+        assert_eq!(message.stream_key, stream_key);
+        assert_eq!(
+            message.result,
+            Err(format!(
+                "Failed to find route to nowhere.com for stream key: {stream_key}"
+            ))
+        );
```
bertllll, line 2535: This is really puzzling me. 

You probably want to assert on the look of the message that arrived to Neighborhood, even though they might be identical. It would be a stronger test. 

However, here from just reading the tests, I don't understand why both actors would need to get this message. I was investigating on that and it seems like Neighborhood doesn't even have a Handler for this message!!! It's really weird. But it had to work out somehow because it does stop the test through the stop condition set at the top. I don't know. This is definitely weird. I was thinking, yeah, Neighborhood might not have the Handler of this message but we provide the Recorder which can handle every single message, still I don't know from where the message flew to it. 

Ohhhh. You know what?? I think it was the SystemKillerActor who allowed to finish the test. I think it's set to wait like 10 seconds probably and then it shut down the system. You can try out. If you have to be waiting really long it'll be an issue. You will want to get rid of the Neighborhood actor from this test and set the stop condition for the `proxy_server_mock`. That will ensure that you will get the message which you want to assert on afterwards. So you will end up really just with one assertion. However, check all these things out really carefully first.

node/src/proxy_server/mod.rs
```diff
@@ -2321,16 +2693,13 @@ mod tests {
...
             subject_addr.try_send(BindMessage { peer_actors }).unwrap();
-            subject_addr
-                .try_send(AddRouteMessage {
-                    stream_key,
-                    route: route_query_response,
-                })
-                .unwrap();
             subject_addr.try_send(msg_from_dispatcher).unwrap();
 
             system.run();
```
dnwiebe, line 2605: I can't leave a comment down where I want to leave it because of GitHub. I want to put this down on the `.await_message_count(1)` below.

After we've seen all the messages we need to see, the `system` should be shut down. It's not your fault this isn't happening, but you can fix it if you put a `System::current().stop();` right after that `.await_message_count(1)`.

Of course, maybe that was left out for some good reason that I can't see. If it makes the test fail or hang, take it out and let me know, and we (and maybe Bert) can come up with something else.

node/src/proxy_server/mod.rs
```diff
@@ -2321,16 +2693,13 @@ mod tests {
...
             );
             subject.stream_key_factory = Box::new(stream_key_factory);
+            subject
+                .stream_key_routes
+                .insert(stream_key, route_query_response);
             let subject_addr: Addr<ProxyServer> = subject.start();
             let mut peer_actors = peer_actors_builder().hopper(hopper_mock).build();
             peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
             subject_addr.try_send(BindMessage { peer_actors }).unwrap();
-            subject_addr
```
dnwiebe, line 2222: I noticed that you removed the processing of this message from the test in favor of manually inserting the results of processing the message. Are you sure the processing of the message is tested elsewhere? If this is the only test of that code, it shouldn't be removed.

node/src/proxy_server/mod.rs
```diff
@@ -2608,17 +2979,45 @@ mod tests {
...
+        expected = "RouteResultMessage Handler: stream key: +dKB2Lsh3ET2TS/J/cexaanFQz4 not found within dns_failure_retries"
     )]
-    fn proxy_server_panics_when_exit_services_are_not_requested_in_non_zero_hop_mode() {
-        let expected_services = vec![ExpectedService::Routing(
-            PublicKey::from(&b"routing_key_1"[..]),
-            make_wallet("routing_wallet_1"),
-            rate_pack(8),
-        )];
-
+    fn route_result_message_handler_panics_when_dns_retries_hashmap_doesnt_contain_a_stream_key() {
```
dnwiebe, line 2882: Once more, I'm a little concerned that there may be a way this panic could be caused by a particular sequence of network operations, where something arrives a little earlier or later than we expect it and doesn't make it into the proper collection. If that can happen, this shouldn't be a panic, because it means an attacker can shut us down from outside.

Can you either A) verify that this imagined sequence of events can't possibly happen, or B) just replace this panic with an error log and continue, to be safe?

Thanks...

node/src/proxy_server/mod.rs
```diff
@@ -2608,17 +2979,45 @@ mod tests {
...
+        let subject = ProxyServer::new(
+            main_cryptde(),
+            alias_cryptde(),
+            true,
+            Some(STANDARD_CONSUMING_WALLET_BALANCE),
+            false,
+        );
+        let subject_addr: Addr<ProxyServer> = subject.start();
+        let mut peer_actors = peer_actors_builder().build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
```
dnwiebe: Try using the `peer_actors_builder()` to build the Proxy Server subs into the `peer_actors` list.

node/src/proxy_server/mod.rs
```diff
@@ -3218,8 +3635,9 @@ mod tests {
         };
         assert_eq!(record, &expected_msg);
 
-        TestLogHandler::new()
-            .exists_log_containing("ERROR: ProxyServer: Failed to find route to server.com");
+        TestLogHandler::new().exists_log_containing(
```
dnwiebe: I notice that you're asserting on this log message in several places.

This may be a nitpick, but I think you only have to assert on it in one place, unless there are ways that it can be materially different. If you assert on it in ten places, then if anyone should ever decide to change the log message, there are ten tests that have to be modified instead of one.

node/src/proxy_server/mod.rs
```diff
@@ -3550,6 +3968,109 @@ mod tests {
...
+        let first_expired_cores_package: ExpiredCoresPackage<ClientResponsePayload_0v1> =
+            ExpiredCoresPackage::new(
+                SocketAddr::from_str("1.2.3.4:1234").unwrap(),
+                Some(make_wallet("irrelevant")),
+                return_route_with_id(cryptde, 1234),
+                first_client_response_payload.into(),
+                0,
+            );
+        let mut peer_actors = peer_actors_builder().build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
```
dnwiebe: Again: use the builder. That's what it's there for.

node/src/proxy_server/mod.rs
```diff
@@ -3839,6 +4378,16 @@ mod tests {
         );
         let stream_key = make_meaningless_stream_key();
         let socket_addr = SocketAddr::from_str("1.2.3.4:5678").unwrap();
+        let mut dns_failure_retries_hash_map = HashMap::new();
```
dnwiebe, line 4266: You probably thought of this already, but it's just occurring to me now. If we try three different routes and can't get to the desired host, do all three of those exit Nodes get reported to the Neighborhood in `AddUnreachableHost` messages, or only the last one? (I think all of them should be; what do you think?)

node/src/proxy_server/mod.rs
```diff
@@ -3905,6 +4454,16 @@ mod tests {
...
```
dnwiebe, line 4347: Not your fault, but I think this is a bad test. The `System::current.stop()` comes right after the CORES-package message, which means that any messages produced by the processing of the CORES-package message (including any improperly-produced `NodeRecordMetadataMessage`) will go into the queue after the Stop message; the `system` will stop before it sends them to their Actors. Therefore, even if the code was misbehaving, this test would claim that it wasn't.

If you don't want to fix this test, I understand; but please either make a card about it (and other tests that may have the same defect) or at least put a //TODO on it so that somebody else can fix it.

Props to Bert for noticing this.

node/src/proxy_server/mod.rs
```diff
@@ -4106,17 +4671,25 @@ mod tests {
...
             .neighborhood(neighborhood_mock)
             .dispatcher(dispatcher_mock)
             .build();
         subject.subs.as_mut().unwrap().update_node_record_metadata =
             peer_actors.neighborhood.update_node_record_metadata;
         subject.subs.as_mut().unwrap().dispatcher = peer_actors.dispatcher.from_dispatcher_client;
-
         let stream_key = make_meaningless_stream_key();
         let socket_addr = SocketAddr::from_str("1.2.3.4:5678").unwrap();
+        let mut dns_failure_retries_hash_map = HashMap::new();
```
dnwiebe, line 4575: After seeing as many as I have of these loads of `dns_failure_retries_hash_map` with an exhausted set of retries, you make me wonder if it might save a fair amount of cognitive effort to have them replaced by a call to an outboard utility function.

node/src/proxy_server/mod.rs
```diff
@@ -4157,6 +4732,367 @@ mod tests {
...
+            alias_cryptde(),
+            false, //meaning ZeroHop
+            Some(STANDARD_CONSUMING_WALLET_BALANCE),
+            false,
+        );
+        let stream_key = make_meaningless_stream_key();
+        let socket_addr = SocketAddr::from_str("1.2.3.4:5678").unwrap();
+        let mut dns_failure_retries_hash_map = HashMap::new();
+        let client_payload = make_request_payload(111, cryptde);
+        dns_failure_retries_hash_map.insert(
```
dnwiebe, line 4645: If you're testing zero-hop mode, do you really want to preload with an exhausted retry count? It seems to me that zero-hop mode should never retry a DNS failure (what's the point?), so there should probably be some sort of special-casing involved when it comes to populating the `dns_failure_retries_hash_map`. Wouldn't a test like this be a perfect place to exercise that special-casing code, rather than bypassing it by preloading?

node/src/proxy_server/mod.rs
```diff
@@ -4157,6 +4732,367 @@ mod tests {
...
+    fn handle_dns_resolve_failure_zero_hop() {
+        let system = System::new("handle_dns_resolve_failure_zero_hop");
+        let (dispatcher_mock, _, dispatcher_recording_arc) = make_recorder();
+        let (neighborhood_mock, _, neighborhood_recording_arc) = make_recorder();
+        let cryptde = main_cryptde();
+        let this_node_public_key = cryptde.public_key();
+        let mut subject = ProxyServer::new(
+            cryptde,
+            alias_cryptde(),
+            false, //meaning ZeroHop
```
dnwiebe, line 4637: Thank you very kindly for the comment.

node/src/proxy_server/mod.rs
```diff
@@ -4157,6 +4732,367 @@ mod tests {
...
+                return_route_with_id(cryptde, 1234),
+                dns_resolve_failure.into(),
+                0,
+            );
+        let mut peer_actors = peer_actors_builder()
+            .neighborhood(neighborhood_mock)
+            .build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
+
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
```
dnwiebe, line 4785: `BindMessage` isn't part of the Act section; it should be clustered up with the Arrange part.

node/src/proxy_server/mod.rs
```diff
@@ -4157,6 +4732,367 @@ mod tests {
...
+            route_query_message_response.unwrap().unwrap(),
+            route_query_response_expected
+        );
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "No entry found inside dns_failure_retries hashmap for the stream_key: +dKB2Lsh3ET2TS/J/cexaanFQz4"
+    )]
+    fn handle_dns_resolve_failure_panics_when_there_is_no_entry_in_the_hashmap_for_the_stream_key()
```
dnwiebe: Is this a duplicate test?

node/src/proxy_server/mod.rs
```diff
@@ -4157,6 +4732,367 @@ mod tests {
...
+                SocketAddr::from_str("1.2.3.4:1234").unwrap(),
+                Some(make_wallet("irrelevant")),
+                return_route_with_id(cryptde, 1234),
+                dns_resolve_failure.into(),
+                0,
+            );
+        let mut peer_actors = peer_actors_builder()
+            .neighborhood(neighborhood_mock)
+            .build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
```
dnwiebe: Use the builder for Proxy Server the way you're doing with Neighborhood!

node/src/proxy_server/mod.rs
```diff
@@ -4846,34 +5782,43 @@ mod tests {
...
-            Err(MailboxError::Timeout),
-        );
+        subject.resolve_message(tth_args, proxy_server_sub, Err(MailboxError::Timeout));
 
-        TestLogHandler::new().exists_log_containing("ERROR: resolve_route_query_response_handles_error: Neighborhood refused to answer route request: MailboxError(Message delivery timed out)");
+        System::current().stop();
+        system.run();
+        let proxy_server_recording = proxy_server_recording_arc.lock().unwrap();
+        let message = proxy_server_recording.get_record::<RouteResultMessage>(0);
+        let expected_error_message = "Neighborhood refused to answer route request: MailboxError(Message delivery timed out)";
```
dnwiebe, line 5696: I don't think a MailboxError means anything timed out; I think a MailboxError means the destination mailbox is gone because the Actor died. I'm not absolutely 100% sure of this, but I think it's the case. If it is, this is the wrong message.

I should be clearer: I _am_ absolutely sure that a MailboxError doesn't mean something timed out, because the whole point of Actors is asynchronicity, which means you don't wait on stuff.

Come to think of it, it might also mean the mailbox you were trying to send to is full.

node/src/proxy_server/mod.rs
```diff
@@ -4934,6 +5879,97 @@ mod tests {
...
+            true,
+            Some(STANDARD_CONSUMING_WALLET_BALANCE),
+            false,
+        );
+        subject.stream_key_factory = Box::new(stream_key_factory);
+        let subject_addr: Addr<ProxyServer> = subject.start();
+        let mut peer_actors = peer_actors_builder()
+            .neighborhood(neighborhood_mock)
+            .build();
+        peer_actors.proxy_server = ProxyServer::make_subs_from(&subject_addr);
```
dnwiebe: Agh!

node/src/proxy_server/mod.rs
```diff
@@ -4934,6 +5879,97 @@ mod tests {
         );
     }
 
+    #[test]
+    fn new_http_request_creates_new_entry_inside_dns_retries_hashmap() {
```
dnwiebe, line 5765: This sort of test is what I had in mind above when I commented on the zero-hop test.

node/src/proxy_server/mod.rs
```diff
@@ -4934,6 +5879,97 @@ mod tests {
...
+                    assert_eq!(dns_retry.unsuccessful_request, expected_payload);
+                }),
+            })
+            .unwrap();
+        System::current().stop();
+        system.run();
+    }
+
+    #[test]
+    fn hostname_works() {
```
dnwiebe, line 5901: I think I'd like to see an example with just `http` (not `https`), and I'd like to see the really wild examples cause errors, rather than returning host names everybody knows are wrong.

node/src/stream_handler_pool.rs
utkarshg6: This is something that Syther and I concluded when we went through a deep dive of the codebase and looked into the every usage and decided that it's appropriate to put a comment over here.

I believe it'll definitely help any developer who would look into this piece of code because they want to know where exactly it's coming from. Also, if the usage changes, the developer will feel the necessity to change the comment appropriately to save them from any confusion.

node/src/proxy_server/mod.rs
utkarshg6: This TODO was just an inspiration for a new card. It's not something that should be within the scope of the current card. It'll be appropriate if we have a discussion whether this comment is worth a new card or it'll be more appropriate to discard this idea and remove this TODO.

node/src/proxy_server/mod.rs
utkarshg6: I believe `warning!()` is more appropriate.

node/src/proxy_server/mod.rs
utkarshg6: Yes, it can be triggered by the outside network, but not at this place.

Some bad Node can send an Expired Cores package with a fake stream key and it may cause the Node to panic and crash. But this happens inside the other case and not this one. For this case a panic is safe to use. Here are the following reasons to justify this statement:

1) The `stream_key` is created as soon as the client generates a new request and the Node adds a new entry inside the HashMap as soon as that happens.
2) By the time this code executes we are certain that there must be an entry inside the HashMap for the desired `stream_key`, if that's not the case there must be some miscommunication between the client and the Node.
3) Also, the `RouteResultMessage` contains information from `Neighbourhood` that's intended for `ProxyServer`. Since, no information is dependent on the information triggered from outside the network, it's completely safe to conclude that this panic can't be triggered from outside the network.

node/src/proxy_server/mod.rs
utkarshg6: It's quite similar to another comment, where there's a similar panic.

This panic is fatal. A bad Node can easily crash a Node by sending an ExpiredCoresPackage with a made up stream key. I'd recommend if we. change it to `error!()` over here.

node/src/proxy_server/mod.rs
utkarshg6: First of all, if you feel that it'd be more clearer to rename it to `AddRouteResultMessage`, then that's a request that we can easily do. 👍 

I do think that the Error has its importance. It is caused by the Neighbourhood and it means that due to some reason the Route wasn't found, and it explains that reason with this error.

It's not fatal, but it's an error.

Syther and I will try to do as you mentioned of logging the error and getting rid of the Result wrapping. Although, if we find a contrary argument why we shouldn't be doing that then we'll let you know. As of now, your idea seems more appropriate.

node/src/proxy_server/mod.rs
utkarshg6: This is something that I proposed to Syther that we should implement. I first request him to add this and now requesting him to remove it, will mean just adding more trouble in Syther's life.

Here are some reasons why I asked him for such a request.
1) The `type_id!` macro can also be used while debugging where a developer wants to just print the type id while testing to make sure, it's what they are expecting.
2) If the developer wants to do the above without the use of this macro, it's a bit hard to remember what exact code they must write to get the type id. You may look into the implementation of the macro, you'll also find this macro easier to use rather than the original code.
3) This macro was the first macro that Syther wrote and it'll be better if we keep it as it'll be more beneficial for everyone to remember his efforts.

Let me know if it's appropriate to keep it. 🙏

utkarshg6: I do think that this method has vast implications throughout the codebase. It was used seldomly because we only used it for the newly written tests and the old tests that might benefit from them are untouched. We can surely replace the long and hard to read code in other tests with this quite simple and easy to use function.

I do think `.stop_on_message()` function was older than the `.system_stop_conditions()` as it was the original idea which was later expanded into the `StopConditions`. As per my experience writing tests, almost each and every test that requires a StopCondition later in the life of the test, can simply use this method.

This is my humble request to keep it, as I intend to use it inside the tests that I'll write in the future. Also, it's something that we can use to make the old tests easier to read.

node/src/proxy_server/mod.rs
utkarshg6: I think this concern isn't as critical as it may appear. Here are the reasons behind this statement - https://github.com/MASQ-Project/Node/pull/326#discussion_r1305364710

node/src/proxy_server/mod.rs
utkarshg6: Yes, we are doing that for each and every exit Node that fails to resolve the DNS.

node/src/proxy_server/mod.rs
utkarshg6: We're panicking at two different places in case we're not able to find an entry inside the Hashmap.

The two places are:

1) Handler for `RouteResultMessage` - This message contains information from Neighbourhood intended for ProxyServer.
2) Inside `handle_dns_resolve_failure()` - This is triggered when we receive an ExpiredCoresPackage containing a DNS Failure.

I do think that the second case is a security risk and it must be changed from a panic to an error log. This test is indeed part of the second case.

node/src/proxy_server/mod.rs
Syther007: This is now fixed,
Adding a comment here to remind my-self to make sure this issue isn't happening with any other tests.

node/src/stream_handler_pool.rs
Syther007: We have done some investigation into this.
First, `NodeQueryMessage` doesn't have its own Handler anywhere in the code base.
Only `DispatcherNodeQueryMessage` allows us to send a response back, if we intend to use `NodeQueryMessage` doe that mean we don't need to send a response back.

Maybe a new card is needed for this?

node/src/stream_handler_pool.rs
Syther007: msg is being used further down in the code.
`.clone()` is needed

node/src/proxy_server/mod.rs
Syther007: @bertllll This source_addr is part of TryTransmitToHopperArgs, can I ask your advise on what we could rename this to?

node/src/proxy_server/mod.rs
Syther007: Completed changing `panic!` to `error!`

node/src/proxy_server/mod.rs
Syther007: @dnwiebe 
This type `Result<Option<RouteQueryResponse>, MailboxError>` is tightly coupled with the futures `.then()` function.
The closure indented for the `.then()` function receives this type as an argument due to this reason we see this type in different places within our code base. 

Do we need to type cast this type into a different one?

node/src/proxy_server/mod.rs
Syther007: I've added a new card for this issue. and added your comments.
https://github.com/MASQ-Project/MASQ-Node-issues/issues/743

node/src/proxy_server/mod.rs
Syther007: I've added a new card for this issue. and added your comments.
https://github.com/MASQ-Project/MASQ-Node-issues/issues/743

node/src/proxy_server/mod.rs
utkarshg6: We've been very consistent with the name `dns_failure` while implementing this card. Keeping it the same will help in maintaining consistency. Also, this variable stores the client request, which is a form of caching. Since the cached data of the dns failure is stored inside the variable, it seems more appropriate to keep the name `dns_failure` over here.

node/src/proxy_server/mod.rs
Syther007: We have renamed `RouteResultMessage` to `AddRouteResultMessage`

node/src/proxy_server/mod.rs
Syther007: @dnwiebe

node/src/proxy_server/mod.rs
Syther007: We had found 3 other tests, testing `AddRouteMessage` that is now know as `AddRouteResultMessage`

node/src/proxy_server/mod.rs
Syther007: @dnwiebe We have added a new error log in the case that `NodeRecordMetadataMessage` wont be sent.
And now asserting on this.

node/src/proxy_server/mod.rs
Syther007: I like to solve this by using the Hashmap macro @utkarshg6 has newly created.
Ticket link: https://github.com/MASQ-Project/MASQ-Node-issues/issues/745

node/src/proxy_server/mod.rs
Syther007: We have tried to remove this insert.
When doing so we receive `NodeRecordMetadataMessage` but not `TransmitDataMsg`
what should the indented behaviour be when it comes to zero-hope mode and a DNS failure?

node/src/proxy_server/mod.rs
Syther007: We have remove this line from this test as it turned out it wasn't needed.
Also found about another 20ish occurrences and have remove also.

However there we a few tests that still needed this code. 
And we were unable to use the builder method because it requires a recorder as an argument.

node/src/proxy_server/mod.rs
Syther007: This MailboxError has a variant named Timeout and is part of the Actix library.
  Due to that we cant modify this.

node/src/proxy_server/mod.rs
Syther007: 😢 we cant do it

node/src/proxy_server/mod.rs
Syther007: I've added a new card for this issue. and added your comments.
https://github.com/MASQ-Project/MASQ-Node-issues/issues/743

node/src/proxy_server/mod.rs
dnwiebe: I can think of two SocketAddrs that this might be. It might be the SocketAddr of the browser connection that initiated this DNS attempt, or it might be the SocketAddr of the relay Node we're going to send the retry to. If it is the former, you might call it `client_addr`. If the latter, maybe `neighbor_addr` or `relay_addr` would be more appropriate.

node/src/proxy_server/mod.rs
dnwiebe: I'm pretty sure there is another card for this now; we should be able to remove the //TODO.

node/src/proxy_server/mod.rs
dnwiebe: I like Bert's idea: can you find the test(s) where the message(s) _are_ sent to the other Actor(s) and assert that your new error log is not logged there?

node/src/proxy_server/mod.rs
dnwiebe: I don't know exactly what zero-hop mode does here, but it seems reasonable that it might special-case `subject.dns_failure_retries` into a state that looks like all the retries have already been consumed, so as not to stimulate any new retries.

If that's the case, you don't want to replace `subject.dns_failure_retries` in this test, because that behavior is what this test should be testing, not what it should be setting up.

dnwiebe (approved):

PR #334, opened 2023-08-24
GH-692: Review 1

utkarshg6:


masq/src/commands/set_configuration_command.rs
```diff
@@ -146,10 +144,9 @@ mod tests {
         ]);
 
         let err_msg = result.unwrap_err();
-        assert!(err_msg.contains(
-            "error: Found argument '--invalid-arg' \
-             which wasn't expected, or isn't valid in this context"
-        ));
+        assert!(err_msg.contains("Found argument"));
```
dnwiebe: These could still use second and third parameters like `, "{}", err_msg)` so that the failure messages would tell what the error message (that didn't contain the required substrings) looked like. This way, all you'll get is a naked notification that the test failed, with no details.

PR #336, opened 2023-09-06
GH-716: Our StreamKeys are insecure

### LINKED ISSUE #716: Our StreamKeys are insecure
Part of the infrastructure of the MASQ network is a series of virtual data streams that run from browser to server and server to browser. Each of these virtual streams consists of several real TCP streams: from the browser to the originating Node, from the originating Node to the first relay Node, several more TCP streams through the relay Nodes, from the last relay Node to the exit Node, and from the exit Node to the server--and back again, since the virtual streams are bidirectional.

In order to tell these virtual streams apart, we have something called a `StreamKey` that is generated and assigned to each stream when it is created in response to the browser opening a new TCP stream to the originating Node. The originating Node has a collection of virtual streams keyed by these `StreamKey`s representing all the streams its browser has open to servers around the Internet, and the exit Node has a collection of virtual streams also keyed by `StreamKey`s that go from every server it's connected to back to the originating Nodes that have requested data from those servers.

When a browser opens a new TCP stream to what it thinks is a server but is actually an originating Node, the originating Node creates a new `StreamKey` for the virtual stream it's about to bring into existence. To create the `StreamKey`, it hashes together (using SHA-1) two pieces of information: its own public key, and the `SocketAddr` of the browser's end of the new TCP stream.

For example, if the originating Node's public key was  `gBviQbjOS3e5ReFQCvIhUM3i02d1zPleo1iXg_EN6zQ`, and the browser opened a stream from 127.0.0.1:51436, then the `StreamKey` for that stream would be 67728a4ec9205ce9cc88f22ade0c93498e425ccc.

But there's a problem.

The exit Node is not supposed to know anything about the identity of the originating Node. It can't directly derive anything about the originating Node's identity from the `StreamKey` it created, but it does have an important indirect source of information: its neighborhood database. That, combined with the fact that the browser's IP address is always 127.0.0.1 and the port number is always an ephemeral port between 49152 and 65535, makes for an attack vector.

If the exit Node wants to know which originating Nodes it's serving, it can do this:
```
fn find_public_key_from_stream_key (stream_key: StreamKey) -> Option<(PublicKey, u16)> {
    for node_record in neighborhood_database {
        for ephemeral_port in 49152u16..65535u16 {
            let candidate_key = StreamKey::new (node_record.public_key(), SocketAddr::new(localhost(), ephemeral_port))
            if (candidate_key == stream_key) {
                return Some((node_record.public_key(), ephemeral_port))
            }
        }
    }
    None // originating Node isn't in neighborhood_database
}
```
The malefactors who create the evil exit Node don't even have to write `StreamKey::new()`: it's already part of the codebase!

__Assignment:__ Find a way to make `StreamKey`s more secure without disrupting existing data or Gossip protocols between Nodes.

__Observation:__ Since nobody is ever going to try to extract from a `StreamKey` the data that was used to create it (as opposed to creating new `StreamKey`s and comparing them), it doesn't really matter what that data is, as long as it's sufficiently unique to prevent unintentional collisions.  (Translation: maybe you could just pile in a bunch of random data having nothing to do with public keys or `SocketAddr`s. How much? Enough to assure uniqueness. How much is that? Well...think about it. If we're going to generate random data anyway, do we even need a hashing function?)

__Another observation:__ Maintaining uniqueness inside an originating Node will be pretty easy, because the same `StreamKey` generator will be generating all those `StreamKey`s. Maintaining uniqueness inside an exit Node might be a little harder, since its `StreamKey`s will have been generated by different generators in all the originating Nodes that are using its services. There should be some assurance that the fourth `StreamKey` I generate can't be the same as the fourth `StreamKey` you generate, in case we're both using the same exit Node.

__Yet another observation:__ UUIDs were invented specifically to address _almost_ exactly this kind of problem, and `crates.io` has at least one library that generates them. The _almost_ part is because UUIDs can be parsed into fields that might give attackers too much information. One set of fields, for example, comes from a time value, which would tell an attacker how old the stream is. (Is this a problem? Maybe.) Another field could allow the extraction of the MAC address of the generating machine's network interface card...although a different version of UUID allows that to be replaced with a random number. If further investigation reveals UUIDs to be insufficiently secure, we could hash them.

utkarshg6:


node/src/proxy_server/mod.rs
```diff
@@ -4361,11 +4346,9 @@ mod tests {
             false,
         );
         let unaffected_socket_addr = SocketAddr::from_str("2.3.4.5:6789").unwrap();
-        let unaffected_stream_key =
-            StreamKey::new(main_cryptde().public_key().clone(), unaffected_socket_addr);
+        let unaffected_stream_key = StreamKey::default();
         let affected_socket_addr = SocketAddr::from_str("3.4.5.6:7890").unwrap();
-        let affected_stream_key =
-            StreamKey::new(main_cryptde().public_key().clone(), affected_socket_addr);
+        let affected_stream_key = StreamKey::default();
```
dnwiebe: Functionally, there's nothing wrong with this test; but it would be clearer to read if you used your `make_meaningful_stream_key()` function here with different parameters. If you don't know exactly what `StreamKey::default()` does, it looks like you're setting `unaffected_stream_key` and `affected_stream_key` to the same value.

node/src/proxy_server/mod.rs
```diff
@@ -4484,11 +4467,9 @@ mod tests {
...
```
dnwiebe: Here too.

node/src/proxy_server/mod.rs
```diff
@@ -4594,11 +4575,9 @@ mod tests {
     fn handle_stream_shutdown_msg_does_not_report_to_counterpart_when_unnecessary() {
         let mut subject = ProxyServer::new(main_cryptde(), alias_cryptde(), true, None, false);
         let unaffected_socket_addr = SocketAddr::from_str("2.3.4.5:6789").unwrap();
-        let unaffected_stream_key =
-            StreamKey::new(main_cryptde().public_key().clone(), unaffected_socket_addr);
+        let unaffected_stream_key = StreamKey::default();
         let affected_socket_addr = SocketAddr::from_str("3.4.5.6:7890").unwrap();
-        let affected_stream_key =
-            StreamKey::new(main_cryptde().public_key().clone(), affected_socket_addr);
+        let affected_stream_key = StreamKey::default();
```
dnwiebe: And here...

node/src/sub_lib/stream_key.rs
```diff
@@ -97,83 +95,42 @@ type HashType = [u8; sha1::DIGEST_LENGTH];
...
-    fn mismatched_keys_and_matching_addrs_make_mismatched_stream_keys() {
-        let one_key = PublicKey::new(&b"These are the times"[..]);
-        let another_key = PublicKey::new(&b"that try men's souls"[..]);
-        let addr = SocketAddr::from_str("2.3.4.5:6789").unwrap();
+    fn stream_keys_are_unique() {
+        let mut stream_keys_set = HashSet::new();
 
-        let one = StreamKey::new(one_key.clone(), addr);
-        let another = StreamKey::new(another_key, addr);
+        for i in 1..=1_000_000 {
```
dnwiebe: I think a million is a little high. Probably 100 or 1000 is fine. The stream key only has to be unique within a set of streams to or from the same entity (browser, Node, server), and you're not going to see more than 1000 streams in a set like that.

node/src/test_utils/mod.rs
```diff
@@ -174,10 +173,17 @@ impl Waiter {
 }
 
 pub fn make_meaningless_stream_key() -> StreamKey {
```
dnwiebe: If you moved these two out of here and made them `#[cfg(test)]`-only constructors for `StreamKey`, then you wouldn't have to make the `hash` field of `StreamKey` public, and it could be completely opaque to its clients.

dnwiebe (approved):

dnwiebe (approved):

PR #338, opened 2023-09-08
GH-651 Review-1 Changes

Syther007:


multinode_integration_tests/docker/Dockerfile
```diff
@@ -17,8 +17,8 @@ RUN apt-get update && \
     apt-get install -y procps
 COPY generated/port_exposer /usr/local/bin/port_exposer
 
-ENV SUDO_UID 1000
-ENV SUDO_GID 1000
+ENV SUDO_UID 1001
```
dnwiebe: Change this back before you merge.

dnwiebe (commented):
Resume at proxy_server/mod.rs:527 red

node/src/proxy_server/mod.rs
```diff
@@ -1095,24 +1094,18 @@ impl IBCDHelper for IBCDHelperReal {
                 .dns_failure_retries
                 .insert(stream_key, dns_failure_retry);
         }
-
-        let tth_args = TryTransmitToHopperArgs {
-            main_cryptde: proxy.main_cryptde,
+        let retire_stream_key_sub_opt = if retire_stream_key {
+            Some(proxy.out_subs("ProxyServer").stream_shutdown_sub.clone())
```
dnwiebe: Not your fault, but `proxy` would be much better as `proxy_server`. Even `server` would be (a little) better, because "server" is more descriptive of its role than "proxy" is.

node/src/proxy_server/mod.rs
```diff
@@ -1095,24 +1094,18 @@ impl IBCDHelper for IBCDHelperReal {
                 .dns_failure_retries
                 .insert(stream_key, dns_failure_retry);
         }
-
-        let tth_args = TryTransmitToHopperArgs {
-            main_cryptde: proxy.main_cryptde,
+        let retire_stream_key_sub_opt = if retire_stream_key {
```
dnwiebe: Maybe it would be better to move this logic into the `TryTransmitToHopperArgs::new()` constructor and pass in just the boolean flag?

dnwiebe (commented):
Resume at proxy_server/mod.rs:1250 green

node/src/proxy_server/mod.rs
```diff
@@ -4503,6 +4492,9 @@ mod tests {
         let neighborhood_recording = neighborhood_recording_arc.lock().unwrap();
         let record_opt = neighborhood_recording.get_record_opt::<NodeRecordMetadataMessage>(0);
         assert_eq!(record_opt, None);
+        TestLogHandler::new().exists_log_containing(&format!(
+            "ERROR: {test_name}: A bad exit node lied to us about the DNS failure"
```
dnwiebe: Can you make this message a bit more specific? For example, we'd like to know that the issue is a stream key that doesn't exist. When we eventually get around to malefactor banning, each ban will probably have a message associated with it telling why the malefactor was banned; think about looking at a database table full of malefactor bans from months or years ago, and use a message that would make clear what the malefactor's crime was.

Syther007: @dnwiebe it appears this error message could be wrong.
Will bring this up in the Dev meeting.

PR #339, opened 2023-09-12
GH-747: transactions received from unknown wallets

bertllll:


node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -120,20 +117,19 @@ impl PayableDao for PayableDaoReal {
         let overflow_update_clause = "update payable set \
                 balance_high_b = :balance_high_b, balance_low_b = :balance_low_b where wallet_address = :wallet";
 
+        let last_paid = to_time_t(timestamp);
```
utkarshg6: I can't put a comment any higher, but if this timestamp is pointing to a time when it was `last_paid`, then I believe it'll be more appropriate to keep the parameter name similar, for example, `last_paid_timestamp`.

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -173,15 +169,18 @@ impl PayableDao for PayableDaoReal {
                     balance_high_b = :balance_high_b, balance_low_b = :balance_low_b, last_paid_timestamp = :last_paid, \
                     pending_payable_rowid = null where pending_payable_rowid = :rowid";
 
-            let conn = Left(self.conn.as_ref());
-            Ok(self.big_int_db_processor.execute(conn, BigIntSqlConfig::new(
+            let i64_rowid = checked_conversion::<u64, i64>(fgp.rowid);
+            let last_paid = to_time_t(fgp.timestamp);
```
utkarshg6: I was struggling to figure out what `fgp` stands for, if we can use some other term, it might help us in the future.

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -125,26 +123,30 @@ impl ReceivableDao for ReceivableDaoReal {
         let overflow_update_clause = "update receivable set balance_high_b = :balance_high_b, balance_low_b = :balance_low_b \
         where wallet_address = :wallet";
 
+        let last_received = to_time_t(timestamp);
```
utkarshg6: There is a similar comment for `more_money_payable`, can we also change the parameter of this fn to `last_received_timestamp`.

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -279,43 +281,71 @@ impl ReceivableDaoReal {
...
+        logger: &Logger,
+    ) {
+        if Self::check_row_presence(conn, &received_payment.from) {
+            panic!(
+                "Update for received payment with {} wei was run without producing a data \
+            change, despite the account for wallet address {} is present",
+                received_payment.wei_amount, received_payment.from
+            )
+        } else {
+            info!(
```
utkarshg6, line 406: It appears to me, that this change is too big.. do we just want to log it and that's it, do we expect the user to figure out the solution to this problem or do we want to try it solve ourselves?

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -364,42 +394,6 @@ impl ReceivableDaoReal {
             feeder.limit_clause
         )
     }
-
-    fn more_money_received_pretty_error_log(
```
utkarshg6, line 368: Is it moved somewhere or are we not using this fn anymore?

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -723,92 +744,223 @@ mod tests {
...
-            \"Error from invalid update command for receivable table and change of -123456789123456789 \
-             wei to 'wallet_address = 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' with error 'no such table: receivable'\")\
-            \n\
-            Block #    Wallet                                     Amount            \n\
-            1234567890 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 123456789123456789\n\
-            2345678901 0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb 234567891234567891\n\
-            3456789012 0xcccccccccccccccccccccccccccccccccccccccc 345678912345678912\n\
-            TOTAL                                                 703703592703703592"
-        ));
+        let thread_handle = thread::spawn(move || {
```
utkarshg6: Do we really need to spawn a new thread for this one, I was hoping that we can test the panic directly using `#[should_panic]`?

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -723,92 +744,223 @@ mod tests {
...
+        TestLogHandler::new().exists_no_log_containing(&format!("INFO: {test_name}: "));
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Update for received payment with 1000000000 wei was run without \
+    producing a data change, despite the account for wallet address 0x0000000000000000000000000\
+    0000000626c6168 is present"
+    )]
+    fn log_or_panic_is_fatal_when_the_row_is_missing() {
```
utkarshg6: Can we directly keep the name something like `panics_when_the_row_is_missing`, because `log_or_panic_is_fatal` sounds more like that this function crashes whenever a row is missing, instead this function is actually doing the right thing of panicking.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -21,23 +20,23 @@ pub struct BigIntDbProcessor<T: TableNameDAO> {
 impl<'a, T: TableNameDAO> BigIntDbProcessor<T> {
     pub fn execute(
         &self,
-        conn: Either<&dyn ConnectionWrapper, &Transaction>,
```
utkarshg6, line 24: Good Job! 👏

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -147,11 +133,13 @@ impl<T: TableNameDAO> UpdateOverflowHandler<T> for UpdateOverflowHandlerReal<T>
...
 impl<T: TableNameDAO + Debug> UpdateOverflowHandlerReal<T> {
     fn execute_update<'a>(
-        conn: Either<&dyn ConnectionWrapper, &Transaction>,
+        conn: &dyn ConnectionWrapper,
         config: &BigIntSqlConfig<'a, T>,
         execute_params: &[(&str, &dyn ToSql)],
     ) {
-        match BigIntDbProcessor::<T>::prepare_statement(conn, config.overflow_update_clause)
+        match conn
+            .prepare(config.overflow_update_clause)
```
utkarshg6: Nicely done! 🎉

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -264,6 +252,9 @@ macro_rules! impl_of_extended_params_marker{
 
 impl_of_extended_params_marker!(i64, &str, Wallet);
 
+type DisplayableParamPair<'a> = (&'a str, &'a dyn ExtendedParamsMarker);
+type RusqliteParamPair<'a> = (&'a str, &'a dyn ToSql);
```
utkarshg6: I'd prefer if we use different structs instead of the other two types, and also give the elements of tuple a better name:

For example:

```rust
struct DisplayableParamPair<'a> {
    param1: &'a str,
    param2: &'a dyn ExtendedParamsMarker,
}
```

Also, I'd be happier if we use some name for a lifetime instead of using the generic keyword `a`, it does proves the relation but it keeps things ambiguous.

A suggestion for the names of the lifetimes are `'display_pair`, and `'rusqlite_pair` respectively.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -1294,47 +1352,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn update_alone_works_also_for_transaction_instead_of_connection() {
```
utkarshg6, line 1298: Is there a new test instead of this one?

node/src/database/db_initializer.rs
```diff
@@ -132,15 +132,15 @@ impl DbInitializerReal {
...
-        self.create_payable_table(conn);
-        self.create_pending_payable_table(conn);
-        self.create_receivable_table(conn);
-        self.create_banned_table(conn);
+        Self::create_config_table(conn);
+        Self::initialize_config(conn, external_params);
+        Self::create_payable_table(conn);
+        Self::create_pending_payable_table(conn);
+        Self::create_receivable_table(conn);
+        Self::create_banned_table(conn);
```
utkarshg6, line 140: Great! 👍

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -279,43 +281,71 @@ impl ReceivableDaoReal {
...
+        } else {
+            info!(
+                logger,
+                "Received a transaction with {} wei from address {} that does not belong to any of \
+                the known debtors. Ignoring",
+                received_payment.wei_amount,
+                received_payment.from
+            )
         }
-        match xactn.commit() {
```
utkarshg6, line 314: Why don't we need to commit now?

utkarshg6 (commented):
Please take a look at the comments when you have some time. 🙏

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: I think it should've been `log_or_panic_if_fatal`. But yeah I might be able to make it better.

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: There is nothing more we can do here... and the user will know better possibly. By the highest chances, it was him who sent the transaction (in general we want to treat it as a funding tx which means that we will ignore those; and we log it because, at least, we might want to make the user aware that this is happening) :man_shrugging:

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: That's the point of the main change here. We used to roll back all the db operations if we ran into an error half the way. So this log was supposed to show the user all the transactions that we were about to write persistently even though we eventually had to give up. 

Now we don't need this big machinery because we agreed with Dan on panicking in case of any error different form the error caused by the transaction from an unknown address.

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: We will...plans have changed. Now, after a discussion with Dan, we want to make use of the `commit()` byproduct which is a rollback (for all txs) which would happen before we update the start block. That way we can recover from panics without serious accountancy errors.

PR #341, opened 2023-09-15
GH 731 review

czarte:


masq_lib/src/multi_config.rs
```diff
@@ -61,24 +62,41 @@ impl<'a> MultiConfig<'a> {
         schema: &App<'a, 'a>,
         vcls: Vec<Box<dyn VirtualCommandLine>>,
     ) -> Result<MultiConfig<'a>, ConfiguratorError> {
+        let mut computed_value_names = HashSet::new();
         let initial: Box<dyn VirtualCommandLine> =
             Box::new(CommandLineVcl::new(vec![String::new()]));
-        let merged: Box<dyn VirtualCommandLine> = vcls
-            .into_iter()
-            .fold(initial, |so_far, vcl| merge(so_far, vcl));
+        let vlc_copy = &vcls;
```
dnwiebe: I think you can get rid of this isolation block and just do something like this:
```
vcls.iter().for_each(|vcl| {
    ...stuff...
}
```
The `iter()` won't take ownership of `vcls` the way `into_iter()` does, but it means that `vcl` will be a reference rather than an owned object. Since you're just reading `vcl` in this loop, that shouldn't be a problem.

masq_lib/src/multi_config.rs
```diff
@@ -61,24 +62,41 @@ impl<'a> MultiConfig<'a> {
...
+        {
+            vlc_copy.into_iter().for_each(|vcl| {
+                vcl.vcl_args().iter().for_each(|vcl_arg| {
+                    match vcl.is_computed() {
+                        true => computed_value_names.insert(vcl_arg.name().to_string()),
+                        false => computed_value_names.remove(&vcl_arg.name().to_string())
```
dnwiebe: I can't say, with the knowledge in my head, whether removing this value from the `computed_value_names` is correct. Can you come up with a few sensible, realistic tests involving data that can be both specified and computed, and run it through this code to see if the result makes sense?

I'm not suggesting that this code doesn't _work;_ I'm unsure whether the situation it creates is the one we want.

Example: if the user specifies the `data-directory` without a chain, do we add a computed value with the chain appended? If so, shouldn't that computed value win over the specified value? (I don't know.)

masq_lib/src/multi_config.rs
```diff
@@ -61,24 +62,41 @@ impl<'a> MultiConfig<'a> {
...
+                    };
+                })
+            });
+        }
+        let merged: Box<dyn VirtualCommandLine> = vcls.into_iter().fold(initial, |so_far, vcl | -> Box<dyn VirtualCommandLine> {
```
dnwiebe: You might not need to specify the return type for the closure. If you do, go ahead, but it seems to me that the definition of `fold()` should take care of that for you.

masq_lib/src/multi_config.rs
```diff
@@ -61,24 +62,41 @@ impl<'a> MultiConfig<'a> {
         schema: &App<'a, 'a>,
         vcls: Vec<Box<dyn VirtualCommandLine>>,
     ) -> Result<MultiConfig<'a>, ConfiguratorError> {
+        let mut computed_value_names = HashSet::new();
         let initial: Box<dyn VirtualCommandLine> =
```
dnwiebe, line 65: Could you move this down so that it's next to where it's used? It'd be easier to read.

masq_lib/src/multi_config.rs
```diff
@@ -139,6 +157,13 @@ impl<'a> MultiConfig<'a> {
         ConfiguratorError::required("<unknown>", &format!("Unfamiliar message: {}", e.message))
     }
 
+    pub fn is_user_specified(&self, value_name: &str) -> bool {
```
dnwiebe, line 157: In the old code, there's a macro that looks like this:
```
#[macro_export]
macro_rules! value_user_specified_m {
    ($m:ident, $v:expr, $t:ty) => {{
        let user_specified = $m.occurrences_of($v) > 0;
        let matches = $m.arg_matches_ref();
        match value_t!(matches, $v, $t) {
            Ok(v) => (Some(v), user_specified),
            Err(_) => (None, user_specified),
        }
    }};
}
```
That seems incompatible with this code. Do the two pass the same tests?

Further observation: The macro above does not seem to be used anywhere in the codebase anymore. (Check to make sure I'm right about this.) If it's not, there's no need to keep it around, and once it's deleted its meaning can be redefined as desired.

masq_lib/src/multi_config.rs
```diff
@@ -139,6 +157,13 @@ impl<'a> MultiConfig<'a> {
...
+        match self.computed_value_names.contains(value_name)  {
```
dnwiebe: This `match` could be replaced by 
```
!self.computed_value_names.contains(value_name)
```

masq_lib/src/multi_config.rs
```diff
@@ -139,6 +157,13 @@ impl<'a> MultiConfig<'a> {
...
+            true => false,
+            false => true
+        }
+    }
+
     pub fn occurrences_of(&self, parameter: &str) -> u64 {
```
dnwiebe, line 161: At this point, I think we can remove `occurrences_of()` (at least as a public function) and replace all its occurrences with calls to `is_user_specified()` since everywhere it's used, the return value is compared to zero.

masq_lib/src/multi_config.rs
```diff
@@ -263,6 +289,24 @@ pub fn merge(
     })
 }
 
+pub struct  ComputedVcl {
+    vcl_args: Vec<Box<dyn VclArg>>,
+}
+
+impl VirtualCommandLine for ComputedVcl {
+    fn vcl_args(&self) -> Vec<&dyn VclArg> {
+        vcl_args_to_vcl_args(&self.vcl_args)
```
dnwiebe, line 299: Not your fault but this might be clearer if it was named `get_vcl_args()`. That would probably make the one below `get_args()`. However, there's also an argument for letting sleeping dogs lie.

masq_lib/src/multi_config.rs
```diff
@@ -283,6 +327,33 @@ impl From<Vec<Box<dyn VclArg>>> for CommandLineVcl {
     }
 }
 
+impl ComputedVcl {
```
dnwiebe, line 331: You said you'd be rearchitecting this, so I'll keep quiet until I see the result.

masq_lib/src/multi_config.rs
```diff
@@ -360,6 +432,16 @@ impl VirtualCommandLine for ConfigFileVcl {
         vcl_args_to_args(&self.vcl_args)
     }
 }
+/*
```
dnwiebe: Commented-out code

node/src/daemon/setup_reporter.rs
```diff
@@ -495,8 +495,8 @@ impl SetupReporterReal {
                 Some(command_line) => command_line,
                 None => vec![],
             };
-            let (config_file_path, user_specified, _data_directory, _real_user) =
-                determine_fundamentals(dirs_wrapper, &app, &command_line)?;
+            let (config_file_path, user_specified, _data_directory, _data_directory_specified, _real_user, _real_user_specified, _preorientation_args) =
```
dnwiebe: There are so many items in this tuple now; is it now justified to turn it into a `struct`?

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
     fn configure(&self, multi_config: &MultiConfig) -> Result<T, ConfiguratorError>;
 }
 
-pub fn determine_fundamentals(
+fn config_file_real_user_data_dir_from_enumerate(configs: Vec<String>) -> (String, String, String) {
+    let config_match = argument_from_enumerate(configs.clone(), "--config-file".to_string()).unwrap_or("".to_string());
```
dnwiebe: Does this have to be `clone`d, or could it be just a reference? Also,  you might be able to get away with typing the second parameter as `&str` so that you don't have to `.to_string()` it.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
     fn configure(&self, multi_config: &MultiConfig) -> Result<T, ConfiguratorError>;
 }
 
-pub fn determine_fundamentals(
+fn config_file_real_user_data_dir_from_enumerate(configs: Vec<String>) -> (String, String, String) {
```
dnwiebe: "Enumerate" is a verb. Did you mean "enumeration?" Also, Bert suggests a return type of `(Option<String>, Option<String>, Option<String>)`.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
     dirs_wrapper: &dyn DirsWrapper,
     app: &App,
     args: &[String],
-) -> Result<(PathBuf, bool, PathBuf, RealUser), ConfiguratorError> {
-    let orientation_schema = App::new("MASQNode")
-        .arg(chain_arg())
-        .arg(real_user_arg())
-        .arg(data_directory_arg(DATA_DIRECTORY_HELP))
-        .arg(config_file_arg());
+) -> Result<(PathBuf, bool, PathBuf, bool, RealUser, bool, Box<dyn VirtualCommandLine>), ConfiguratorError> {
```
dnwiebe: This tuple is getting awfully long, don't you think?

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
+    let env_args = Box::new(EnvironmentVcl::new(&app)).args();
+    let args_to_vec = args.to_vec();
```
bertllll: This name just confuses people, especially because it is not an important piece of information that this particular method was used to form the data. Instead call it according to the state it got from the look at it afterwards, like: `vec_of_args`.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
     app: &App,
     args: &[String],
-) -> Result<(PathBuf, bool, PathBuf, RealUser), ConfiguratorError> {
-    let orientation_schema = App::new("MASQNode")
-        .arg(chain_arg())
-        .arg(real_user_arg())
-        .arg(data_directory_arg(DATA_DIRECTORY_HELP))
-        .arg(config_file_arg());
+) -> Result<(PathBuf, bool, PathBuf, bool, RealUser, bool, Box<dyn VirtualCommandLine>), ConfiguratorError> {
+    let env_args = Box::new(EnvironmentVcl::new(&app)).args();
```
bertllll: I would like more if you built this variable first inside the function you call "create_preorientation_args". It doesn't need to come in from outside. Especially not, if you already supply the "app" argument to the same fn, which you do need to get the `env_args`.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
+    let args_to_vec = args.to_vec();
+    let pre_orientation_args = match create_preorientation_args(env_args, args_to_vec.clone(), &app) {
```
bertllll: It feels fishy that you need to clone things that often. Do always your best to use only a slice of Vec instead of the whole Vec. ...` create_preorientation_args(env_args, &args_to_vec, &app)`...

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    content: Vec<String>
+}
+
+impl CombinedVcl {
+    fn len(&self) -> u32 {
+        *&self.content.as_slice().iter().count() as u32
+    }
+}
+
+fn create_preorientation_args(envargs: Vec<String>, argstovec: Vec<String>, app: &App) -> Result<Box<dyn VirtualCommandLine>, ConfigFileVclError> {
```
bertllll: Does this really need to say "preorientation" which is a word so vague from just a look at? I think it would help much better if you mentioned somewhat that this will serve to determine the location of the config file. 

Also, I realized that you've adopted the naming style with the word `orientation` ("orientation_schema") from the original code but you might not understand enough why it was chosen. I don't know the full story either but it probably refers to "getting oriented about where the config file lies", I'm not sure if preorientation wouldn't drift too away from what idea had borne. Please think about keeping the code clear for the future devs and in order to do that never be shy to express the functionality of the function with more meaningful words.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    let (env_config_file, env_data_dir, env_real_user) = config_file_real_user_data_dir_from_enumerate(envargs);
+    let (cmd_config_file, cmd_data_dir, cmd_real_user) = config_file_real_user_data_dir_from_enumerate(argstovec);
+    let mut combined_vcl: CombinedVcl = CombinedVcl { content: vec![] };
+    let combine_vcl = | name: String, vcl: &mut CombinedVcl, cmd_str: &String, env_str: &String | {
```
bertllll: This would pollute the space here less if it is a separate standalone function. Notice it captures no outer variable so we don't need a closure here, a classical function would do well.

node/src/server_initializer.rs
```diff
@@ -115,6 +118,7 @@ impl ResultsCombiner for RunModeResult {
     }
 }
 
+#[derive(Debug)]
 pub struct GatheredParams<'a> {
```
bertllll: This surprises me, didn't we say that GatheredParams should vanish? After the change in the MultiConfig (making it smarter and error-prone) we can finally return it alone and therefore the rest of the GatheredParams loses its purpose. For me, this clearly means that GatheredParams should be deleted.

node/src/server_initializer.rs
```diff
@@ -37,34 +38,36 @@ pub struct ServerInitializerReal {
 impl ServerInitializer for ServerInitializerReal {
     fn go(&mut self, streams: &mut StdStreams<'_>, args: &[String]) -> RunModeResult {
         let params = server_initializer_collected_params(self.dirs_wrapper.as_ref(), args)?;
+        let real_user = value_m!(params, "real-user", RealUser).unwrap();
```
bertllll: unwrap() is forbidden in production

This and the next line...twice.

node/src/server_initializer.rs
```diff
@@ -37,34 +38,36 @@ pub struct ServerInitializerReal {
...
             .combine_results(
                 self.bootstrapper
                     .as_mut()
-                    .initialize_as_privileged(&params.multi_config),
+                    .initialize_as_privileged(&params),
             );
 
         self.privilege_dropper
-            .chown(&params.data_directory, &params.real_user);
+            .chown(Path::new(data_directory.as_str()), &real_user);
```
bertllll, line 54: Why do you create the type `Path` when you want to return a `&str` but you have already a `String` in the variable (check it out). It is unnecessary!

node/src/server_initializer.rs
```diff
@@ -37,34 +38,36 @@ pub struct ServerInitializerReal {
...
 
-        self.privilege_dropper.drop_privileges(&params.real_user);
+        self.privilege_dropper.drop_privileges(&real_user);
 
-        result
+            result
```
bertllll: Can you try to pull the variable back to the left and see if the formatter will still be reacting on it?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1056,12 +1344,12 @@ mod tests {
...
-            .unwrap()
-            .data_directory
-            .to_string_lossy()
-            .to_string();
+            .unwrap();
 
-        assert_eq!(result, expected.unwrap());
+        assert_eq!(
+            value_m!(result, "data-directory", String).unwrap(),
+            expected.unwrap()
```
bertllll, line 1511: Isn't that a bit dangerous if you unwrap() and Option? Usually it is. 
It means this either does not need to be inside an Option (therefore it's always present) or you should be more careful and offer a way to deal with the possible None value. I think the first statement is right.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                    value_m!(env_multiconfig, "config-file", String).unwrap(),
+                    pwd.join(PathBuf::from( "generated/config.toml")).to_string_lossy().to_string()
+                )
+            },
+            (_, _) => ()
+        }
+    }
+
+    #[test]
+    fn server_initializer_collected_params_combine_vlcs_properly() {
```
bertllll, line 1134: ...combineS...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        running_test();
+        let home_dir = ensure_node_home_directory_exists( "node_configurator_standard","server_initializer_collected_params_combine_vlcs_properly");
+        let data_dir = &home_dir.join("data_dir");
+        let config_file = File::create(&home_dir.join("config.toml")).unwrap();
+        let current_directory = current_dir().unwrap();
+        let data_dir_sys = create_dir_all(PathBuf::from(home_dir.to_string_lossy().as_ref().to_owned() + "/data_dir/MASQ/polygon-mainnet"));
```
bertllll: This should be written with `join()`. 
```
home_dir
.join(data_dir)
.join(MASQ)
.join(polygon-mainnet)
```
Looks much simpler.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
         assert_eq!(config.crash_point, CrashPoint::Panic);
     }
 
+    fn fill_up_config_file(mut config_file: File) {
+        {
```
bertllll, line 863: What about this extra level of curly brackets? Please delete the pair.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        let home_dir = ensure_node_home_directory_exists( "node_configurator_standard","server_initializer_collected_params_combine_vlcs_properly");
+        let data_dir = &home_dir.join("data_dir");
+        let config_file = File::create(&home_dir.join("config.toml")).unwrap();
+        let current_directory = current_dir().unwrap();
+        let data_dir_sys = create_dir_all(PathBuf::from(home_dir.to_string_lossy().as_ref().to_owned() + "/data_dir/MASQ/polygon-mainnet"));
+        {
+            fill_up_config_file(config_file);
+            match data_dir_sys {
+                Ok(..) => {
+                    let config_file_vcl = File::create(PathBuf::from(home_dir.to_string_lossy().as_ref().to_owned() + "/data_dir/MASQ/polygon-mainnet").join("config.toml")).unwrap();
```
bertllll: Again, don't do this like this. Notice that home_dir is a PathBuf anyway and offers a lot of convenient methods. So, don't convert it to the string representation.

node/src/node_configurator/node_configurator_standard.rs
bertllll, line 1134: I'm sorry but this huge test can hardly read. I think you will need to spend more time refactoring it into smaller logical units (probably put in separate helper functions). Also, it's hard to distinguish the `act` of this test from the setup part and the assertions. Could you please make it look like three different parts?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                    fill_up_config_file(config_file_vcl);
+                }
+                Err(e) => panic!("unable to create config file: {}", e)
+            }
+        }
+
+        let env_vec_array = vec![
+            ("MASQ_CONFIG_FILE", "./generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home/config.toml"),
```
bertllll: I don't understand that this way of stating the path works for all three OS platforms? Does it?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+            //("MASQ_DATA_DIRECTORY", home_dir.to_str().unwrap()),
```
bertllll: Left over unneeded code.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+            #[cfg(not(target_os = "windows"))]
+            ("MASQ_REAL_USER", "9999:9999:booga"),
+        ];
+        env_vec_array.clone().into_iter()
```
bertllll: I'm surprised that you need to clone a vector of `&str`s. They act as references already. You should not need the clone. Use `.iter()` instead of `.into_iter()`. (I noticed you do this very often. Break the habit please)

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+            .for_each (|(name, value)| std::env::set_var (name, value));
+        let args = ArgsBuilder::new();
+        let args_vec: Vec<String> = args.into();
```
bertllll: This is probably an empty Vec even after this step, isn't it? If so please skip all this unnecessary stuff and write `vec![]` for short.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+        let mut env_data_directory = false;
+        for (item, _vaue) in env_vec_array.to_vec().as_slice() {
+            if item == &"MASQ_DATA_DIRECTORY" { env_data_directory = true; }
```
bertllll: Why would you want a condition like this in a test. You should always refrain from using conditional code in tests. It tastes wrong. Justify this or better, make it then clearer with a comment. However, think in first place why do you need to do this?  

I think it is for case the person running these tests on his device has the environment variable set on his system? But maybe rather because your various tests might collide with each other, setting this variable each in different moments (but sometimes when it overlaps). You need to know that there is a dev helper tool devised to prevent this. Look for `EnvironmentGuard`, put it at the beginning of your test according alike the places where you can find it already being used, that's what you want to have in your test anytime if it manipulates with the environment variables. Remember to use it now and also in the future.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        }
+
+        let result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let env_multiconfig = result.unwrap();
```
bertllll: Cannot you please rearrange this into a single line? You can call the unwrap() just after the act.

node/src/node_configurator/node_configurator_standard.rs
bertllll, line 1134: I've already spent more time with this test and I think the name really but really does not make it simpler to understand what the test is supposed to an evidence for. Think about giving it a better name or at worst you can add a comment under the name, describing its partial goals (because it seems there are many).

I think it should reflect that there are actually two different scenarios in the test (two calls of the tested function). Make sure you will express the difference between these two. 

Well, having been thinking even longer, I think the test would be clear if were **two tests**. Can you please divide it in case you cannot come to a clear reason why those two scenarios must be tested together? I couldn't check very closely but it left a feeling on me that _they don't refer to each other._

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+
+        let args = ArgsBuilder::new()
+            .param("--data-directory", current_directory.join(Path::new("generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home")).to_string_lossy().to_string().as_str())
+            .param("--real-user", "1001:1001:cooga");
+        let args_vec: Vec<String> = args.into();
+        let mut cmd_data_directory = false;
+        for item in args_vec.clone().as_slice() {
+            if item == &"--data-directory".to_string() { cmd_data_directory = true; }
```
bertllll: Hmm, again, what is this strange condition for? Cannot you at least explain it in a comment if it's so vital?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        }
+        let params = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let result2 = params.as_ref().expect("REASON");
```
bertllll: what do you mean by "REASON"... better message?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        let vcl_multiconfig = result2;
```
bertllll: This is funny you wasted two lines to state almost the same thing three times. Please make the assignment to `vcl_multiconfig` directly on a single line. Also, `params.as_ref().expect("REASON")` could be written with `.unwrap()` alone.

node/src/node_configurator/node_configurator_standard.rs
bertllll: I also noticed that a string like this one is used in the assertions. Can you make a variable out of it and reuse it here and down there as well?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+            ("MASQ_REAL_USER", "9999:9999:booga"),
+        ];
+        env_vec_array.clone().into_iter()
+            .for_each (|(name, value)| std::env::set_var (name, value));
+        let args = ArgsBuilder::new();
+        let args_vec: Vec<String> = args.into();
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+        let mut env_data_directory = false;
```
bertllll: This variable name should be aligned with the fact it holds a boolean. Something like `is_env_data_dir_set` ?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        }
+
+        let result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let env_multiconfig = result.unwrap();
+
+        let args = ArgsBuilder::new()
+            .param("--data-directory", current_directory.join(Path::new("generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home")).to_string_lossy().to_string().as_str())
+            .param("--real-user", "1001:1001:cooga");
+        let args_vec: Vec<String> = args.into();
+        let mut cmd_data_directory = false;
```
bertllll: This variable name doesn't make you expect a boolean, please modify it so.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        }
+        let params = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let result2 = params.as_ref().expect("REASON");
+        let vcl_multiconfig = result2;
+
+        assert_eq!(
+            value_m!(env_multiconfig, "config-file", String).unwrap(),
+            current_directory.join("generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home/config.toml").to_string_lossy().to_string()
+        );
+        match env_multiconfig.is_user_specified("--data-directory") {
```
bertllll: This kind of assertion where it checks things only sometimes should never take place in a test. 

First of all, do all you can do to prevent such behavior in your test. Hopefully the `EnvironmentGuard` I advised you will be enough...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+            true => {
+                assert_eq!(
+                    value_m!(env_multiconfig, "data-directory", String).unwrap(),
+                    "generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home".to_string()
+                )
+            }
+            false => {
+                println!("data-directory is not user specified in Environment");
```
bertllll: According to what I wrote just a few lines here above, this println! should be a panic. We don't want to tolerate these easements in the test. It should be a legitimate failure if it happens.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        #[cfg(not(target_os = "windows"))]
+        match env_multiconfig.is_user_specified("--real-user") {
+            true => {
+                assert_eq!(
+                    value_m!(env_multiconfig, "real-user", String).unwrap(),
+                    "1002:1002:/home/wooga".to_string()
+                )
+            }
+            false => {
+                println!("real-user is not user specified in Environment");
```
bertllll: Again, a legitimate failure. Use a panic and change other things so that the test doesn't behave unanticipatedly.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                ()
+            }
+        }
+
+        #[cfg(not(target_os = "windows"))]
+        match vcl_multiconfig.is_user_specified("--real-user") {
+            true => {
+                match env_multiconfig.is_user_specified("--real-user") {
```
bertllll: What?? You allow two different paths and you claim both be right. This is far from a firm test. Again, `EnvironmentGuard` should help you out.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                    false => {
+                        assert_eq!(
+                            value_m!(vcl_multiconfig, "real-user", String).unwrap(),
+                            "1002:1002:/home/wooga".to_string()
+                        )
+                    }
+                }
+            }
+            false => {
+                println!("real-user is not user specified in Command-line");
```
bertllll: Hmm, this should be undoubtly a panic. 

Okay, let's rephrase what you should do. You need to get rid of the sometimes-present factor at your parameters and so the flags for ENV_DATA_DIR and CMD_DATA_DIR. 

There must be a solid path thorough the test without multiple possibilities!!

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                ()
+            }
+        }
+        println!("env_data_directory {:#?}", env_data_directory);
```
bertllll: Left over debug fn.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        match vcl_multiconfig.is_user_specified("--data-directory") {
+            true => {
+                match env_data_directory && !cmd_data_directory {
```
bertllll: I would repeat myself here. This whole branching structure need to be rewritten to work without conditional truths.

bertllll (commented):
Check the functions `create_preorientation_args` and `server_initializer_collected_params`, seen just partly, and the two first tests.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
+    let data_dir_match = argument_from_enumerate(configs.clone(), "--data-directory".to_string()).unwrap_or("".to_string());
+    let real_user_match = argument_from_enumerate(configs, "--real-user".to_string()).unwrap_or("".to_string());
+    (config_match, data_dir_match, real_user_match)
+}
+
+fn argument_from_enumerate(configs: Vec<String>, needle: String) -> Option<String> {
+    let mut arg_match = None;
+    for (i, arg) in configs.iter().enumerate() {
+        if arg.as_str() == needle {
+            arg_match = Some(configs[i + 1].to_string())
```
bertllll: How can you be sure that every parameter has a value or not more than one.  (There is a big set of various args among which the ones you are searching here are just a small part. So I believe the problem is realistic.) If that goes true then your algorithm will hopelessly fail. You could easily end up remembering a value that would be actually a parameter name or for example a second value. I think it creates space for later developed errors.  

On the other hand, it is possible that you are confident this does not threaten. I'm okay if you just think a couple of mins about this.

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
 
-pub fn determine_fundamentals(
+fn config_file_real_user_data_dir_from_enumerate(configs: Vec<String>) -> (String, String, String) {
+    let config_match = argument_from_enumerate(configs.clone(), "--config-file".to_string()).unwrap_or("".to_string());
+    let data_dir_match = argument_from_enumerate(configs.clone(), "--data-directory".to_string()).unwrap_or("".to_string());
+    let real_user_match = argument_from_enumerate(configs, "--real-user".to_string()).unwrap_or("".to_string());
+    (config_match, data_dir_match, real_user_match)
+}
+
+fn argument_from_enumerate(configs: Vec<String>, needle: String) -> Option<String> {
```
bertllll: The use of the enumerate() fn has nothing to do with the purpose of this traversing algorithm. So there is no need to mention it in the name of this function. What about something similar to "find_optional_arg_and_return_value" or "return_arg_value_from_set_if_present"

node/src/node_configurator/mod.rs
```diff
@@ -29,19 +27,37 @@ pub trait NodeConfigurator<T> {
...
+    let mut arg_match = None;
+    for (i, arg) in configs.iter().enumerate() {
+        if arg.as_str() == needle {
+            arg_match = Some(configs[i + 1].to_string())
+        }
+    };
+    arg_match
+}
+
+pub fn determine_user_specific_data (
```
bertllll: From my view, all the input is user specific data, because its given by a certain unique person. 
If not like that, I think I understand what you hand in mind. 
What about "determine_users_data_sources_and_os_identity" or "determine_input_location_and_user".

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
-    let multi_config = make_new_multi_config(&orientation_schema, vec![Box::new(orientation_vcl)])?;
-    let config_file_path = value_m!(multi_config, "config-file", PathBuf)
-        .expect("config-file parameter is not properly defaulted by clap");
-    let user_specified = multi_config.occurrences_of("config-file") > 0;
-    let (real_user, data_directory_path, chain) =
-        real_user_data_directory_path_and_chain(dirs_wrapper, &multi_config);
-    let data_directory = match data_directory_path {
-        Some(data_dir) => data_dir,
-        None => data_directory_from_context(dirs_wrapper, &real_user, chain),
+    let (config_file, data_dir, mut real_user) = config_file_real_user_data_dir_from_enumerate(orientation_vcl.args());
```
bertllll: Again, the word "enumerate" tells the reader nothing about the purpose of this function. It might be saying something to you as the author, but other people don't care. 

Maybe you could drop "_from_enumerate".

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    let config_user_specified = !config_file.is_empty();
```
bertllll: Here we have a good example of why the version with the treatment of these as optional (With the use of Option<T>) would look neater. It would probably be `config_file_opt.is_some()` whilst if one sees `config_file.is_empty()`, the first thing they would think of is that the file _contains no data_.  Which is quite far from what we want. I do think you should use the Option type for your strings as it is most idiomatic in Rust.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    let data_directory_specified = !data_dir.is_empty();
+    let real_user_specified = !real_user.is_empty();
+    if real_user.is_empty() {
+        real_user = RealUser::new(None, None, None).populate(dirs_wrapper).to_string();
```
bertllll: This would look much nicer if you wrote it like 
```
let real_user = real_user_opt.unwrap_or_else(RealUser::new(None, None, None).populate(dirs_wrapper).to_string());
```

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    }
+    let chain = match argument_from_enumerate(orientation_vcl.args(), "--chain".to_string()) {
```
bertllll: Do you see the commonalities at this code block shared with what the fn `config_file_real_user_data_dir_from_enumerate` does within its guts? 
It uses heavily this fn `argument_from_enumerate` and both this part of the code and the big function take `orientation_vcl` as the source of the searched data. I think it would be reasonable to handle the chain parameter together with the others. More code of the same look would be at one place.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+        Some(chain) => {
+            Chain::from(chain.as_str())
+        },
+        None => DEFAULT_CHAIN
     };
-    let config_file_path = if config_file_path.is_relative() {
-        data_directory.join(config_file_path)
-    } else {
-        config_file_path
+    let real_user_split: Vec<&str> = real_user.split(":").collect();
```
bertllll: I think all the five lines starting with this one should become an outboard function. It is obviously lower level than the code that surrounds it.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+}
 
-    Ok((config_file_path, user_specified, data_directory, real_user))
+struct CombinedVcl {
+    content: Vec<String>
+}
+
+impl CombinedVcl {
+    fn len(&self) -> u32 {
+        *&self.content.as_slice().iter().count() as u32
```
bertllll: You certainly don't need to iterate through and use that many fn calls. Simply use the convenient method that Vec offers, that is `my_vec.len()`.

Also, in the Rust ecosystem it's a standard to return a `usize` in cases of counting elements etc. You would sound more idiomatic by using it.

node/src/node_configurator/mod.rs
bertllll: I'm not a fan of a solution like this where you make a String value mutable just to be able to modify it.
I'd rather pick the way of "shadowing values" by defining a new variable with the same name again and giving it the more up to date value. 

Seeing `mut` at a variable makes it always unclear concerning at what places it is mutated. Hopefully at the least possible, but when you see it first you don't know yet and I become cautious.

node/src/node_configurator/mod.rs
bertllll: Also, in idiomatic Rust, we don't write `&String` but always `&str` instead.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+        else if !env_str.is_empty() {
+            vcl.content.push(name);
+            vcl.content.push(env_str.to_string());
+        }
+        else {
+            vcl.content.push(name);
+            vcl.content.push("".to_string());
+        }
+    };
+    combine_vcl("--data-directory".to_string(), &mut combined_vcl, &cmd_data_dir, &env_data_dir);
```
bertllll: Supply just &str and convert it to String inside, it will be more symmetrical compared to the code which you already have inside and which also uses `to_string()` calls. If you do alike stuff close together it's almost always a good idea.

node/src/node_configurator/mod.rs
```diff
@@ -54,23 +70,115 @@ pub fn determine_fundamentals(
...
+    combine_vcl("--config-file".to_string(), &mut combined_vcl, &cmd_config_file, &env_config_file);
+    combine_vcl("--real-user".to_string(), &mut combined_vcl, &cmd_real_user, &env_real_user);
+
+    // In case we define config file in ENV and we can find real-user in that file, therefore we need
+    // to specify also data-directory in Environment or Command line, to be Node able figure out the
+    // config file location, when config file is specified without absolute path or current directory eg.: "./" or "../"
```
bertllll: I'm not so good at English either, but let's try this:

In situations where the config file path features in ENV but is not absolute (eg.: "./" or "../") and the real user parameter is in there too it is necessary to have the data directory set in ENV or on the command line to make the Node able to correctly recognize the config file location.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -139,32 +137,33 @@ fn collect_externals_from_multi_config(
...
+) -> Result<MultiConfig<'a>, ConfiguratorError> {
     let app = app_node();
+    let (config_file_path,
+        config_user_specified,
+        data_directory,
+        data_directory_specified,
+        real_user,
+        real_user_specified,
+        pre_orientation_args) = determine_user_specific_data(dirs_wrapper, &app, args)?;
+    let mut full_multi_config_vec: Vec<Box<dyn VirtualCommandLine>> = vec![
```
bertllll: This variable's name is confusing. Look where it goes in the end, this siunds as if this was already the final product. The truth is this is only a feed of the constructor of a multi config. So I'd expect something like `compete_multi_config_args`.

I was thinking that giving it the name "full" or "complete" might not be so appropriate because you still need to "fill it with more parameters" and only after it we can call it really "complete". But that's a detail.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -139,32 +137,33 @@ fn collect_externals_from_multi_config(
...
 
-    let (config_file_path, user_specified, data_directory, real_user) =
-        determine_fundamentals(dirs_wrapper, &app, args)?;
-
-    let config_file_vcl = match ConfigFileVcl::new(&config_file_path, user_specified) {
-        Ok(cfv) => Box::new(cfv),
-        Err(e) => return Err(ConfiguratorError::required("config-file", &e.to_string())),
+    let mut fill_specified_or_unspecified_box = |key: String, value: String, specified: bool | {
+        match specified {
+            true => full_multi_config_vec.push(Box::new(CommandLineVcl::new(vec!["".to_string(), key, value,]))),
```
bertllll: Yeah, the need of putting the empty string at the beginning, always, is truly ugly. I believe it can be fixed without too big efforts.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -139,32 +137,33 @@ fn collect_externals_from_multi_config(
...
-    )?;
-    let config_file_path =
-        value_m!(full_multi_config, "config-file", PathBuf).expect("defaulted param");
-    Ok(GatheredParams::new(
-        full_multi_config,
-        config_file_path,
-        real_user,
-        data_directory,
-    ))
+    fill_specified_or_unspecified_box("--config-file".to_string(), config_file_path.to_string_lossy().to_string(), config_user_specified);
```
bertllll: When you have the special structs holding the parameter and the user specified flag:

a) you can supply it here as a single thing, and get less params for this function.
b) you can go even further and give the struct another field holding the name of the parameter eg.: "--config-file".

If you go also with b) then you can make this function a method of that struct. It would look really compact.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -139,32 +137,33 @@ fn collect_externals_from_multi_config(
...
+            pre_orientation_args
+        ];
 
-    let (config_file_path, user_specified, data_directory, real_user) =
-        determine_fundamentals(dirs_wrapper, &app, args)?;
-
-    let config_file_vcl = match ConfigFileVcl::new(&config_file_path, user_specified) {
-        Ok(cfv) => Box::new(cfv),
-        Err(e) => return Err(ConfiguratorError::required("config-file", &e.to_string())),
+    let mut fill_specified_or_unspecified_box = |key: String, value: String, specified: bool | {
```
bertllll: Again, nobody really cares what data structure serves as the medium (whether a Box or a Vec), he can read it from the code easily. To express the purpose is always helpful, though. 

Maybe: `sort_out_by_whether_user_specified`.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -139,32 +137,33 @@ fn collect_externals_from_multi_config(
...
-        value_m!(full_multi_config, "config-file", PathBuf).expect("defaulted param");
-    Ok(GatheredParams::new(
-        full_multi_config,
-        config_file_path,
-        real_user,
-        data_directory,
-    ))
+    fill_specified_or_unspecified_box("--config-file".to_string(), config_file_path.to_string_lossy().to_string(), config_user_specified);
+    fill_specified_or_unspecified_box("--data-directory".to_string(), data_directory.to_string_lossy().to_string(), data_directory_specified);
+    fill_specified_or_unspecified_box("--real-user".to_string(), real_user.to_string(), real_user_specified);
```
bertllll: Again, a function call usually look better if without other function calls in the place of its arguments.  You can enhance it here if you supply `&str` instead of the first argument. You will simply call the `to_string()` method inside the function.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                .write_all(b"payment-thresholds = \"3333|55|33|646|999|999\"\n")
+                .unwrap();
+            config_file
+                .write_all(b"scan-intervals = \"111|100|99\"\n")
+                .unwrap()
+        }
+    }
+
+    #[test]
+    fn server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_path_started_by_dot() {
```
bertllll: "handleS"

You are mentioning the config file twice in the name, you can put the two parts together and get a simpler name.

bertllll: You will want the Environment guard here too because this tests refers to the environment setting.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        env_vec_array.clone().into_iter()
+            .for_each (|(name, value)| std::env::set_var (name, value));
+        let args = ArgsBuilder::new();
+        let args_vec: Vec<String> = args.into();
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+
+        let result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let env_multiconfig = result.unwrap();
```
bertllll, line 929: Why these two lines if it can be easily done in one?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+
+        match env_multiconfig.is_user_specified("--data-directory") {
+            true => (),
+            false => {
+                assert_eq!(
+                    value_m!(env_multiconfig, "data-directory", String).unwrap(),
+                    "/home/wooga/data_dir/MASQ/polygon-mainnet".to_string()
+                )
+            }
+        }
```
bertllll: Yeah these assertions are really bad. Get rid of the variability of the test's behavior. You need it to behave always the same and you need it to match a single, concrete result (again...`EnvironmentGuard`)

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+    }
+
+    #[test]
+    fn server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_path_started_by_dot() {
+        running_test();
+        let home_dir = ensure_node_home_directory_exists( "node_configurator_standard","server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_path_started_by_dot");
+        let data_dir = &home_dir.join("data_dir");
+        let config_file_relative = File::create(PathBuf::from("./generated").join("config.toml")).unwrap();
+        fill_up_config_file(config_file_relative);
```
bertllll, line 912: The test would be easier to understand if the reader saw what (supposed as different) real user value defines there. Maybe you could make it an argument your `fill_up_config_file` function and supply it from here. Then the reader can check both values easily. These seem important for this test.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+                    value_m!(env_multiconfig, "config-file", String).unwrap(),
+                    pwd.join(PathBuf::from( "generated/config.toml")).to_string_lossy().to_string()
+                )
+            },
+            (_, _) => ()
+        }
+    }
+
+    #[test]
+    fn server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_data_directory() {
```
bertllll: "handleS"

The name stated like this isn't really clear can you restate it a bit?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+        let env_multiconfig = result.unwrap();
+
+        match env_multiconfig.is_user_specified("--data-directory") {
+            true => {
+                assert_eq!(
+                    &value_m!(env_multiconfig, "data-directory", String).unwrap(),
+                    &current_directory.to_string_lossy().to_string()
+                )
+            },
+            false => ()
```
bertllll: This conditionality is wrong. Mentioned at the other tests...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +736,295 @@ mod tests {
...
+    fn server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_path_started_by_dot() {
+        running_test();
+        let home_dir = ensure_node_home_directory_exists( "node_configurator_standard","server_initializer_collected_params_handle_config_file_from_environment_and_real_user_from_config_file_with_path_started_by_dot");
+        let data_dir = &home_dir.join("data_dir");
+        let config_file_relative = File::create(PathBuf::from("./generated").join("config.toml")).unwrap();
+        fill_up_config_file(config_file_relative);
+        let env_vec_array = vec![
+            ("MASQ_CONFIG_FILE", "./generated/config.toml"),
+            #[cfg(not(target_os = "windows"))]
+            ("MASQ_REAL_USER", "9999:9999:booga"),
```
bertllll, line 916: What role does this parameter plays here? If none big one, you should remove it. If some real than think about what will happen for this test when running on Windows... It will be missing.

Bu the way, Windows doesn't need to deal with this business around the real-user, never. Maybe your tests, and now I really talk in a plural, should not run on Windows completely, or some of these??? Do think about that please.

masq_lib/src/multi_config.rs
czarte: `value_user_specified_m!` macro is definitely not comparable to `is_user_specified` method in `MultiConfig`.  `value_user_specified_m!` macro is returning true even if the value is `computed`.

masq_lib/src/multi_config.rs
czarte: I'm afraid it is not possible. `is_user_specified` is tracking only `user_specific_data`, that's means `data_directory`, `real_user` & `config_file`  so far. If we want to scale it to all data, we should add more config variables to this process or make the ComputedVcl global for all variables. That should go to the new card in my point of view.

czarte: Next thing is `is_user_specified` fn in `unprivileged_parse_args_configuration.rs` - this function could be replaced with `occurrences_of` or we should consider the comment above and solve it within the new card. That depends on the purpose of usage `is_user_specified` in `UnprivilegedParseArgsConfiguration::unprivileged_parse_args` where we use it to fill up `unprivileged_config
            .blockchain_bridge_config
            .blockchain_service_url_opt`
 and `unprivileged_config.blockchain_bridge_config.gas_price`

PR #345, opened 2023-09-21
GH-747: review two

bertllll:


node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -118,35 +120,41 @@ impl ReceivableDao for ReceivableDaoReal {
...
         Self::multi_row_update_from_received_payments(
-            self.conn.as_mut(),
-            &self.big_int_db_processor,
+            &mut *self.conn,
+            &*self.big_int_db_processor,
             &self.logger,
             timestamp,
             &payments,
         )
+        .unwrap_or_else(|e| {
```
dnwiebe: If this fails to record incoming payments, I think it should either panic (if the situation is unrecoverable) or return an error value (if the caller might be able to recover it).

dnwiebe (commented):
Resume at receivable_dao.rs:284 green

PR #346, opened 2023-09-23
GH-600-review-5

masqrauder:
This may not truly be review number 5 but there have been several before and 5 is sufficiently unique going forward.

### DIFF
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -336,7 +336,7 @@ impl BlockchainBridge {
             Err (e) => panic! ("Cannot retrieve start block from database; payments to you may not be processed: {:?}", e)
         };
         let max_block_count = match self.persistent_config.max_block_count() {
-            Ok(mbc) => mbc,
+            Ok(Some(mbc)) => mbc,
             _ => u64::MAX,
         };
         let end_block = match self.blockchain_interface.get_block_number() {
@@ -393,7 +393,7 @@ impl BlockchainBridge {
                 if let Some(max_block_count) = self.extract_max_block_count(e.clone()) {
                     debug!(self.logger, "Writing max_block_count({})", max_block_count);
                     self.persistent_config
-                        .set_max_block_count(max_block_count)
+                        .set_max_block_count(Some(max_block_count))
                         .map_or_else(
                             |_| {
                                 warning!(self.logger, "{} update max_block_count to {}. Scheduling next scan with that limit.", e, max_block_count);
@@ -1325,7 +1325,7 @@ mod tests {
             )))
             .get_block_number_result(LatestBlockNumber::Ok(U64::from(1234u64)));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(100_000))
+            .max_block_count_result(Ok(Some(100_000)))
             .start_block_result(Ok(5)); // no set_start_block_result: set_start_block() must not be called
         let mut subject = BlockchainBridge::new(
             Box::new(blockchain_interface),
@@ -1620,7 +1620,7 @@ mod tests {
             )));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1711,7 +1711,7 @@ mod tests {
             .get_block_number_result(latest_block_number);
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
@@ -1778,7 +1778,7 @@ mod tests {
             .get_block_number_result(Ok(0u64.into()));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(10000u64))
+            .max_block_count_result(Ok(Some(10000u64)))
             .start_block_result(Ok(6))
             .set_start_block_params(&set_start_block_params_arc)
             .set_start_block_result(Ok(()));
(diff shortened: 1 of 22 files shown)


node/src/blockchain/blockchain_bridge.rs
```diff
@@ -521,7 +531,7 @@ impl BlockchainBridge {
 
     pub fn extract_max_block_count(&self, error: BlockchainError) -> Option<u64> {
         let regex_result =
-            Regex::new(r".* (max:|more than|allowed for your plan:|limited to) (?P<max_block_count>\d+)(.*| blocks at once\.)")
+            Regex::new(r".* (max: |allowed for your plan: |is limited to |block range limit \()(?P<max_block_count>\d+).*")
```
dnwiebe, line 534: This strikes me as probably being narrowly tailored to a particular blockchain client service. Do we only ever use that service in real life, or are there others?

Furthermore, is there a way to tell that an error means the max block count is too big even if we don't get a specific message we can pull apart with a regex to find out what the maximum acceptable size is? If we could identify a range-too-big message (as opposed to an invalid-request message or a my-connection-is-down message or whatever), but the regex failed on it, we could just cut the max block count in half (or down to one-tenth, or whatever) and try again, right?  It seems as though eventually we should get it down to a size that works.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -2035,7 +2045,9 @@ mod tests {
 
     #[test]
     fn extract_max_block_range_from_pokt_error_response() {
-        let result = BlockchainError::QueryFailed("RPC error: Error { code: ServerError(-32064), message: \"You cannot query logs for more than 100000 blocks at once.\", data: None }".to_string());
+        // [{"jsonrpc":"2.0","id":4,"result":"0x2400ee1","error":{}},{"jsonrpc":"2.0","id":5,"error":{"code":-32001,"message":"Relay request failed validation: invalid relay request: eth_getLogs block range limit (100000 blocks) exceeded"}}]
```
dnwiebe: Commented-out code

node/src/blockchain/blockchain_interface.rs
```diff
@@ -344,7 +344,7 @@ where
             )
             .build();
 
-        let fallback_end_block_number = match end_block {
+        let fallback_start_block_number = match end_block {
             BlockNumber::Number(eb) => eb.as_u64(),
             _ => {
                 if let BlockNumber::Number(start_block_number) = start_block {
```
dnwiebe, line 350: I can't put this comment any lower, but it should be on the `panic!()` below. Can you expand and elaborate this error message? If I got it, I wouldn't have any idea what was wrong.

node/src/blockchain/blockchain_interface.rs
```diff
@@ -988,6 +985,7 @@ mod tests {
         let subject =
             BlockchainInterfaceWeb3::new(transport, event_loop_handle, TEST_DEFAULT_CHAIN);
 
+        let end_block_nbr = 1024u64;
```
dnwiebe, line 988: I'm puzzled. Obviously, `end_block_nbr` has been part of this test since before these changes, because it's used in the last assert. But I don't see where it was set before this line (absent until these changes) came into existence. Do you know where it came from before?

node/src/db_config/persistent_configuration.rs
```diff
@@ -416,7 +416,14 @@ impl PersistentConfiguration for PersistentConfigurationReal {
     }
 
     fn max_block_count(&self) -> Result<u64, PersistentConfigError> {
-        self.simple_get_method(decode_u64, "max_block_count")
+        match self.get("max_block_count") {
```
dnwiebe: I thought I saw somewhere else that the type of `max_block_count` in `PersistentConfiguration` was `Option<u64>`, and that it's initialized in a new database to `None`. Here, though, it seems non-optional. What am I missing?

Also, I saw a new test for the existing `set_max_block_count()` method, but I didn't see any new tests to drive the new logic here.

node/src/node_configurator/configurator.rs
```diff
@@ -554,6 +554,11 @@ impl Configurator {
             "earningWalletAddressOpt",
         )?;
         let start_block = Self::value_required(persistent_config.start_block(), "startBlock")?;
+        let max_block_count_opt =
+            match Self::value_required(persistent_config.max_block_count(), "maxBlockCount") {
+                Ok(value) => Some(value),
+                _ => None,
```
dnwiebe: Really? An error retrieving from the database (probably meaning the database is corrupt) is converted into an innocent `None` value? Are we sure we want to do this? Almost everywhere else in the code, a corrupt database triggers an instant panic, so that we don't keep running with a bad database and perhaps damage it further.

node/src/node_configurator/configurator.rs
```diff
@@ -2647,6 +2657,7 @@ mod tests {
                 "0x0123456789012345678901234567890123456789".to_string(),
             )))
             .start_block_result(Ok(3456))
+            .max_block_count_result(Ok(100000))
```
dnwiebe: I've seen a number of tests about the `Some()` side of the `max_block_count_opt` retrieval; is there at least one test of the `None` side?

node/src/blockchain/blockchain_bridge.rs
bertllll: Without fawning over Dan, this is an interesting idea. :) 

If you want Steve, you could make it into a card so that we avoid further delays.

node/src/blockchain/blockchain_interface.rs
masqrauder: This is the only supported usage of the enum BlockNumber `BlockNumber::Number(eb)`. The other 3 enum values are useless and unsupported here: `BlockNumber::Latest`, `BlockNumber::Earliest`, `BlockNumber::Pending`. This is a `panic!()` intended for the developer who calls `retrieve_transactions()` and passes in one of the unsupported values of the enum. It has to be a real block number. There's an earlier initial call to find out what the Latest block number is so it can be passed to this method. You may remember that originally I had the parameter `start_block` defined as `u64` however it seemed too strange and unexpected to have `start_block: u64` while `end_block: BlockNumber` so I defined them the same but protected from using the wrong one.

Should I change the message to something like:
`panic!("Developer: don't pass start_block as BlockNumber::Latest, BlockNumber::Earliest, or BlockNumber::Pending; they are not supported");
?

node/src/blockchain/blockchain_interface.rs
masqrauder: I may have refactored to allow the test to reuse the value in more than one place... part of the call to `subject.retrieve_transactions()` as `BlockNumber::Number(end_block_nbr.into())` and the value appears as part of the request payload `"fromBlock":"0x2a","toBlock":"0x400"` though not as obvious.

node/src/blockchain/blockchain_interface.rs
dnwiebe: I see. Thanks for the explanation! How about "`start_block must be BlockNumber::Number`"? That double colon helps me understand that the problem is one of not having the right enum value.

node/src/blockchain/blockchain_interface.rs
dnwiebe: Never mind: I see it now. `end_block_nbr` used to be just `42`. I swear the last time I looked at this code, the "`1 + end_block_nbr`" in the last assertion wasn't green. But it is now, and it's obviously been changed from "`42 + 1`", so it all makes sense now. Never mind. Sorry.

node/src/blockchain/blockchain_bridge.rs
masqrauder: Removed

node/src/db_config/persistent_configuration.rs
masqrauder: I'll comment on this in the next review I'm about to open. I did some rework here.

PR #349, opened 2023-10-01
GH-751: Hot fix for wrong books keeping at transactions confirmation 

bertllll:


node/src/accountant/scanners/mod.rs
```diff
@@ -319,23 +319,32 @@ impl PayableScanner {
...
-            .map(
-                |((rowid_opt, hash), pending_payable)| PendingPayableTriple {
-                    recipient: &pending_payable.recipient_wallet,
-                    hash,
-                    rowid_opt,
-                },
-            )
+            .sorted_by(|(_, hash_a), (_, hash_b)| Ord::cmp(&hash_a, &hash_b));
+
+        rowid_pairs_sorted_by_hashes
```
dnwiebe, line 353: Should there be some code in here to verify that the two lists of hashes are identical, and panic if they're not?

dnwiebe (approved):

dnwiebe (approved):

PR #350, opened 2023-10-04
GH-747: transactions received from unknown wallets 

bertllll:


utkarshg6 (commented):
First 13 files completed. No issues found.

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -89,7 +87,7 @@ impl PayableDaoFactory for DaoFactoryReal {
 #[derive(Debug)]
 pub struct PayableDaoReal {
     conn: Box<dyn ConnectionWrapper>,
-    big_int_db_processor: BigIntDbProcessor<Self>,
+    big_int_db_processor: BigIntDatabaseProcessorReal<Self>,
```
utkarshg6: I liked the old name better, because it's short and the short form `db` can easily be expanded to `database` in mind because it's very popular.

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -118,31 +122,55 @@ impl ReceivableDao for ReceivableDaoReal {
...
 
-    fn more_money_received(&mut self, timestamp: SystemTime, payments: Vec<BlockchainTransaction>) {
-        self.try_multi_insert_payment(timestamp, &payments)
-            .unwrap_or_else(|e| self.more_money_received_pretty_error_log(&payments, e))
+    fn more_money_received(
+        &mut self,
+        timestamp: SystemTime,
+        received_payments: &[BlockchainTransaction],
+    ) -> Box<dyn TransactionWrapper + '_> {
+        match self
```
utkarshg6: I don't understand what's the purpose of `match` over here. The way I understand how futures work is that we provide some code to execute in case of error inside `map_err()` and the success case inside `and_then()`. Since, we've already implemented both of these functions, I don't see what is the `match` keyword exactly matching on?

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -265,57 +293,104 @@ impl ReceivableDao for ReceivableDaoReal {
...
     pub fn new(conn: Box<dyn ConnectionWrapper>) -> ReceivableDaoReal {
         ReceivableDaoReal {
             conn,
-            big_int_db_processor: BigIntDbProcessor::default(),
+            big_int_db_processor: Box::new(BigIntDatabaseProcessorReal::default()),
             logger: Logger::new("ReceivableDaoReal"),
         }
     }
 
-    fn try_multi_insert_payment(
```
utkarshg6, line 280: Don't know yet, what happened to this function, let me know...

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -265,57 +293,104 @@ impl ReceivableDao for ReceivableDaoReal {
...
+            Err(ReceivableDaoError::RusqliteError(format!(
+                "Update for received payment with {amount} wei ran without producing a data change, \
+                despite a record for the wallet {address} exists"
+            )))
+        } else {
+            info!(
+                logger,
+                "Detected an inward transfer of {amount} wei in the block {block_number}. Address \
+                {address} does not match any of the tracked debtors. Ignoring"
+            );
```
utkarshg6: Maybe rephrase this which is closely related to the name of the card, something like:


```
Received {amount} wei from unknown debtor {address} in the block {block_number}.
```

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -265,57 +293,104 @@ impl ReceivableDao for ReceivableDaoReal {
...
+            Ok(())
         }
     }
 
+    // TODO you'd better check if there is a single row or more (meaning obviously terribly wrong)
```
utkarshg6: Maybe you want to work on this todo in this card. If so, maybe you can use the following SQL query:

```sql
SELECT COUNT(*) as row_count
FROM receivable
WHERE wallet_address = ?;
```

utkarshg6 (commented):
Continue on receivable_dao.rs:441

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -363,27 +438,29 @@ impl ReceivableDaoReal {
         )
     }
 
-    fn more_money_received_pretty_error_log(
-        &self,
+    fn more_money_received_roll_back_error_log(
```
utkarshg6: If I were writing this code, I would've migrated this function into a separate file and I've kept the nested functions outside this function. I find that way easier to read, but it's up to you if you want to do that.

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -481,10 +563,10 @@ mod tests {
     #[should_panic(
         expected = "Overflow detected with 340282366920938463463374607431768211455: cannot be converted from u128 to i128"
     )]
-    fn try_multi_insert_payment_handles_error_of_number_sign_check() {
+    fn multi_row_update_from_received_payments_handles_error_of_number_sign_check() {
```
utkarshg6: I'm worried if these test renames would cause trouble to someone else who have been working on this file in their card. In case someone is, then I'd suggest we keep the old names of these test renames as a comment. If not, feel free to ignore this comment.

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -705,8 +809,10 @@ mod tests {
             },
         ];
 
-        subject.more_money_received(payment_time, transactions);
+        let mut txn = subject.more_money_received(payment_time, &transactions);
 
+        txn.commit().unwrap();
+        drop(txn);
```
utkarshg6: I don't see a return type for `more_money_received()`. I'm confused how did you end up receiving the `txn` out of it and then managed to call functions on it, and ultimately drop it.

Secondly, I don't know what's the purpose of doing this exercise.

utkarshg6 (commented):
Continue on receivable_dao.rs:816

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -724,92 +830,349 @@ mod tests {
...
+            actual_record_3.balance_wei,
+            second_initial_balance as i128 - 9999
+        );
+        assert_eq!(
+            to_time_t(actual_record_3.last_received_timestamp),
+            to_time_t(time_of_change)
+        );
+        let log_handler = TestLogHandler::new();
+        log_handler.exists_log_containing(&format!(
+            "INFO: {test_name}: Detected an inward transfer of 2222 wei in the block 4446. \
```
utkarshg6: I don't know whether it's a good suggestion, but instead of logging shouldn't we create a new table to store txns from unknown wallets.

utkarshg6 (commented):
Continue on receivable_dao.rs:1104

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -724,92 +830,349 @@ mod tests {
...
+        // written into the database. In order to be faithful to the various distinct operations
+        // that take place in our code, we do care really a lot about the terminal stage of the
+        // transaction. If we don't make it to the 'commit' call for an issue having stepped in before,
+        // the staged changes will not engrave into the database. Which is exactly you can watch in
+        // this test.
+        //
+        // Another proof about the correct course of events is that even though we hadn't prepared
+        // a result for the mocked 'commit' method, the test didn't panic because of that. Besides
+        // that, if it had been committing each transaction separately we would've come across at
+        // least one call of 'commit' (and crashed), yet we didn't.
```
utkarshg6: I've rephrased this comment so that I can understand `TransactionWrapperMock` in a better way. If you'd like you can use the rephrased words in the comment. :pray:

If I've confused something, please correct me.

```rust
// This test uses the TransactionWrapperMock, a powerful tool that 
// simulates errors thrown from Statements in our monitored transaction, 
// playing a key role in the test. It functions as a typical mock, 
// but it goes beyond by simulating how a transaction behaves in reality, 
// especially regarding commit time or the object being dropped 
// without an explicit 'commit' call, which should leave the 
// database unmodified (using the roll-back strategy).

// Why is this important? 
// The mocked wrapper has its own transaction, 
// usable for multiple initial calls within the same piece of code. 
// If it interacts with a real transaction, all changes would be persistently 
// written to the database. Considering the various distinct operations 
// in our code, we particularly care about the final stage of the transaction. 
// If an issue prevents reaching the 'commit' call, the staged changes 
// won't be written to the database, as demonstrated in this test.

// Another confirmation of the correct sequence of events is that 
// even without a result prepared for the mocked 'commit' method, 
// the test doesn't panic. If it were committing each transaction 
// separately, we would encounter at least one 
// 'commit' call (resulting in a crash), but we didn't.
```

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -724,92 +830,349 @@ mod tests {
...
     }
 
     #[test]
-    fn receivable_account_status_works_when_account_doesnt_exist() {
+    fn verify_possibly_unknown_wallet_is_fatal_when_the_row_for_this_wallet_exists() {
```
utkarshg6: Maybe rename this test name to:

```rust
fn receivable_dao_throws_err_when_a_row_for_unknown_wallet_already_exists_in_the_db() {
    ...
}
```

node/src/accountant/db_access_objects/receivable_dao.rs
```diff
@@ -724,92 +830,349 @@ mod tests {
...
+        init_test_logging();
+        let test_name =
+            "verify_possibly_unknown_wallet_is_fatal_when_the_row_for_this_wallet_exists";
+        let home_dir = ensure_node_home_directory_exists("receivable", test_name);
+        let mut conn = DbInitializerReal::default()
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let wallet = make_wallet("blah");
```
utkarshg6, line 1142: let's rename this `wallet` to `unknown_wallet`.

utkarshg6 (commented):
Continue on `big_int_db_processor.rs`

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -1,43 +1,57 @@
...
 use std::fmt::{Debug, Display, Formatter};
 use std::iter::once;
 use std::marker::PhantomData;
 use std::ops::Neg;
 
+pub trait BigIntDatabaseProcessor<T>: Debug + Send
+where
+    T: TableNameDAO,
+{
+    fn execute<'a>(
```
utkarshg6: If I were writing this I might've tried to prevent using traits here. I do feel when we use traits, they end up biting us later. Let me know what you think of using a trait here.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -50,19 +64,22 @@ impl<'a, T: TableNameDAO> BigIntDbProcessor<T> {
     }
 }
 
-impl<T: TableNameDAO + 'static> Default for BigIntDbProcessor<T> {
-    fn default() -> BigIntDbProcessor<T> {
+impl<T: TableNameDAO + 'static> Default for BigIntDatabaseProcessorReal<T> {
```
utkarshg6: Also, I think, we can definitely use `db` for database here. It's a well known short form, so it'll be useful here.

`BigIntDatabaseProcessorReal` -> `BigIntDbProcessorReal`
`BigIntDatabaseProcessor` -> `BigIntDbProcessor`

utkarshg6 (commented):
Continue at `big_int_db_processor.rs:91`

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -71,12 +88,15 @@ impl<T: TableNameDAO> BigIntDbProcessor<T> {
     }
 }
 
-pub trait UpdateOverflowHandler<T>: Debug + Send {
-    fn update_with_overflow<'a>(
+pub trait UpdateOverflowHandler<T>: Debug + Send
+where
```
utkarshg6, line 98: I think, we can keep a pure function instead of a trait. It'll make it easier to maintain.

utkarshg6 (commented):
Continue at `big_int_db_processor.rs:169`

node/src/accountant/db_big_integer/mod.rs
```diff
@@ -2,5 +2,5 @@
 
 pub mod big_int_db_processor;
 pub mod big_int_divider;
-#[cfg(test)]
-mod test_utils;
+
+pub mod test_utils;
```
utkarshg6, line 5: I don't know why it was made public. Maybe it could be prevented. I think it would be good if we manage to do that.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -146,14 +166,17 @@ impl<T: TableNameDAO> UpdateOverflowHandler<T> for UpdateOverflowHandlerReal<T>
...
     ) {
-        match BigIntDbProcessor::<T>::prepare_statement(conn, config.overflow_update_clause)
-            .execute(execute_params)
-            .expect("logic broken given the previous non-overflow call accepted right")
+        match BigIntDatabaseProcessorReal::<T>::prepare_statement(
+            conn,
+            config.overflow_update_clause,
+        )
+        .execute(execute_params)
+        .expect("logic broken given the previous non-overflow call accepted right")
```
utkarshg6: Maybe you can use this statement, in case you like it...

```
If the previous non-overflow call was accepted, this should not happen. There is some flaw in logic.
```

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -253,28 +276,33 @@ impl<'a, T: TableNameDAO> BigIntSqlConfig<'a, T> {
     }
 }
 
-//to be able to display things that implement ToSql
-pub trait ExtendedParamsMarker: ToSql + Display {}
+// This is a native Rusqlite format for data supply to their interface
+type RusqliteParamPair<'params> = (&'params str, &'params dyn ToSql);
+// This is a representation of the above, plus providing a way to display the values (not possible natively)
+type DisplayableRusqliteParamPair<'params> = (&'params str, &'params dyn DisplayableParamValue);
```
utkarshg6: Since, we can't define what the first and second item of the tuple represent in `RusqliteParamPair` because it's part of the native format. I'd suggest, if we improve that by providing field names to the `DisplayableRusqliteParamPair` by changing it to struct.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -253,28 +276,33 @@ impl<'a, T: TableNameDAO> BigIntSqlConfig<'a, T> {
...
+
+macro_rules! implement_displayable_param_value {
     ($($implementer: ty),+) => {
-        $(impl ExtendedParamsMarker for $implementer {})+
+        $(impl DisplayableParamValue for $implementer {})+
     }
 }
 
-impl_of_extended_params_marker!(i64, &str, Wallet);
+implement_displayable_param_value!(i64, &str, Wallet);
```
utkarshg6: I'd prefer if we directly implement it over the types. It'll remove half the lines.

```rust
impl DisplayableParamValue for i64 {}
impl DisplayableParamValue for &str {}
impl DisplayableParamValue for Wallet {}
```

utkarshg6 (commented):
Continue at `big_int_db_processor.rs:296`

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -338,30 +366,36 @@ impl<'a> SQLParamsBuilder<'a> {
...
+        val: &'params dyn DisplayableParamValue,
     },
 }
 
-impl<'a> From<KnownKeyVariants<'a>> for (&'static str, &'static str, &'a dyn ExtendedParamsMarker) {
-    fn from(variant: KnownKeyVariants<'a>) -> Self {
+impl<'params> From<KeyVariants<'params>>
+    for (
+        &'static str,
+        &'static str,
```
utkarshg6: Instead of using static lifetimes, it would be better if we use the `String` type. That way the reference won't stay for the whole life of the Node and we'll get rid of the lifetimes (at least static).

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -384,10 +418,10 @@ impl Display for ByteMagnitude {
     }
 }
 
-pub struct SQLParams<'a> {
-    table_unique_key_name: &'a str,
+pub struct SQLParams<'params> {
```
utkarshg6, line 429: In case we can get rid of the lifetime `params` here, it would make the whole code lot easier to read. I believe, it's safe to use the `String` in-place of `&str`. We just need to find a good way to get rid of it from the `params_except_wei_change`.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -408,61 +442,59 @@ impl StdNumParamFormNamed {
     }
 }
 
-pub struct Param<'a> {
-    value_pair: (&'a str, &'a dyn ExtendedParamsMarker),
-    participates_in_overflow_clause: bool,
+pub enum Param<'params> {
+    BothClauses(DisplayableRusqliteParamPair<'params>),
+    MainClauseLimited(DisplayableRusqliteParamPair<'params>),
```
utkarshg6: Maybe you can just say `MainClause` and `BothClauses` here. Also, if we can get rid of the lifetime, the `enum` will become easier to read.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -408,61 +442,59 @@ impl StdNumParamFormNamed {
...
+            Param::BothClauses(pair) => pair,
+            Param::MainClauseLimited(pair) => pair,
         }
     }
 
-    fn as_rusqlite_params(&'a self) -> (&'a str, &'a dyn ToSql) {
-        (self.value_pair.0, &self.value_pair.1 as &dyn ToSql)
+    fn as_rusqlite_param_pair(&'params self) -> RusqliteParamPair<'params> {
+        let pair = self.param_pair();
+        (pair.0, &pair.1 as &dyn ToSql)
```
utkarshg6: If it was a struct instead of a tuple, it would be easier to understand what the first and the second element of the tuple mean. Also, a reference can be removed if we use a Box.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -408,61 +442,59 @@ impl StdNumParamFormNamed {
...
 }
 
-impl<'a> From<&'a WeiChangeAsHighAndLowBytes> for [(&'a str, &'a dyn ToSql); 2] {
-    fn from(wei_change: &'a WeiChangeAsHighAndLowBytes) -> Self {
+impl<'params> From<&'params WeiChangeAsHighAndLowBytes> for [RusqliteParamPair<'params>; 2] {
+    fn from(wei_change: &'params WeiChangeAsHighAndLowBytes) -> Self {
         [
             (wei_change.high.name.as_str(), &wei_change.high.value),
             (wei_change.low.name.as_str(), &wei_change.low.value),
         ]
```
utkarshg6, line 521: I think it deserves a type of it's own with four fields.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -481,35 +513,56 @@ pub enum WeiChange {
...
 
 #[derive(Debug, PartialEq, Eq)]
-pub struct BigIntDbError(pub String);
+pub enum BigIntDatabaseError {
+    General(String),
+    RowChangeMismatch {
+        row_key: String,
+        detected_count_changed: usize,
+    },
+}
```
utkarshg6, line 596: This was a good change. 🎉

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -654,17 +746,26 @@ mod tests {
                 low: StdNumParamFormNamed::new(":balance_low_b".to_string(), 9223372036854321124)
             }
         );
-        assert_eq!(result.params_except_wei_change[0].value_pair.0, ":some_key");
         assert_eq!(
-            result.params_except_wei_change[0].value_pair.1.to_string(),
+            result.params_except_wei_change[0].param_pair().0,
```
utkarshg6: I think there are some repeated assertions, if we can eliminate one of them, it would be easier to understand the test, and I believe the assertions will still be strong enough to break the test in case something in the production changes.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -736,10 +837,10 @@ mod tests {
...
-                Param::new(update_positive_2, true),
-                Param::new(update_negative_1, false),
-                Param::new(update_positive_1, true),
-                Param::new(update_negative_2, false),
+                Param::BothClauses(update_positive_2),
+                Param::MainClauseLimited(update_negative_1),
+                Param::BothClauses(update_positive_1),
+                Param::MainClauseLimited(update_negative_2),
             ],
         };
```
utkarshg6, line 926: I can't put this comment any lower than this, but this is related to the assertions below. I don't know why the result in such an order. In case there's a reason for that, maybe add a comment in the code.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -880,15 +981,18 @@ mod tests {
     #[derive(Debug, Default)]
     struct UpdateOverflowHandlerMock {
```
utkarshg6: I think this could be moved into a different file for `test_utils`.

node/src/accountant/db_big_integer/big_int_db_processor.rs
```diff
@@ -1512,14 +1616,14 @@ mod tests {
                 .build(),
         );
 
-        let result = BigIntDbProcessor::<DummyDao>::default()
+        let result = BigIntDatabaseProcessorReal::<DummyDao>::default()
```
utkarshg6: Maybe it's a good idea to use `Db` instead of `Database`.

node/src/accountant/db_big_integer/test_utils.rs
```diff
@@ -14,11 +23,38 @@ pub(in crate::accountant::db_big_integer) mod restricted {
...
         TestKey {
             var_name: "name",
             sub_name: ":name",
             val,
         }
     }
 }
+
+#[derive(Default, Debug)]
+pub struct BigIntDatabaseProcessorMock {
```
utkarshg6: In case the `execute()` function can be tested directly, then we can safely remove this mock.

node/src/accountant/mod.rs
```diff
@@ -1541,7 +1557,7 @@ mod tests {
             actual_prepared_adjustment
                 .original_setup_msg
                 .protected_qualified_payables,
-            protect_payables_in_test(affordable_accounts.clone())
+            initial_unadjusted_accounts
```
utkarshg6, line 1561: Previously it was adjusted accounts (inside `affordable_accounts`), now it is  unadjusted payments (inside ` initial_unadjusted_accounts`). I don't understand what exactly changed in the production code.

node/src/accountant/mod.rs
```diff
@@ -1862,13 +1882,21 @@ mod tests {
             from: make_wallet("wallet1"),
             wei_amount: 10000,
         };
-        let more_money_received_params_arc = Arc::new(Mutex::new(vec![]));
+        let transaction_id = ArbitraryIdStamp::new();
+        let transaction = TransactionWrapperMock::default()
```
utkarshg6: Maybe rename this variable to `transaction_wrapper` to avoid confusion.

node/src/accountant/mod.rs
```diff
@@ -1887,6 +1916,18 @@ mod tests {
         assert_eq!(
             *more_money_received_params,
             vec![(now, vec![expected_receivable_1, expected_receivable_2])]
+        );
+        let commit_params = commit_params_arc.lock().unwrap();
+        assert_eq!(*commit_params, vec![()]);
```
utkarshg6, line 1923: Just a suggestion, if you'd like we can just assert on the length of the vector, which will imply how many times the `commit()` function has been called. This empty `()` doesn't offer any value.

node/src/accountant/scanners/mod.rs
```diff
@@ -850,21 +860,12 @@ impl Scanner<RetrieveTransactions, ReceivedPayments> for ReceivableScanner {
...
-                .payments
-                .iter()
-                .fold(0, |so_far, now| so_far + now.wei_amount);
-            self.receivable_dao
-                .as_mut()
-                .more_money_received(message.timestamp, message.payments);
-            self.financial_statistics
-                .borrow_mut()
-                .total_paid_receivable_wei += total_newly_paid_receivable;
+            self.update_account_balances_start_block_and_stats(&message, logger)
```
utkarshg6: Maybe you can write a function named `update_start_block()` and another named `update_balances_and_stats()`. That way you can call the `update_start_block()` inside the if clause and the combination of both functions in the else clause.

node/src/accountant/scanners/mod.rs
```diff
@@ -894,10 +897,75 @@ impl ReceivableScanner {
...
             financial_statistics,
         }
     }
 
+    fn update_start_block_while_no_received_transactions_found(
+        &mut self,
+        new_start_block: u64,
+        logger: &Logger,
+    ) {
+        info!(
```
utkarshg6: I think this `info` should stay at a higher level and only the logic should stay inside this method.

node/src/accountant/scanners/mod.rs
```diff
@@ -894,10 +897,75 @@ impl ReceivableScanner {
...
+            logger,
+            "No new received payments were detected during the scanning process."
+        );
+        match self
+            .persistent_configuration
+            .set_start_block(new_start_block)
+        {
+            Ok(()) => debug!(logger, "Start block updated to {}", new_start_block),
+            Err(e) => panic!(
```
utkarshg6: I think it should be a warning or error log instead of a panic. This may lead the Node to die even though the condition is not fatal.

node/src/accountant/scanners/mod.rs
```diff
@@ -894,10 +897,75 @@ impl ReceivableScanner {
...
+        msg: &ReceivedPayments,
+        logger: &Logger,
+    ) {
+        let mut txn = self
+            .receivable_dao
+            .as_mut()
+            .more_money_received(msg.timestamp, &msg.payments);
+
+        let new_start_block = msg.new_start_block;
+        match self
```
utkarshg6, line 918: This code repetition can be avoided if we write functions as mentioned in a comment few comments above this one.

node/src/accountant/scanners/mod.rs
```diff
@@ -894,10 +897,75 @@ impl ReceivableScanner {
...
+                "Attempt to set new start block to {} failed due to: {:?}",
+                new_start_block, e
+            ),
+        }
+
+        match txn.commit() {
+            Ok(_) => {
+                debug!(logger, "Updated start block to: {}", new_start_block)
+            }
+            Err(e) => panic!("Commit of received transactions failed: {:?}", e),
```
utkarshg6, line 933: I'm unsure about this, but I think we can prevent the panic here too by just using a warning or an error log.

node/src/accountant/scanners/mod.rs
```diff
@@ -894,10 +897,75 @@ impl ReceivableScanner {
...
+        }
+        // At this moment the txn is useless; but we cannot have it consumed by its method
+        // because it qualifies as a trait object, therefore the 'self' must always be
+        // referenced
+        drop(txn);
```
utkarshg6: I don't quite understand the commented out lines.

First confusion, what does it mean by `self` must always be referenced.

Second confusion, why is it important for a trait object to be dropped like that.

node/src/accountant/scanners/mod.rs
```diff
@@ -1118,9 +1188,15 @@ mod tests {
         let pending_payable_dao_factory = PendingPayableDaoFactoryMock::new()
             .make_result(PendingPayableDaoMock::new())
             .make_result(PendingPayableDaoMock::new());
+        let receivable_dao = ReceivableDaoReal::new(Box::new(ConnectionWrapperMock::new()));
         let receivable_dao_factory =
-            ReceivableDaoFactoryMock::new().make_result(ReceivableDaoMock::new());
+            ReceivableDaoFactoryMock::new().make_boxed_result(Box::new(receivable_dao));
```
utkarshg6: Why is it inside a box again?

node/src/accountant/scanners/mod.rs
```diff
@@ -2947,35 +3033,83 @@ mod tests {
...
+        assert_eq!(*set_start_block_params, vec![4321]);
         TestLogHandler::new().exists_log_containing(&format!(
             "INFO: {test_name}: No new received payments were detected during the scanning process."
         ));
     }
 
+    #[test]
+    #[should_panic(expected = "Attempt to set new start block to 6709 failed due to: \
+    UninterpretableValue(\"Illiterate database manager\")")]
+    fn no_transactions_received_but_start_block_setting_fails() {
```
utkarshg6, line 3041: I've added a comment above, where I think it should be a warn or an error log instead of a panic. In case you decide to go that way, this test will need to be changed.

node/src/accountant/scanners/mod.rs
```diff
@@ -3005,16 +3140,99 @@ mod tests {
...
+        assert_eq!(*commit_params, vec![()]);
         TestLogHandler::new().exists_log_matching(
             "INFO: receivable_scanner_handles_received_payments_message: The Receivables scan ended in \\d+ms.",
         );
     }
 
+    #[test]
+    #[should_panic(expected = "Attempt to set new start block to 7890123 failed due to: \
+    DatabaseError(\"Fatigue\")")]
+    fn received_transactions_processed_but_start_block_setting_fails() {
```
utkarshg6, line 3136: This one too, it should be a log instead of a panic.

node/src/accountant/scanners/mod.rs
```diff
@@ -3005,16 +3140,99 @@ mod tests {
...
+
+        subject.finish_scan(msg, &Logger::new(test_name));
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "Commit of received transactions failed: SqliteFailure(Error { code: \
+    InternalMalfunction, extended_code: 0 }, Some(\"blah\"))"
+    )]
+    fn transaction_for_balance_start_block_updates_fails_on_its_commit() {
```
utkarshg6, line 3172: This too.

node/src/accountant/test_utils.rs
```diff
@@ -434,7 +441,11 @@ impl ReceivableDaoFactoryMock {
     }
 
     pub fn make_result(self, result: ReceivableDaoMock) -> Self {
-        self.make_results.borrow_mut().push(Box::new(result));
+        self.make_boxed_result(Box::new(result))
+    }
+
+    pub fn make_boxed_result(self, result: Box<dyn ReceivableDao>) -> Self {
```
utkarshg6: I think this function is unnecessary, we can directly do this inside the `make_result` function. While reading the code, we'll have all the logic in one place (because it's small) and that'll make it easier to understand what's going on.

node/src/accountant/test_utils.rs
```diff
@@ -1393,6 +1425,18 @@ impl PayableThresholdsGaugeMock {
     }
 }
 
+pub fn trick_rusqlite_with_read_only_conn(
```
utkarshg6, line 1424: Is it a hack?

node/src/actor_system_factory.rs
```diff
@@ -451,6 +451,7 @@ impl ActorFactory for ActorFactoryReal {
         let pending_payable_dao_factory = Box::new(Accountant::dao_factory(data_directory));
         let receivable_dao_factory = Box::new(Accountant::dao_factory(data_directory));
         let banned_dao_factory = Box::new(Accountant::dao_factory(data_directory));
+        let config_dao_factory = Box::new(Accountant::dao_factory(data_directory)); //TODO can you find a test where you could add this in?
```
utkarshg6: You left a todo. Maybe you need to look at it again.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -338,16 +338,6 @@ impl BlockchainBridge {
                 .retrieve_transactions(start_block, end_block, &msg.recipient);
         match retrieved_transactions {
             Ok(transactions) => {
-                debug!(
-                    self.logger,
-                    "Write new start block: {}", transactions.new_start_block
-                );
-                if let Err(e) = self
-                    .persistent_config
-                    .set_start_block(transactions.new_start_block)
```
utkarshg6, line 347: The start block was previously set here but now it's been set inside the scanners. I don't get it why did you decided to make this change?

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
```diff
@@ -0,0 +1,314 @@
...
+use crate::blockchain::blockchain_interface::BlockchainError;
+use crate::sub_lib::wallet::Wallet;
+use futures::Future;
+use std::rc::Rc;
+use web3::contract::{Contract, Options};
+use web3::transports::Batch;
+use web3::types::BlockNumber;
+use web3::{BatchTransport, Web3};
+
+pub struct RPCHelpersWeb3<T>
```
utkarshg6: I don't know why this code is under this card. I'd appreciate if we keep this code inside a different card. This will help me and @Syther007 to work with a sigh of relief that another merge in master won't introduce new calls to the `wait()` function to fix.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
```diff
@@ -0,0 +1,314 @@
...
+    // TODO waiting for GH-707 (note: consider to query the balances together with the id)
+    _batch_web3: Rc<Web3<Batch<T>>>,
+    contract: Contract<T>,
+}
+
+impl<T> RPCHelpers for RPCHelpersWeb3<T>
+where
+    T: BatchTransport,
+{
+    fn get_transaction_fee_balance(&self, wallet: &Wallet) -> ResultForBalance {
```
utkarshg6: This is for a future reference. If we use the return types that are easier to use, by that I means returning a plain Result of U256 instead of a data structure inside the web3 library. It'll make our job easier in terms of writing tests, then we won't need to worry about to find a way to create these types inside tests.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
```diff
@@ -0,0 +1,314 @@
...
+                wallet.address(),
+                None,
+                Options::default(),
+                None,
+            )
+            .map_err(|e| BlockchainError::QueryFailed(format!("{} for wallet {}", e, wallet)))
+            .wait()
+    }
+
+    fn get_transaction_id(&self, wallet: &Wallet) -> ResultForNonce {
```
utkarshg6: And a better name would be `get_transaction_nonce` here, the `id` is confusing, it can be misunderstood.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
```diff
@@ -0,0 +1,314 @@
...
+        assert_eq!(
+            bodies[0]["params"][0]["to"].to_string(),
+            format!("\"{:?}\"", contract_address),
+        );
+        assert_eq!(bodies.len(), 1)
+    }
+
+    #[test]
+    #[should_panic(expected = "No address for an uninitialized wallet!")]
+    fn get_masq_balance_returns_err_for_an_invalid_wallet() {
```
utkarshg6: I think a lot of these tests are an overhead. We're testing the `web3` library over here, instead of some new logic that we've written.

node/src/blockchain/blockchain_interface/rpc_helpers.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+use crate::blockchain::blockchain_interface::BlockchainResult;
+use crate::sub_lib::wallet::Wallet;
+use web3::types::U256;
+
+pub trait RPCHelpers {
```
utkarshg6: Let's find a common ground and migrate this code to a new card, because if this code says under this card, it'll cause conflicts with GH-744, and those conflicts won't be easy to resolve due to multiple calls to the `wait()` function.

I hope this comment aligns with your original idea to eliminate all the calls to the `wait()` function.

node/src/database/db_migrations/db_migrator.rs
```diff
@@ -125,16 +124,16 @@ impl DbMigratorReal {
...
+        transaction
+            .prepare(&format!(
                 "UPDATE {} SET value = {} WHERE name = 'schema_version'",
                 name_of_given_table, update_to
-            ),
-            [],
-        )?;
+            ))
+            .expect("internal rusqlite error")
+            .execute([])?;
```
utkarshg6, line 136: What does this empty box (`[]`) do?

node/src/database/db_migrations/migrator_utils.rs
```diff
@@ -98,14 +101,14 @@ impl<'a> DBMigrationUtilities for DBMigrationUtilitiesReal<'a> {
 }
 
 struct DBMigDeclaratorReal<'a> {
```
utkarshg6, line 100: Maybe rename the name of the lifetime here, or remove it if possible.

node/src/database/db_migrations/migrator_utils.rs
```diff
@@ -176,9 +179,16 @@ pub struct StatementWithRusqliteParams {
 }
 
 impl StatementObject for StatementWithRusqliteParams {
-    fn execute(&self, transaction: &Transaction) -> rusqlite::Result<()> {
+    fn execute(&self, transaction: &dyn TransactionWrapper) -> rusqlite::Result<()> {
```
utkarshg6: The argument name is `transaction`, I believe we should rename it to `transaction_wrapper`.

node/src/database/rusqlite_wrappers.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    fn execute(&self, query: &str, params: &[&dyn ToSql]) -> Result<usize, Error> {
+        self.transaction
+            .as_ref()
+            .expectv("rusqlite transaction")
+            .execute(query, params)
+    }
+
+    fn commit(&mut self) -> Result<(), Error> {
```
utkarshg6, line 162: Instead of making the `transaction` an option, maybe you can consume the whole object when you commit it.

node/src/database/test_utils/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::database::db_initializer::{DbInitializer, InitializationError};
+use crate::database::rusqlite_wrappers::{ConnectionWrapper, TransactionWrapper};
+use rusqlite::{Error, Statement};
+use std::cell::RefCell;
+use std::fmt::Debug;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex};
+
+#[derive(Debug, Default)]
+pub struct ConnectionWrapperMock<'a> {
```
utkarshg6: Either we should rename the lifetime, or try to get rid of it. I personally prefer the latter.

node/src/database/test_utils/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    prepare_params: Arc<Mutex<Vec<String>>>,
+    prepare_results: RefCell<Vec<Result<Statement<'a>, Error>>>,
+    transaction_results: RefCell<Vec<Result<Box<dyn TransactionWrapper + 'a>, Error>>>,
```
utkarshg6: Instead of a dynamic dispatch, if you only use the `TransactionWrapperMock` then you can get rid of the `unsafe` on Line 23 in this file. Since it's a test utility, I believe it'll be ideal to do that.

node/src/database/test_utils/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+impl<'a> ConnectionWrapper for ConnectionWrapperMock<'a> {
+    fn prepare(&self, query: &str) -> Result<Statement, Error> {
+        self.prepare_params
+            .lock()
+            .unwrap()
+            .push(String::from(query));
+        self.prepare_results.borrow_mut().remove(0)
+    }
+
+    fn transaction<'b>(&'b mut self) -> Result<Box<dyn TransactionWrapper + 'b>, Error> {
```
utkarshg6: Either rename this lifetime to a more appropriate name (suggestion: `tx`) or maybe eliminate it if possible.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+            .as_ref()
+            .unwrap()
+            .produce_statement(prod_code_query)
+    }
+
+    fn execute(&self, _query: &str, _params: &[&dyn ToSql]) -> Result<usize, Error> {
+        unimplemented!("not needed yet")
+    }
+
+    fn commit(&mut self) -> Result<(), Error> {
```
utkarshg6, line 100: I believe this `commit()` function can be re-written using the if-else and it'll make it easier to read.

```rust
    fn commit(&mut self) -> Result<(), Error> {
        self.commit_params.lock().unwrap().push(());
        let next_result = self.commit_results.borrow_mut().remove(0);
        if next_result.is_ok() {
            if let Some(prepared_results) = self.prepare_results_opt.as_mut() {
                if let Either::Left(for_both) = &mut prepared_results.setup {
                    return for_both.commit_prod_calls();
                }
            }
        }

        next_result
    }
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+                Either::Right(_) => next_result,
+            },
+            _ => next_result,
+        }
+    }
+
+    arbitrary_id_stamp_in_trait_impl!();
+}
+
+// TODO curate this text (probably not up to date)
```
utkarshg6: Here's the rephrased version of this text. I hope you'd like to use this one.
 
```
// TODO: Update this text (it might be outdated)
// Trying to store a rusqlite 'Statement' inside the TransactionWrapperMock and then placing this
// combination into the ConnectionWrapperMock wasn't effective. This approach led to complex problems
// related to object lifetimes. We tried many times to fix these issues, but success depended on
// creating numerous new explicit lifetimes. Some of these lifetimes required strict hierarchical
// relationships with others.
// Eventually, we decided to take a different approach. The most feasible solution was to reduce
// the excessive use of borrowed references.
// To achieve this, we had to make the mock smarter by giving it its own database connection. This
// connection acts as a ready-to-use source for creating native rusqlite Statements. This was necessary
// because we couldn't find any alternative to letting the rusqlite library build the Statement
// using their standard methods. Additionally, our attempts to create a StatementWrapper failed due
// to solid compilation problems. These problems arose from the way generic arguments were spread
// across the original Statement implementation, which we couldn't replicate in our wrapper. Also,
// we couldn't bypass these generics. In Rust, a trait with generics in its methods isn't valid,
// while we generally use trait objects.
 
// With that said, we're relieved to have a working solution now. The most challenging aspect of
// this mock system is the 'prepare' method. Usually, you won't need an error to occur in this
// method because the production code often handles the results from 'prepare' using 'expect'. As
// you might know, we don't require using 'expect' for writing new tests. Therefore, there's little
// benefit in causing an error that would only trigger a panic due to 'expect'.
 
// From the above, it's clear that we don't need to worry about generating errors in the 'prepare'
// method's return. Our focus should be on the Statement produced by this method. Although it might
// seem trivial at first, the 'prepare' method can significantly influence the result of the next
// function call made using this Statement. Fortunately, despite some challenges, we can indirectly
// shape this to meet our requirements.
// This indirectly prepared Statement can then lead to a useful error, simplifying our test writing.
// For instance, consider how this applies to methods like 'execute', 'query_row', or 'query_map'.
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+// Statement coming out from this method. Even though looking perhaps a little nonsensical at first,
+// the 'prepare' method can have a strong impact on the result returned from the next, upstream
+// function call, to happen on top of this Statement, and which, luckily for us (despite certain
+// difficulties), can be indirectly shaped this way towards our requirements.
+// This distantly prepared Statement can then cause a certainly useful error, easing our test
+// writing. For some examples let's consider the following methods 'execute', 'query_row' or
+// 'query_map'.
+
+#[derive(Debug)]
+struct SetupForStubbed<S> {
```
utkarshg6: We're using `String` in-place of the generic `S`. I believe it's safe to just use the `String` here and remove the generic.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+    stubbed: SetupForStubbed<Option<String>>,
+}
+
+impl SetupForBoth {
+    fn commit_prod_calls(&mut self) -> Result<(), Error> {
+        let txn_opt = self.prod_code_calls_transaction_opt.take();
+        let mut txn = txn_opt
+            .expect("Error: missing transaction in setup for both prod code and stubbed calls");
+        txn.commit()
+    }
```
utkarshg6, line 215: I saw an `opt` created out of opt and it seems confusing. To prevent this confusion, I suggest if we use a `match` here.


```rust
    fn commit_prod_calls(&mut self) -> Result<(), Error> {
        match self.prod_code_calls_transaction_opt.take() {
            Some(mut txn) => txn.commit(),
            None => {
                panic!("Error: missing transaction in setup for both prod code and stubbed calls")
            }
        }
    }
 ```

File: node/src/database/test_utils/transaction_wrapper_mock.rs [hunk 64]
```diff
@@ -0,0 +1,394 @@
...
+}
+
+impl Drop for SetupForBoth {
+    fn drop(&mut self) {
+        // Making sure that the referenced transaction will deconstruct
+        // before the connection it was pointing to
+        drop(self.prod_code_calls_transaction_opt.take())
```
utkarshg6: I still don't understand why we need to explicitly do that. If we can prevent it's usage as well as it's implementation we can avoid some confusions.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+#[derive(Debug)]
+pub struct PrepareMethodResults {
+    calls_counter: RefCell<usize>,
+    setup: Either<SetupForBoth, SetupForStubbed<String>>,
+}
+
+impl PrepareMethodResults {
+    pub fn new_with_both_prod_code_and_stubbed_calls(
+        prod_code_calls_conn: Box<dyn ConnectionWrapper>,
+        stubbed_calls_conn: Box<dyn ConnectionWrapper>,
```
utkarshg6: Maybe `stubbed` is not the word you're looking for. As I looked over the Internet it means `the short part of something that is left after the main part has been used`.

Now, if this call was the short part, what was the main part, that's confusing to me.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+                stubbed: SetupForStubbed {
+                    stubbed_calls_conn,
+                    stubbed_calls_optional_statements_literals: vec![],
+                },
+            };
+
+            let conn = unsafe { ptr.as_mut().unwrap() };
+            let txn = conn.transaction().unwrap();
+
+            setup.prod_code_calls_transaction_opt = Some(txn);
```
utkarshg6: Using raw pointer is risky, and you're using the unsafe code. We have a way around this thing, by using the `Box::leak()`. This below mentioned code isn't perfect and maybe you can find a way to do that for this dynamic dispatch.

```rust
    let setup = {
        let conn = Box::leak(mem::replace(&mut prod_code_calls_conn, Box::new(/* Create the new connection */)));
        let mut setup = SetupForBoth {
            prod_code_calls_conn_used_for_both: false,
            prod_code_calls_conn: conn.clone(),
            prod_code_calls_transaction_opt: None,
            requested_preceding_prod_code_calls: 0,
            stubbed: SetupForStubbed {
                stubbed_calls_conn,
                stubbed_calls_optional_statements_literals: vec![],
            },
        };

        let txn = conn.transaction().unwrap();
        setup.prod_code_calls_transaction_opt = Some(txn);

        Either::Left(setup)
    };
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+            stubbed_calls_optional_statements_literals: vec![],
+        });
+
+        Self {
+            calls_counter: RefCell::new(0),
+            setup,
+        }
+    }
+
+    pub fn count_of_initial_prod_code_calls(mut self, prod_code_calls: usize) -> Self {
```
utkarshg6: This function name should be renamed to `set_initial_prod_code_call_count`, it's easier to understand. The name you've used suggest as if it's just a getter function.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+        let dual_setup = self.setup_for_both_or_panic_ref_mut();
+        if dual_setup.requested_preceding_prod_code_calls == 0 {
+            dual_setup.requested_preceding_prod_code_calls = prod_code_calls
+        } else {
+            panic!("Use only single call of \"number_of_prod_code_calls!\"")
```
utkarshg6: The panic message should also be changed to this:

```rust
panic!("Use only a single call to set the initial production code call count!");
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+        Self {
+            calls_counter: RefCell::new(0),
+            setup,
+        }
+    }
+
+    pub fn count_of_initial_prod_code_calls(mut self, prod_code_calls: usize) -> Self {
+        let dual_setup = self.setup_for_both_or_panic_ref_mut();
+        if dual_setup.requested_preceding_prod_code_calls == 0 {
+            dual_setup.requested_preceding_prod_code_calls = prod_code_calls
```
utkarshg6: Also, change this filed name to this:

```rust
requested_preceding_prod_code_calls --> initial_prod_code_call_count
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+            panic!("Use only single call of \"number_of_prod_code_calls!\"")
+        }
+        self
+    }
+
+    pub fn prod_code_calls_conn_used_for_both(mut self) -> Self {
+        let dual_setup = self.setup_for_both_or_panic_ref_mut();
+        dual_setup.prod_code_calls_conn_used_for_both = true;
+        self
+    }
```
utkarshg6: I want you to make some renames here too:

```rust
pub fn mark_prod_code_calls_as_used_for_both(mut self) -> Self {
    let dual_setup = self.get_setup_for_both_as_mut_or_panic();
    dual_setup.is_prod_code_calls_conn_used_for_both = true;
    self
}
```

also change the name of the function used in this function:

```rust
fn get_setup_for_both_as_mut_or_panic(&mut self) -> &mut SetupForBoth {
    match self.setup.as_mut() {
        Either::Left(both) => both,
        Either::Right(_) => Self::dual_setup_mismatch(),
    }
}
```

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    fn dual_setup_mismatch() -> ! {
+        panic!(
+            "{}",
+            "This mock was construct as intended for the setup with both prod code and stubbed \
+            calls. Reconsider that step, if you think you will be using only the stubbed calls"
+        )
+    }
```
utkarshg6: This function is unnecessary, we can store this string inside a constant, and use it wherever we want to panic with this message.

It'll be easier to see at those places that the certain function will panic, at the moment this information is abstracted away.

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    fn setup_for_both_or_panic_ref(&self) -> &SetupForBoth {
```
utkarshg6: This should be renamed to:

```rust
get_setup_for_both_as_ref_or_panic
```

This name conveys that
1. It's a getter function
2. It resembles the `as_ref()` function that rust offers
3. It tells that the code may panic

node/src/database/test_utils/transaction_wrapper_mock.rs
```diff
@@ -0,0 +1,394 @@
...
+        match self.setup.as_ref() {
+            Either::Left(both) => both,
+            Either::Right(_) => Self::dual_setup_mismatch(),
+        }
+    }
+
+    fn setup_for_both_or_panic_ref_mut(&mut self) -> &mut SetupForBoth {
```
utkarshg6: This should be renamed to:

```rust
get_setup_for_both_as_mut_or_panic
```

Similarly, This name conveys that
1. It's a getter function
2. It resembles the `as_mut()` function that rust offers
3. It tells that the code may panic

utkarshg6 (commented):
Continue on `transaction_wrapper_mock.rs:256`

node/src/db_config/config_dao.rs
```diff
@@ -41,6 +41,12 @@ pub trait ConfigDao {
     fn get_all(&self) -> Result<Vec<ConfigDaoRecord>, ConfigDaoError>;
     fn get(&self, name: &str) -> Result<ConfigDaoRecord, ConfigDaoError>;
     fn set(&self, name: &str, value: Option<String>) -> Result<(), ConfigDaoError>;
+    fn set_through_provided_transaction(
```
utkarshg6: Let's rename this to `set_config_through_transaction`

node/src/db_config/config_dao.rs
```diff
@@ -65,23 +71,38 @@ impl ConfigDao for ConfigDaoReal {
...
+        Self::set(prepare_stm, name, value)
     }
 }
 
 impl ConfigDaoReal {
     pub fn new(conn: Box<dyn ConnectionWrapper>) -> ConfigDaoReal {
         ConfigDaoReal { conn }
     }
+
+    fn set<'a, P>(prepare: P, name: &str, value: Option<String>) -> Result<(), ConfigDaoError>
```
utkarshg6: A method with same name is already there, let's rename it to something else.

node/src/db_config/config_dao.rs
```diff
@@ -212,25 +227,22 @@ mod tests {
...
             true,
         );
+
         subject
             .set(
                 "consuming_wallet_private_key",
                 Some("Two wrongs don't make a right, but two Wrights make an airplane".to_string()),
             )
             .unwrap();
-
```
utkarshg6, line 231: We need a blank line here, to separate the act, maybe you removed it by mistake, also you can eliminate the blank line above the assert statements.

node/src/db_config/config_dao.rs
```diff
@@ -279,4 +285,46 @@ mod tests {
...
+            result_before_commit,
+            ConfigDaoRecord::new(
+                "schema_version",
+                Some(&CURRENT_SCHEMA_VERSION.to_string()),
+                false
+            )
+        );
+        txn.commit().unwrap();
+        let result_after_commit = assert_dao.get("schema_version").unwrap();
+        assert_ne!(CURRENT_SCHEMA_VERSION, 3);
```
utkarshg6: Let's move this line to the top of the test. It would make it easier to read.

node/src/db_config/config_dao_null.rs
```diff
@@ -283,4 +295,41 @@ mod tests {
...
+    fn set_works_simple() {
+        let subject = ConfigDaoNull::default();
+
+        subject.set("param1", Some("value1".to_string())).unwrap();
+        subject.set("schema_version", None).unwrap();
+        subject
+            .set("schema_version", Some("456".to_string()))
+            .unwrap();
+
+        let subject_data_sorted = subject.data.iter().sorted().collect::<Vec<(_, _)>>();
```
utkarshg6, line 309: Let's consider if the `set()` was modifying the above mentioned changes, the assert would still succeed because we just took the same reading twice.

It would be a stronger assertion if we move the `subject_data_sorted` before the act.

node/src/db_config/config_dao_null.rs
```diff
@@ -283,4 +295,41 @@ mod tests {
...
+            .set_through_provided_transaction(&mut txn, "param1", Some("value1".to_string()))
+            .unwrap();
+        subject
+            .set_through_provided_transaction(&mut txn, "schema_version", None)
+            .unwrap();
+        subject
+            .set_through_provided_transaction(&mut txn, "schema_version", Some("456".to_string()))
+            .unwrap();
+
+        let subject_data_sorted = subject.data.iter().sorted().collect::<Vec<(_, _)>>();
```
utkarshg6: This test has the same weakness as the test before this.

utkarshg6 (commented):
Done!

node/src/accountant/mod.rs
bertllll: If you do that people could wonder what are the values inside the vector and why I didn't assert on them. This way it's understandable that the only intention was to know how many times it was called.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
bertllll: Yeah. This is design inherited from Substratum. I don't know...
Maybe you're right but I would like to do it you rather because I don't have that feeling for it. 

It's not important to me, but I quite like these results of multiple methods are like a family, where the root of each is the BlockcahinError on the right side, next to the ok results. At least this was a good idea. I think there are like two levels of these structures where you're complaining basically about the upper one only.

I'm not going to do anything with this in this card though...

node/src/database/test_utils/transaction_wrapper_mock.rs
bertllll: Nope. The whole point is I need to store two things that reference to each other inside two fields of a struct. That's kind of impossible thing to do unless you invalidate the reference somehow, that's why I need the raw pointer and still I have to be careful about implementing the rest. 

This is also where all those "drops" etc come from, preventing OS errors. I've rewritten it hugely so that it's is clearer and safer, and made comments which believe are understandable.

node/src/db_config/config_dao.rs
bertllll: But there are three methods that can be considered part of the test's target, including the get methods.

node/src/db_config/config_dao.rs
bertllll: It doesn't reflect the fact, the transaction is "foreign", not provided by this local connection belonging to this ConfigDAO, but coming from outside. Lately I devised a phrase "guest_transaction", I think, it is good.

node/src/db_config/config_dao_null.rs
bertllll: Right! I must have overlooked that haha. This assertion makes no sense.

node/src/db_config/config_dao_null.rs
bertllll: Thank you!

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: Replaced by a different method

node/src/accountant/db_big_integer/big_int_db_processor.rs
bertllll: I did it in order to acquire power in the position of testing. It allows mocking.

I think it isn't that terribly complicated.

node/src/accountant/db_big_integer/big_int_db_processor.rs
bertllll: This is an older piece of code, also written for me to allow strict testing

node/src/accountant/db_big_integer/mod.rs
bertllll: Because there is a mock that I use under a different module. It's okay, I think

node/src/accountant/db_big_integer/big_int_db_processor.rs
bertllll: I wrote the code with the goal to provide an easy-to-use interface from outside. This is the price, kind of. You're not supposed to come in contact with this code much. The second face of this is that I had some performance concerns, using references, hoping to make it more efficient. But that part is debatable. Anyway, when I was learning Rust extensively, I realised the community favour referencing and sometimes they say that using the owned objects - if not necessary - makes weak code. I don't know. 

If I see some string to be hardcoded I'm reluctant to convert it to String because I know it will never change and so the piece of data of our program, in memory, will be there whether I choose to upgrade it to dynamically approached object or if I just reference to that memory slot directly. Therefore I think it doesn't create a new space of strings hanging in the memory all the time. It has ever been there as part of our binaries and therefore no need to create it again.

node/src/accountant/mod.rs
bertllll: This is a correction of a mistake I did before. We say "affordable accounts" or "payments" about the result of the adjustment, however here we are looking at sort of the input of the process so we should not use the word affordable now, they definitely are not affordable because that's the whole point of the PaymentAdjuster: if we've got right here, it means, we cannot afford to pay these out.

node/src/accountant/mod.rs
bertllll: To me, the fact it is a transaction wrapped in something doesn't change that it is still the transaction inside what matters. And because a wrapper by itself is not supposed to add functionality I consider these two terms interchangeable. 


But I will do it for you :-)

node/src/database/rusqlite_wrappers.rs
bertllll: Not possible with trait objects, but you will see once you start reviewing my new changes...

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: It's not Futures, just an ordinary Result. I use the pallet of methods the standard lib offers for it, that is also `map` and `map_err`. I've changed that anyway to make you a bit happier buddy.

node/src/database/db_migrations/migrator_utils.rs
bertllll: You know what is a wrapper, right? It should do just what the wrapped thing inside already does. 

I'm used to seeing: conn or connection wherever we use the ConnectionWrapper. It has never been weird to me to interchange the two things completely, because it actually is almost an identical thing. Plus, we can avoid unnecessarily long names.

node/src/database/db_migrations/migrator_utils.rs
bertllll: I checked that out, trying to knock some of the lifetimes off by using a Box instead but it started to be clear to me it wouldn't be trivial. I would like to do it either throughly or not at all, nothing between. So, I have an offer: if you do think this is an issue write a small card and maybe mark it as `good first issue`. Although the person will probably break his teeth at this meal. It's one file so he could get over it.

node/src/database/db_migrations/db_migrator.rs
bertllll: it means no arguments given after a statement has been formed. If you have a statement with for example question marks then you can give the values into the vector slice (or an array)

node/src/blockchain/blockchain_bridge.rs
bertllll: The whole card is about it. We need to set it after we record payments made in our direction, not before like we used to. And that ripped some code here and caused new code to show up in the Accountant.

node/src/accountant/test_utils.rs
bertllll: In tests it is relative what you should still call a hack and what is proper path of helping yourself get the test working right.
It is a decent way here: giving the code a read only conn while the code is suppose to make changes in the database. Testing database operations is quite uneasy in general so you are then happy to use this "trick".

node/src/accountant/scanners/mod.rs
bertllll: I need to oppose to you.  It's been a long-living rule we've agreed on concerning the database: if it fails at any place where an error is not expected to be a legitimate option we should immediately kill the Node - because we don't know what such a bad thing happened with the database - preventing even further damage of the data. You can ask Dan.

node/src/accountant/scanners/mod.rs
bertllll: The same thing I referred to before, this is a rule we've set up in the team.

node/src/accountant/scanners/mod.rs
bertllll: I explained to you in the comment attached to your original complaint.

node/src/accountant/scanners/mod.rs
bertllll: Nope. Shouldn't.

node/src/accountant/scanners/mod.rs
bertllll: Nope

node/src/accountant/scanners/mod.rs
bertllll: it uses a completely different function call inside. Definitely not repetition. I'll think a while about the names but otherwise I cannot serve you.

node/src/accountant/scanners/mod.rs
bertllll: I told you up there that this is not any repetition.

node/src/accountant/db_access_objects/receivable_dao.rs
bertllll: Nobody bore this out, showing interest, so I chose not to do it.

node/src/blockchain/blockchain_interface/rpc_helpers.rs
bertllll: This was a mistake. I got two versions of the same things during a merge. I deleted the outdated files.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
bertllll: This file is gone now. 

But the truth is this test the blockchain interface using web3 code inside - an argument could be made that some of the methods of the trait of this interface are almost pure wrappers but still there isn't anything fundamentally wrong about testing wrappers, bear in mind that this is the prod code while probably vast majority of the higher level code uses the mock version. Then I ask who would test what is inside the real implication? What if there is mud instead of code?

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
bertllll: The file is gone now

node/src/blockchain/blockchain_interface/blockchain_interface_web3/rpc_helpers_web3.rs
bertllll: I've solved this. Thanks for pointing me to this problem. It's been a duplicate of a newer version that got in with a merge.

utkarshg6 (approved):

utkarshg6 (approved):

PR #351, opened 2023-10-05
GH-600: Review 6

masqrauder:
### Highlights
* `max_block_count` is now `Option<u64>` with `PersistenceConfig` making it consistent with other optional values and its usage in the UI config message for displaying values
* Database migration now initializes `max_block_count` to an empty String which can be used in a future card to reset the value back to 'unlimited'
* Added tests to cover the `None` cases for `max_block_count`
* Updated existing tests to handle `max_block_count` as `Option<u64>`

### Details
* This commit changed `PersistenceConfig::max_block_count()` and `PersistenceConfig::set_max_block_count()` to work with `Option<u64>` based on Review 5 feedback. It now panics on error reading the value unless there's a bad format error because the value is empty string. This change will more easily allow the future card allowing a reset of `max_block_count` in the event of changing to a different blockchain service url; I expect it can be done without requiring a schema migration.

* I better understand now how `PersistenceConfig` has been designed to not allow storing a null value in the database when passed `None`. Before `max_block_count`, `None` passed to PeristenceConfig was rejected and not stored. I could have required a `u64` (not a `Option<u64>`) and used `u64::MAX` as the 'unlimited' value but I didn't like it. Also a consideration is that the dao doesn't accept None and doesn't support updating the config value to null. I prefer this the most because decoding a null is automatically converted to a None without treating it as an error. The second best option is using an empty String as the 'unlimited' value. Since the dao encodes u64 as a String to store, when passed `None` `set_max_block_count()` converts `None` to an empty String and stores it. I also made the migration consistent to initialize `max_block_count` to empty String ('') instead of null. When the empty string value doesn't decode successfully the match arm uses a guard to verify the empty String otherwise it returns the error.

### DIFF
--- a/node/src/database/db_migrations/migrations/migration_8_to_9.rs
+++ b/node/src/database/db_migrations/migrations/migration_8_to_9.rs
@@ -10,7 +10,7 @@ impl DatabaseMigration for Migrate_8_to_9 {
         declaration_utils: Box<dyn DBMigDeclarator + 'a>,
     ) -> rusqlite::Result<()> {
         declaration_utils.execute_upon_transaction(&[
-            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', null, 0)",
+            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', '', 0)",
         ])
     }
 
@@ -51,7 +51,7 @@ mod tests {
         let connection = result.unwrap();
         let (mp_value, mp_encrypted) = retrieve_config_row(connection.as_ref(), "max_block_count");
         let (cs_value, cs_encrypted) = retrieve_config_row(connection.as_ref(), "schema_version");
-        assert_eq!(mp_value, None);
+        assert_eq!(mp_value, Some("".to_string()));
         assert_eq!(mp_encrypted, false);
         assert_eq!(cs_value, Some("9".to_string()));
         assert_eq!(cs_encrypted, false);
--- a/node/src/database/db_migrations/migrations/migration_8_to_9.rs
+++ b/node/src/database/db_migrations/migrations/migration_8_to_9.rs
@@ -10,7 +10,7 @@ impl DatabaseMigration for Migrate_8_to_9 {
         declaration_utils: Box<dyn DBMigDeclarator + 'a>,
     ) -> rusqlite::Result<()> {
         declaration_utils.execute_upon_transaction(&[
-            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', '', 0)",
+            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', null, 0)",
         ])
     }
 
@@ -51,7 +51,7 @@ mod tests {
         let connection = result.unwrap();
         let (mp_value, mp_encrypted) = retrieve_config_row(connection.as_ref(), "max_block_count");
         let (cs_value, cs_encrypted) = retrieve_config_row(connection.as_ref(), "schema_version");
-        assert_eq!(mp_value, Some("".to_string()));
+        assert_eq!(mp_value, None);
         assert_eq!(mp_encrypted, false);
         assert_eq!(cs_value, Some("9".to_string()));
         assert_eq!(cs_encrypted, false);
--- a/node/src/db_config/persistent_configuration.rs
+++ b/node/src/db_config/persistent_configuration.rs
@@ -416,23 +416,11 @@ impl PersistentConfiguration for PersistentConfigurationReal {
     }
 
     fn max_block_count(&self) -> Result<Option<u64>, PersistentConfigError> {
-        match self.get("max_block_count") {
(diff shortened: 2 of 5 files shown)


node/src/database/db_migrations/migrations/migration_8_to_9.rs
```diff
@@ -10,7 +10,7 @@ impl DatabaseMigration for Migrate_8_to_9 {
         declaration_utils: Box<dyn DBMigDeclarator + 'a>,
     ) -> rusqlite::Result<()> {
         declaration_utils.execute_upon_transaction(&[
-            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', null, 0)",
+            &"INSERT INTO config (name, value, encrypted) VALUES ('max_block_count', '', 0)",
```
bertllll: This seems wrong to me. We do use null values. Null was right here. Let me elaborate within the PersistentConfig implementation.

node/src/db_config/persistent_configuration.rs
```diff
@@ -415,19 +415,25 @@ impl PersistentConfiguration for PersistentConfigurationReal {
         self.simple_set_method("start_block", value)
     }
 
-    fn max_block_count(&self) -> Result<u64, PersistentConfigError> {
+    fn max_block_count(&self) -> Result<Option<u64>, PersistentConfigError> {
         match self.get("max_block_count") {
             Ok(max_block_count) => match decode_u64(max_block_count) {
```
bertllll: Here you're supposed to take advantage of the return type of `decode_u64`. It actually returns a result of an `Option`.
In that case it's preferred that you'd treat it simply by returning None immediately and so you can solve this by deleting this whole line:

```
Err(TypedConfigLayerError::BadNumberFormat(value)) if value.is_empty() => Ok(None),
```
The first match close should take care for the rest.

node/src/db_config/persistent_configuration.rs
```diff
@@ -415,19 +415,25 @@ impl PersistentConfiguration for PersistentConfigurationReal {
...
 
-    fn set_max_block_count(&mut self, value: u64) -> Result<(), PersistentConfigError> {
-        self.simple_set_method("max_block_count", value)
+    fn set_max_block_count(&mut self, value: Option<u64>) -> Result<(), PersistentConfigError> {
+        self.simple_set_method(
+            "max_block_count",
+            match encode_u64(value) {
+                Ok(Some(mbc)) => mbc,
+                _ => "".to_string(),
+            },
```
bertllll: This should also be done differently. If you have a test for this already, it's going to be easy though. I believe you should simply pass the value in without using a match in the place. The apparatus of ConfigDao will know the way to handle it. Just write:

```
  self.simple_set_method( "max_block_count", value)
 ```

File: node/src/node_configurator/configurator.rs [hunk 4]
```diff
@@ -2620,6 +2622,119 @@ mod tests {
         assert_eq!(*past_neighbors_params, vec!["password".to_string()])
     }
 
+    #[test]
+    fn configuration_handles_retrieving_max_block_count_none() {
```
bertllll: Hmm. Matter-of-factly, if you look at the trails in the method, there is always just one, no matter whether the value for the max_block_count evaluates to None or Some(). That said, you have already tested this new piece of code by the two preexisting tests where you've added the parameter in. I consider this test groundless.

You either should delete it or if you do want to keep it, I would recommend to rewrite it into a more general variation of itself where you'd be claiming on testing all optional parameters when they return None. I'd probably understand that but otherwise this test shouldn't stay. (It's also noticeably inconsistent with the other picks of tests where you could have the same kind of argument, which you might have had in your mind, for every other optional value in there but yet such tests don't exist.)

node/src/db_config/persistent_configuration.rs
masqrauder: I wish it worked like you wrote it; it would be so simple (sorry for the terrible formatting):

```stdout
error[E0277]: `std::option::Option<u64>` doesn't implement `std::fmt::Display`
   --> src/db_config/persistent_configuration.rs:429:51
    |
429 |         self.simple_set_method("max_block_count", value)
    |              -----------------                    ^^^^^ `std::option::Option<u64>` cannot be formatted with the default formatter
    |              |
    |              required by a bound introduced by this call
    |
    = help: the trait `std::fmt::Display` is not implemented for `std::option::Option<u64>`
    = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
note: required by a bound in `persistent_configuration::PersistentConfigurationReal::simple_set_method`
   --> src/db_config/persistent_configuration.rs:559:29
    |
559 |     fn simple_set_method<T: Display>(
    |                             ^^^^^^^ required by this bound in `persistent_configuration::PersistentConfigurationReal::simple_set_method`
```

node/src/node_configurator/configurator.rs
masqrauder: ok, deleted

PR #353, opened 2023-10-07
GH-600: Review 7

masqrauder:
I tried to address all the feedback from review 6. If I missed anything please let me know.

### DIFF
--- a/node/src/db_config/persistent_configuration.rs
+++ b/node/src/db_config/persistent_configuration.rs
@@ -416,23 +416,11 @@ impl PersistentConfiguration for PersistentConfigurationReal {
     }
 
     fn max_block_count(&self) -> Result<Option<u64>, PersistentConfigError> {
-        match self.get("max_block_count") {
-            Ok(max_block_count) => match decode_u64(max_block_count) {
-                Ok(mbc_opt) => Ok(mbc_opt),
-                Err(e) => Err(PersistentConfigError::from(e)),
-            },
-            Err(e) => Err(PersistentConfigError::from(e)),
-        }
+        Ok(decode_u64(self.get("max_block_count")?)?)
     }
 
     fn set_max_block_count(&mut self, value: Option<u64>) -> Result<(), PersistentConfigError> {
-        match encode_u64(value) {
-            Ok(mbc_opt) => match self.dao.set("max_block_count", mbc_opt) {
-                Ok(_) => Ok(()),
-                Err(e) => Err(PersistentConfigError::from(e)),
-            },
-            Err(e) => Err(PersistentConfigError::from(e)),
-        }
+        Ok(self.dao.set("max_block_count", encode_u64(value)?)?)
     }
 
     fn set_wallet_info(
--- a/node/src/db_config/persistent_configuration.rs
+++ b/node/src/db_config/persistent_configuration.rs
@@ -419,7 +419,6 @@ impl PersistentConfiguration for PersistentConfigurationReal {
         match self.get("max_block_count") {
             Ok(max_block_count) => match decode_u64(max_block_count) {
                 Ok(mbc_opt) => Ok(mbc_opt),
-                Err(TypedConfigLayerError::BadNumberFormat(value)) if value.is_empty() => Ok(None),
                 Err(e) => Err(PersistentConfigError::from(e)),
             },
             Err(e) => Err(PersistentConfigError::from(e)),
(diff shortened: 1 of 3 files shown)


node/src/db_config/persistent_configuration.rs
```diff
@@ -419,21 +419,20 @@ impl PersistentConfiguration for PersistentConfigurationReal {
...
-                Ok(Some(mbc)) => mbc,
-                _ => "".to_string(),
+        match encode_u64(value) {
+            Ok(mbc_opt) => match self.dao.set("max_block_count", mbc_opt) {
+                Ok(_) => Ok(()),
+                Err(e) => Err(PersistentConfigError::from(e)),
             },
-        )
+            Err(e) => Err(PersistentConfigError::from(e)),
+        }
```
masqrauder: @bertllll I tried to apply all your suggestions except for this one. This is the way I was able to pass a `None`. I hope you find it acceptable. `self.dao.set()` can take an `Option<String>` and will store `None` as null. `PersistentConfiguration::simple_set_method()` would need to check if the `value` parameter is already an `Option<String>` and might be `None` and simply pass that to `self.dao.set()` If `value: T` is a `None ` value of `Option<String>`, then `value.to_string()` fails because `Option<String>` does not implement the required `Display` trait. If I'm doing something wrong here please let me know.

node/src/node_configurator/configurator.rs
```diff
@@ -2583,7 +2583,7 @@ mod tests {
                 clandestine_port: 1234,
                 chain_name: "ropsten".to_string(),
                 gas_price: 2345,
-                max_block_count_opt: Some(100000),
+                max_block_count_opt: None,
```
masqrauder, line 2586: @dnwiebe Had feedback I hadn't covered the usage of `None` so now there's one for `Some(100_000u64)` and this one for `None.

node/src/db_config/persistent_configuration.rs
bertllll: I see. It seems that I was wrong. Sorry. It was me who wrote the "simple_set_method()" which was a naive attempt to standardize at least some of the methods because I noticed a repetitive pattern in there. Unfortunately, there are many methods that cannot be treated so simply. I know about the limitations but when I suggested you those few things I believed it was right to do. It was before I understood that the "simple_set_method()" could've been more flexible if I'd designed it with the generic argument inside an Option, then you could use it even for you optional value. In fact, I think one day in the future my patient with my past mistakes will reach the limits and will refine it (which might mean giving up on unnaturally much pushing people to use this little code "if their implementation is simple enough to use it". I don't like this kind of additional consideration that's hard to answer straightly and that has nothing to do with elegance in the end.)

I think you did a good job. The only detail I came to in my mind is that you could simplify the match statements and the PersistendConfigError::from() expressions by use of the question mark. It's kind of calling for it but I know that this can differ from person to another. It's more about a personal style than readability. So I'm okay with this.

node/src/node_configurator/configurator.rs
bertllll: Yes, it kind of makes sense but I also think this solution is sub-optimal. Somebody who wouldn't know about these narrow intentions could easily change this value on a sudden occasion and this fragile warranty will break up with almost no chance to track it back. I know this isn't really your card, this is a long time system issue, but now I think we should've proceeded by creating the single test claiming to test all optional parameters when they evaluate to `None`. It would've laid down a path for any time when we need to deal with this in the future because hopefully the test would make the coder have to modify it with every new parameter and he should first go look at the title of the test. If written clearly, it should help the coder understand that he's supposed to set his newly added parameter with `None` in this test. (While he should use Some in the others). This could bring us more assurance that the None won't lose because the purpose of the test will stand above as a sign of undesirability in the removal of it.

I don't know I feel like I would want too much at the moment if I was insisting on this. So it's in your hands. I'm gonna give you an approval now (and another if you decided to write such a test for the `Configuration` command covering all those parameters with optional values).

bertllll (approved):

node/src/db_config/persistent_configuration.rs
masqrauder: Can I really use the question mark for both match arms? Perhaps it's easier to see with the RustRover/IntelliJ hints:
image

The inner error match arm is converting `ConfigDaoError` and outer error match arm is converting `TypedConfigLayerError`. If it can be simplified I'm happy to do so.

node/src/node_configurator/configurator.rs
masqrauder: I kind of thought thats what the test below was doing but you recommended deleting it. I can restore it of course but I need some guidance on how to improve it where you thought it was lacking. If the test name `configuration_handles_retrieving_max_block_count_none` is too specific I could change it to `configuration_handles_retrieving_all_possible_none_values` or something.

PR #355, opened 2023-10-11
GH-728: Finish the password-change logic

### LINKED ISSUE #728: Finish the password-change logic
When the password is changed from the UI, a lot of correct things happen, but not enough of them.

In general, when it receives a `UiChangePasswordRequest`, the `Configurator` actor sends a `NewPasswordMessage` to each of the actors listed in its `new_password_subs` field. That's good.

The problem is that when the `Configurator` receives a `BindMessage`, it populates the `new_password_subs` field with only the `Neighborhood`'s recipient. This means that when the password changes, only the `Neighborhood` finds out about the change, and every other actor that needs encrypted data from the `PersistentConfiguration` keeps the old password. That means the Node goes crazy and has to be restarted whenever the password is changed.

__Tasks:__
1. Look through the `PersistentConfiguration` and find all the methods that require a password.
2. Find all the actors that call any of those methods.
3. Implement a handler for the `NewPassword` message in each of those actors, and make sure the `BindMessage` contains a `NewPassword` recipient for each of those actors.
4. In the `Configurator`'s `BindMessage` handler, add those new recipients to the `Configurator`'s `new_password_subs` list.
5. Write a test (probably a single-hop `_integration` test) that changes a Node's database password and then verifies that every actor that uses the database password is using the new password properly to access the database.

utkarshg6:


node/src/accountant/mod.rs
```diff
@@ -164,6 +165,19 @@ impl Handler<BindMessage> for Accountant {
     }
 }
 
+impl Handler<ConfigChangeMsg> for Accountant {
+    type Result = ();
+
+    fn handle(&mut self, msg: ConfigChangeMsg, _ctx: &mut Self::Context) -> Self::Result {
```
dnwiebe, line 171: It's not clear to me why the earning wallet is an `Rc`. Except for this code right here, it seems like it ought to be read-only, and any scanner that needs it ought to be happy with a clone.

node/src/accountant/mod.rs
```diff
@@ -1208,6 +1225,63 @@ mod tests {
         assert_eq!(financial_statistics.total_paid_payable_wei, 0);
     }
 
+    #[test]
+    fn accountant_handles_config_change_msg() {
```
dnwiebe, line 1257: I don't really like conditional code (the `match` in this case) in assertions; it can produce unexpected behavior that makes for false positive tests.

I'd like to see the expectation passed into `assert_handling_of_config_change_msg` along with the `change`. You could pass an assertion closure, right? Then the higher-level test would be dictating exactly what the code under test was supposed to do, rather than depending on the utility function to figure it out.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -106,6 +107,18 @@ impl SkeletonOptHolder for RetrieveTransactions {
     }
 }
 
+impl Handler<ConfigChangeMsg> for BlockchainBridge {
```
dnwiebe: After some cogitation, I'm thinking that BlockchainBridge probably shouldn't have copies of the wallets; the Accountant should be the source of truth for them, and whenever the BlockchainBridge needs them, they should be supplied in a message from the Accountant. This isn't part of your card, but could you take a look and see how difficult it would be to accomplish that? If it's more than trivial, can you create a card? I foresee an issue here, when the user changes the wallets in the middle of a scan.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -644,6 +660,63 @@ mod tests {
         );
     }
 
+    #[test]
+    fn blockchain_bridge_handles_config_change_msg() {
```
dnwiebe: Same pattern change requested as in the Accountant

node/src/neighborhood/mod.rs
```diff
@@ -2991,52 +2974,60 @@ mod tests {
...
-        let expected_after_route = Route::round_trip(
-            segment(&[&o, &r, &e], &Component::ProxyClient),
-            segment(&[&e, &r, &o], &Component::ProxyServer),
-            cryptde,
-            Some(expected_new_wallet.clone()),
-            1,
-            Some(TEST_DEFAULT_CHAIN.rec().contract),
-        )
-        .unwrap();
+    fn neighborhood_handles_config_change_msg() {
```
dnwiebe, line 2956: Same pattern as in Accountant

node/src/neighborhood/mod.rs
```diff
@@ -3070,9 +3061,9 @@ mod tests {
     }
 
     #[test]
-    fn min_hops_can_be_changed_during_runtime_using_configuration_change_msg() {
+    fn min_hops_change_triggers_node_to_ui_broadcast_message() {
         init_test_logging();
-        let test_name = "min_hops_can_be_changed_during_runtime_using_configuration_change_msg";
+        let test_name = "min_hops_can_be_changed_during_runtime_using_config_change_msg";
```
dnwiebe: I think the `test_name` is wrong...

node/src/node_configurator/config_change_subs.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2019-2024, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+use crate::sub_lib::neighborhood::ConfigChangeMsg;
+use actix::Recipient;
+
+pub trait ConfigChangeSubs {
```
dnwiebe: It seems as though what we really need here is a set of filters that, when presented with a `BindMessage` and a `ConfigChange`, will produce a `Vec` of `Recipient`s from the `BindMessage` that need to know about that `ConfigChange`.

What if you had a function with a signature like this? Would it allow you to have the `ConfigChange` -> `Vec<Recipient>` pattern in only one place?
```
fn config_change_recipients_for(msg: &BindMessage, config_change: ConfigChange) -> Vec<Recipient<ConfigChangeMsg>>)
```
The problem with this, of course, is that you have to pass in example `ConfigChange` values, which must have parameter values, which don't exist at the time this function would be called.

Since all you have to know at this point is which value of the enum you're looking at, and you don't care about its parameter value, you could pass in nonsense parameter values (`ConfigChange::NewPassword("".to_string())`), but that's a hack and might be misleading.

Another option would be to have three functions, not just one, that had signatures like this:
```
fn password_change_recipients_for(msg: &BindMessage) -> Vec<Recipient<ConfigChangeMsg>>
```
There'd probably be a little duplication of code there (although most of it could be factored out), but you would have mostly squeezed out the duplication of _knowledge,_ which is what's irritating me at the moment.

node/src/node_configurator/configurator.rs
```diff
@@ -113,9 +123,10 @@ impl Configurator {
             Box::new(PersistentConfigurationReal::new(Box::new(config_dao)));
         Configurator {
             persistent_config,
-            new_password_subs: None, // GH-728
             node_to_ui_sub_opt: None,
-            configuration_change_msg_sub_opt: None,
+            update_min_hops_subs_opt: None,
```
dnwiebe: Since these three fields are closely related, it might be clearer to put them inside a `struct` to make that clearer.
```
let configurator = Configurator {
// ...
    config_change_subs: Option<ConfigChangeSubs> {
        update_min_hops_subs,
        // ...
    }
// ...
}
```

node/src/node_configurator/configurator.rs
```diff
@@ -701,23 +720,15 @@ impl Configurator {
...
         );
-        match Self::unfriendly_handle_set_configuration(
-            msg,
-            context_id,
-            &mut self.persistent_config,
-            configuration_change_msg_sub_opt,
-            logger,
-        ) {
-            Ok(message_body) => message_body,
+        match self.unfriendly_handle_set_configuration(msg) {
```
dnwiebe: As Bert pointed out, all the other `unfriendly_handle_xxx` methods return `MessageBody`s; perhaps for consistency this one should as well.

dnwiebe (commented):
Resume at configurator.rs:860 green

node/src/node_configurator/configurator.rs
```diff
@@ -1055,41 +1095,44 @@ mod tests {
...
             .unwrap();
 
         System::current().stop();
         system.run();
-        let change_password_params = change_password_params_arc.lock().unwrap();
+        let neighborhood_recording = neighborhood_recording_arc.lock().unwrap();
+        let expected_configuration_msg = ConfigChangeMsg {
+            change: ConfigChange::UpdatePassword(new_password.to_string()),
+        };
+        let ui_gateway_recording = ui_gateway_recording_arc.lock().unwrap();
```
dnwiebe: Maybe this line should move down a bit.

node/src/node_configurator/configurator.rs
```diff
@@ -1104,15 +1147,91 @@ mod tests {
...
+            body: make_example_generate_wallets_request().tmb(4321),
+        });
+        assert_wallets_synchronisation_among_other_actors(NodeFromUiMessage {
+            client_id: 1234,
+            body: make_example_recover_wallets_request_with_paths().tmb(4321),
+        });
+    }
+
+    fn assert_wallets_synchronisation_among_other_actors(msg: NodeFromUiMessage) {
+        let system = System::new("consuming_wallet_is_updated_when_new_wallet_is_generated");
```
dnwiebe: This System name should be changed

node/src/accountant/mod.rs
utkarshg6: It is used to create the `Scanners`. This `Rc` is shared with `Accountant`, and then the `ReceivableScanner` within `Scanners`.

node/src/node_configurator/config_change_subs.rs
utkarshg6: Instead of that, I added a method to the `PeerActors` that provides all the required config_change_recipients.

dnwiebe (approved):

PR #357, opened 2023-10-11
GH-672: review 8a

bertllll:


node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1410,10 +1414,9 @@ mod tests {
                 },
             ],
         };
-        let rpc_helpers =
-            RPCHelpersMock::default().get_block_number_result(LatestBlockNumber::Err(
-                BlockchainError::QueryFailed("Failed to read the latest block number".to_string()),
-            ));
+        let rpc_helpers = LowerBCIMock::default().get_block_number_result(LatestBlockNumber::Err(
```
utkarshg6, line 1417: I'd guess the full form of BCI, but eventually I'll be wrong. It's upto you in case you'd like to change its name.  🙏

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -12,7 +12,7 @@ use web3::transports::Batch;
 use web3::types::BlockNumber;
 use web3::{BatchTransport, Web3};
 
-pub struct RPCHelpersWeb3<T>
+pub struct LowerBCIWeb3<T>
```
utkarshg6, line 15: Also, this rename..

node/src/blockchain/test_utils.rs
```diff
@@ -321,176 +318,6 @@ pub fn make_fake_event_loop_handle() -> EventLoopHandle {
...
-
-impl<T> BatchPayableToolsFactoryMock<T> {
-    pub fn make_result(self, result: Box<dyn BatchPayableTools<T>>) -> Self {
-        self.make_results.borrow_mut().push(result);
-        self
-    }
-}
-
-#[derive(Default)]
-pub struct BatchPayableToolsMock<T: BatchTransport> {
```
utkarshg6, line 337: A better thing to do would be to make changes in-place rather than moving it into a different file. It would make it easier to see the changes in the diff. Although, I don't want you to do it now, but I wish if you do it in the future PRs.

utkarshg6 (approved):

node/src/blockchain/test_utils.rs
bertllll: I understand your point, but I did do it as considerately as possible. It looks messy in the review but the resulting state is what matters here.  I had to create a fully new hierarchy in the module blockchain because this area has advanced to a really multi-functional module with a lot of distinct parts, I felt it was urgent to make it maintainable within well defined separate files. It's really hard to do this right, and I did try hard, so please don't let you have quick prejudices. :pray:

node/src/blockchain/blockchain_bridge.rs
bertllll: It's been extremely tough for me to come up with a good abbreviation for BlockchainInterface. Can you help me on this? I'm not really good at concise names.

utkarshg6: Aha! Now, since I know the full form, everything makes sense. I think you can just add a comment where it's declared. A suggestion: `LowerBCIWeb3`.

node/src/blockchain/test_utils.rs
utkarshg6: It's alright @bertllll! It wasn't too hard for me to verify it. I'm glad to know that we're on the same page on this. 🙏

PR #359, opened 2023-10-11
GH-672: review 8b

bertllll:


node/src/accountant/db_access_objects/pending_payable_dao.rs
```diff
@@ -65,9 +66,7 @@ impl PendingPayableDao for PendingPayableDaoReal<'_> {
             .filter(|hash| !all_found_records.keys().contains(hash));
         all_found_records
             .iter()
-            .sorted()
             .map(|(hash, rowid)| (Some(*rowid), *hash))
-            .rev()
```
utkarshg6, line 70: I don't see the reason why you'd like to remove these two function calls, `sorted()` and `rev()`. Maybe you had your reasons, but it's unknown to me yet. 🤷‍♂️

node/src/accountant/db_access_objects/pending_payable_dao.rs
```diff
@@ -388,25 +387,61 @@ mod tests {
...
+            subject
+                .insert_new_fingerprints(&[(hash_2, 8901234)], SystemTime::now())
+                .unwrap();
+            {
+                subject
+                    .insert_new_fingerprints(&[(hash_3, 1234567)], SystemTime::now())
+                    .unwrap()
+            }
+            subject.delete_fingerprints(&[1]).unwrap()
+        }
```
utkarshg6, line 448: I don't think that we need to create scope 2 times. As per my opinion, the only use of scope is that if we want to create a variable and delete it within that scope. In our case, we don't need it.

node/src/accountant/scanners/mod.rs
```diff
@@ -1349,11 +1391,139 @@ mod tests {
...
+            (Some(1), hash_1),
+            (Some(3), hash_3),
+            (Some(2), hash_2),
+        ]);
+        let subject = PayableScannerBuilder::new()
+            .pending_payable_dao(pending_payable_dao)
+            .build();
+
+        let (existent, nonexistent) = subject
+            .separate_id_triples_by_existent_and_nonexistent_fingerprints(&pending_payables_ref);
```
utkarshg6, line 1424: We can make the test stronger if we let one entry of id triples with the nonexistent fingerprint. Just a suggestion. 🙏

node/src/accountant/scanners/mod.rs
```diff
@@ -1349,11 +1391,139 @@ mod tests {
...
+        let bb_returned_p_payables_ref = blockchain_bridge_returned_pending_payables
+            .iter()
+            .collect::<Vec<&PendingPayable>>();
+        let rowids_and_hashes_from_fingerprints =
+            vec![(Some(3), hash_1), (Some(5), hash_2), (Some(6), hash_3)];
+
+        PayableScanner::symmetry_check(
+            &bb_returned_p_payables_ref,
+            &rowids_and_hashes_from_fingerprints,
+        )
```
utkarshg6, line 1457: If you'd like you can make this function return a boolean and maybe panic somewhere else. It would provide a strong test for the happy path. In case you want to reuse the code with panic, you may migrate the call to that function with a panic in an outboard function.

node/src/accountant/scanners/scanners_utils.rs
```diff
@@ -219,12 +219,27 @@ pub mod payable_scanner_utils {
         }
     }
 
+    #[derive(Debug, PartialEq, Eq)]
     pub struct PendingPayableTriple<'a> {
```
utkarshg6, line 224: If you want you may call this `PendingPayableMetadata`, in case someone wants to extend this `struct` in the future.

utkarshg6 (approved):

node/src/accountant/db_access_objects/pending_payable_dao.rs
bertllll: This is for emphasis. I could probably replace it with a comment.

node/src/accountant/db_access_objects/pending_payable_dao.rs
bertllll: Believe me, this was well thought out. We reviewed it with Dan and in fact this _IS_ already code of the master branch, here you're seeing changes from the master merge before the Steve's GH-600. 

Long story short, I had needed the sorting just for a test where I assert on a collection but the ordering was different in times (the set of the elements though stay the same anytime). 

I needed to do something similar for the hotfix I was working on. So I chose to delete this because it was too tight to this low level functionality, instead I did anew but outside and in combination with measures that eventually fixed the emergency issue.

node/src/accountant/db_access_objects/pending_payable_dao.rs
bertllll: Solved

node/src/accountant/scanners/mod.rs
bertllll: I first though I could. But the function's guts works a bit unexpectedly (tested in a different test). It will **append** all the values with the fingerprints that haven't been found to the end of all those proper pairs. This also means in the code elsewhere that we will process the proper pairs and than the Node will panic due to the detection of missing fingerprints. In this light, the test is more important as "testing the happy path part alone" because that's the crucial part which will have effect on the database (new db writes).

node/src/accountant/db_access_objects/pending_payable_dao.rs
utkarshg6: Ok, I understand now. 👍

node/src/accountant/scanners/mod.rs
utkarshg6: Ok, I'd like if it was simpler, but I understand that it was needed to be this way. It makes me concerned if it'll cause bugs in the future. For the time being, I'll limit this concern to myself. 😥

PR #362, opened 2023-11-10
GH 731 - review

czarte:


node/tests/utils.rs
```diff
@@ -273,6 +273,8 @@ impl MASQNode {
 
     #[allow(dead_code)]
     pub fn wait_for_exit(&mut self) -> Option<Output> {
+        //TODO Put the body of this function in a background thread and wait on the thread for a few
```
dnwiebe: This comment is now a card (GH-758) and can be removed.

node/src/test_utils/mod.rs
```diff
@@ -643,7 +643,7 @@ pub mod unshared_test_utils {
         let mut app_args = vec!["MASQNode".to_string()];
         app_args.append(&mut slice_of_strs_to_vec_of_strings(&args));
         let arg_matches = app_node().get_matches_from_safe(app_args).unwrap();
-        MultiConfig::new_test_only(arg_matches)
+        MultiConfig::new_test_only(arg_matches, HashSet::new())
```
dnwiebe: If this is the only place this function is called, you can eliminate this second parameter, and just make `HashSet::new()` part of the body of the function.

masq_lib/src/multi_config.rs
```diff
@@ -63,9 +64,20 @@ impl<'a> MultiConfig<'a> {
     ) -> Result<MultiConfig<'a>, ConfiguratorError> {
         let initial: Box<dyn VirtualCommandLine> =
```
dnwiebe, line 64: This should move down near the `fold()` that uses it.

masq_lib/src/multi_config.rs
```diff
@@ -63,9 +64,20 @@ impl<'a> MultiConfig<'a> {
...
+        let mut computed_value_names = HashSet::new();
+        vcls.iter().for_each(|vcl| {
+            vcl.vcl_args().iter().for_each(|vcl_arg| {
+                match vcl.is_computed() {
+                    true => computed_value_names.insert(vcl_arg.name().to_string()),
+                    false => computed_value_names.remove(&vcl_arg.name().to_string()),
+                };
+            })
+        });
+        // TODO pull this out to function to use in determine_user_specific_data
```
dnwiebe: This TODO should either be done or eliminated.

masq_lib/src/multi_config.rs
```diff
@@ -334,7 +406,7 @@ impl EnvironmentVcl {
             .collect();
         let mut vcl_args: Vec<Box<dyn VclArg>> = vec![];
         for (upper_name, value) in std::env::vars() {
-            if (upper_name.len() < 5) || (&upper_name[0..5] != "MASQ_") {
+            if (upper_name.len() < 5) || (&upper_name[0..5] != "MASQ_") || (value == *"") {
```
dnwiebe: This makes me itch. Proposed scenario: A user has a config file that sets `log-level` to `trace`. He doesn't know what the default `log-level` is, but he knows it's much less prolific than `trace`, and he wants to make a run with skinny logs. So he sets the `MASQ_LOG_LEVEL` environment variable to `""`, hoping that it will override the config file's `trace` value with the less-demanding default.

This addition will defeat his efforts.

Is it necessary? Do tests fail if you remove it? If so, why do they fail? I'm curious.

masq_lib/src/multi_config.rs
```diff
@@ -949,6 +1037,40 @@ pub mod tests {
         assert_eq!(subject.args(), command_line);
     }
 
+    #[test]
+    fn command_line_vcl_return_value_from_vcl_args_by_name() {
+        let command_line: Vec<String> = vec![
```
dnwiebe: Isn't there a utility function with a long name that you could use in place of this? Something like `vec_of_slices_to_vec_of_strings()` or something?

masq_lib/src/multi_config.rs
```diff
@@ -949,6 +1037,40 @@ pub mod tests {
...
+            "/nonexistent/directory",
+            "--takes_no_value",
+            "--other_takes_no_value",
+        ]
+        .into_iter()
+        .map(|s| s.to_string())
+        .collect();
+
+        let subject = CommandLineVcl::new(command_line.clone());
+        let existing_value = match subject
```
dnwiebe: I think this `match` at the beginning could be replaced with a `.map()` at the end. Also, this sequence of code that you write once for `existing_value` and once for `non_existing_value` could probably be made into a local function or closure and called twice. If locating a `VclArg` inside a `VirtualCommandLine` by name turns out to be useful enough, it could turn into a local utility function that other tests could use.

node/src/server_initializer.rs
```diff
@@ -37,34 +37,34 @@ pub struct ServerInitializerReal {
 impl ServerInitializer for ServerInitializerReal {
     fn go(&mut self, streams: &mut StdStreams<'_>, args: &[String]) -> RunModeResult {
         let params = server_initializer_collected_params(self.dirs_wrapper.as_ref(), args)?;
```
dnwiebe: Rename `params` to `multi_config`.

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
 
 pub trait NodeConfigurator<T> {
     fn configure(&self, multi_config: &MultiConfig) -> Result<T, ConfiguratorError>;
 }
 
-pub fn determine_fundamentals(
+#[derive(Debug)]
+pub struct UserSpecifiedData {
+    pub(crate) chain: Chain,
+    pub(crate) chain_spec: bool,
```
dnwiebe: These field pairs seem to be duplicated a lot. Should there be a two-field type here instead?

dnwiebe:
```
struct<T> FieldPair {
    item: T,
    user_specified: bool
}
```

dnwiebe:
```
pub struct UserSpecifiedData {
    pub(crate) chain: FieldPair<Chain>,
    pub(crate) real_user: FieldPair<RealUser>,
    // ...
```

dnwiebe (commented):
Resume at `node_configurator/mod.rs`

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+use std::ops::Deref;
 use std::path::{Path, PathBuf};
 
 pub trait NodeConfigurator<T> {
     fn configure(&self, multi_config: &MultiConfig) -> Result<T, ConfiguratorError>;
 }
 
-pub fn determine_fundamentals(
+#[derive(Debug)]
+pub struct UserSpecifiedData {
```
dnwiebe: This isn't really user-specified data; it's _potentially_ user-specified data, with flags to tell us whether it's actually user-specified or not. Maybe there's a better name? Suggestions: ConfigInitializationData, BootstrapData...that's all I can think of right now.

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+    pub(crate) chain_spec: bool,
+    pub(crate) real_user: RealUser,
+    pub(crate) real_user_spec: bool,
+    pub(crate) data_directory: PathBuf,
+    pub(crate) data_directory_spec: bool,
+    pub(crate) config_file: PathBuf,
+    pub(crate) config_file_spec: bool,
+}
+
+fn get_chain_from_vcl(configs: &[Box<dyn VclArg>]) -> (Chain, bool) {
```
dnwiebe: MultiConfig should be used for this.

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+    dirs_wrapper: &dyn DirsWrapper,
+    real_user: &RealUser,
+    chain: &Chain,
+) -> (PathBuf, bool) {
+    match argument_from_vcl(configs, "--data-directory") {
+        Some(data_dir) => match PathBuf::from(data_dir).starts_with("~/") {
+            true => {
+                let home_dir_from_wrapper = dirs_wrapper
+                    .home_dir()
+                    .expect("expexted users home dir")
```
dnwiebe: "Expected" is misspelled"

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+    }
+}
+
+fn get_real_user_from_vcl(
+    configs: &[Box<dyn VclArg>],
+    dirs_wrapper: &dyn DirsWrapper,
+) -> (RealUser, bool) {
+    match argument_from_vcl(configs, "--real-user") {
+        Some(user) => {
+            let real_user_split: Vec<&str> = user.split(':').collect();
```
dnwiebe: This logic seems as though it should go in `impl FromStr for RealUser` rather than here.

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+    dirs_wrapper: &dyn DirsWrapper,
+) -> (PathBuf, bool) {
+    match argument_from_vcl(configs, "--config-file") {
+        Some(config_str) => {
+            let path = match PathBuf::from(config_str).is_relative() {
+                true => {
+                    match PathBuf::from(config_str).file_name().expect("expected file name") == config_str {
+                        true => PathBuf::from(data_directory).join(PathBuf::from(config_str)),
+                        false => match PathBuf::from(config_str).starts_with("./") || PathBuf::from(config_str).starts_with("../") {
+                            true => current_dir().expect("expected curerrnt dir").join(PathBuf::from(config_str)),
```
dnwiebe: "Current" is misspelled"

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
     app: &App,
     args: &[String],
-) -> Result<(PathBuf, bool, PathBuf, RealUser), ConfiguratorError> {
-    let orientation_schema = App::new("MASQNode")
-        .arg(chain_arg())
-        .arg(real_user_arg())
-        .arg(data_directory_arg(DATA_DIRECTORY_HELP))
-        .arg(config_file_arg());
+) -> Result<UserSpecifiedData, ConfiguratorError> {
     let orientation_args: Vec<Box<dyn VclArg>> = merge(
```
dnwiebe: I think this should be replaced with a full-fledged MultiConfig, and that the filtering I put in years ago is superfluous and should be removed. I think the filtering reveals some intention, and might be important in that respect, but MultiConfig isn't designed to do such a thing (and shouldn't be, in my opinion), so although the intention is revealed, the code is necessarily obtuse and confusing. Intention could probably be revealed just as effectively with a comment declaring that only four fields are of interest.

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+    }
+}
+
+fn get_config_file_from_vcl(
+    configs: &[Box<dyn VclArg>],
+    data_directory: &PathBuf,
+    data_directory_def: bool,
+    dirs_wrapper: &dyn DirsWrapper,
+) -> (PathBuf, bool) {
+    match argument_from_vcl(configs, "--config-file") {
```
dnwiebe: We make a PathBuf from this string many times below; perhaps we should make the PathBuf once at the top and use it many times instead?

dnwiebe: Also...what if the string in `--config-file` actually names a directory? Is that handled here?

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+                true => {
+                    match PathBuf::from(config_str).file_name().expect("expected file name") == config_str {
+                        true => PathBuf::from(data_directory).join(PathBuf::from(config_str)),
+                        false => match PathBuf::from(config_str).starts_with("./") || PathBuf::from(config_str).starts_with("../") {
+                            true => current_dir().expect("expected curerrnt dir").join(PathBuf::from(config_str)),
+                            false => match PathBuf::from(config_str).starts_with("~") {
+                                true => {
+                                    let home_dir_from_wrapper = dirs_wrapper
+                                        .home_dir()
+                                        .expect("expexted users home_dir")
```
dnwiebe: "Expected" is misspelled

node/src/node_configurator/mod.rs
```diff
@@ -11,34 +11,218 @@ use crate::database::db_initializer::{DbInitializer, DbInitializerReal};
...
+                                        .to_string();
+                                    let replaced_tilde_dir =
+                                        config_str
+                                            .to_string()
+                                            .replacen('~', home_dir_from_wrapper.as_str(), 1);
+                                    PathBuf::from(replaced_tilde_dir)
+                                }
+                                false => match data_directory_def {
+                                    true => PathBuf::from(data_directory).join(PathBuf::from(config_str)),
+                                    false => panic!("You need to define data-directory to define config file with naked directory 'dirname/config.toml'.")
```
dnwiebe: Can you modify this error message to contain the real strings that came in instead of example strings?

node/src/node_configurator/mod.rs
dnwiebe: Suddenly I'm wondering whether this too could be done with a MultiConfig, given the `value_user_specified_m!` or whatever that macro is. I dunno. If you start modifying the code to do more `MultiConfig` stuff, you will probably be close enough to the issue to decide for yourself.

node/src/node_configurator/mod.rs
```diff
@@ -240,18 +421,27 @@ mod tests {
...
+                .data_directory
+                .to_string_lossy()
+                .to_string(),
+        );
+        assert_eq!(
+            "booga.toml",
+            user_specific_data.config_file.file_name().unwrap()
         );
-        assert_eq!("booga.toml", config_file_path.file_name().unwrap());
-        assert_eq!(true, user_specified);
```
dnwiebe, line 254: Can you add an assert to preserve this old proof that the config file is marked as being user-specified?

node/src/node_configurator/mod.rs
```diff
@@ -272,9 +462,9 @@ mod tests {
 
         assert_eq!(
             "/tmp/booga.toml",
-            &format!("{}", config_file_path.display())
+            &format!("{}", user_specific_data.config_file.display())
         );
-        assert_eq!(true, user_specified);
```
dnwiebe, line 277: Same thing here

node/src/node_configurator/mod.rs
```diff
@@ -295,9 +485,9 @@ mod tests {
 
         assert_eq!(
             r"\tmp\booga.toml",
-            &format!("{}", config_file_path.display())
+            &format!("{}", user_specific_data.config_file.display())
         );
-        assert_eq!(true, user_specified);
```
dnwiebe, line 300: ...and etc...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -461,25 +580,28 @@ mod tests {
...
             "server_initializer_collected_params_can_read_parameters_from_config_file",
         );
         {
             let mut config_file = File::create(home_dir.join("config.toml")).unwrap();
             config_file
-                .write_all(b"dns-servers = \"111.111.111.111,222.222.222.222\"\n")
+                .write_all(b"dns-servers = \"111.111.111.111,222.222.222.222\"\nreal-user = \"1004:1004:/home/gooba\"\n")
                 .unwrap();
         }
+        std::env::set_var("MASQ_DNS_SERVERS", "");
```
dnwiebe: You said you thought this wasn't necessary.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -461,25 +580,28 @@ mod tests {
...
-
-        let gathered_params = server_initializer_collected_params(
-            &directory_wrapper,
-            &slice_of_strs_to_vec_of_strings(&["", "--data-directory", home_dir.to_str().unwrap()]),
-        )
-        .unwrap();
-
-        let multi_config = gathered_params.multi_config;
+        let args = ArgsBuilder::new().param("--data-directory", home_dir.to_str().unwrap());
+        let args_vec: Vec<String> = args.into();
```
dnwiebe, line 551: Are you creating a slice, converting to a Vec, and then back to a slice? I don't know what ArgsBuilder::new() produces.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -461,25 +580,28 @@ mod tests {
...
+        let _clap_guard = ClapGuard::new();
         let home_dir = ensure_node_home_directory_exists(
             "node_configurator_standard",
             "server_initializer_collected_params_can_read_parameters_from_config_file",
         );
         {
             let mut config_file = File::create(home_dir.join("config.toml")).unwrap();
             config_file
-                .write_all(b"dns-servers = \"111.111.111.111,222.222.222.222\"\n")
+                .write_all(b"dns-servers = \"111.111.111.111,222.222.222.222\"\nreal-user = \"1004:1004:/home/gooba\"\n")
```
dnwiebe: I don't think you're asserting on `real-user` here. Do you need it to be set?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+                .write_all(b"payment-thresholds = \"3333|55|33|646|999|999\"\n")
+                .unwrap();
+            config_file
+                .write_all(b"scan-intervals = \"111|100|99\"\n")
+                .unwrap()
+        }
+    }
+
+    #[test]
+    fn multi_config_vcl_is_computed_do_right_job() {
```
dnwiebe: Could you make this name a little clearer?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            ("MASQ_CONFIG_FILE", "./generated/test/node_configurator_standard/multi_config_vcl_is_computed_do_right_job/config.toml"),
+            #[cfg(not(target_os = "windows"))]
+            ("MASQ_REAL_USER", "9999:9999:booga"),
+        ];
+        env_vec_array
+            .clone()
+            .into_iter()
+            .for_each(|(name, value)| std::env::set_var(name, value));
+        let args = ArgsBuilder::new();
+        let args_vec: Vec<String> = args.into();
```
dnwiebe, line 871: Also perhaps unnecessary?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
         assert_eq!(config.crash_point, CrashPoint::Panic);
     }
 
+    fn fill_up_config_file(mut config_file: File) {
+        {
+            config_file
```
dnwiebe, line 812: This would be easier to read if these parameters were in alphabetical order...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+
+        #[cfg(not(target_os = "windows"))]
+        {
+            assert_eq!(
+                value_m!(env_multiconfig, "data-directory", String).unwrap(),
+                "booga/data_dir/MASQ/polygon-mainnet".to_string()
+            );
+            assert_eq!(
+                value_m!(env_multiconfig, "config-file", String).unwrap(),
+                current_dir().expect("expected curerrnt dir")
```
dnwiebe: "Current" is misspelled

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            .data_dir_result(Some(data_dir.to_path_buf()));
+
+        let result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let env_multiconfig = result.unwrap();
+
+        #[cfg(not(target_os = "windows"))]
+        {
+            assert_eq!(
+                value_m!(env_multiconfig, "data-directory", String).unwrap(),
+                "booga/data_dir/MASQ/polygon-mainnet".to_string()
```
dnwiebe: I don't understand where the "`data_dir`" part of this path came from. If it came from the `DirsWrapperMock` object, it seems as though it should have the complicated `home_dir` pasted on the beginning of it.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        }
+        #[cfg(target_os = "windows")]
+        assert_eq!(
+            value_m!(env_multiconfig, "data-directory", String).unwrap(),
+            "generated/test/node_configurator_standard/multi_config_vcl_is_computed_do_right_job/home\\data_dir\\MASQ\\polygon-mainnet".to_string()
+        );
+    }
+
+    #[test]
+    fn server_initializer_collected_params_handle_dot_in_path_and_real_user_in_config_file() {
```
dnwiebe: `dot_in_path`...the path of the config file, right?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        let args = ArgsBuilder::new();
+        let args_vec: Vec<String> = args.into();
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+
+        let result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+        let env_multiconfig = result.unwrap();
+
+        assert_eq!(env_multiconfig.is_user_specified("--data-directory"), false);
```
dnwiebe: These assertions look to me to be very similar--maybe identical--to the assertions in the test above. So I wonder: how are these tests different? Are both required?

dnwiebe (commented):
Resume at node_configurator_standard.rs:1001

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            .param("--config-file", "~/masqhome/config.toml")
+            .param("--data-directory", "~/masqhome");
+        #[cfg(target_os = "windows")]
+        let args = ArgsBuilder::new()
+            .param("--blockchain-service-url", "https://www.mainnet1.com")
+            .param("--config-file", "~/masqhome\\config.toml")
+            .param("--data-directory", "~/masqhome");
+        let args_vec: Vec<String> = args.into();
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(current_dir().expect("expect current directory").join("generated/test/node_configurator_standard/server_initializer_collected_params_handle_tilde_in_path_config_file_from_commandline_and_real_user_from_config_file/home")))
```
dnwiebe: Investigate using the `home_dir` variable here; it might save you some duplication.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            .data_dir_result(Some(data_dir.to_path_buf()));
+        let result_data_dir = current_dir().expect("expect current directory").join("generated/test/node_configurator_standard/server_initializer_collected_params_handle_tilde_in_path_config_file_from_commandline_and_real_user_from_config_file/home/masqhome");
```
dnwiebe: This might be no different from `data_dir`.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        let _guard = EnvironmentGuard::new();
+        let _clap_guard = ClapGuard::new();
+        let home_dir = ensure_node_home_directory_exists( "node_configurator_standard","server_initializer_collected_params_handle_tilde_in_path_config_file_from_commandline_and_real_user_from_config_file");
+        let data_dir = &home_dir.join("masqhome");
+        let _create_data_dir = create_dir_all(data_dir);
+        let config_file_relative = File::create(data_dir.join("config.toml")).unwrap();
+        fill_up_config_file(config_file_relative);
+        let env_vec_array = vec![
+            ("MASQ_IP", "9.8.7.6"),
+            ("MASQ_BLOCKCHAIN-SERVICE-URL", "https://www.mainnet2.com"),
```
dnwiebe: Seems like this should be `MASQ_BLOCKCHAIN_SERVICE_URL` with underscores.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            result_data_dir
+                .join(PathBuf::from("config.toml"))
+                .to_string_lossy()
+                .to_string()
+        );
+        assert_eq!(value_m!(multiconfig, "ip", String).unwrap(), "9.8.7.6");
+        assert_eq!(
+            value_m!(multiconfig, "blockchain-service-url", String).unwrap(),
+            "https://www.mainnet1.com"
+        );
```
dnwiebe, line 1132: I'd like to see an assertion in here on some value that comes in from the config file, as an obvious proof that the config file was found with the tilde. I understand that specifying the config file and not failing means the config file was found, but the test would be clearer with an actual value from the config file.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        assert_eq!(multiconfig.is_user_specified("--data-directory"), true);
+        assert_eq!(
+            &value_m!(multiconfig, "data-directory", String).unwrap(),
+            &home_dir.to_string_lossy().to_string()
+        );
+        #[cfg(not(target_os = "windows"))]
+        assert_eq!(
+            &value_m!(multiconfig, "real-user", String).unwrap(),
+            "1001:1001:cooga"
+        );
```
dnwiebe, line 1176: Again, it'd be nice to see an assertion from the config file. You can probably just copy and paste whatever you put in in the previous test.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "You need to define data-directory to define config file with naked directory 'dirname/config.toml'."
```
dnwiebe: I think this error message should say `'config/config.toml`' based on what the user configured. If you give him a value he didn't configure, it's going to be confusing. I think I mentioned this already somewhere else.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        let args_vec: Vec<String> = args.into();
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+
+        let _result = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
+    }
+
+    #[test]
+    fn server_initializer_collected_params_combine_vlcs_properly() {
```
dnwiebe: Should be `vcls`, not `vlcs`.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            .into_iter()
+            .for_each(|(name, value)| std::env::set_var(name, value));
+        let dir_wrapper = DirsWrapperMock::new()
+            .home_dir_result(Some(home_dir.clone()))
+            .data_dir_result(Some(data_dir.to_path_buf()));
+        let args = ArgsBuilder::new()
+            .param("--data-directory", current_directory.join(Path::new("generated/test/node_configurator_standard/server_initializer_collected_params_combine_vlcs_properly/home")).to_string_lossy().to_string().as_str())
+            .param("--real-user", "1001:1001:cooga");
+        let args_vec: Vec<String> = args.into();
+        let params = server_initializer_collected_params(&dir_wrapper, args_vec.as_slice());
```
dnwiebe, line 1237: Need blank lines around Act

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+        let result = params.as_ref().expect("REASON");
```
dnwiebe: Maybe something better than "REASON", or just `unwrap()`

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -737,6 +859,331 @@ mod tests {
...
+            "node_configurator_standard",
+            "server_initializer_collected_params_combine_vlcs_properly",
+        );
+        let data_dir = &home_dir.join("data_dir");
+        let config_file = File::create(&home_dir.join("config.toml")).unwrap();
+        let current_directory = current_dir().unwrap();
+        fill_up_config_file(config_file);
+
+        let env_vec_array = vec![
+            ("MASQ_CONFIG_FILE", "config.toml"),
```
dnwiebe: `config.toml` is a default. Can you specify something different here to prove that the default isn't leaking through somehow?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
     )
 }
 
+fn extract_values_vcl_fill_multiconfig_vec(
+    config_file_vcl: Box<dyn VirtualCommandLine>,
+    environment_vcl: Box<dyn VirtualCommandLine>,
+    commandline_vcl: Box<dyn VirtualCommandLine>,
+    user_specific_data: UserSpecifiedData,
+) -> (Vec<String>, Vec<String>) {
```
dnwiebe: I still think these `_vcl`s could be replaced with a `MultiConfig.` ...sorry...

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+    let mut unspecified_vec: Vec<String> = vec!["".to_string()];
+    let mut specified_vec: Vec<String> = vec!["".to_string()];
```
dnwiebe: Could you move these down to nearer where they're used?

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+        "--real-user",
+        user_specific_data.real_user.to_string().as_str(),
+        user_specific_data.real_user_spec,
+    );
+
+    let mut fill_specified_or_unspecified_box =
+        |key: &str, value: &str, specified: bool| match value.is_empty() {
+            true => (),
+            false => match specified {
+                true => match specified_vec.contains(&key.to_string()) {
```
dnwiebe: I see some code duplication here. You already have some local functions; you can probably pull out the `match [un]specified_vec` part into another one.

Also, the do-this-if-it-already-exists-or-this-if-it-doesn't logic is already part of `HashMap`, if you decided to use that. Unfortunately, what you need to make a VCL is a vector. However, there might be an easy way to convert a `HashMap` into a `Vec` with pairs. I don't know, but it might be worth looking into to simplify this code.

dnwiebe: Turns out I think you can do it by calling `.iter()` on the `HashMap` and then `fold()` on the result, starting with an empty vector. Every time through the closure you pass to `fold()`, push the two items in the pair onto the vector; that should do it, maybe in two or three lines.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+    );
+
+    let (env_real_user, env_real_user_specified) = extract_value_from_vcl(
+        &*environment_vcl,
+        "--real-user",
+        user_specific_data.real_user.to_string().as_str(),
+        user_specific_data.real_user_spec,
+    );
+
+    let (cmd_real_user, cmd_real_user_specified) = extract_value_from_vcl(
```
dnwiebe: Three operations on `real-user` might be unnecessary here if you pass in a `MultiConfig` instead of three vectors.

dnwiebe (commented):
Resume at node_configurator_standard.rs:253

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+        Box::new(EnvironmentVcl::new(&app)),
+        Box::new(CommandLineVcl::new(commandline_vcl.args())),
+        user_specific_data,
+    );
+    let mut multi_config_args_vec: Vec<Box<dyn VirtualCommandLine>> = vec![
+        Box::new(config_file_vcl),
+        Box::new(environment_vcl),
+        Box::new(commandline_vcl),
+    ];
+    //TODO write test for MultiConfig "try_new" merge line 76
```
dnwiebe: This should probably either be done or removed.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+    if unspecified_vec.len() > 1 {
```
dnwiebe: I think you can get rid of these `if` statements. If a `Vcl` is empty, it won't affect the `MultiConfig`.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -136,35 +137,152 @@ fn collect_externals_from_multi_config(
...
+    );
+
+    let (cmd_real_user, cmd_real_user_specified) = extract_value_from_vcl(
+        &*commandline_vcl,
+        "--real-user",
+        user_specific_data.real_user.to_string().as_str(),
+        user_specific_data.real_user_spec,
+    );
+
+    let mut fill_specified_or_unspecified_box =
```
dnwiebe: I notice that there are a lot of paths through this code. I'm concerned that not all of them may be exercised by tests. If this piece of functionality retains a significant amount of complexity after you switch to `MultiConfig`, I'd recommend that you run the tests at least once with Coverage and make sure that all the paths through the code are covered.

node/src/node_configurator/mod.rs
czarte: Should we look for config.toml on defined directory or return ConfigurationError?

node/src/node_configurator/node_configurator_standard.rs
czarte: this is some machinery I took from other tests, should I investigate it?

node/src/node_configurator/node_configurator_standard.rs
czarte: data_dir is the placeholder in tests for `.local/share` directory

node/src/node_configurator/node_configurator_standard.rs
czarte: this test is testing the scenario, when we specify the `config-file` with dot, assertions just proving that we correct the path correctly and we can read from config file - so `real-user` and `dns-servers` are asserted right way

node/src/node_configurator/node_configurator_standard.rs
czarte: try this on one place and then create card out of it (good first issue)

dnwiebe (approved):

PR #367, opened 2023-12-20
GH-651 Review-3

Syther007:


node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -347,6 +350,15 @@ impl StreamHandlerPoolReal {
             }
             Ok(lookup_ip) => lookup_ip.iter().collect(),
         };
+
+        if ip_addrs[0] == IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)) {
```
dnwiebe: Suggestion: Put this test in an outboard function somewhere with a meaningful name, so that it can be enhanced if it becomes necessary. A good set of starting functionality would be Utkarsh's suggestion of filtering all the quad-zero entries out and seeing if there's anything left. It seems like it shouldn't be too much of a performance hit.

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -347,6 +350,15 @@ impl StreamHandlerPoolReal {
...
+            error!(
```
dnwiebe: First, I don't think this is an error for the exit Node. Even `warn!` seems a little dramatic. How about just `info!`?

Second, not everybody knows what "wildcard IP address" means. Can you modify the message to talk about IP blocking?

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -1218,10 +1230,108 @@ mod tests {
...
+                Box::new(resolver),
+                cryptde,
+                peer_actors.accountant.report_exit_service_provided.clone(),
+                peer_actors.proxy_client_opt.clone().unwrap(),
+                100,
+                200,
+            );
+            let (stream_killer_tx, stream_killer_rx) = unbounded();
+            subject.stream_killer_rx = stream_killer_rx;
+            let inner_clone = Arc::clone(&subject.inner);
```
dnwiebe: Could you just do
```
{
    subject.inner.lock().expect("message").logger = Logger::new(test_name);
}
```
? (Maybe on more than one line for readability?)

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -1218,10 +1230,108 @@ mod tests {
...
+                Box::new(StreamEstablisherFactoryMock {
+                    make_results: RefCell::new(vec![establisher]),
+                });
+
+            run_process_package_in_actix(subject, package);
+        });
+
+        proxy_client_awaiter.await_message_count(2);
+        let proxy_client_recording = proxy_client_recording_arc.lock().unwrap();
+        assert_eq!(
```
dnwiebe, line 1366: As a general philosophical concept, when the originating Node gets a `DnsResolveFailure` message back, it ought to free the same stream resources that are freed by an `InboundServerData` with a `last_data` of `true`. This seems so clear to me that I think we should have caught it in the original iteration from Substratum. Can you look to see what happens when the originating Node gets a DNS failure? If it doesn't free the resources, we should make a card about that. Sending the extra `InboundServerData` package will cost the user needlessly.

Syther007: We have found that the originating Node is reusing the stream key for DNS retries.
Due to this if originating node receives an `ExpiredCoresPackage` with `DnsResolveFailure` it wont remove the stream yet.

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -333,6 +333,13 @@ impl StreamHandlerPoolReal {
         )
     }
 
+    fn filter_wildcard_ips(ip_addrs: Vec<IpAddr>) -> Vec<IpAddr> {
+        ip_addrs
+            .into_iter()
+            .filter(|&ip_addr| ip_addr != IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)))
```
dnwiebe, line 339: You could put this outside the loop, rather than constructing it every time through...

dnwiebe (approved):

PR #372, opened 2024-01-08
Gh 731 review3

czarte:


node/src/accountant/mod.rs
```diff
@@ -4297,7 +4297,7 @@ mod tests {
             let factory = Accountant::dao_factory(data_dir);
             factory.make();
         };
-
+        println!("make_dao_factory_uses_panic_on_migration {:?}", &data_dir);
```
utkarshg6, line 4300: Maybe you want to remove this. :wink:

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -224,25 +213,28 @@ pub fn server_initializer_collected_params<'a>(
 
     let environment_vcl = EnvironmentVcl::new(&app);
     let commandline_vcl = CommandLineVcl::new(args.to_vec());
-    let (unspecified_vec, specified_vec) = extract_values_vcl_fill_multiconfig_vec(
+    let multiconfig_for_values_extraction = make_new_multi_config(
```
utkarshg6, line 216: Maybe removing the `make` from the function name is a good idea. You might notice this word being extensively used in the functions of test utilities. So, maybe `create_new_multi_config` is a better name here.

utkarshg6 (approved):

node/src/node_configurator/node_configurator_standard.rs
czarte: This function is widely used in production code. I'm going to make a card out of it.

PR #374, opened 2024-01-21
GH-606: Initialize start_block to none to use latest block

masqrauder:


masq/src/commands/set_configuration_command.rs
```diff
@@ -35,9 +35,13 @@ impl SetConfigurationCommand {
...
 fn validate_start_block(start_block: String) -> Result<(), String> {
-    match start_block.parse::<u64>() {
-        Ok(_) => Ok(()),
-        _ => Err(start_block),
+    if "none".eq_ignore_ascii_case(&start_block) {
+        Ok(())
+    } else {
+        match start_block.parse::<u64>() {
+            Ok(_) => Ok(()),
+            _ => Err(start_block),
```
bertllll: I'd like to see a longer err message that would contain also either the parse err or an assistive instruction that it needs to be a valid integer.

masq/src/commands/set_configuration_command.rs
```diff
@@ -126,6 +130,7 @@ mod tests {
     fn validate_start_block_works() {
```
bertllll, line 142: This is actually a bad test because if the function returns an err we want to know if it is an appropriate one, or its message meaningful. 

I'd like to see two-sides assertion with Ok(()) or Err("The error message").

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -301,8 +301,9 @@ impl BlockchainBridge {
...
     fn handle_retrieve_transactions(&mut self, msg: RetrieveTransactions) -> Result<(), String> {
         let start_block_nbr = match self.persistent_config.start_block() {
-            Ok (sb) => sb,
-            Err (e) => panic! ("Cannot retrieve start block from database; payments to you may not be processed: {:?}", e)
+            Ok(Some(sb)) => sb,
+            Ok(None) => u64::MAX,
+            Err(e) => panic!("Cannot retrieve start block from database; payments to you may not be processed: {:?}", e)
         };
         let max_block_count = match self.persistent_config.max_block_count() {
             Ok(Some(mbc)) => mbc,
```
bertllll, line 289: I'm having a hard time to accept that we ignore the error. I know it isn't dangerous if we don't have it, on the other hand, at all places I remember we treat such an error from our communication with the database manager, specifically, we always go and panic.

An argument is, though, that there is a card that should make this reaction implicit, ripping the Result structure from all the methods of PersistentConfiguration. Now, advice is worth gold. I don't feel like needing to insist on this.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -314,25 +315,31 @@ impl BlockchainBridge {
...
             }
             Err(e) => {
-                info!(
-                    self.logger,
-                    "Using 'latest' block number instead of a literal number. {:?}", e
-                );
                 if max_block_count == u64::MAX {
+                    info!(
+                        self.logger,
+                        "Using 'latest' block number instead of a literal number. {:?}", e
```
bertllll, line 310: You first need to think about what to do with the error. Please don't just throw it away. I think logging would be fair, we do it all the time. 

The way from here, I think, goes by dividing this message into two logs (or even three, according to the differences I mentioned about the if statement). One for the error, maybe perceived as WARN and another less sever with this little informative message (or actually two like that).  

If you want to keep an info log, which means one that "we want the user to see", it maybe should have more information to give so that they understand the context. Like what bigger operation is going on of which this is a smaller part.

Remember there is a rule that all info logs will eventually display to the user directly in the UI, not just write to the file.

I do think this kind of information is useful for us developers rather than anybody else so I'd choose just a debug log instead (or actually two debug logs). The fine think about this is you wouldn't have to make assertions on them.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -314,25 +315,31 @@ impl BlockchainBridge {
             .get_block_number()
         {
             Ok(eb) => {
-                if u64::MAX == max_block_count {
+                if u64::MAX == max_block_count || u64::MAX == start_block_nbr {
```
bertllll: You use this boolean condition a couple of times again and again. Maybe you could create a variable before here, with a nice name, that would store the resolution of this expression. Then use the variable where you need to. 

I think I'd find it easier to understand since I wouldn't have to check if those are different conditions or the same one all the time.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -314,25 +315,31 @@ impl BlockchainBridge {
...
                     BlockNumber::Number(eb)
                 } else {
                     BlockNumber::Number(eb.as_u64().min(start_block_nbr + max_block_count).into())
                 }
             }
             Err(e) => {
-                info!(
-                    self.logger,
-                    "Using 'latest' block number instead of a literal number. {:?}", e
-                );
```
bertllll, line 307: I think this info log was wrong.

First, I don't like it used to generalise the the two possible situation after that, strictly speaking, we wouldn't be using the 'latest' option, which has usually a different meaning than, "the furthest the blockchain service provider allows us to request at the moment". I'm afraid, if the logs is a good idea at all, we should have two logs for these two different situations.

Second, I understand there has been questions about whether we should care of the error from the blockchain interface, maybe no, but what is definitely wrong is grafting an error message into an info log. The two pieces of information don't belong together.    

I'm pointing to these issues because I'm going to comment on the logging here again and this is important to prevent.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1418,18 +1425,19 @@ mod tests {
...
+        );
         let blockchain_interface_mock = BlockchainInterfaceMock::default()
             .retrieve_transactions_params(&retrieve_transactions_params_arc)
             .retrieve_transactions_result(Ok(expected_transactions.clone()))
             .lower_interface_results(Box::new(lower_interface));
         let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
         let persistent_config = PersistentConfigurationMock::new()
-            .max_block_count_result(Ok(Some(10000u64)))
-            .start_block_result(Ok(6))
+            .max_block_count_result(Ok(None))
```
bertllll, line 1314: I don't like this kind of testing, where you pick a random test and modify it so that it intercept the place you're making changes in. This is dirty, because it can easily happen that somebody will come and change this value or even delete the whole test. Why would they not realise they might be exposing some logic out of the test coverage? Because they cannot, the test clearly states what it tests and there is no mention about the null value for the max_block_count. It is not the focus of this test, obviously. 

That's why I have to ask you to create an extra test that will actually be very specific about the focus on this value.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1484,7 +1492,9 @@ mod tests {
             }
         );
 
-        TestLogHandler::new().exists_log_containing("INFO: BlockchainBridge: Using 'latest' block number instead of a literal number. QueryFailed(\"Failed to read the latest block number\")");
+        TestLogHandler::new().exists_log_containing(
+            "INFO: BlockchainBridge: Using 'latest' block number instead of a literal number.",
```
bertllll: This may change and will no problem anymore but it is a mistake that you cut away the tail of the message. If it is how the prod code behaves, that is a message written or printed including the error, than the assertion must also be complete, with the error shown at the end of it.

node/src/test_utils/database_version_0_sql.txt
```diff
@@ -7,12 +7,12 @@ insert into config (name, value, encrypted) values ('consuming_wallet_public_key
 insert into config (name, value, encrypted) values ('earning_wallet_address', null, 0)
 insert into config (name, value, encrypted) values ('schema_version', '0', 0)
 insert into config (name, value, encrypted) values ('seed', null, 0)
-insert into config (name, value, encrypted) values ('start_block', 8688171, 0)
+insert into config (name, value, encrypted) values ('start_block', null, 0)
```
bertllll: This is definitely wrong. You cannot time-travel. This file should never be changed again. It is a fingerprint of past history. It documents what the database used to be like before all the migrations we have written since. Please reverse this. (I'm slightly disconcerted that this was okay in other tests and did not manifest with failed assertions. But I guess it is possible.)

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn migration_from_9_to_10_is_properly_set() {
+        init_test_logging();
+        let dir_path = ensure_node_home_directory_exists(
+            "db_migrations",
+            "migration_from_9_to_10_is_properly_set",
+        );
+        create_dir_all(&dir_path).unwrap();
+        let db_path = dir_path.join(DATABASE_FILE);
+        let _ = bring_db_0_back_to_life_and_return_connection(&db_path);
+        let subject = DbInitializerReal::default();
```
bertllll: We have a habit that there are two steps in these test. In the test setup phase, we first use the subject to migrate from 0 to the schema version one step before yours. That way we know that we are really testing only the step between 9 and 10. I quite like it more. Please implement it too.

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        let result = subject.initialize_to_version(
+            &dir_path,
+            10,
+            DbInitializationConfig::create_or_migrate(make_external_data()),
+        );
```
bertllll: There should be a blank line but that's really a detail.

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
...
+        let connection = result.unwrap();
+        let (mp_value, mp_encrypted) = retrieve_config_row(connection.as_ref(), "start_block");
```
bertllll: I think these names had been given based on testing different parameters (probably "mapping_protocol"), please adjust them so that they fit to the right side of the expressions.

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
...
+#[allow(non_camel_case_types)]
+pub struct Migrate_9_to_10;
+
+impl DatabaseMigration for Migrate_9_to_10 {
+    fn migrate<'a>(
+        &self,
+        declaration_utils: Box<dyn DBMigDeclarator + 'a>,
+    ) -> rusqlite::Result<()> {
+        declaration_utils.execute_upon_transaction(&[
+            &"INSERT INTO config (name, value, encrypted) VALUES ('start_block', null, 0) ON CONFLICT DO NOTHING",
```
bertllll: Careful here, you need two tests! Only that way the condition that if there is value in place it will keep it will be exercised. This test is missing in your submitted work.

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
...
+            DbInitializationConfig::create_or_migrate(make_external_data()),
+        );
+        let connection = result.unwrap();
+        let (mp_value, mp_encrypted) = retrieve_config_row(connection.as_ref(), "start_block");
+        let (cs_value, cs_encrypted) = retrieve_config_row(connection.as_ref(), "schema_version");
+        assert_eq!(mp_value, None);
+        assert_eq!(mp_encrypted, false);
+        assert_eq!(cs_value, Some("10".to_string()));
+        assert_eq!(cs_encrypted, false);
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
```
bertllll: We don't do this. We usually assert on logs only if they are added inside the frame of the migration and only if they are more sever than DEBUG. 

I think this shouldn't be kept, it adds no value for the test. We expect the migration step to be successful if the assertions on the changes implied by the way the migration is written all pass.

node/src/database/config_dumper.rs
```diff
@@ -503,11 +500,8 @@ mod tests {
         assert_value("pastNeighbors", "masq://polygon-mainnet:QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVowMTIzNDU@1.2.3.4:1234,masq://polygon-mainnet:QkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWjAxMjM0NTY@2.3.4.5:2345", &map);
         assert_value("neighborhoodMode", "consume-only", &map);
         assert_value("schemaVersion", &CURRENT_SCHEMA_VERSION.to_string(), &map);
-        assert_value(
-            "startBlock",
-            &Chain::PolyMainnet.rec().contract_creation_block.to_string(),
-            &map,
-        );
+        assert!(map.contains_key("startBlock"));
+        assert_none("startBlock", &map);
```
bertllll: Please rewrite the definition of assert_none (which is alone a nice idea), so that you can assert all needed by that call. You always have two lines but could do in one which would be more symmetrical to the very convenient looking usages of "assert_value".

node/src/database/config_dumper.rs
```diff
@@ -679,6 +670,10 @@ mod tests {
         assert_eq!(actual_value, expected_value);
     }
 
+    fn assert_none(key: &str, map: &Map<String, Value>) {
+        assert!(!map.get(key).is_none());
+    }
+
```
bertllll, line 681: I have two notes on this. 

First, I somewhat cannot believe this assertion stands right. I think it tests something completely different than what the name of this helper function tries to convey. 

It only tests that "a certain parameter (and its row) _IS_ present in the db". That is far from what you intended to get. 

Let me suggest a solution:
You should simply reach the value from the set by using 'get(key)' but don't stop there, now you got an enum with variants representing possible values that can take part in a JSON. Look at the body of "assert_value" above your function. Do you see 'Value::String()'? That's one of the variants. Another one is 'Value::Null'. Your assertion should only make sure that you get exactly this Value when you ask for your parameter.

node/src/node_configurator/configurator.rs
```diff
@@ -2040,7 +2045,49 @@ mod tests {
         let (_, context_id) = UiSetConfigurationResponse::fmb(response.body.clone()).unwrap();
         assert_eq!(context_id, 4444);
         let check_start_block_params = set_start_block_params_arc.lock().unwrap();
-        assert_eq!(*check_start_block_params, vec![166666]);
+        assert_eq!(*check_start_block_params, vec![Some(166666)]);
+        TestLogHandler::new().exists_log_containing(&format!(
+            "DEBUG: {}: A request from UI received: {:?} from context id: {}",
+            test_name, msg, context_id
```
bertllll: We try not to assert on Debug logs 🤷‍♂️

node/src/node_configurator/configurator.rs
```diff
@@ -549,7 +549,8 @@ impl Configurator {
             persistent_config.earning_wallet_address(),
             "earningWalletAddressOpt",
         )?;
-        let start_block = Self::value_required(persistent_config.start_block(), "startBlock")?;
+        let start_block_opt =
+            Self::value_not_required(persistent_config.start_block(), "startBlock")?;
```
bertllll, line 555: You haven't provided any assertion for this where the value was None.

node/src/node_configurator/configurator.rs
```diff
@@ -817,11 +818,15 @@ impl Configurator {
         string_number: String,
         config: &mut Box<dyn PersistentConfiguration>,
     ) -> Result<(), (u64, String)> {
-        let block_number = match string_number.parse::<u64>() {
-            Ok(num) => num,
-            Err(e) => return Err((NON_PARSABLE_VALUE, format!("start block: {:?}", e))),
+        let block_number_opt = if "none".eq_ignore_ascii_case(&string_number) {
```
bertllll, line 790: I'm sorry to be adding in more work for you but I need to be thorough. I'll bet the use of this function (from a library) isn't tested. 

Once you use something that can behave variously or produce different outcomes you will need to test it separately. 

I recommend to test this small function only, on values like "noNe, none, NONE", always passing the same. Using a PersistentConfigurationMock and assertions on the params of set_start_block(), making sure it is supplied by the 'None' value properly.

node/src/database/db_migrations/migrations/migration_9_to_10.rs
```diff
@@ -0,0 +1,71 @@
+use crate::database::db_migrations::db_migrator::DatabaseMigration;
+use crate::database::db_migrations::migrator_utils::DBMigDeclarator;
+
```
bertllll: Ohhhhhhh. This took me a good amount of time before I realised. That what I want to tell you is probably not going to be good news to you.

It will affect a lot of comments that I'd made before this moment.  

Thinking of schema of the database, you must admit that with your card there has been _no_ change. Not even a fingernail. 

The data types in the columns for the config table and the values of the parameters have ever been declared as an optional TEXT.

Which is exactly what you are using. 

Now, let's thing about what older databases could have had in their row for the start_block. Well, it was always a number. (And even if there was a way how it could yield 'Null', it would still be completely in accordance with your design.

That means _you don't need any db migration at all_. For that reason, you should make sure all the related code will go away. I'm sorry. 

Please don't forget delete this whole file.

bertllll (commented):
There is an important comment at the beginning of the file "migration_9_to_10.rs".

masq/src/commands/set_configuration_command.rs
masqrauder: I think this now does what you mean... 2 separate test methods one for Err cases and this one for Ok cases.

node/src/blockchain/blockchain_bridge.rs
masqrauder: My opinion of the logging is that an error from the blockchain service provider doesn't have to be considered an error for the user if the application has a suitable resolution for it. Warning seemed too severe but Info didn't seem like too much. I understand there's a natural mismatch between logging the error at Info level, however, I think it would be worse to just ignore the error in case we receive a response we didn't expect. I can change it to DEBUG and remove the assertions for it.

node/src/blockchain/blockchain_bridge.rs
masqrauder: I had added this test in GH-600 and thought I had broken the behavior and cause the name to no longer match. I realize now the error it mentions is due to the blockchain response and not the database. Adding another test for the persistence related error.

node/src/test_utils/database_version_0_sql.txt
masqrauder: Reverted

node/src/database/db_migrations/migrations/migration_9_to_10.rs
masqrauder: Removed migration

node/src/node_configurator/configurator.rs
masqrauder: `eq_ignore_ascii_case()` looks to me like it's part of the Rust standard library, isn't it? Normally I try not to write tests that only test external library code already having its own tests. Do you agree?

node/src/node_configurator/configurator.rs
masqrauder: Is this where that is checked or do I need something more?: https://github.com/MASQ-Project/Node/blob/GH-606/node/src/node_configurator/configurator.rs#L2090

node/src/accountant/scanners/mod.rs
```diff
@@ -938,13 +939,20 @@ impl ReceivableScanner {
...
+                match txn.commit() {
+                    Ok(_) => debug!(logger, "The new_start_block ({}) is not greater than the current_start_block ({}). \
+                    Committed received transactions.", &new_start_block, &current_start_block),
+                    Err(e) => panic!("Commit of received transactions failed: {:?}", e),
+                }
+            }
+        } else {
+            match txn.commit() {
+                Ok(_) => debug!(logger, "Committed received transactions."),
+                Err(e) => panic!("Commit of received transactions failed: {:?}", e),
```
masqrauder: These are the additional 2 paths that needed the `txn.commit()` calls

multinode_integration_tests/docker/blockchain/entrypoint.sh
```diff
@@ -1,3 +1,16 @@
...
+# While on the transaction fee it's a choice done by us, with the latter, there probably isn't any other solution given
+# the mechanism how the deployment of the blockchain smart contract generates the entire token supply only on
+# the account of the contract owner's wallet from where it must be sent out to other wallets if needed.
+
+node /app/ganache-core.docker.cli.js \
+    -h 0.0.0.0 \
+    -p 18545 \
+    --networkId 2 \
+    --verbose \
+    --mnemonic "timber cage wide hawk phone shaft pattern movie army dizzy hen tackle lamp absent write kind term toddler sphere ripple idle dragon curious hold"
```
bertllll, line 8: I'm not consistent on how we should finish this branch regarding these files in the docker folder. 

Do you think it is not reasonable to put the master alike code in here? Especially the comment speaks about things that cannot be found on your branch, instead the GH-711 has it. Deleting the comment is necessary, therefore I'm suggesting to undo the changes in both files:
```
multinode_integration_tests/docker/blockchain/entrypoint.sh
multinode_integration_tests/docker/blockchain/Dockerfile
```

multinode_integration_tests/src/masq_real_node.rs
```diff
@@ -1018,7 +1018,7 @@ impl MASQRealNode {
 
         args.push("test_node_image");
         let mut command = Command::new("docker", Command::strings(args));
-        command.stdout_or_stderr()?;
+        println!("{}", command.stdout_or_stderr()?);
```
bertllll: I'm unsure if we want to keep this debugging feature? I was not the person using it so I have a limited picture of how useful it could be. But it makes the impression on me that it was intended as temporary. 

Could you either cultivate it somehow, maybe with a literal part of the final string introducing what the printed stuff is about? Or simply do a deletion.

multinode_integration_tests/src/mock_blockchain_client_server.rs
```diff
@@ -241,8 +241,10 @@ impl MockBlockchainClientServer {
                         let mut requests = requests_arc.lock().unwrap();
                         requests.push(body);
                     }
-                    let response = responses.remove(0);
-                    Self::send_body(conn_state, response);
+                    if !responses.is_empty() {
+                        let response = responses.remove(0);
+                        Self::send_body(conn_state, response);
+                    }
```
bertllll, line 247: Are you sure this is something we want to have? I don't think so. Let me check if I understand what is going on here: this is a mock server which is supposed to serve requests as if sent to a real blockchain service. And I do think each such a test should also contain the second part, where it awaits a response. The response needs to be set up in the mock before the actual test starts off, if this step is left out it is actually an error of the programmer and therefore it's adequate to halt the test if we happen to find out there's been a request but we can't reach in for a prepared response. That's actually this very moment which you bypassed with the extra if statement. 

Correct me if I can't see the improvement beyond your modification. However, unless you do, I'd definitely revert these lines.

multinode_integration_tests/src/mock_blockchain_client_server.rs
```diff
@@ -600,13 +602,13 @@ mod tests {
...
         let requests = subject.requests();
-        assert_eq! (requests, vec! [
+        assert_eq!(requests, vec![
             "POST / HTTP/1.1\r\ncontent-type: application/json\r\nuser-agent: web3.rs\r\nhost: 172.18.0.1:32768\r\ncontent-length: 308\r\n\r\n{\"jsonrpc\":\"2.0\",\"method\":\"eth_getLogs\",\"params\":[{\"address\":\"0x59882e4a8f5d24643d4dda422922a870f1b3e664\",\"fromBlock\":\"0x3e8\",\"toBlock\":\"latest\",\"topics\":[\"0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef\",null,\"0x00000000000000000000000027d9a2ac83b493f88ce9b4532edcf74e95b9788d\"]}],\"id\":0}".to_string()
         ])
     }
 
     fn connect(port: u16) -> TcpStream {
-        let deadline = Instant::now().add(Duration::from_secs(1));
+        let deadline = Instant::now().add(Duration::from_secs(5));
```
bertllll: Is the increase necessary? Hmm. I think it's unimportant and so I'd rather go back on this. I can see why it could be 1500 millis or 2000 millis perhaps but 5 seconds seems weird.

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -435,7 +432,7 @@ fn expire_payables(path: PathBuf) {
     statement.execute([]).unwrap();
 
     let mut config_stmt = conn
-        .prepare("update config set value = '0' where name = 'start_block'")
+        .prepare("update config set value = '0' where name = 'start_block' and (value is null or value < 1)")
```
bertllll: Is the part saying `value < 1` really important to have? I feel like we don't have to worry there. Even hypothetically, if it was negative it would be good to check if it kills the Node. Just my thoughts, however, I'll leave this up to you. This doesn't bother me in particular.

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -64,7 +61,7 @@ fn verify_bill_payment() {
     assert_balances(
         &contract_owner_wallet,
         &blockchain_interface,
-        "99998043204000000000",
+        "99998381140000000000",
```
bertllll, line 64: Yeah, I'm afraid, that if you follow my advice in the `docker` folder and you put the old version of `ganache` back in these asserted values will start causing failures again, and on the contrary, the old values will make the assertions work fine.

node/src/accountant/scanners/mod.rs
```diff
@@ -938,13 +939,20 @@ impl ReceivableScanner {
                     ),
                 }
                 match txn.commit() {
-                    Ok(_) => {
```
bertllll: Nope. You did not understand my objections. I cannot comment on the exact line but I will leave it here. 

You must not have an operation setting the start_block at any other place than here and only here, that is, after making sure any inbound payments are properly recorded. Now go and look into your code in `node/src/blockchain/blockchain_bridge.rs` at the method `handle_retrieve_transactions`. You'll see what I'm telling is not desirable. You must throw away the `set_start_block` functionality. Which will hopefully also affect some tests.

However that's not all, after it, you need to take care of this area in turn. I noticed that the actor message which introduces the operations in this big function, has the field `new_start_block` declared as an `Option` of `u64`. Please check out the code in blockchain_bridge and look for where you feed this field with values. There is just one place and it always uses Some() wrapping the actual block number around. It's constantly in use. That's apparently wrong. 

We can reason that at the time when we're constructing the `ReceivedPayments` msg there we must already have the start_block defined, which is implied by the preceding operations in there (that you incorporate by one of your past cards). All I'm saying is that field does not need to be optional. However, I'm tired to death at the moment and I'd want you to double-check if I did the survey right. 

Once we have an agreement about it, we need to move over here and fix it so that `if let Some(new_start_block) = msg.new_start_block {` makes sense in this new perspective. The field is not optional anymore. So what can we do? Basically, we've got rid off one of the else clauses and you can erase the lines 953-955. That's great news.

node/src/accountant/scanners/mod.rs
```diff
@@ -938,13 +939,20 @@ impl ReceivableScanner {
...
-                    }
+                    Ok(_) => debug!(logger, "Updated start block to: {}", new_start_block),
                     Err(e) => panic!("Commit of received transactions failed: {:?}", e),
                 }
             } else {
-                warning!(logger, "The new_start_block ({}) is less than the current_start_block ({}). This is not a problem but by checking we avoid rescanning the same blocks again later.", &new_start_block, &current_start_block);
+                match txn.commit() {
+                    Ok(_) => debug!(logger, "The new_start_block ({}) is not greater than the current_start_block ({}). \
+                    Committed received transactions.", &new_start_block, &current_start_block),
+                    Err(e) => panic!("Commit of received transactions failed: {:?}", e),
```
bertllll: Another and even less cool thing I'm witnessing is that I can't see any new tests that support this new code. There should be at least one test modified for the happy path (as I expect there might've been some sort of happy path before too), or a fully-new test, if somebody did a floppy job. Plus, there definitely should be one brand new test that concentrates on the situation when the committing results in an error and so a panic. 

These things are a must.  

Please make this right. 
You'll need to use this special version of a database transaction: `TransactionSafeWrapper`. Make it by using `TransactionSafeWrapper::new_with_builder`. Either get some inspiration from tests that have it in use or just let the code navigate you through since the builder is designed to be the only way of how you can construct a mock version which you can control. And you do want to.

You need to assert by `commit_params` that this function was called, that for the happy path. In the sad path, you are just obligated to ensure the mock will return an error during that function call. So choose an error to be spit out, by setting it up with `commit_result`.    

You're still possessing quite a lot of work to do. Please keep yourself better-focused this time.

multinode_integration_tests/src/mock_blockchain_client_server.rs
masqrauder: I only changed this because `responses` was empty and causing a panic. I thought it was hiding the real test failure. I can revert it if you prefer.

multinode_integration_tests/src/masq_real_node.rs
masqrauder: I think RustRover was complaining that it was an unused result. I figured this was harmless in test code.

multinode_integration_tests/docker/blockchain/entrypoint.sh
masqrauder: Master does not have `-h 0.0.0.0` however `BlockchainServer::wait_until_ready();` watches for 10 seconds to find it with `output.contains("Listening on 0.0.0.0:18545")`. It seems right to be explicit about something the test relies on to succeed. But I can revert it if you prefer.

multinode_integration_tests/src/mock_blockchain_client_server.rs
masqrauder: I can revert this. I may have been trying to resolve macOS multinode failures that were hopeless to solve here.

multinode_integration_tests/tests/verify_bill_payment.rs
masqrauder: I'll revert it. It was a fleeting thought to use the same thing in production code to prevent undoing progress by storing a smaller number for the config value. Then I realized we don't have the ability to have a custom where clause to it.

node/src/accountant/scanners/mod.rs
masqrauder: I agree setting the start block number in a single place is the right way to go.

Where is the single best place for calling `set_start_block`? Let me ask a few questions to show my thought process to arrive at `blockchain_bridge` as the correct answer.

Which _impl_ 'knows' the most about the block number and the range of blocks we're allowed to query with each call? The receivable scanner only knows `new_start_block` because `blockchain_bridge` sends it in the message. What happens if we receive no payments? Why is `blockchain_bridge` sending a message with empty transactions? Isn't that processing cost we don't need to incur? What if the receivable scanner is slow and it hasn't written the `new_start_block` before the `blockchain_bridge` fetches another set of transactions? It will use the old `start_block_nbr` again, correct? Do we know if this is an impossible scenario?

I have an answer for `ReceivedPayments.new_start_block`: Remove it like it was before November 2023.
If we have `set_start_block` in its rightful place in `blockchain_bridge` we don't always have to send the `ReceivedPayments` message to the `Accountant`. Only when there are transactions to scan. If we don't send `new_start_block` at all that totally cleans up trying to `set_start_block` in `finish_scan` and `handle_new_received_payments`. Let the `Accountant` worry about `ReceivedPayments` and let the `BlockchainBridge` worry about the `start_block`.

The reason it's `Option<u64>` is because `PersistentConfiguration` declares:
```Rust
fn start_block(&self) -> Result<Option<u64>, PersistentConfigError>;
fn set_start_block(&mut self, value_opt: Option<u64>) -> Result<(), PersistentConfigError>;
```
making this call very simple:
```Rust
self.persistent_configuration.set_start_block(msg.new_start_block)
```

node/src/accountant/scanners/mod.rs
masqrauder: I mentioned this in the call on Sunday. Why don't I already find `TransactionSafeWrapper` in use in `accountant/scanners/mod.rs` I didn't introduce `txn.commit()` for the first time.

I tried to trace even the happy path using the added logging but I didn't find it.

node/src/accountant/scanners/mod.rs
masqrauder: Ok. If I could strike through the previous comments I would, but GitHub markdown doesn't support it. This has percolated in my subconscious and I realize I am wrong and you are right. All you had to mention is that we have to commit the `new_start_block` in the same transaction with the `ReceivedPayments`. I will also make the `ReceivedPayments` message have `new_start_block: u64` no `Option` needed.

node/src/accountant/scanners/mod.rs
masqrauder: If you haven't seen my [comment](https://github.com/MASQ-Project/Node/pull/374/commits/d93e62cc8ca4e8d381553e49c9e23816ad7fc174#r1776159174) just before this one, please read.

I will provide adequate coverage for these changes and hopefully eliminate the extra paths. so it will be as it was before GH-606 in this part of the code except for passing `Some(msg.new_start_block)` to be persisted.

I do still think there is benefit for not sending the `ReceivedPayements` message when the blockchain transactions are empty. In that case I think it would be more efficient to have `blockchain_bridge` store the `new_start_block` but only if there are no transactions to send. But if you feel it's bad design to store the value in 2 places even though it's protected from conflict by the aformentioned condition and the message sending overhead is not too costly, I'll implement it how you decide. Please tell me.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1467,29 +1665,33 @@ mod tests {
...
+        //         new_start_block: 7,
+        //         response_skeleton_opt: Some(ResponseSkeleton {
+        //             client_id: 1234,
+        //             context_id: 4321
+        //         }),
+        //     }
+        // );
+        let test_log_handler = TestLogHandler::new();
+        test_log_handler.exists_log_containing("DEBUG: BlockchainBridge: Write new start block: 7");
+        test_log_handler
```
masqrauder: @bertllll Depending on your response, I will either uncomment the code in this test to make sure the `ReceivedPayments` message is always sent even if the transactions are empty so the `start_block_nbr` is updated in `accountant/scanner/mod`. The alternative is to not send the message when there are no transactions and update the `start_block_nbr` directly in `blockchain_bridge` and save the message sending overhead.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1684,31 +1665,33 @@ mod tests {
...
+        //     &ReceivedPayments {
+        //         timestamp: received_payments.timestamp,
+        //         payments: vec![],
+        //         new_start_block: 7,
+        //         response_skeleton_opt: Some(ResponseSkeleton {
+        //             client_id: 1234,
+        //             context_id: 4321
+        //         }),
+        //     }
+        // );
```
masqrauder: I was expecting to see my latest [comment](https://github.com/MASQ-Project/Node/pull/374/files#r1781920860) here on this commit but it doesn't appear because I made it from the main PR files view.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1639,13 +1627,10 @@ mod tests {
                 transactions: vec![],
```
bertllll, line 1627: This is not related to the current card, but I noticed that the line 1623 (GitHub doesn't let me leave the comment there in place) is silly. It says that when we query for knowing what number is of the head block at the moment we receive a response saying it's just 0.  The problem is it never can be so if we had already saved in the database the number 6 for the next position to start from. 

Then I realised the test doesn't care because the whole blockchain interface is mocked and that questioned method is only called in the real version. 

The bottom line is, that you probably should get rid of the whole entry for `lower_interface`. That means also the `lower_interface_results` are something what can get gone without any impact.

multinode_integration_tests/docker/blockchain/entrypoint.sh
bertllll: No probs. I'm no longer concerned about this. You can leave it in if you want to.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -1046,7 +1051,37 @@ mod tests {
...
+                transactions: vec![]
+            })
+        );
+    }
+
+    #[test]
+    fn blockchain_interface_retrieve_transactions_start_and_end_blocks_can_be_latest() {
+        let port = find_free_port();
+        let _test_server = TestServer::start (port, vec![
+            br#"[{"jsonrpc":"2.0","id":1,"result":"error"},{"jsonrpc":"2.0","id":2,"result":[{"address":"0xcd6c588e005032dd882cd43bf53a32129be81302","blockHash":"0x1a24b9169cbaec3f6effa1f600b70c7ab9e8e86db44062b49132a4415d26732a","data":"0x0000000000000000000000000000000000000000000000000010000000000000","logIndex":"0x0","removed":false,"topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef","0x0000000000000000000000003f69f9efd4f2592fd70be8c32ecd9dce71c472fc","0x000000000000000000000000adc1853c7859369639eb414b6342b36288fe6092"],"transactionHash":"0x955cec6ac4f832911ab894ce16aa22c3003f46deff3f7165b32700d2f5ff0681","transactionIndex":"0x0"}]}]"#.to_vec()
```
bertllll: I just don't understand what drove you to make the first transaction in the batch return an error. What does it imply in relation to the test? If it's actual the crucial peace of data in the response which set the right path in the code to go along it, maybe we should make it more obvious to the potential reader knowing about the context even less than for example me.  

Maybe you could do some gymnastics
like this: 

```Rust
let error_causing_to_pick_fallback_value = "error";
 let _test_server = TestServer::start (port, vec![
concat!(br#"[{"jsonrpc":"2.0","id":1,"result":"#,  error_causing_to_pick_fallback_value, br#"},{"jsonrpc":"2.0","id":2,"result":[{"address":"0xcd6c588e005032dd882cd43bf53a32129be81302" ..."#
```

node/src/node_configurator/configurator.rs
bertllll: I agree with your take on this but it isn't really addressing what I was poking into. I can go along with you that sometimes my opinions are extreme, as well as here: I just pointed out in my old comment that your only test doesn't cover the behaviour which was added by incorporating the `eq_ignore_ascii_case`. It really is extreme because my reasoning stem from "if somebody went and rewrote the place with some different code, yet accepting "none", there is no guarantee it would still be code supporting also the other letter-casing. Here it is slightly hilarious, but it raises the question whether it is important that this extra little functionality is in there.

Okay, restating myself: my concern was that the test cannot tell apart, if the production code uses 
```
let block_number_opt = if "none".eq_ignore_ascii_case(&string_number)...
```
or simply
```
let block_number_opt = if  string_number == "none" {...
```

You see what I mean?

bertllll (commented):
I'll be happy to approve your changes when you have had a look at these minor issues and tried to work them out.

bertllll (approved):

PR #383, opened 2024-02-12
Review GH-632

czarte:


node/src/sub_lib/utils.rs
```diff
@@ -251,6 +258,41 @@ pub struct MessageScheduler<M: Message> {
     pub delay: Duration,
 }
 
+#[cfg(target_os = "windows")]
+pub fn wsa_startup_init() {
+    let lp_vendor: *mut u8 = 1 as *mut u8;
+    let wsdata: *mut windows_sys::Win32::Networking::WinSock::WSADATA =
+        &mut windows_sys::Win32::Networking::WinSock::WSADATA {
+            wVersion: 2.2 as u16,
+            wHighVersion: 2.2 as u16,
```
dnwiebe: I believe the number you want here is 0x0202: a two-byte value with a 2 in the first byte and a 2 in the second byte. What I think you're getting here is 0x0002, which translates to version 2.0, because the floating-point value 2.2 will be truncated to 2.0 when it's converted to an integer.

Maybe the syntax you're after is `wVersion: 0x0202u16`.

node/src/sub_lib/utils.rs
```diff
@@ -251,6 +258,41 @@ pub struct MessageScheduler<M: Message> {
...
+            iMaxSockets: 0,
+            iMaxUdpDg: win_cfg::SO_MAX_MSG_SIZE as u16,
+            lpVendorInfo: lp_vendor as win_cfg::PSTR,
```
dnwiebe, line 272: These three elements are all obsolete. Could you set them all to zero and provide `// obsolete` comments (or something equally intention-revealing) to alleviate future concerns about values?

Also, that will let you get rid of `lp_vendor` as a mutable pointer to absolute address 0x0000000000000001, which is a bit alarming.

node/src/sub_lib/utils.rs
```diff
@@ -251,6 +258,41 @@ pub struct MessageScheduler<M: Message> {
...
+    let wsa_startup_init: i32 = unsafe { wsa_startup_call(2.2 as u16, wsdata) };
+
+    match wsa_startup_init {
+        0 => {},
+        10091 => panic!("WSAStartup: The underlying network subsystem is not ready for network communication. "),
+        10092 => panic!("WSAStartup: The version of Windows Sockets support requested is not provided by this particular Windows Sockets implementation."),
+        10036 => panic!("WSAStartup: A blocking Windows Sockets 1.1 operation is in progress."),
+        10067 => panic!("WSAStartup: A limit on the number of tasks supported by the Windows Sockets implementation has been reached. "),
+        10014 => panic!("WSAStartup: The lpWSAData parameter is not a valid pointer."),
+        _ => panic!("WSAStartup: WSAStartup returned unimplemented error")
```
dnwiebe: Can you put the error number in the panic message? It'll probably be difficult to get this message, which means that if a user shows us a log with this message in it, we'll kick ourselves for not actually having the number in it.

node/src/sub_lib/utils.rs
```diff
@@ -251,6 +258,41 @@ pub struct MessageScheduler<M: Message> {
...
+            wVersion: 2.2 as u16,
+            wHighVersion: 2.2 as u16,
+            iMaxSockets: 0,
+            iMaxUdpDg: win_cfg::SO_MAX_MSG_SIZE as u16,
+            lpVendorInfo: lp_vendor as win_cfg::PSTR,
+            szDescription: [1u8; 257usize],
+            szSystemStatus: [1u8; 129usize],
+        } as *mut windows_sys::Win32::Networking::WinSock::WSADATA;
+
+    let wsa_startup_init: i32 = unsafe { wsa_startup_call(2.2 as u16, wsdata) };
```
dnwiebe: 0x0202u16 again

PR #423, opened 2024-03-13
GH-777: Improve caching in GitHub workflow

### LINKED ISSUE #777: Improve Caching in the GitHub workflow
It turns out that while caching the cargo files in the workflow, we are using the `key: ${{ runner.os }}-cargo` to uniquely identify and restore the cache.

This has cache invalidation issues, as it doesn't take into account changes in dependencies. To solve it we can introduce the hash of `Cargo.lock` to uniquely identify the cache based on the lockfile.

Here's a better alternative: `key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}`

utkarshg6:


FinsaasGH:
This has been encompassed by https://github.com/MASQ-Project/Node/pull/452

PR #424, opened 2024-03-13
Gh-728: Review One

utkarshg6:


node/src/accountant/scanners/mid_scan_msg_handling/payable_scanner/msgs.rs
```diff
@@ -2,23 +2,27 @@
...
 pub struct QualifiedPayablesMessage {
     pub protected_qualified_payables: Obfuscated,
+    pub consuming_wallet_opt: Option<Wallet>,
     pub response_skeleton_opt: Option<ResponseSkeleton>,
 }
 
 impl QualifiedPayablesMessage {
     pub(in crate::accountant) fn new(
         protected_qualified_payables: Obfuscated,
+        consuming_wallet_opt: Option<Wallet>,
```
dnwiebe, line 20: Maybe if there's no consuming wallet, the payables scan could be short-circuited here, or near here, rather than waiting for the BlockchainBridge to notice.

node/src/accountant/scanners/mod.rs
```diff
@@ -114,6 +112,7 @@ where
 {
     fn begin_scan(
         &mut self,
+        wallet_opt: Option<Wallet>,
```
dnwiebe, line 115: Suggestion: we could have some or all of the parameters for `begin_scan` in a `struct` named in the `Scanner` trait's generic type list, and that `struct` could contain parameters specific to the particular scan under consideration.

This may or may not be a case where "the juice isn't worth the squeeze." If you don't like the idea, we can abandon it or leave it for later in case it becomes obviously necessary.

node/src/accountant/scanners/mod.rs
```diff
@@ -226,8 +226,11 @@ impl Scanner<QualifiedPayablesMessage, SentPayables> for PayableScanner {
                     qualified_payables.len()
                 );
                 let protected_payables = self.protect_payables(qualified_payables);
-                let outgoing_msg =
-                    QualifiedPayablesMessage::new(protected_payables, response_skeleton_opt);
+                let outgoing_msg = QualifiedPayablesMessage::new(
```
dnwiebe, line 229: So somewhere in here (or, better, at the beginning of the method) would be the place where you popped the error if there isn't any consuming wallet?

node/src/accountant/scanners/mod.rs
```diff
@@ -2920,17 +2946,40 @@ mod tests {
...
         let is_scan_running = receivable_scanner.scan_started_at().is_some();
         assert_eq!(is_scan_running, true);
         assert_eq!(result, Err(BeginScanError::ScanAlreadyRunning(now)));
     }
 
+    #[test]
+    #[should_panic(expected = "Accountant failed to provide the earning wallet")]
+    fn receivable_scanner_panics_when_accountant_fails_to_provide_the_earning_wallet() {
+        let now = SystemTime::now();
+        let receivable_dao = ReceivableDaoMock::new()
```
dnwiebe, line 2973: Bert doesn't think we need the `receivable_dao` in this test.

node/src/accountant/test_utils.rs
```diff
@@ -96,6 +96,7 @@ pub fn make_payable_account_with_wallet_and_balance_and_timestamp_opt(
 
 pub struct AccountantBuilder {
     config: Option<BootstrapperConfig>,
+    consuming_wallet: Option<Wallet>,
```
dnwiebe, line 99: `consuming_wallet_opt`...and `logger_opt` below

node/src/accountant/test_utils.rs
```diff
@@ -1579,7 +1582,7 @@ impl NullScanner {
 }
 
 pub struct ScannerMock<BeginMessage, EndMessage> {
-    begin_scan_params: Arc<Mutex<Vec<()>>>,
+    begin_scan_params: Arc<Mutex<Vec<Option<Wallet>>>>,
```
dnwiebe, line 1585: Why just one parameter here? Are the others not being asserted? [insert thundercloud emoji]

node/src/node_configurator/configurator.rs
```diff
@@ -2377,7 +2315,7 @@ mod tests {
         init_test_logging();
         let test_name = "handle_set_configuration_throws_err_for_invalid_min_hops";
         let mut subject = make_subject(None);
-        subject.update_min_hops_subs_opt = Some(Box::new(ConfigChangeSubsNull));
+        // subject.update_min_hops_subs_opt = Some(Box::new(ConfigChangeSubsNull));
```
dnwiebe, line 2318: Commented-out code

node/src/accountant/mod.rs
```diff
@@ -578,6 +572,15 @@ impl Accountant {
         info!(self.logger, "Accountant bound");
     }
 
+    fn handle_config_change_msg(&mut self, msg: ConfigChangeMsg) {
+        if let ConfigChange::UpdateWallets(wallet_pair) = msg.change {
+            self.earning_wallet = wallet_pair.earning_wallet;
+            self.consuming_wallet_opt = Some(wallet_pair.consuming_wallet);
+        } else {
+            trace!(self.logger, "Unexpected message received: {:?}", msg);
```
dnwiebe, line 580: If this condition is of concern, it ought to be stronger than a `trace!`. If it's commonplace, and happens whenever the UI sends a particular kind of message, there probably shouldn't be any log at all here.

If you want the log for testing, the text shouldn't contain alarms like "Unexpected." The message actually is expected; it's just ignored because it's irrelevant. Maybe "Ignored irrelevant message: {:?}"?

node/src/accountant/scanners/mod.rs
utkarshg6: Initially, I proceeded in the direction of creating structs for the parameters of the `begin_scan` function for individual scanner. While implementing the trait, I became curious what if, we decided to retry transactions that were pending for too long, then the `PendingPayable` Scanner will also require the `consuming_wallet`. Thus, the list of parameters for all the Scanners will become the same. 

So, Instead of creating individual structs I decided to remove the `Option<T>`.

PR #429, opened 2024-03-31
GH-606: Review 2

masqrauder:
* Removed database migration that is unnecessary.
* Added `database_version_9_sql.txt` to create the database at version 9 with a null `start_block` defined.
* I left the `set_configuration_command` to be resolved with the card where we remove the ability to set the start block number from the masq set configuration commands.


node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1435,6 +1435,7 @@ mod tests {
             .retrieve_transactions_params(&retrieve_transactions_params_arc)
             .retrieve_transactions_result(Ok(expected_transactions.clone()))
             .lower_interface_results(Box::new(lower_interface));
+        let set_start_block_params_arc = Arc::new(Mutex::new(vec![]));
```
bertllll: This probably isn't an issue now but it strikes how did you manage to add this line while no other line in the test has changed. It seems like it must have been completely missing before then. Weird stuff, given I can see these added lines at many places in your PR. 

EDIT: Regardless it made the review hard to verify, I think, together with the previous commit (the first one from the only two of them) it can be read us complete and proper code.

node/src/test_utils/database_utils.rs
```diff
@@ -40,6 +40,27 @@ pub fn bring_db_0_back_to_life_and_return_connection(db_path: &Path) -> Connecti
     conn
 }
 
+pub fn bring_db_9_back_to_life_and_return_connection(db_path: &Path) -> Connection {
```
bertllll, line 43: I cannot properly review code if there is some code missing. I haven't seen this used anywhere.

node/src/test_utils/database_version_9_sql.txt
```diff
@@ -0,0 +1,18 @@
+create table config (name text not null, value text, encrypted integer not null)
```
bertllll: Seems like this database state does not need to be preserved. I couldn't even find where you were going to use it.

EDIT: Yeah, I'm confident now that this entire file is unnecessary code with no use.

node/src/test_utils/database_utils.rs
bertllll: I double-checked, this function has no use anymore. Should be deleted.

masqrauder:

PR #432, opened 2024-04-02
GH-783 Fix tests failing on non standard environment

czarte:


node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1071,14 +1070,25 @@ mod tests {
...
+            current_dir()
+                .unwrap()
+                .join(&base_dir)
+                .to_string_lossy()
+                .to_string(),
+        );
+        let _create_data_dir = create_dir_all(&data_dir);
         let config_file_relative = File::create(data_dir.join("config.toml")).unwrap();
         fill_up_config_file(config_file_relative);
         let env_vec_array = vec![
```
bertllll, line 1089: I cannot write it lower than here but I continue to think that the DirsWrapperMock doesn't belong in here, it should only be the real one. The code should figure out home_dir and data_dir on its own, genuinely. It would be possible that the test wouldn't pass for that reason of what the tests claims but just because you determined it by these supplied values in the Mock. If I'm wrong and the test fails then... I'd be disappointed... but I think we'll have to think that out even further.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -1071,14 +1070,25 @@ mod tests {
...
         let _guard = EnvironmentGuard::new();
         let _clap_guard = ClapGuard::new();
-        let home_dir = home_dir().expect("expectexd home dir");
-        let data_dir = &home_dir.join("masqhome");
-        let _create_data_dir = create_dir_all(data_dir);
+        let base_dir = ensure_node_home_directory_exists(
+            "node_configurator_standard",
+            "tilde_in_config_file_path_from_commandline_and_args_uploaded_from_config_file",
+        );
+        let home_dir = base_dir.clone();
```
bertllll: I might have overlooked, it is possible, but it seems to me that you actually don't need to create this "base_dir" first and _then_ the "home_dir". I might be able to write directly:

```
            "node_configurator_standard",
            "tilde_in_config_file_path_from_commandline_and_args_uploaded_from_config_file",
        );
```

node/src/daemon/setup_reporter.rs
```diff
@@ -2053,14 +2055,20 @@ mod tests {
...
-                .join("config.toml"),
-        )
-        .unwrap();
+        let config_file_path = home_dir()
+            .expect("expect home dir")
+            .join("masqhome")
+            .join("config.toml");
+        let mut config_file = File::create(&config_file_path).unwrap();
         config_file
             .write_all(b"blockchain-service-url = \"https://www.mainnet.com\"\n")
```
bertllll, line 2060: I mentioned it at the other test already, please try eliminating the `DirsWrapperMock`.

bertllll (approved):

PR #436, opened 2024-04-21
GH-606-Review-3

masqrauder:
I think I've addressed all the feedback from the first and second reviews. I responded to comments in #429. I may understand what happened last time with the previous PR; I suspect I have to be careful using rebase when creating the baseline branch for the next review. Rebasing disturbs the hashes and I have to update my rebased branch to a new hash if it changed.


node/src/database/config_dumper.rs
```diff
@@ -670,8 +667,14 @@ mod tests {
...
-    fn assert_none(key: &str, map: &Map<String, Value>) {
-        assert!(!map.get(key).is_none());
+    fn assert_null(key: &str, map: &Map<String, Value>) {
+        assert!(map.contains_key(key));
+        match map
+            .get(key)
+            .unwrap_or_else(|| panic!("record for {} is missing", key))
+        {
+            value => assert!(value.is_null()),
+        }
```
bertllll: This match should be replaced by simple:

```
let value = map.get(key).unwrap_or_else(|| panic!("record for {} is missing", key));
assert!(value.is_null())
```

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -309,33 +309,39 @@ impl BlockchainBridge {
...
+                        start_block_nbr + max_block_count,
+                        e
+                    );
                     BlockNumber::Number((start_block_nbr + max_block_count).into())
                 }
             }
         };
-        let start_block = if u64::MAX == start_block_nbr {
+        let start_block = if use_latest_block {
             end_block
```
bertllll, line 325: I found an issue you must look at. Given your new test uses only the mock version of the blockchain interface (in order to call `.retrieve_transactions()`), it didn't have a chance to dive in that function. If you consider the parameters your code from here supplies the function with it would most likely cause a panic soon after. 

Notice that you pick `end_block` also to define the value of `start_block`. Because `end_block` is `BlockNumber::Latest` by this time, `start_block` becomes so too. This goes all aligned with the test where you also do params assertion on that method. 

However, go and look inside there, you will bump soon into:

```
 let fallback_start_block_number = match end_block {
            _ => {
                if let BlockNumber::Number(start_block_number) = start_block {
                    start_block_number.as_u64() + 1u64
                } else {
                    panic!("start_block of Latest, Earliest, and Pending are not supported");
                }
            }
        };
```
I read that code so that the panic is reached because we supplied both the `start_block` and `end_block` equal to `BlockNumber::Latest`.

masqrauder:

PR #449, opened 2024-06-16
GH-606-Review-4

masqrauder:
Summary (from AI Assistant)

Changes
1. Changed accountant::mod.rs
Adjusted the field new_start_block in struct ReceivedPayments to be of type `Option<u64>`.
Altered multiple unit tests to reflect the new field type.
Enforced an additional condition to only update the `new_start_block` if its value is greater than the current start block maintained in the persistent configuration.
2. Modified accountant::scanners::mod.rs
Similar to accountant::mod.rs, an additional condition was added before updating the `new_start_block`.
`new_start_block` is now checked in two methods to ensure it isn't updated with lower values which would lead to reprocessing of previously processed blocks.
Various unit tests adjusted to incorporate the changes related to `new_start_block`.
Note: The commit involves improvements and code quality changes and does not appear to add any new features or remove existing ones.


node/src/accountant/scanners/mod.rs
```diff
@@ -859,15 +859,23 @@ impl Scanner<RetrieveTransactions, ReceivedPayments> for ReceivableScanner {
...
+                if new_start_block > current_start_block {
+                    match self
+                        .persistent_configuration
+                        .set_start_block(msg.new_start_block)
+                    {
+                        Ok(()) => debug!(logger, "Start block updated to {}", &new_start_block),
+                        Err(e) => panic!(
+                            "Attempt to set new start block to {} failed due to: {:?}",
+                            &new_start_block, e
+                        ),
```
bertllll: If we do want this if statement to be with us I'd prefer adding an else clause with only a debug log in it. Even though I'm uncertain of why should we allow for these counts being less than the stored value, it seems like a design error if it happens and such we be treatable with a panic. However, even if it could've have happened some day what you're suggesting here then it would not prevent the whole procedure from finishing as it's supposed to. Therefore it's not really that concerning. 

I'm also considering if it should be a WARN log (with an inllustrative message) instead of just DEBUG log, because the testers and users are taught to catch and report anything above the INFO level...so even if it somehow happened there is a chance we would notice. Not sure, but would you say I'm overthinking?

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -351,40 +351,43 @@ impl BlockchainBridge {
...
-                    debug!(self.logger, "No new receivable detected");
+                if let BlockNumber::Number(new_start_block_number) = transactions.new_start_block {
+                    debug!(
+                        self.logger,
+                        "Write new start block: {}",
+                        new_start_block_number.as_u64()
+                    );
+                    if let Err(e) = self
+                        .persistent_config
+                        .set_start_block(Some(new_start_block_number.as_u64()))
```
bertllll: This must be some kind of a merge issue. Now we should always set the start block in Accountant (Why? Because it must be done after all the retrieved transactions are written into the books. Otherwise we're risking impaired state of the books if we restart the Node following a panic or the OS's death, with the start block already advanced for the next scan but having lost some payment information, so with missing inaccurate figures in the database)

I'm sorry but it takes possibly all the code in this file, even the tests. What is useful, though, is the code you wrote in the Accountant. There it is okay.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -183,11 +183,14 @@ where
...
                                 logger,
-                                "Discovered transaction max block nbr: {}",
+                                "Discovered transaction max block nbr: {:?}",
                                 transaction_max_block_number
                             );
                             Ok(RetrievedBlockchainTransactions {
-                                new_start_block: 1u64 + transaction_max_block_number,
+                                new_start_block: transaction_max_block_number
+                                    .map_or(BlockNumber::Latest, |nsb| {
+                                        BlockNumber::Number((1u64 + nsb).into())
```
bertllll: Maybe could you please rewrite this as 

```rust

.map(|nsb|{
     BlockNumber::Number((1u64 + nsb).into()
})
.unwrap_or(BlockNumber::Latest)

```

This is more common, almost nobody uses these fancy functions like .map_or() combining multiple steps in a single call. I'm not saying it's bad on the whole but sometimes it's not the clearest way to go.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -137,12 +137,12 @@ where
                 let response_block_number = match block_request.wait() {
```
bertllll: Could you please use the suffix `_opt` for this variable and any similarly created? As now the typing is different, and it is actually optional.

bertllll: On the other hand, I haven't come across a line where this variable could resolve as None. It seems to me it can never be None. If you agree with me, you may want to simplify it, it would also help with the arguments in `find_transactions_largest_block_number()`.  Which I'd also appreciate.

node/src/database/config_dumper.rs
```diff
@@ -669,12 +669,10 @@ mod tests {
...
         assert!(map.contains_key(key));
-        match map
+        let value = map
             .get(key)
-            .unwrap_or_else(|| panic!("record for {} is missing", key))
-        {
-            value => assert!(value.is_null()),
-        }
+            .unwrap_or_else(|| panic!("record for {} is missing", key));
+        assert!(value.is_null())
```
bertllll: Maybe you could provide a message to this macro which prints if it fails that would print the presented value if it was NOT null. It can make fixing failing tests easier.

node/src/db_config/persistent_configuration.rs
```diff
@@ -568,12 +568,14 @@ impl PersistentConfigurationReal {
...
+        value: Option<T>,
         txn: &mut TransactionSafeWrapper,
     ) -> Result<(), PersistentConfigError> {
-        Ok(self
-            .dao
-            .set_by_guest_transaction(txn, parameter_name, Some(value.to_string()))?)
+        Ok(self.dao.set_by_guest_transaction(
+            txn,
+            parameter_name,
+            value.map(|v| v.to_string()).or(None),
```
bertllll: You don't have to say 

```rust
.or(None)

```

if it is Option and you already provided what happens if it is Some. Only None can be the other value.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
masqrauder: ok

node/src/accountant/scanners/mod.rs
masqrauder: Maybe I'm the one overthinking. I occurred to me there might be a minor race condition (not something at the level I would expect Rust to catch) and I just thought it would be bad if blockchain_bridge had just advanced the block number and the accountant was still processing a previous block of transactions and  blindly updated the `start_block` to something old. This is just to prevent inadvertent retreat. I can certainly add an else clause with a WARN log.

node/src/blockchain/blockchain_bridge.rs
masqrauder: I don't think this is a merge issue. For example, if blockchain_bridge scans 10,000 blocks from 100 to 10,100 and the accountant processes payments received at blocks 4242, and 5758; blockchain_bridge may have persisted that it found all the incoming payments between 100 and 10,100 making it ready to start at 10,101 for the next scan. However, previously if accountant blindly records 4242 then 5758, blockchain_bridge might rescan blocks. By checking to avoid storing a smaller block number, accountant lowers the risk of repeat scans.

I'm reluctant to make such a big change in my card since it's gone on far too long already. I agree with you that it does make sense to record the block number according to what the accountant has successfully recorded. I still think there are some initialization times where we should set and advance the starting block number here in blockchain_bridge while we await our first receivable. If we start and stop the node without receiving any payments the start block number won't advance. When we do restart it will rescan the blocks since the last received payment.

I still argue what the accountant records should be idempotent. It should be safe to reprocess the same transaction block and it can be if we record the sending address, the block (or tx hash), and the amount, we'll never accidentally double count anything. It also would not be harmful to re-scan since it might occur despite our best efforts to prevent it. A database confiscated in a raid won't have any additional information that what is already on the public blockchain -- I agree we are providing a tidy collection of all the 'incriminating evidence' without all the noise. However, the overwhelming benefit is that the books are far less likely to become imbalanced. With the current implementation I suspect frequently declaring bankruptcy will be common.

node/src/db_config/persistent_configuration.rs
masqrauder: Oh right. I'm surprised Rust or RustRover didn't make that recommendation.

masqrauder:

PR #450, opened 2024-06-19
GH-800: Drain the Pipeline

### LINKED ISSUE #800: Drain the Pipeline
As things stand at the time of this writing, when the exit Node is producing a stream of data that the originating Node decides it no longer wants to pay for, the originating Node removes the relevant StreamKey from its ProxyServer's StreamKeyed HashMaps and sends a `ClientRequestPayload_0v1` package to the exit Node with `last_data = true`. The intent is to have the exit Node's ProxyClient break its connection with the server so that the stream of data stops.

However, even if this is accomplished, there will still be a few CORES packages that are already on their way that will arrive at the originating Node's ProxyServer after the `last_data = true` `ClientRequestPayload_0v1` is sent. Since the ProxyServer has removed all trace of that StreamKey from its HashMaps already, though, it will have no idea how to pay for those last few CORES packages, and so the exit Node and the relay Nodes--which moved that data in good faith--will remain unpaid, and eventually the originating Node will be banned.

The originating Node needs to pay for the good-faith work of the exit and relay Nodes, even though it meant to refuse those last few packets, and even though the stream to the browser will be closed by the time they arrive, so they'll just be discarded.

__Task:__
Modify the ProxyServer so that when it directs an incoming data stream to be terminated, it doesn't immediately purge the StreamKey from its HashMaps, but instead schedules an actor message containing the StreamKey to be sent to itself after a reasonable number of seconds. When it receives _that_ message, it should purge the StreamKey.

__Note:__
There may be a temptation to design the solution so that the StreamKey stays alive as long as it's being used, and only dies after it has been inactive for awhile. Avoid this: it's an attack surface that will be exploited by evil Nodes that will continue sending data after they've been told to shut up, because they can continue to "earn" money that way. Applying a definite time limit after the `last_data = true` will not prevent this kind of attack, but it will limit its effectiveness.

As long as incoming data is being consumed by the browser, the spoofing capability of an evil Node is limited, because the browser will detect any TLS data manipulation and protest. However, once the browser shuts down a stream and can no longer see what's on it, an evil Node can send whatever it wants.

### LINKED ISSUE #801: ProxyClient should heed instruction to shut up
The way the ProxyClient presently works is that when a TCP stream to the server is established, it is immediately split into a Read half and a Write half. The Write half is stored in `StreamHandlerPoolRealInner::stream_writer_channels`, keyed by its `StreamKey`, and the Read half is spawned off into its own Future that listens for data from the server and sends Actor messages to the ProxyClient when it arrives.

When the server closes such a connection, the Read half of the stream detects the closure and informs the ProxyClient via its `stream_killer` channel. The ProxyClient then sends a `last_data = true` CORES package to the originating Node, and everything is taken care of in order.

When the originating Node sends a `last_data = true` CORES package, the ProxyClient looks up the Write half of the stream in `stream_writer_channels` and instructs it to close the stream. When the server sees that its client wants to close the stream, IN MOST CASES it will acknowledge the closure, which the Read half of the stream will detect, as above, and everything will be taken care of in order.

However, sometimes the server is in the middle of sending lots and lots of data (for example, when the user is watching a video stream) and might not get around to checking whether the ProxyClient has requested closure of the stream for hours. In that case, the Read half of the stream will continue to receive the data and pass it on in Actor messages to the ProxyClient, which will continue to provide (and charge for) exit services for that data. The routing Nodes will continue to provide (and charge for) routing services for the data, and the originating Node will have nowhere to send it, since the browser has shut down the stream. Not only that: the originating Node, having retired the `StreamKey`, will have no idea whom to pay for all that unwanted data, and so will go rapidly delinquent with the exit Node and the relay Nodes and will be banned.

The problem here is that there is no way for the ProxyClient, when it receives a `last_data = true` CORES package from the originating Node, to tell the Read half of the stream to shut down and stop passing data from the server.

__Task:__ Create some kind of communication method between the ProxyClient (or perhaps its `StreamHandlerPool`, if that is more appropriate) to allow the ProxyClient to instruct the Read half of the stream associated with a particular `StreamKey` to shut down and abandon its stream from the server. Probably this will involved modifying the `poll()` method of the `StreamReader` to check for that communication either before or after every time it handles incoming data from the stream.

Whatever this communication method ends up being, modify the ProxyClient to employ it when it receives a `last_data = true` CORES package from the originating Node.

__Note:__ We have run into situations before, when closing virtual streams, where too many `last_data = true` CORES packages are sent. For example: the browser closes its stream, so the originating Node sends `last_data` to the exit Node. The exit Node has its Write half close the stream to the server, whereupon the server acknowledges by closing its end of the stream as well. The exit Node's read half detects this closure and informs the ProxyClient, at which point the Proxy Client sends a `last_data` CORES package back to the originating Node to tell it the stream has been closed. This last CORES package is 1) unnecessary, 2) a needless expense, and 3) unrecognized and thrown away by the originating Node because its `StreamKey` has already been retired on that end. When you design your solution for the __Task:__ above, don't make this problem any worse.

__Extra Credit:__ See if you can find a way to clean up the needless-`last_data` issue above for _all_ stream closing scenarios, not just browser-initiated ones.

utkarshg6:


node/src/proxy_client/stream_reader.rs
```diff
@@ -386,18 +402,49 @@ mod tests {
...
+        init_test_logging();
+        let test_name = "stream_reader_shuts_down_when_it_receives_the_shutdown_signal";
+        let (shutdown_tx, shutdown_rx) = unbounded();
+        let mut subject = make_subject();
+        subject.shutdown_signal = shutdown_rx;
+        subject.logger = Logger::new(test_name);
+
+        shutdown_tx.send(()).unwrap();
+
+        assert_eq!(subject.poll(), Ok(Async::Ready(())));
```
dnwiebe: I think `subject.poll` should be the Act here, and `shutdown_tx.send()` should be part of Arrange.

node/src/proxy_server/mod.rs
```diff
@@ -58,6 +58,8 @@ use tokio::prelude::Future;
 pub const CRASH_KEY: &str = "PROXYSERVER";
 pub const RETURN_ROUTE_TTL: Duration = Duration::from_secs(120);
 
+pub const STREAM_KEY_PURGE_DELAY: Duration = Duration::from_secs(5);
```
dnwiebe: Think about this value. It's not only the time allowed for straggling packets to make it to their destination, it's also the time allowed for evil exit Nodes to pad connections with garbage that they'll get paid for. Remember,  the browser has already shut this stream down, so the user will never see any of the straggling packets or their consequences. Therefore an evil exit Node, once it gets notification that a stream is to be shut down, can immediately start sending big packets full of random trash back to us, and we'll accept and pay for them for this long. This parameter needs fairly careful tuning for the sake of security.

dnwiebe: A suggestion might be to put in logging logic (if it doesn't already exist) to log whenever we receive one of these straggler packets that we're going to throw away, and then the testers can provide us logs to look at so that we can see how late stragglers are in real life, and tune this parameter to a balance between getting banned as a deadbeat and paying for trash from evil exit Nodes.

node/src/proxy_server/mod.rs
```diff
@@ -603,6 +633,10 @@ impl ProxyServer {
     }
 
     fn purge_stream_key(&mut self, stream_key: &StreamKey) {
+        debug!(
```
dnwiebe, line 665: We should probably check here to see if the stream key has already been purged (by a previous delayed message) and abort if it has.

node/src/proxy_server/mod.rs
```diff
@@ -603,6 +633,10 @@ impl ProxyServer {
...
+            self.logger,
+            "Retiring stream key {}: no more data", &stream_key
+        );
         let _ = self.keys_and_addrs.remove_a(stream_key);
         let _ = self.stream_key_routes.remove(stream_key);
         let _ = self.tunneled_hosts.remove(stream_key);
```
dnwiebe, line 671: Suggestion: Combine these last two `HashMaps` into one, with the `StreamKey` indexing a single structure that has both route and tunneled host in it. Add a third field that's a boolean flag that says whether the stream is in the process of waiting for stragglers. Use this flag in `handle_client_response_payload()` above to make sure each dying stream gets only one `StreamKeyPurge` message, no matter how many straggling packets come in for it.

node/src/proxy_server/mod.rs
```diff
@@ -3576,32 +3612,50 @@ mod tests {
...
-        let peer_actors = peer_actors_builder().dispatcher(dispatcher).build();
+        let peer_actors = peer_actors_builder()
+            .dispatcher(dispatcher)
+            .proxy_server(proxy_server)
+            .build();
         subject_addr.try_send(BindMessage { peer_actors }).unwrap();
 
         subject_addr.try_send(first_expired_cores_package).unwrap();
-        subject_addr.try_send(second_expired_cores_package).unwrap(); // should generate log because stream key is now unknown
+        subject_addr.try_send(second_expired_cores_package).unwrap(); // should shedule a new message
```
dnwiebe: schedule

node/src/proxy_server/mod.rs
```diff
@@ -3576,32 +3612,50 @@ mod tests {
...
+        let scheduled_msg =
+            proxy_server_recording.get_record::<MessageScheduler<StreamKeyPurge>>(0);
+        assert_eq!(transmit_data_msg.endpoint, Endpoint::Socket(socket_addr));
+        assert_eq!(transmit_data_msg.last_data, true);
+        assert_eq!(transmit_data_msg.data, b"16 bytes of data".to_vec());
+        assert_eq!(
+            scheduled_msg,
+            &MessageScheduler {
+                scheduled_msg: StreamKeyPurge { stream_key },
+                delay: Duration::from_secs(5)
```
dnwiebe: Should this be 5, or should it be a reference back to that manifest constant? Remember, we'll probably be tuning that constant a bit; we don't want this test to fail every time we change it.

node/src/proxy_server/mod.rs
```diff
@@ -3643,12 +3698,45 @@ mod tests {
...
+                System::current().stop();
+            }),
+        };
+        proxy_server_addr
+            .try_send(MessageScheduler {
+                scheduled_msg: post_assertions_msg,
+                delay: Duration::from_millis(stream_key_purge_delay_in_millis + offset_in_millis),
+            })
+            .unwrap();
+        system.run();
```
dnwiebe, line 3893: If this test fails, does the test panic, or does this `system.run()` just return an unsuccessful code? If the latter, this test will always pass. If you're absolutely sure that the test will panic if it fails, okay, but if it were me, I'd modify one of those `assert!()`s (maybe by adding or removing a `!`) to make sure the test fails and run it to see if I was right.

dnwiebe: Also, this is a very clever test. I hope it's not too clever for folks to figure out. If you think it might be, a few judicious comments might help clarify it. Maybe a banner comment at the top with a little ASCII timeline diagram explaining what you're doing with those -100ms and +100ms messages.

node/src/proxy_server/mod.rs
dnwiebe: Further suggestion: That flag that makes sure each `StreamKey` gets only one `StreamKeyPurge` message? Make it a `struct` that contains both that boolean flag and also two timestamps: one that carries the instant the terminal `ExpiredCoresPackage` was received, and one that carries the instant the most recent straggler was received. That way, when the `StreamKeyPurge` message is finally processed, you can create a log saying something like, "For StreamKey <blah>, stragglers kept arriving for <blah> milliseconds after it was ordered closed." That will help immensely with tuning the delay for the real world.

node/tests/connection_shutdown_test.rs
```diff
@@ -0,0 +1,80 @@
...
+        true,
+        false,
+        true,
+    );
+    let (server_write_error_tx, server_write_error_rx) = unbounded();
+    let server_port = find_free_port();
+    let join_handle = thread::spawn(move || {
+        endless_write_server(server_port, server_write_error_tx);
+    });
+    let mut stream = TcpStream::connect(SocketAddr::from_str("127.0.0.1:80").unwrap()).unwrap();
```
dnwiebe: Can you call this `client_stream`, or maybe even `browser_stream` so that it's clearer why we're connecting specifically to port 80 rather than to some `find_free_port()` port?

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -29,20 +29,28 @@ use tokio::prelude::future::{err, ok};
 use trust_dns_resolver::error::ResolveError;
 use trust_dns_resolver::lookup_ip::LookupIp;
 
+// TODO: This should be renamed to ProxyClientStreamHandlerPoolReal (or something more concise)
+// to differentiate it from the other StreamHandlerPool, which, unlike this, is an actor.
```
dnwiebe: Tune this comment a little: we don't want to name the trait `XxxXxxReal`.

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -729,11 +787,10 @@ mod tests {
             let peer_actors = peer_actors_builder().proxy_client(proxy_client).build();
             let resolver = ResolverWrapperMock::new()
                 .lookup_ip_success(vec![IpAddr::from_str("2.3.4.5").unwrap()]);
```
dnwiebe: If you wanted to set `peer_addr` a little earlier, you could make this `peer_addr.ip()` instead of a literal address...

node/src/proxy_client/stream_handler_pool.rs
```diff
@@ -1735,6 +1919,49 @@ mod tests {
         );
     }
 
+    #[test]
+    fn clean_up_dead_streams_logs_when_the_shutdown_channel_is_down() {
```
dnwiebe: Maybe a little higher-level test name? Something about the `StreamReader` being down, rather than a low-level reference to the shutdown channel?

node/src/proxy_server/mod.rs
utkarshg6: It shows this once it fails: `assertion failed: proxy_server.keys_and_addrs.is_empty()`.

node/src/proxy_server/mod.rs
utkarshg6: I realized that once the fields are added to the struct, they will be saved inside Option<T>, which will add an extra check whenever we retrieve the value from any field.

node/src/proxy_server/mod.rs
```diff
@@ -436,6 +438,39 @@ impl ProxyServer {
...
+            .expect("ProxyServer is dead");
+    }
+
+    fn log_straggling_packet(&self, stream_key: &StreamKey, old_timestamp: &SystemTime) {
+        let duration_since = SystemTime::now()
+            .duration_since(*old_timestamp)
+            .expect("time calculation error");
+        debug!(
+            self.logger,
+            "Straggling packet received for a stream key {:?} with a delay of {:?}",
```
dnwiebe: I'd like to see a packet length in this log to make forensics easier.

node/src/proxy_server/mod.rs
```diff
@@ -436,6 +438,39 @@ impl ProxyServer {
...
+                delay: self.stream_key_purge_delay,
+            })
+            .expect("ProxyServer is dead");
+    }
+
+    fn log_straggling_packet(&self, stream_key: &StreamKey, old_timestamp: &SystemTime) {
+        let duration_since = SystemTime::now()
+            .duration_since(*old_timestamp)
+            .expect("time calculation error");
+        debug!(
```
dnwiebe: This is probably more serious than a debug situation. It's not an error situation, because we're going to handle it, but it could qualify as a precursor to a situation involving unpaid bills and delinquency banning.

We're going to pay the bill for this particular packet, but the fact that we're receiving straggling packets means that we might be in a situation where the exit Node is evil and will ignore the `last_data = true` packet we've sent. In that case, once the `stream_key_purge_delay` is over and we throw away the stream, the evil exit Node will probably keep sending data, which will result in charges from the exit Node and the relay Nodes. If that happens, and we look through the logs to investigate it, we'll want to see these logs. Therefore, these should be either INFO or WARN level; I'm leaning toward WARN.

(These logs shouldn't happen on every connection: only connections actively carrying data that are closed by the browser instead of the server. That happens, but comparatively rarely.)

node/src/proxy_server/mod.rs
```diff
@@ -495,15 +530,12 @@ impl ProxyServer {
...
-                        .schedule_stream_key_purge
-                        .try_send(MessageScheduler {
-                            scheduled_msg: StreamKeyPurge { stream_key },
-                            delay: self.stream_key_purge_delay,
-                        })
-                        .expect("ProxyServer is dead");
+                    match self.stream_key_ttl.get(&stream_key) {
+                        None => self.schedule_stream_key_purge(stream_key),
+                        Some(old_timestamp) => {
+                            self.log_straggling_packet(&stream_key, old_timestamp)
```
dnwiebe: This isn't where we want to log straggling packets. A "straggling" packet with `last_data = true` is exactly the correct, polite, and classy response to our browser closing an active stream.

It's the continuing packets with `last_data = false` that are the real stragglers that might precede an evil-exit-Node attack, and that we want to see.

node/src/proxy_server/mod.rs
```diff
@@ -3618,8 +3650,8 @@ mod tests {
             .build();
         subject_addr.try_send(BindMessage { peer_actors }).unwrap();
 
-        subject_addr.try_send(first_expired_cores_package).unwrap();
```
dnwiebe, line 3582: I can't get this comment any closer to the context; but I just noticed that this test isn't realistic, because it has the ProxyServer receiving two consecutive `last_data = true` CORES packages on the same connection, and that can't happen in real life.

In some cases, this won't matter, because the tests won't be about the `last_data` flag and don't care about its value; but I bet this test does. You might want to set the flag to `false`, make that second CORES package mutable, and then reach into it and flip its flag to `true` after it's created.

node/src/proxy_server/mod.rs
```diff
@@ -3635,13 +3667,52 @@ mod tests {
             scheduled_msg,
             &MessageScheduler {
                 scheduled_msg: StreamKeyPurge { stream_key },
-                delay: Duration::from_secs(5)
+                delay: STREAM_KEY_PURGE_DELAY
             }
-        )
+        );
+        TestLogHandler::new()
```
dnwiebe: The delay-the-purge machinery shouldn't even kick in here, since we're getting notification that the far end is shutting down, rather than sending notification to the far end that the near end is shutting down. That's when we ought to delay the purge. Here, we ought to purge immediately*.

 * Ultimately, after the entity beyond the exit Node drops the far end of the connection, we want to wait until the client drops the near end of the connection to purge the StreamKey. That extra bit of life for the StreamKey won't mean anything in HTTP, but it might be important for other protocols. If this is not what we're currently doing, though, that modification is for another card, not this one.

node/src/proxy_server/mod.rs
```diff
@@ -3635,13 +3667,52 @@ mod tests {
...
+        );
+        let stream_key = StreamKey::make_meaningless_stream_key();
+        let timestamp = SystemTime::now()
+            .checked_add(Duration::from_secs(10))
+            .unwrap();
+        let _ = subject.log_straggling_packet(&stream_key, &timestamp);
     }
 
     #[test]
     fn handle_client_response_payload_purges_stream_keys_for_terminal_response() {
```
dnwiebe: You know this already, but just to mark this location: this is about receiving `last_data = true` rather than sending it, so the delayed-purge logic shouldn't come into play here.

node/src/proxy_server/mod.rs
```diff
@@ -3793,7 +3849,7 @@ mod tests {
                 assert!(!proxy_server.stream_key_routes.is_empty());
                 assert!(!proxy_server.tunneled_hosts.is_empty());
                 TestLogHandler::new().exists_log_containing(&format!(
-                    "DEBUG: {test_name}: Last data received for stream key vYMR621EYFzvxTJqpzrGKr3KMbc, \
+                    "DEBUG: {test_name}: Last data received for stream key zpi6SfvohvBgqq9y8zLgzIhxPS4, \
```
dnwiebe: Instead of having the literal value of `stream_key` here, can you make this assert refer to the variable `stream_key` itself, so that if the method of making a meaningful stream key changes in the future this test won't fail?

Also line 3870 below.

Further observations about this message.

"Received" is a little ambiguous, since the ProxyServer can receive data either from the client or from a relay Node. How about something like "Client closed stream referenced by stream key <blah>?"

Also, I'm pretty sure the `tunneled_hosts` map contains names of server hosts, which are the _targets_ of tunnels, rather than tunnels themselves. I think having the host name in the message is useful, but we shouldn't say that the host is being tunneled through: that's not right. Furthermore, the fact that you have the hostname in the message makes me think how useful it would be to have even more information in the message. What information? Well...the next most useful thing after the hostname might be the public key of the exit Node. After that, it'd probably be the IP address of the first relay Node, which you could probably get from the route in `stream_key_routes`. If you still felt ambitious after that, you could try including the entire route in the message; although that's probably pushing it a little.

And finally...here in this assertion, you should probably have a value retrieved from the `tunneled_hosts` map, rather than a literal "hostname."

node/src/proxy_server/mod.rs
```diff
@@ -3690,11 +3690,70 @@ mod tests {
...
+        subject.handle_client_response_payload(expired_cores_package);
+
+        assert!(subject.keys_and_addrs.is_empty());
+        assert!(subject.stream_key_routes.is_empty());
+        assert!(subject.tunneled_hosts.is_empty());
+    }
+
+    #[test]
+    fn proxy_server_schedules_stream_key_purge_once_shutdown_order_is_received_for_stream() {
         /*
```
dnwiebe, line 3768: Could you add a box here about the event of ordering the stream to shut down? Also, it would be nice if the (currently) first and third boxes made reference to the assertion messages they match. So...something like:

1. (0ms) - Stream shutdown is ordered
2. (stream_key_purge_delay_in_millis - offset_in_millis) - Pre-purge assertion message finds records
3. (stream_key_purge_delay_in_millis) - Stream is purged
4. (stream_key_purge_delay_in_millis + offset_in_millis) - Post-purge assertion message finds no records

node/src/proxy_server/mod.rs
utkarshg6: I added some of the changes mentioned under the card.

On the other hand, it seems that extracting the other values may not be feasible at this point.

```rust
match self
    .stream_key_routes
    .get(&stream_key)
    .unwrap()
    .expected_services
{
    ExpectedServices::OneWay(expected_services) => match expected_services[0] {
        ExpectedService::Routing(_, _, _) => {}
        ExpectedService::Exit(_, _, _) => {}
        ExpectedService::Nothing => {}
    },
    ExpectedServices::RoundTrip(_, _, _) => {}
}
```

node/src/proxy_server/mod.rs
```diff
@@ -3849,33 +3856,36 @@ mod tests {
                 assert!(!proxy_server.stream_key_routes.is_empty());
                 assert!(!proxy_server.tunneled_hosts.is_empty());
                 TestLogHandler::new().exists_log_containing(&format!(
-                    "DEBUG: {test_name}: Last data received for stream key zpi6SfvohvBgqq9y8zLgzIhxPS4, \
-                    which was tunneling through host Some(\"hostname\"). It will be purged after 500ms."
+                    "DEBUG: {test_name}: Client closed stream referenced by stream key {:?}, \
+                    which was tunneling to the host Some(\"hostname\"). \
```
dnwiebe: This won't block approval, but in case you felt energetic, this could be made friendlier.

dnwiebe (approved):

PR #451, opened 2024-06-23
GH-606-Review-5

masqrauder:
GH-606: Apply PR review 4 feedback changes

I think I responded to all the view comments. If I missed any please let me know.


node/src/accountant/scanners/mod.rs
```diff
@@ -942,6 +944,8 @@ impl ReceivableScanner {
                     }
                     Err(e) => panic!("Commit of received transactions failed: {:?}", e),
                 }
+            } else {
+                warning!(logger, "The new_start_block ({}) is less than the current_start_block ({}). This is not a problem but by checking we avoid rescanning the same blocks again later.", &new_start_block, &current_start_block);
```
bertllll: If I remember correctly, there are plans that the start_block parameter will be eventually hidden from the user. So I like what you did here and if not best then it will go away and solve this tricky caveat on its own. :-)

node/src/db_config/persistent_configuration.rs
```diff
@@ -571,11 +571,9 @@ impl PersistentConfigurationReal {
         value: Option<T>,
```
bertllll: I know this isn't your fault. Therefore don't feel obligated to do the fixing but it would be nice to have this argument named as `value_opt`.

bertllll (approved):

node/src/db_config/persistent_configuration.rs
masqrauder: Ok changed all `value: Option<...` to `value_opt: Option<...` in the file.

bertllll: Thanks :pray:

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -320,30 +330,38 @@ fn make_init_config(chain: Chain) -> DbInitializationConfig {
...
     assert_eq!(
         token_balance,
-        web3::types::U256::from_dec_str(expected_token_balance).unwrap(),
-        "Actual TokenBalance {} doesn't match with expected {}",
+        expected_token_balance,
+        "Actual TokenBalance {} exceeds the expected {} by {}",
         token_balance,
-        expected_token_balance
+        expected_token_balance,
+        token_balance - token_balance
```
masqrauder: @bertllll did you mean this instead?

```suggestion
        token_balance - expected_token_balance
```

masqrauder:

PR #472, opened 2024-08-13
Gh 784 fixing review

czarte:


.github/workflows/ci-matrix.yml
```diff
@@ -42,6 +42,12 @@ jobs:
           key: ${{ runner.os }}-cargo
       - name: Build ${{ matrix.target.os }}
         run: |
+          echo "before ci/all disc diagnostics ------>"
```
bertllll: Can you please make these `echo`s use the same style as we use usually in such places. You can find inspiration maybe even in this file but mainly in the `ci/xxx.sh` scripts. Thank you :pray:

ci/all.sh
```diff
@@ -45,4 +45,10 @@ cd "$CI_DIR/../automap"
 ci/all.sh "$PARENT_DIR"
 echo "***                                             AUTOMAP TAIL                                          ***"
 echo "*********************************************************************************************************"
+echo "*********************************************************************************************************"
+echo "***                                             IP COUNTRY HEAD                                       ***"
```
bertllll, line 49: Not wanting to be a nitpicker really, but the other headers are all centered to a common axis. Could you please preserve that?

ip_country/Cargo.toml
```diff
@@ -21,3 +23,6 @@ path = "src/main.rs"
 [lib]
 name = "ip_country_lib"
 path = "src/lib.rs"
+
+[profile.release]
+codegen-units = 1
```
bertllll, line 26: This is still uncertain if it helps at all. I have a feeling it doesn't and it may be therefore removed from here. 

What we should do instead is to tell Finsaas about this (cited from the web):

> Setting codegen-units to 1 gives even better code quality than thin local LTO, but takes longer. Some authors of binary rust crates always use that setting for release builds because they are willing to accept the extra compile times for the highest code quality.
> 
> I wonder how many of the authors who don't use it do so because they don't know about it.
> 
> Personally, i think release builds should use codegen-units=1 by default. The meaning of a release build is "take as long as needed to build the fastest possible code". Users shouldn't have to keep track of some set of additional settings needed to achieve that. If authors want to sacrifice performance to get faster builds, or have experimentally confirmed that there is no performance impact, then they can still set codegen-units to something else

ip_country/ci/all.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: copyright ...should be 2024 as well as at any other file in the IP_COUNTRY folder

ip_country/ci/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: Here too

ip_country/ci/integration_tests.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: here too

ip_country/ci/license.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -e
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: here too

ip_country/ci/lint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: here too

ip_country/ci/run_integration_tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: here too

ip_country/ci/unit_tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -xev
+# Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
```
bertllll, line 2: here too

ip_country/src/bit_queue.rs
```diff
@@ -21,6 +29,10 @@ impl BitQueue {
         (self.byte_queue.len() * 8) - self.back_blank_bit_count - self.front_blank_bit_count
     }
 
+    pub fn is_empty(&self) -> bool {
+        ((self.byte_queue.len() * 8) - self.back_blank_bit_count - self.front_blank_bit_count) == 0
```
bertllll, line 33: Look at the fn above this one. If you do need `is_empty()`, implement it as:

```
pub fn is_empty(&self)  -> bool  {
       self.len() == 0
}

File: ip_country/src/ip_country.rs [hunk 12]
```diff
@@ -273,4 +277,34 @@ Line 17: Invalid (AddrParseError(Ip)) IP address in CSV record: 'BOOGA'"#
...
+    fn write_error_test() {
+        let mut subject = BitQueue::new();
+        let output = &mut ByteArrayWriter::new();
+
+        subject.add_bits(0b11011, 5);
+        subject.add_bits(0b00111001110011100, 17);
+        subject.add_bits(0b1, 1);
+
+        output.reject_next_write(Error::new(ErrorKind::WriteZero, "Bad file Descriptor"));
+        let result = generate_country_data("ipv4_country_data", subject, output).unwrap_err();
```
bertllll, line 432: This is the "act" of the test.  It should stand insulated by the blank lines from both sides.

ip_country/src/ip_country.rs
```diff
@@ -273,4 +277,34 @@ Line 17: Invalid (AddrParseError(Ip)) IP address in CSV record: 'BOOGA'"#
...
+
+    #[test]
+    fn write_error_test() {
+        let mut subject = BitQueue::new();
+        let output = &mut ByteArrayWriter::new();
+
+        subject.add_bits(0b11011, 5);
+        subject.add_bits(0b00111001110011100, 17);
+        subject.add_bits(0b1, 1);
+
```
bertllll, line 430: Lines from 285 to 289 seem doing more than what we need. 

if there is the NEED of pushing some bits into the queue first, then you can probably do the same by just a single call. For example by concatenating all the bit literals together and reevaluating the correct total amount of bits handled. In this case 20 is the total. 

if there is NO NEED of this filling stage, then just delete all three lines. 

Plus, at any rate, the blank lines of the ln.285 and 289, don't belong in there.

ip_country/src/ip_country.rs
```diff
@@ -273,4 +277,34 @@ Line 17: Invalid (AddrParseError(Ip)) IP address in CSV record: 'BOOGA'"#
...
+        stdout.reject_next_write(Error::new(ErrorKind::WriteZero, "Bad file Descriptor"));
+
+        let result = ip_country(vec![], stdin, stdout, stderr);
+
+        assert_eq!(result, 1);
+        let stdout_string = String::from_utf8(stdout.get_bytes()).unwrap();
+        let stderr_string = String::from_utf8(stderr.get_bytes()).unwrap();
+        assert_eq!(stderr_string, "Error generating Rust code: Custom { kind: WriteZero, error: \"Bad file Descriptor\" }");
+        assert_eq!(stdout_string, "\n            *** DO NOT USE THIS CODE ***\n            It will produce incorrect results.\n            The process that generated it found these errors:\n\nError generating Rust code: Custom { kind: WriteZero, error: \"Bad file Descriptor\" }\n\n            Fix the errors and regenerate the code.\n            *** DO NOT USE THIS CODE ***\n");
+    }
```
bertllll, line 450: Now I'm confused. It seems to me that these two tests with green code test almost the same thing but only at a different level. 

If you agree, please remove one... probably the first placed.

ip_country/src/test_dbip_country.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub fn ipv6_country_data() -> (Vec<u64>, usize) {
+    (
+        vec![
+            0x3000040000400007, 0x00C0001400020000, 0xB400150000000700, 0xA004AF8B024C0E23,
+            0x0000000400BC2051,
+        ],
+        302
+    )
+}
```
bertllll, line 22: What the hack is this file for??

It probably should not be here.

If I'm looking at a product of some test of code generating a rust source file which necessarily needs to be performed, then the `ip_country` crate should have its own `generated` folder (like in Node) and it should be generated inside there.

Again, if you need it, you should make the extra effort and parameterize the tested code so it place the product, this file in such a directory which would be specified by an argument of the function. 

---

However, I read some more code in the meantime and it seems like this file is only some kind of static file that provides small set of data for various tests to prevent long runs on those. Eh. That's not clean. 
If this is true, you would do better deleting the confusing comment at the top and writing your own comment explaining this is not generated by machinery. This file should probably be marked with `![cfg(test)]` somewhere at the beginning. 

Does the ip_country crate already have a file or even a folder that maintains publicly accessible test code functions? If so, this could possibly be placed in there.

node/tests/ui_gateway_test.rs
```diff
@@ -166,7 +166,7 @@ fn daemon_does_not_allow_node_to_keep_his_client_alive_integration() {
     let assertion_lookup_pattern_2 =
         |_port_spec_ui: &str| "Received shutdown order from client 1".to_string();
     let second_port = connected_and_disconnected_assertion(2, assertion_lookup_pattern_2);
-    //TODO Create card to Change infrastructure and add node_client.send here and it should return result - we send the suhtdown request in loop until it not returns
+    //TODO Card #806 "create looping utility with node_client.send": Create a test utility with a node_client.send(shutdown_request) and keep sending the shutdown request in a loop until the Node doesn't response anymore.
```
bertllll, line 169: You can possibly leave just the name of the name of the card here, including the numeric marking `806`. The rest can be read when you find the card. 

Or better, something saying: "Test utility to easily verify the Node's termination"

ip_country/src/country_block_stream.rs
```diff
@@ -53,40 +53,39 @@ impl IpRange {
...
-                let (low_range, hi_range) = (u32::from(*low), u32::from(*hi));
-                let ip_num = u32::from(ip);
-                if ip_num < low_range {
-                    Ordering::Greater
-                } else if ip_num > hi_range {
-                    Ordering::Less
-                } else {
-                    Ordering::Equal
-                }
+                Self::compare_with_range::<u32, Ipv4Addr>(ip, *low, *hi)
```
bertllll, line 59: Are you sure it's clever to use this abbreviation `hi` for `high` everywhere? I don't think it makes much sense. I tried a test and asked the chat GPT which told me what I'd thought: it's not a generally known abbreviation and it actually advises against due to easy confusion. 

I'd be happy if you could use a regex search in the search&replace functionality of your IDE, probably also limiting the scope to only the Node/ip_country/.. location.

ip_country/src/country_block_stream.rs
```diff
@@ -147,27 +146,29 @@ impl CountryBlock {
         Ok(ip_addr)
     }
 
-    fn validate_ip_addresses(start_ip: IpAddr, end_ip: IpAddr) -> Result<(), String> {
+    fn validate_ips_are_sequential<SingleIntegerIPRep, IP>(start: IP, end: IP) -> Result<(), String>
+    where
+        SingleIntegerIPRep: From<IP> + PartialOrd,
+        IpAddr: Display,
```
bertllll, line 152: I might be wrong but... my feeling is you can delete this ln. 152, with no negative effect.

ip_country/src/country_block_stream.rs
```diff
@@ -408,4 +409,18 @@ mod tests {
...
+        expected = "Mismatch ip (1.2.3.4) and range (V6(::1:2:3:4, ::4:3:2:1)) versions"
+    )]
+    fn in_range_panics_on_v4_v6_missmatch() {
+        let subject = IpRange::V6(
+            Ipv6Addr::from_str("0:0:0:0:1:2:3:4").unwrap(),
+            Ipv6Addr::from_str("0:0:0:0:4:3:2:1").unwrap(),
+        );
+
+        let ip = Ipv4Addr::from_str("1.2.3.4").unwrap();
+        let _result = subject.ordering_by_range(IpAddr::V4(ip));
```
bertllll, line 424: Ideally, the arrangement should be:

```
fn  in_range_panics_on_v4_v6_missmatch()   {
        let subject = IpRange::V6(
            Ipv6Addr::from_str("0:0:0:0:1:2:3:4").unwrap(),
            Ipv6Addr::from_str("0:0:0:0:4:3:2:1").unwrap(),
        );
        let ip = Ipv4Addr::from_str("1.2.3.4").unwrap();

        let _result = subject.ordering_by_range(IpAddr::V4(ip));
}
```

The "act" that we decided to put always alone in these tests is the function call which contains the exercised logic, or in case of panic, we can say it's the line that panics.

ip_country/src/country_finder.rs
```diff
@@ -19,88 +28,40 @@ pub struct CountryCodeFinder {
...
-            Ok(index) => Some(country_finder[index].country.clone()),
-            _ => None,
-        };
-        let country = match country {
-            Some(country_inner) => match country_inner.iso3166.as_str() {
-                "ZZ" => None,
-                _ => Some(country_inner),
-            },
-            None => None,
+        let country_finder: &Vec<CountryBlock> = match ip_addr {
```
bertllll, line 44: I actually think we should recast this `&Vec<CountryBlock>` into `&[CountryBlock]`. It makes only little difference but it's an idiom in Rust to work with "slices" rather than "references to a vector". 

Of course, you will need to change the static typing also at some other places deeper along the call but it's not gonna be a lot.

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
 
     #[test]
-    fn real_test_ipv4_with_sk_isp() {
-        let timestart = SystemTime::now();
+    fn real_test_ipv4_with_sk_ip() {
+        let _ = CountryCodeFinder::find_country(
+            &COUNTRY_CODE_FINDER,
+            IpAddr::from_str("213.81.185.100").unwrap(), // dig www.zoznam.sk A
+        )
+        .unwrap();
```
bertllll, line 199: Why do you invoke the function twice? While you also ignore the first result. I think You don't want lines from 195 to 199.

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
+        let time_start = SystemTime::now();
         let result = CountryCodeFinder::find_country(
             &COUNTRY_CODE_FINDER,
             IpAddr::from_str("213.81.185.100").unwrap(), // dig www.zoznam.sk A
         )
         .unwrap();
-        let timeend = SystemTime::now();
+        let time_end = SystemTime::now();
```
bertllll, line 206: Also, the arrangement should be:

```
.....fn name......
        let time_start = SystemTime::now();

        let result = CountryCodeFinder::find_country(
            &COUNTRY_CODE_FINDER,
            IpAddr::from_str("213.81.185.100").unwrap(), // dig www.zoznam.sk A
        )
        .unwrap();

        let time_end = SystemTime::now();
        assert_eq!(result.free_world, true);
        ...
```

Notice position of the blank lines

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
         assert_eq!(result.name, "Slovakia".to_string());
-        let duration = timeend.duration_since(timestart).unwrap().as_secs();
-        println!("duration: {}", duration);
-        assert!(duration <= 5);
+        let duration = time_end.duration_since(time_start).unwrap();
+        assert!(
+            duration.as_secs() < 1,
+            "Duration of the search was too long: {} ms",
+            duration.as_millis()
+        );
```
bertllll, line 216: What sense does this assertion make if we test above a test dataset (it contains only a few entries!). 

Did Dan advised this to you? Well, I'd like you to discuss again while the answer is either way. This sounds like testing unreal scenario and therefore unnecessary. You must decide if we want to test on the original dataset or not at all.

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
         assert_eq!(result.name, "United States".to_string());
-        let duration = timeend.duration_since(timestart).unwrap().as_secs();
-        println!("duration: {}", duration);
-        assert!(duration <= 5);
+        let duration = time_end.duration_since(time_start).unwrap();
+        assert!(
+            duration.as_secs() < 1,
+            "Duration of the search was too long: {} ms",
+            duration.as_millis()
+        );
```
bertllll, line 242: This test suffers from those three same issues as the previous one.

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
     }
 
     #[test]
     fn deserialize_country_blocks_ipv4_and_ipv6_adn_fill_vecs() {
```
bertllll, line 246: `fn country_blocks_for_ipv4_and_ipv6_are_deserialized_filled_into_vecs() {`

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
+        assert!(
+            duration_deserialize.as_secs() < 15,
+            "Duration of the deserialization was too long: {} ms",
+            duration_deserialize.as_millis()
+        );
+        assert!(
+            duration_fill.as_secs() < 8,
+            "Duration of the filling the vectors was too long: {} ms",
+            duration_fill.as_millis()
+        );
```
bertllll, line 272: Duration 15 and 8 seconds?? Isn't it using only the very short dataset?

ip_country/src/test_dbip_country.rs
```diff
@@ -0,0 +1,22 @@
+
+// GENERATED CODE: REGENERATE, DO NOT MODIFY!
```
bertllll, line 2: Is this really also automatically generated file? Doesn't the comment lie?

ip_country/src/country_finder.rs
bertllll, line 246: This whole test should be performed probably in short by simply testing the constructor for CountryCodeFinder, because it's a function that contains exactly the same logic which you're  strangely testing here separately. Seems clumsy to me.

ip_country/src/country_finder.rs
```diff
@@ -225,77 +191,84 @@ mod tests {
...
-        let duration = timeend.duration_since(timestart).unwrap().as_secs();
-        let duration_fill = timeend_fill
-            .duration_since(timestart_fill)
-            .unwrap()
-            .as_secs();
-        println!("duration deserialize: {}", duration);
-        assert!(duration < 8);
-        println!("duration fill_vec: {}", duration_fill);
-        assert!(duration_fill < 8);
+        let duration_deserialize = time_end.duration_since(time_start).unwrap();
```
bertllll, line 261: I've got big reservations for the assertion part of this test. Why does it only test times? Why don't you check that the produced vectors take in the expected values, or at least in the expected count (plus maybe checking that each item is different)??

ip_country/src/country_finder.rs
```diff
@@ -110,7 +71,7 @@ mod tests {
             CountryCodeFinder::new(ipv4_country_data(), ipv6_country_data());
     }
 
-    pub(crate) fn ipv4_country_data() -> (Vec<u64>, usize) {
+    pub fn ipv4_country_data() -> (Vec<u64>, usize) {
```
bertllll, line 74: Are you sure about the second returned value in the tuple? Is it or it isn't important that it corresponds to the count of the vector's items? (I'm aware it's test data but still)

ip_country/src/country_finder.rs
```diff
@@ -123,7 +84,7 @@ mod tests {
         )
     }
 
-    pub(crate) fn ipv6_country_data() -> (Vec<u64>, usize) {
+    pub fn ipv6_country_data() -> (Vec<u64>, usize) {
```
bertllll, line 87: Ehhhh. 

There are two issues here.

First, you should reconsider two distinct places I'm driving at and make a choice of keeping just one. Either go with these two functions `ipv4_country_data`, `ipv6_country_data` defined here, make use of the file `ip_country/test_dbip_country`. We're discussing here a pair of interchangeable approaches. In other words, you probably shouldn't define a new static COUNTRY_CODE_FINDER_TEST if you already have COUNTRY_CODE_FINDER compiled with the #[cfg(test)] attribute and providing qualitatively the same values as the other one. Given you're already using the trick with overriding via conditional compilation for which you have two attributes that replace each other based on the `cargo` command we're running (whether with or without `test`) and which you can find at the beginning of this file, I might go with that one more happily and extra few lines above, extended by _TEST at the end of its name. However, you might not be able to, as I will elaborate in the next paragraph. 

Second, I wish you were eager to discuss the time measurement for which you apparently wrote the tests located below. The question you should have an answer for is how important is that in our matter to check the real times of the look-ups. Whether and how we want to keep a track of how much it can take the Node to finish its bootstrap.  If a partial answer to it is yes, we do need such a test, you CAN'T use the aforementioned technique with the `cfg` attributes because in order to have realistic times you'd have to be testing it with the complete set of data which we can expect the Node would operate with, not just a test sentinel containing about one six-thousandth of it. You can't gen any relevant opinion on what you are after over such data!

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1625,6 +1625,38 @@ mod tests {
...
+    #[should_panic(expected = "Writing max_block_count failed: QueryFailed")]
+    fn handle_retrieve_transactions_panics_if_writing_max_block_count_failed() {
+        let retrieve_transactions_params_arc = Arc::new(Mutex::new(vec![]));
+        let earning_wallet = make_wallet("somewallet");
+        let latest_block_number = LatestBlockNumber::Ok(1024u64.into());
+        let lower_interface =
+            LowBlockchainIntMock::default().get_block_number_result(latest_block_number);
+        let blockchain_interface =
+            BlockchainInterfaceMock::default()
+                .retrieve_transactions_params(&retrieve_transactions_params_arc)
```
bertllll, line 1638: This line and line 1631 is not needed because you don't do any assertions except waiting for the panic.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1625,6 +1625,38 @@ mod tests {
         let _ = subject.handle_retrieve_transactions(retrieve_transactions);
     }
 
+    #[test]
+    #[should_panic(expected = "Writing max_block_count failed: QueryFailed")]
```
bertllll, line 1629: You need to fix the code. There is a mistake now. You are printing the error from the RPC of the blockchain_interface. Instead, you should speak of the error that occurred during writing the persistent database. That error prepared on line 1644.

Go to the production code and fix it. But you need to take care of this panic message too...because it cannot stay like this.

node/src/neighborhood/mod.rs
```diff
@@ -545,13 +545,16 @@ impl Neighborhood {
...
             .metadata
-            .node_location_opt = node_location.clone();
-        self.neighborhood_database.root_mut().inner.country_code =
-            node_location.expect("expected Node location").country_code;
+            .node_location_opt = node_location_opt.clone();
+        self.neighborhood_database.root_mut().inner.country_code_opt = Some(
+            node_location_opt
+                .expect("expected Node location")
+                .country_code,
+        );
```
bertllll, line 557: You can rewrite the last operation as `self.neighborhood_database.root_mut().inner.country_code_opt = node_location_opt.map(|nl|nl.country_code)` by using map on Option

node/src/neighborhood/mod.rs
```diff
@@ -4351,15 +4350,18 @@ mod tests {
...
+            Some(
+                subject
+                    .neighborhood_database
+                    .root()
+                    .metadata
+                    .node_location_opt
+                    .as_ref()
+                    .unwrap()
+                    .country_code
+                    .clone()
```
bertllll, line 4363: Not sure why the `.clone()` is there... Can you try removing it?

Also, it's worth some thought whether CountryCode should be `Copy`, allowing copies without explicit calls written by the programmer. If yes, then I'd recommend to search for its occurrences (at least roughly) and take the calls of `.clone()` out.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -51,18 +51,19 @@ impl NeighborhoodDatabase {
...
-            get_node_location(Some(node_addr.ip_addr))
-        } else {
-            None
-        };
+        let location_opt = get_node_location(Some(
+            neighborhood_mode
+                .node_addr_opt()
+                .unwrap_or_default()
+                .ip_addr(),
+        ));
```
bertllll, line 59: This green pile of code ISN'T semantically the same as the pink one from before. Can you explain why? Like why should it be preserved this way? 

The difference in my eyes lies in the ".unwrap_or_default()" call. You're stating that at case of the node addr being None, you will initialize some default (whatever it is for this type) and then you will return the ip addr of this default. I'm concerned because I can imagine it could return an unexpected value. It has me thinking: Is that code-flow path with None actually tested anywhere??

node/src/test_utils/neighborhood_test_utils.rs
```diff
@@ -36,11 +36,11 @@ lazy_static! {
         ),
         (
             IpAddr::V4(Ipv4Addr::new(99, 99, 99, 99)),
-            "AU".to_string(),
+            "FR".to_string(),
             true
         ),
         (
-            IpAddr::V4(Ipv4Addr::new(180, 223, 223, 123)),
+            IpAddr::V4(Ipv4Addr::new(3, 3, 3, 3)),
```
bertllll, line 43: Why did you modify this to quad three? I don't like that already three examples in this collection have all four segments making the IpAddr identical. Which is far from "randomly provided test data".

node/src/neighborhood/node_record.rs
```diff
@@ -77,13 +77,10 @@ impl NodeRecord {
...
         node_record_inputs: NodeRecordInputs,
     ) -> NodeRecord {
-        let mut country = String::default();
-        match node_record_inputs.location.as_ref() {
-            Some(node_location) => {
-                country = node_location.country_code.clone();
-            }
-            None => {}
-        };
+        let country = node_record_inputs
```
bertllll, line 80: `country_opt` would be more appropriate...

node/src/neighborhood/node_location.rs
```diff
@@ -5,10 +5,10 @@ use ip_country_lib::country_finder::{CountryCodeFinder, COUNTRY_CODE_FINDER};
 use std::net::IpAddr;
 
 #[allow(dead_code)]
```
bertllll, line 7: Really, is this lint necessary? I think our code must have multiple references to this structure and therefore it is "alive".

node/src/neighborhood/node_location.rs
```diff
@@ -5,10 +5,10 @@ use ip_country_lib::country_finder::{CountryCodeFinder, COUNTRY_CODE_FINDER};
...
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Eq)]
```
bertllll, line 8: Even the PartialEq trait could be implemented via a `derive` macro. It should probably compare both fields inside it, and so the impl at line 14 appears even wrong.

node/src/neighborhood/node_location.rs
```diff
@@ -17,19 +17,14 @@ impl PartialEq<Self> for NodeLocation {
...
-            match country {
-                Some(country) => Some(NodeLocation {
-                    country_code: country.iso3166.to_string(),
-                    free_world_bit: country.free_world,
-                }),
-                None => None,
-            }
+            let country_opt = CountryCodeFinder::find_country(&COUNTRY_CODE_FINDER, ip_addr);
+            country_opt.map(|country| NodeLocation {
+                country_code: country.iso3166.to_string(),
```
bertllll, line 25: This should probably be `.clone()` instead of `.to_string()`? I assume that the value is already String. It isn't a change of function but it would be semantically more correct, reflecting the very operation.

node/src/neighborhood/node_location.rs
```diff
@@ -17,19 +17,14 @@ impl PartialEq<Self> for NodeLocation {
...
+                free_world_bit: country.free_world,
```
bertllll, line 26: I had to think a while about whether I should comment on whether there is any distinct difference between the first structure we are taking the values from, I guess it's `CountryCode`, and the `NodeLocation`. It suddenly looks here as if they held exactly the same information. 

I still think the two structures are better separate though.

Still, you can consider an approach like naming the variables `node_location` but using only the data type of `CountryCode`. I don't know if it would be worth pursuing further. Probably not.

node/src/neighborhood/node_location.rs
```diff
@@ -66,28 +60,27 @@ mod tests {
...
-        let mut node_record = make_node_record(1111, true);
-        let country_record = pick_country_code_record(3333);
-        node_record.metadata.node_location_opt = Some(NodeLocation {
-            country_code: country_record.1.clone(),
-            free_world_bit: country_record.2,
-        });
-        node_record.metadata.node_addr_opt =
-            Some(NodeAddr::new(&country_record.0, &[8000 % 10000]));
-        node_record.inner.country_code = country_record.1;
+        let node_record = make_node_record(3333, true);
```
bertllll, line 67: I didn't pay enough attention to this last time, obviously, but now I'm quite sure this is an unwanted test whose application seems rather awkward. 

Basically, we don't want to write tests for test utils UNLESS it has some complex logic under the hood or it needs to be extremely reliable and we aren't sure it is. As for this scenario, I don't see any of these two concerns in place so I'd like you to delete the whole test. 

Also, next time, if you're writing test for code that is not actually defined in that file, you should stop and find it a better "home". If necessary, you CAN write a test even in files like `test_utils`, but it's very rare to have to do it. Again, it's what we have here.

node/src/neighborhood/node_location.rs
```diff
@@ -66,28 +60,27 @@ mod tests {
                 free_world_bit: true
             }
         );
-        assert_eq!(
```
bertllll, line 69: I was more vigilant today perhaps and that means I noticed this test has no real value. 

The only NEW functionality you could be planning to test, and I could imagine, is the `get_node_location()` call. 
There is no reason for ASSIGNING the result of the previously mentioned function to a field of the default for NodeRecordMetadata and test if it contains it afterwards. It is as if you were trying to test if the assignment operation in Rust works. Of course it does. 

Conclusion: we don't want to keep this test.

node/src/neighborhood/node_location.rs
```diff
@@ -98,37 +91,25 @@ mod tests {
     }
 
     #[test]
-    fn create_gossip_node_with_addr_and_country_code_reveal_results_in_node_with_addr_adn_free_world_bit(
+    fn node_record_from_gossip_with_addr_and_country_is_populated_with_right_addr_and_free_world_bit(
```
bertllll, line 94: Sorry that I cannot comment on that directly, but I found out there is unneeded `        expected_node_record.resign();` in the test called `can_create_a_node_record_from_a_reference`. Please remove that line. The sign doesn't change from change in the metadata only.

bertllll, line 94: More importantly, I again think that we're dealing with a redundant test here... 
This test proves exactly the same thing as the one mentioned in the previous comment. It makes no sense to me to test `NodeRecord::try_from(<gossip>)` in this file. It's use of much different code than the simple code this file holds. 

Therefore, believe me or not, I'd like to delete even this one.

node/src/neighborhood/node_location.rs
```diff
@@ -98,37 +91,25 @@ mod tests {
...
-            gossip_result.node_addr_opt.unwrap(),
-            node.node_addr_opt().unwrap()
-        );
-        assert_eq!(node_record.inner.country_code, "US")
+        original_node_record.metadata.last_update = result_node_record.last_updated();
+        assert_eq!(result_node_record, original_node_record)
     }
 }
 
 #[allow(dead_code, unused_imports)]
```
bertllll, line 112: If we do want to preserve this module with the benchmarks, then you should probably replace this attribute above simply with the standard #[cfg(test)].

However, I realized that if we do that it will never SEE the fully populated country code database with all the 60K-lines data. 

Honestly, this is where we come to the need for a critical decision what would be right to do here. I do require to discuss this together with Dan.

The concern is that the benchmark makes an impression of a legit measurement but is fake. It seems like we ARE interested in the times of look-ups in the database. However, we test it on DIFFERENT data set, almost anecdotally   extremely restricted, that of course gives back lovely tiny measures of almost no time. But by these wanna-be tests we are laying ourselves uncovered.

If we decide collectively that we want to keep some kind of benchmarks in we should design it more carefully as it may not apply to a unit test anymore given the high expected time lapses. It may be quite a challenge. We'll see. **Dan's oversight required though**.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -350,6 +341,28 @@ impl DebutHandler {
...
+        root_node
+            .add_half_neighbor_key(debut_node_key.clone())
+            .unwrap_or_else(|e| {
+                panic!(
+                    "Couldn't add debuting {} as a half neighbor: {:?}",
+                    debut_node_key, e
+                )
+            });
+        let gnr =
+            GossipNodeRecord::from((root_node.inner.clone(), root_node.node_addr_opt(), cryptde));
```
bertllll, line 360: If you create a variable `node_addr_opt` first, you can write:

```
        let root_node_addr_opt = root_node.node_addr_opt();
        let gnr =
            GossipNodeRecord::from((root_node.inner, root_node_addr_opt, cryptde));
```

Notice that I spared the single .clone() call. ;-) Maybe not much, but if you think like this, you can optimize using just a little change. It's another way how you get better code, and all together it makes a big difference in overall.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1655,18 +1667,14 @@ mod tests {
     }
 
     #[test]
-    fn two_debut_in_same_time_with_empty_db_handled_by_try_accept_debut() {
+    fn two_parallel_progressing_debuts_handled_by_try_accept_debut_without_introduction() {
```
bertllll, line 1670: Please, this reads slightly better 

two_parallel_debuts_in_progress_handled_by_try_accept_debut_without_introduction

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -1679,43 +1687,28 @@ mod tests {
...
-        //     debuted_node_five.public_key(),
-        // );
-        // let agr_man = &AccessibleGossipRecord::from(&new_debutant);
-        // let more_apropriate_neighbor =
-        //     DebutHandler::find_more_appropriate_neighbor(&subject, &dest_db, agr_man);
+            counter_debut,
+            GossipAcceptanceResult::Reply(
+                debut_reply.clone(),
+                dest_public_key.clone(),
+                dest_node_addr.clone()
```
bertllll, line 1709: All these three .clone() calls above have no sound implication. They should be removed.

ip_country/src/country_block_serde.rs
```diff
@@ -118,34 +126,38 @@ CountryBlock {
  */
 
 pub struct CountryBlockSerializer {
-    prev_start_ipv4: Ipv4Addr,
-    prev_end_ipv4: Ipv4Addr,
-    bit_queue_ipv4: BitQueue,
-    prev_start_ipv6: Ipv6Addr,
-    prev_end_ipv6: Ipv6Addr,
-    bit_queue_ipv6: BitQueue,
+    ipv4: VersionedIPSerializer<Ipv4Addr, u8, 4>,
```
bertllll, line 129: Now I got an interesting idea with aliasing. If you define two types you could get rid of specifying the argument types all the time.

```
type Ipv4Serializer = VersionedIPSerializer<Ipv4Addr, u8, 4>;
type Ipv6Serializer = VersionedIPSerializer<Ipv6Addr, u16, 8>;
```
Now you can search through the code, find occurrences of VersionedIPSerializer<x,x,x> and replace it by one of those two aliases. 
Until now, what was brought in on the advice to use my refactoring techniques didn't look much aesthetic, and I did see it but was going to ignore this "side effect". However, if you give it a nice facade by using these two aliases, it'll be so much better.

ip_country/src/country_block_serde.rs
```diff
@@ -154,133 +166,321 @@ impl CountryBlockSerializer {
...
     }
 }
 
-pub trait CountryBlockDeserializer {
-    fn next(&mut self) -> Option<CountryBlock>;
+// Rust forces public visibility on traits that come to be used for type boundaries in any public
+// interface (as in structs). This is how we can meet these requirements while those implementations
+// of these traits will be ineffective from farther than this file. It works for prevention of
+// namespace pollution in this kind of trait to be implemented on massively common types such as
+// Ipv4Addr or Ipv6Addr
```
bertllll, line 290: Please replace the whole comment by this one which I gave some corrections.  

// Rust forces public visibility on traits that come to be used as type boundaries in any public
// interface. This is how we can meet the requirements while the implementations of such 
// traits becomes ineffective from farther than this file. It works as a form of prevention to
// namespace pollution for such kind of trait to be implemented on massively common types,
// here namely Ipv4Addr or Ipv6Addr

ip_country/src/country_block_serde.rs
```diff
@@ -154,133 +166,321 @@ impl CountryBlockSerializer {
...
-                        self.prev_record.start,
-                        Ipv4Addr::new(255, 255, 255, 255),
-                    ),
-                    country: Country::from(self.prev_record.country_idx),
-                })
-            }
-        }
+        self.next_generic()
+    }
+}
```
bertllll, line 409: I like how it fits nicely the Iterator impl. Looks cool. :)

ip_country/src/country_block_serde.rs
```diff
@@ -154,133 +166,321 @@ impl CountryBlockSerializer {
...
+
+impl CountryBlockDeserializer<Ipv6Addr, u16, 8> {
+    pub(crate) fn new(country_data: (Vec<u64>, usize)) -> Self {
```
bertllll, line 412: Just detail, but `(crate)` is redundant. You know why I always make fuss about this? :D Because I learnt a long time ago, that this is the default, not sure when it could ever be useful. I understand only things like `(super)`, `(self::super)`, `(in crate::high_module_name)`, but not `(crate)` alone, because it literally means in every place in this Rust crate. 

But I later noticed there are more places in this file saying `(crate)`. Could you please look them up and delete the extra wording? Thanks

ip_country/src/country_block_serde.rs
```diff
@@ -154,133 +166,321 @@ impl CountryBlockSerializer {
...
     empty: bool,
 }
 
-impl CountryBlockDeserializer for CountryBlockDeserializerIpv4 {
+pub trait DeserializerPublic {
+    fn new(country_data: (Vec<u64>, usize)) -> Self;
+    fn next(&mut self) -> Option<CountryBlock>;
+}
+
+impl CountryBlockDeserializer<Ipv4Addr, u8, 4> {
```
bertllll, line 398: Please use the trick with type aliases also here, getting two no-spec-types-requiring entities in return.
With names like:
`Ipv4CountryBlockDeserializer` and `Ipv6CountryBlockDeserializer` 


Caution: replace only the occurrences with the already specific usage (it shows the types), don't in any declaration of a generic function where these types are still treated as unspecific.

ip_country/src/country_block_serde.rs
```diff
@@ -154,133 +166,321 @@ impl CountryBlockSerializer {
...
+impl<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize>
+    CountryBlockDeserializer<IPType, SegmentNumRep, SEGMENTS_COUNT>
+where
+    Self: DeserializerPrivate<IPType>,
+    IPType: IPIntoSegments<SegmentNumRep, SEGMENTS_COUNT>
+        + PlusMinusOneIP
+        + From<[SegmentNumRep; SEGMENTS_COUNT]>
+        + Copy
+        + Debug
+        + Display,
```
bertllll, line 469: I'm a bit surprised we need both Debug and Display to print an IP. I don't know if you really wanna do that, it may or may not be a hassle, but I'd try to remove the Display and also at other places in the file. I couldn't find where we would've needed it.

ip_country/src/country_block_serde.rs
```diff
@@ -341,119 +528,94 @@ impl CountryBlockDeserializer for CountryBlockDeserializerIpv6 {
...
-    let old_data = u32_from_ipv4(ip_addr);
-    let new_data = old_data.overflowing_sub(1).0;
-    Ipv4Addr::from(new_data)
+impl<IPType: Debug, SegmentNumRep, const SEGMENTS_COUNT: usize> Debug
+    for VersionedIP<IPType, SegmentNumRep, SEGMENTS_COUNT>
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("VersionedIP {{ ip: {:?} }}", self.ip))
+    }
 }
```
bertllll, line 578: I just realized there would be a simpler alternative to this: 

```
#[derive(Debug)]
struct VersionedIP<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize> 
where 
   IPType: Debug,
   SegmentNumRep: Debug 
{
    ip: IPType,
    segment_num_rep: PhantomData<SegmentNumRep>,
}

```

Explanation: the derive macro can be used because we added that our generics implements Debug and although I had doubted it turned out PhantomData already implements Debug. So it should work.

I really like this better, I'm sorry if it's gonna cause you troubles in your tests. You may want to delete one for your custom Debug impl.

ip_country/src/country_block_serde.rs
```diff
@@ -341,119 +528,94 @@ impl CountryBlockDeserializer for CountryBlockDeserializerIpv6 {
...
-fn u32_from_ipv4(ip_addr: Ipv4Addr) -> u32 {
-    let octets = ip_addr.octets();
-    let mut bit_data = 0u32;
-    octets.into_iter().for_each(|octet| {
-        bit_data <<= 8;
-        bit_data |= octet as u32;
-    });
-    bit_data
+#[derive(Debug)]
+struct StreamRecord<IPType: Display, SegmentNumRep, const SEGMENTS_COUNT: usize> {
```
bertllll, line 581: I think this Display after IPType is redundant.

ip_country/src/country_block_serde.rs
```diff
@@ -341,119 +528,94 @@ impl CountryBlockDeserializer for CountryBlockDeserializerIpv6 {
...
 }
 
-fn plus_one_ipv6(ip_addr: Ipv6Addr) -> Ipv6Addr {
-    let old_data = u128_from_ipv6(ip_addr);
-    let new_data = old_data.overflowing_add(1).0;
-    Ipv6Addr::from(new_data)
+impl<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize>
+    StreamRecord<IPType, SegmentNumRep, SEGMENTS_COUNT>
+where
+    IPType: From<[SegmentNumRep; SEGMENTS_COUNT]> + Display,
```
bertllll, line 589: I think so also here

ip_country/src/country_block_serde.rs
```diff
@@ -341,119 +528,94 @@ impl CountryBlockDeserializer for CountryBlockDeserializerIpv6 {
...
+    <SegmentNumRep>::Error: Debug,
+{
+    fn new(
+        differences: Vec<Difference>,
+        mut segments: [SegmentNumRep; SEGMENTS_COUNT],
+        country_idx: usize,
+    ) -> StreamRecord<IPType, SegmentNumRep, SEGMENTS_COUNT> {
+        differences.into_iter().for_each(|d| {
+            segments[d.index] = SegmentNumRep::try_from(d.value)
+                .expect("Difference noted in a bigger number than which the IP segment can contain")
```
bertllll, line 600: Please modify the message as follows 

Difference noted in a bigger number than which  ---> Difference represented by a bigger number than which

ip_country/src/ip_country.rs
```diff
@@ -116,9 +119,151 @@ fn write_value(
...
+            FakeStreamHolder {
+                stdin: ByteArrayReader::new(&[0; 0]),
+                stdout: ByteArrayWriter::new(),
+                stderr: ByteArrayWriter::new(),
+            }
+        }
+    }
+
+    impl FakeStreamHolder {}
+
```
bertllll, line 265: Everything added from here up to the beginning of the test module is wrong. Let's go with what I suggested in the parallel branch, when you pick up on my commit ending that branch this all will have been take care of. We do need to prevent this hacky way of fixes. That's all right, I know, you didn't mean it badly. I'm glad though we can go on without duplicating the code.

ip_country/src/ip_country.rs
```diff
@@ -75,22 +73,27 @@ fn generate_country_data(
     output: &mut dyn io::Write,
 ) -> Result<(), io::Error> {
     let bit_queue_len = bit_queue.len();
-    write!(output, "\n")?;
```
bertllll, line 78: I cannot comment on this directly but line 64 contains a TODO with a suggestion tightly coupled with the card being reviewed which normally is not an acceptable state. Because it's not clear either to me what Dan talks about in this comment I advise you to consult together the future of the referred idea.

ip_country/src/country_finder.rs
czarte: this is because we want to isolate time for selecting from already constructed array. so we want to avoid measure duration of construction of vector of `CountryBlocks`

ip_country/src/country_finder.rs
czarte: Those are tests with real data.

ip_country/src/country_finder.rs
czarte: same as above

ip_country/src/country_finder.rs
czarte: No it uses real data

ip_country/src/test_dbip_country.rs
czarte: Comment does not lie. This is manually generated by ip_country binary, to establish set of ips for tests.

ip_country/src/country_finder.rs
czarte: but in that way we will be unable to track those parts of the deserialization separately, as was Dan's proposal.

ip_country/src/country_finder.rs
czarte: It was ment to be performance test. If you insist, I can add more assertions.

ip_country/src/country_finder.rs
czarte: This is probably question for Dan. I did not created or reviewd this piece of code. And I assume, that nuber is right. But I can add test for this. It is really simple.

ip_country/src/country_finder.rs
czarte: I do not understand this comment. We can discuss this on the call.

node/src/blockchain/blockchain_bridge.rs
czarte: any usggestions?

czarte: with this I should probably need some help. We were fixing this test with Dan, and the result is this. 

If I change the `.retrieve_transactions_result(Err(BlockchainError::QueryFailed` (line1639) with an RPC error 
to `Ok(<RetrievedBlockchainTransactions { new_start_block: 0, transactions: vec![] }` There is error 'Accountant not bound' and basically I wrote the test fot the panic on `handle_retrieve_transactions in blockchain_bridge.rs::361` where we handling `BlockchainError`

node/src/neighborhood/mod.rs
czarte: Yes I did, there is error: `move occurs because value has type `std::string::String`, which does not implement the `Copy` trait`

czarte: when I add `Copy` to derive on NodeLocation, it complains Trait `Copy` is not implemented for `String`, should I implement it?

node/src/neighborhood/neighborhood_database.rs
czarte: I was confident, there is couple of tests testing None value on `node_location_opt` in inner of `NodeRecord`. For be safe I created new test `can_create_node_record_with_node_location_opt_none()`

node/src/test_utils/neighborhood_test_utils.rs
czarte: We are not testing the randomly provided data, but I needed the FR country code, for specific test, to match output from the real data set.

node/src/neighborhood/node_location.rs
czarte: if it compares both members, wouldn't it be Eq, instead of PartialEq?

node/src/neighborhood/node_location.rs
czarte: probably some residuum

node/src/neighborhood/node_location.rs
czarte: Im not quite sure, what do you mean by this comment. Can we disucss it on the call?

node/src/neighborhood/node_location.rs
czarte: So you mean we do no t want to test `try_from` for NodeRecord? Moreover this test is actually testing the persistance of `country_code` in Gossip. Do you really wan to delete it?

node/src/neighborhood/gossip_acceptor.rs
czarte: this is lack of `Copy` trait for `Gossip_0v1`, which leads down to the rabbit hole for all members of `GossipNodeRecord`, do we really want to implement the `Copy` trait so much in depth just for this one test?

ip_country/src/country_block_serde.rs
czarte: sure, It is always added by IDE, I will pay more attention on this in future :)

ip_country/src/country_block_serde.rs
czarte: display is needed for `fn get_record_generic()` .. tried to remove but then was unsatified Trait bound occurring

czarte: impl Debug solves this, I could remove that

ip_country/src/country_finder.rs
bertllll: Well, I'm quite sure the number tells how many blocks are there so I'm confused why we are obviously supplying a wrong one, as it would amount to 5 correctly. 

There isn't a point in adding a test. This is a piece of test code and simple one, so it doesn't require testing. We just should push ourselves to provide sensible values if possible. 

By the way, this one and the next function, they don't need to be public. Things within the `tests` module should never be public and it rather points out bad arrangement of something if it needs to be.

ip_country/src/country_finder.rs
bertllll: Okay, I've edited the comment so it makes more sense. You were right it wasn't particularly clear what I meant.

ip_country/src/country_finder.rs
bertllll: While I can see some sense in what you're saying, there would probably be a neater and more centralized way of making sure that the data set assigned to the static variable has been already initialized by that time when we are to let it go on performing an actual test which is sensitive to this. You probably could count up more tests out here that will need to be treated carefully in this regard. Therefore I'm thinking of how we may improve it without the clumsy duplication.  

My proposal consists of maintaining two different country blocks databases during and purely for the test period. I want to stress the need to take such actions or similar because otherwise we'd never be able to assess what are the time parameters of the use of the fully-fledged database. Until now all the tests that are performed are stubbed by a dataset that is ridiculously small. However it is perfectly reasonable for some tests it is clearly unacceptable for the other as those placed in this file which attempts to assert on reasonable look-up times. For those, despite there might be just a few of them, a realistic dataset must be kept on hand. As long as we properly label such code with test-only compilation directives, it will serve only the purposes of the tests and will not pollute the production code.     

```
// This piece of code will remain the same as it is the prod code and we need to avoid this
// initialization in all those places many tests that use the call for finding the Node's location.

#[cfg(not(test))]
lazy_static! {
    pub static ref COUNTRY_CODE_FINDER: CountryCodeFinder = {
        let finder = CountryCodeFinder::new(
            crate::dbip_country::ipv4_country_data(),
            crate::dbip_country::ipv6_country_data()
        )
    };
}


// Here, there is a difference. We will have two data sets for two different kind of tests.
// The more frequent of them will use the extremely shrunken data set from the file `test_dbip_country.rs`.
// Then there a few where, as is my understanding, you want to realistically look into the times for the ip 
// look-ups of the about 30 thousand of records for ipv4 or ipv6. There, it is crucial to manipulate 
// data that is same as in the reality. That's why we get the other variable
// TEST_REALISTIC_COUNTRY_CODE_FINDER above which the second kind of those tests should be
// performed.  

#[cfg(test)]
lazy_static! {
    pub static ref COUNTRY_CODE_FINDER: CountryCodeFinder = {
        let finder = CountryCodeFinder::new(
            crate::test_dbip_country::ipv4_country_data(),
            crate::test_dbip_country::ipv6_country_data()
        )};

    pub static ref REALISTIC_COUNTRY_CODE_FINDER_FOR_BENCHMARK: CountryCodeFinder = {
        let finder = CountryCodeFinder::new(
            crate::dbip_country::ipv4_country_data(),
            crate::dbip_country::ipv6_country_data()
        )};
}

#[cfg(test)]
pub mod c_c_f_initialization {
    pub const BENCHMARK_COUNTRY_CODE_FINDER_INITIALIZED: CCFInitializationIndication =   CCFInitializationIndication::default();
    
    #[derive(Default)]
    pub struct BCCFInitializationIndication{
         initialized: Mutex<bool>
    }
    
    impl BCCFInitializationIndication {
         fn check_initialized(&self){
             let is_initialized = self.initialized.lock().expect("mutex was poisoned");
             if !is_initialized {
                  REALISTIC_COUNTRY_CODE_FINDER_FOR_BENCHMARK.claim_init_access();
                  is_initialized.replace(true);
             }
         }
    }
}

// Probably in a different file, there could be this method as part of the impl block of CountryCodeFinder:

impl CountryCodeFinder {
     fn claim_init_access(&self){
         // Intentionally empty. The only purpose is to provoke
         // an initialization of this static variable. 
     } 
} 

// In the tests where you need the real data initialized, all you need to do is:

#[test]
fn some_test(){
    // Will either initialize the data or will wait until other test initialize it instead or the other test is done
    // with its work over this data.  
    COUNTRY_CODE_FINDER_INITIALIZED.check_initialized()
} 
 
```

ip_country/src/country_finder.rs
bertllll: Nope. I don't think so. If the comprehend comment from above here is not enough to explain, we can talk about it live. I wouldn't be opposed to consult it with Dan as an advisor.

ip_country/src/country_finder.rs
bertllll: I think you're not right.

.github/workflows/ci-matrix.yml
```diff
@@ -51,6 +57,11 @@ jobs:
         with:
           name: Node-${{ matrix.target.name }}
           path: results
+      - name: diagnostics
+        if: failure()
+        run: |
+          echo "final disc diagnostics ------>"
+          df -h ~
```
bertllll: What are you planning to do with these new lines? Do you think they should stay in or they rather played their role only in debugging that is already over now?

ip_country/src/country_finder.rs
bertllll: You don't test them separately anyway so what are you complaining about? 

If you want to test them separately to be able to put forth how much time is needed for IpV4 and how much for IpV6, do as you please, it's also possible. I was just commenting on that your test represented exactly what is inside the constructor. If you'd shown an intention to test them separately I'd probably not have suggested such a change because it would've lacked sense. True.

ip_country/src/country_finder.rs
bertllll: Yes, I thought it would be a nice enhancement as you haven't done it in any of those tests too. 

Also, if you want to put the stress on testing the performance it should somehow be seen in the name of the test. The name is quite general.

ip_country/src/country_finder.rs
bertllll: Again, I think you're not describing what the code really does. You should think more of the consequences of the switch implemented on the compilation level for the functions holding the populated vector of compressed country blocks. There is the dispatch since which you can't ever test a dataset comparable to the intended in the reality.

node/src/neighborhood/gossip_acceptor.rs
bertllll: I think I didn't realize from just looking at the diffs that these are probably only references, I regarded them as owned values.

Then this conversation has reached the end.

node/src/neighborhood/node_location.rs
bertllll: I don't think that is the definition of the difference between Eq and PartialEq. I don't even remember the details as it's complicated but all I need to know for my work is that PartialEq makes asserting on values possible and therefore we need it on types we assert on. I only think the Clippy told you to add Eq, it wasn't probably your idea. 

What I don't really like, though, is your exclusive (or `custom` if put differently) implementation of PartialEq which is in about 90% cases needed only into the assertions. And we do want to test the structures wholly, if we can, and here it seems like we can but you decided not to by writing your own impl ignoring a part of the structure.

node/src/neighborhood/node_location.rs
bertllll: This was just an idea for a polemic. 

Basically, you have two data structures of different names containing the exact information. It draws questions whether we necessarily need to have both.

node/src/test_utils/neighborhood_test_utils.rs
bertllll: You mean the staged data approximating the reality?? Hmm. Maybe then you should've chosen less monotonic ip addresses then four same values. But whatever, after all, this isn't the biggest issue I've found.

node/src/neighborhood/neighborhood_database.rs
bertllll: Man, returning here, I don't really like this new piece of green code. 

It is counter-intuitive to me. If our root Node doesn't have an Ip address, it probably only can signify that it is an originate-only Node. Then why, for the God's sake, would you supply Some(<made-up IP address>) to form the record. Why don't we simply leave None in there? It feels much better. 

This is as if you knew you were an originate-only Node, but you'd deliberately shared a garbage address with other Nodes through Gossip. Doesn't it feel so bad to you as well?

This is the moment when it's completely justifiable to use None. 

But before you change things blindly please think twice as you should know more context than me. I'm working with incomplete information, still it makes me think this is a sure thing. (I wondered what the default for a NodeAddress - which I don't think we want to use here - could have been.)

ip_country/src/test_dbip_country.rs
bertllll: Manually generated? 

I think this definitely isn't generated by any binary as the other official file is. So the comment is out of place here. While it was written for the original file (the prod code), it makes no sense to have it here.

node/src/blockchain/blockchain_bridge.rs
bertllll: OK. I reviewed this case again and must admit my assumption from the first time was wrong. I didn't realize there is a sequence of steps where we probably query the blockchain logs and the process fails with a particular error which our code is able to handle (Steve's modification), if it is recognized as an error reasoned from requesting too many blocks at once, we don't want to treat it fatally, instead, correcting our knowledge of the maximum range of blocks to be queried together is the next step to happen, but in this test this last step, saving the parsed number of blocks, is supposed to fail. Then you get finally your fatal error followed by a panic. 

And I know understand the asserted message corresponds to the staged error on the line 1644. 

Yeah, my bad. This looks kind of okay. 

(PS: The other suggestion meant for this test still holds, though.)

node/src/neighborhood/mod.rs
bertllll: Oh...I see. Then that's it. It cannot be Copy. At least we have this cleared up now.

node/src/neighborhood/node_location.rs
bertllll: I want you to learn critical thinking. You should find the answers hopefully with just clues given by another person. I didn't do a perfect job writing the comment but I think it still could at least direct you to go checking what I brought up. That is, the act in this test is actually a function call of a function that has nothing to do with this file. I wouldn't think it was untested as it is code that had been around long before you joined the team. Therefore, the easiest thing to do is to go to the file where this function is declared and search for tests that back it. You'll eventually find some. Now, if anything changed about what happens inside the call, it should only be reflected in the assertions of those tests unlike a new one. If the old tests have accommodated any additional things or changed behavior, then you're fine. This is your all concern. 

So before anything else, instead of jumping on writing a new test right away, check the old ones and see if you could enhance them to test all you need. It usually works pretty well.

node/src/neighborhood/neighborhood_database.rs
czarte: youre right, I changed it to:
```
let location_opt = match neighborhood_mode.node_addr_opt() {
            Some(node_addr) => get_node_location(Some(node_addr.ip_addr()), &COUNTRY_CODE_FINDER),
            None => None,
};
```

ip_country/src/country_finder.rs
czarte: I tested separately deserialization and filling the vectors. That was suggestion from Dan and I saw it reasonable. If  want to suggest some more assertion, please let me know what do you see appropriate, because I can't see the benefit of asserting on separate deserialization of ipv4 and ipv6 country blocks, because these can vary and it does not bother us too much. We want to know how efficient is process of whole deserialization vs flling the vecs, because it affects the loading time of the Node. And we will be allways deserializing both version.

The reason, why we track down the time for deserialization and filling the vectors, in separate duration timestamps is, that if the filling of vetors could rise to some considerable duration, we can optimize this process by constructing enough memory space right on the compilation time.

node/src/neighborhood/node_location.rs
czarte: I think my critical thinking is quite ok. What I need to learn is to write efficient and reasonable Test Driven Code. Thank for advice. I have found test in `node_record.rs::398` where I added assetion on `inner.country_code_opt` and `metadata.node_location_opt.free_world_bit`, this redundant test un `node_location.rs` is deleted.

node/src/neighborhood/node_location.rs
czarte: Ok I deleted the implementation and used annotation.

PR #513, opened 2024-09-17
Gh 784 review three head

czarte:


ip_country/src/country_block_serde.rs
```diff
@@ -174,27 +182,41 @@ impl CountryBlockSerializer {
...
+            FinalBitQueue {
+                bit_queue: bit_queue_ipv6,
+                block_count: block_count_ipv6,
+            },
+        )
     }
 }
 
-struct VersionedIPSerializer<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize> {
+struct VersionedIPSerializer<IPType: Debug, SegmentNumRep: Debug, const SEGMENTS_COUNT: usize> {
```
bertllll, line 202: Not necessary, this is only a suggestion: Regarding how you added the Debug constraints, I like more the where clause construct. Provides better arrangement.

ip_country/src/country_block_serde.rs
```diff
@@ -283,11 +308,11 @@ where
...
 }
 
-// Rust forces public visibility on traits that come to be used for type boundaries in any public
-// interface (as in structs). This is how we can meet these requirements while those implementations
-// of these traits will be ineffective from farther than this file. It works for prevention of
-// namespace pollution in this kind of trait to be implemented on massively common types such as
-// Ipv4Addr or Ipv6Addr
+// Rust forces public visibility on traits that come to be used as type boundaries in any public
+// interface. This is how we can meet the requirements while the implementations of such
+// traits becomes ineffective from farther than this file. It works as a form of prevention to
```
bertllll, line 313: Instead of "traits becomes ineffective from farther than this file.", let's write "traits become ineffective beyond this file".

ip_country/src/country_block_serde.rs
```diff
@@ -552,12 +584,17 @@ where
...
+where
+    IPType: Debug,
+    SegmentNumRep: Debug,
+{
     ip: IPType,
     segment_num_rep: PhantomData<SegmentNumRep>,
 }
 
-impl<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize>
+impl<IPType: Debug, SegmentNumRep: Debug, const SEGMENTS_COUNT: usize>
```
bertllll, line 597: Maybe a where clause also here? I'd probably do it.

ip_country/src/country_block_serde.rs
```diff
@@ -569,25 +606,17 @@ impl<IPType, SegmentNumRep, const SEGMENTS_COUNT: usize>
...
-    for VersionedIP<IPType, SegmentNumRep, SEGMENTS_COUNT>
-{
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.write_fmt(format_args!("VersionedIP {{ ip: {:?} }}", self.ip))
-    }
-}
-
 #[derive(Debug)]
-struct StreamRecord<IPType: Display, SegmentNumRep, const SEGMENTS_COUNT: usize> {
+struct StreamRecord<IPType: Debug, SegmentNumRep: Debug, const SEGMENTS_COUNT: usize> {
```
bertllll, line 610: Where clause?

ip_country/src/country_finder.rs
```diff
@@ -37,89 +29,38 @@ impl CountryCodeFinder {
...
+            Ok(index) => country_blocks[index].country.clone(),
             _ => Country::try_from("ZZ").expect("expected Country"),
         };
         match country.iso3166.as_str() {
             "ZZ" => None,
             _ => Some(country),
         }
     }
+
+    pub fn init(&self) {}
```
bertllll, line 49: This function deserves an explanation. Please write a comment inside the body or above. Along these lines: "This should provoke lazy_static to perform the value initialization"

You can also give the function a better name, perhaps: "`ensure_init()`".

ip_country/src/country_finder.rs
```diff
@@ -128,17 +69,18 @@ mod tests {
...
+            &COUNTRY_CODE_FINDER,
             IpAddr::from_str("0.0.5.0").unwrap(),
         );
         let time_end = SystemTime::now();
 
         assert_eq!(result, None);
         let duration = time_end.duration_since(time_start).unwrap();
         assert!(
-            duration.as_secs() < 1,
+            duration.as_millis() < 1,
```
bertllll, line 83: 1 ms sounds risky to me. Maybe at least `<= 1`, allowing at least 1 ms.

ip_country/src/country_finder.rs
```diff
@@ -192,74 +134,77 @@ mod tests {
...
         );
     }
 
     #[test]
     fn real_test_ipv6_with_google() {
-        let _ = CountryCodeFinder::find_country(
-            &COUNTRY_CODE_FINDER,
-            IpAddr::from_str("2607:f8b0:4009:814::2004").unwrap(), // dig www.google.com AAAA
-        )
-        .unwrap();
```
bertllll, line 225: You should replace this pink code by `COUNTRY_CODE_FINDER.ensure_init();`.

ip_country/src/country_finder.rs
```diff
@@ -192,74 +134,77 @@ mod tests {
...
-            duration.as_secs() < 1,
+            duration.as_millis() < 1,
             "Duration of the search was too long: {} ms",
             duration.as_millis()
         );
     }
 
     #[test]
-    fn deserialize_country_blocks_ipv4_and_ipv6_adn_fill_vecs() {
+    fn country_blocks_for_ipv4_and_ipv6_are_deserialized_filled_into_vecs() {
```
bertllll, line 182: `country_blocks_for_ipv4_and_ipv6_are_deserialized_filled_into_vecs` is not grammatical...

Let's see: `country_blocks_for_ipv4_and_ipv6_are_deserialized_and_inserted_into_vecs`

ip_country/src/country_finder.rs
```diff
@@ -192,74 +134,77 @@ mod tests {
...
-        let deserializer_ipv4 = CountryBlockDeserializer::<Ipv4Addr, u8, 4>::new(
-            crate::dbip_country::ipv4_country_data(),
-        );
-        let deserializer_ipv6 = CountryBlockDeserializer::<Ipv6Addr, u16, 8>::new(
-            crate::dbip_country::ipv6_country_data(),
-        );
-        let time_end = SystemTime::now();
 
+        let deserializer_ipv4 =
+            CountryBlockDeserializer::<Ipv4Addr, u8, 4>::new(dbip_country::ipv4_country_data());
```
bertllll, line 186: You can use the type alias you've been using for this structure already.

ip_country/src/country_finder.rs
```diff
@@ -192,74 +134,77 @@ mod tests {
...
+        let deserializer_ipv6 =
+            CountryBlockDeserializer::<Ipv6Addr, u16, 8>::new(dbip_country::ipv6_country_data());
```
bertllll, line 188: Same here.

ip_country/src/country_finder.rs
```diff
@@ -192,74 +134,77 @@ mod tests {
...
+
+        let time_end = SystemTime::now();
         let time_start_fill = SystemTime::now();
```
bertllll, line 191: Maybe `filling` instead of `fill` :man_shrugging:

ip_country/src/country_finder.rs
```diff
@@ -1,28 +1,20 @@
...
+        dbip_country::ipv4_country_data(),
+        dbip_country::ipv6_country_data()
     );
 }
 
 pub struct CountryCodeFinder {
-    ipv4: Vec<CountryBlock>,
-    ipv6: Vec<CountryBlock>,
+    pub ipv4: Vec<CountryBlock>,
+    pub ipv6: Vec<CountryBlock>,
```
bertllll, line 17: The need for this public visibility surprises me a little... can you assure yourself that taking it out would truly break anything somewhere else?

Feels like if this was required it could be pointing to a wrong design (exposing the guts while nobody outside probably really need to know the details.)

node/src/neighborhood/node_location.rs
```diff
@@ -1,28 +1,24 @@
...
-impl PartialEq<Self> for NodeLocation {
-    fn eq(&self, other: &Self) -> bool {
-        self.country_code == other.country_code
-    }
-}
-
-pub fn get_node_location(ip_opt: Option<IpAddr>) -> Option<NodeLocation> {
+pub fn get_node_location(
+    ip_opt: Option<IpAddr>,
+    country_code_finder: &CountryCodeFinder,
```
bertllll, line 15: I'm not sure where you needed to pull this variable out through the function's interface. As far as I remember you only supply `&COUNTRY_CODE_FINDER` at all those places. It makes a question if it should be inline or not. I'll leave the decision for you as you understand better the code and you might remember also usages (I assume it'd have been in a test) with a reference to a CountryCodeFinder. The problem is I cannot even imagine where you could and why would you get down to constructing a special CountryCodeFinder of somehow different qualities than what the central, official one offers.

czarte: Thank you, this could go back to use only `COUNTRY_CODE_FINDER`, because we use everywhere only real data

ip_country/src/country_finder.rs
czarte: I'm not sure, that is better name, but as you wish :)

PR #536, opened 2024-10-09
Gh 468 review2

czarte:


masq/src/command_factory.rs
```diff
@@ -258,6 +264,33 @@ mod tests {
         );
     }
 
+    #[test]
+    fn factory_produces_exit_location() {
```
dnwiebe, line 268: I'd like to see more tests here. This doesn't give me

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+
+const EXIT_LOCATION_ABOUT: &str =
+    "If you activate exit-location preferences, all exit Nodes in countries you don't specify will be prohibited: \n\
+    that is, if there is no exit Node available in any of your preferred countries, you'll get an error. However, \
+    if you just want to make a suggestion, and you don't mind Nodes in other countries being used if nothing is available \
+    in your preferred countries, you can specify --fallback-routing, and you'll get no error unless there are no exit Nodes \
+    available anywhere.\n\n\
+    Here are some example commands:\n\
+        masq> exit-location                                                       // disable exit-location \n\
+        masq> exit-location --fallback-routing                                    // disable exit-location \n\n\
```
dnwiebe, line 72: These comments sound like these commands remove the entire concept of exit locations. Maybe "disable exit-location preferences?"

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+        masq> exit-location --country-codes \"CZ,PL|SK\" --fallback-routing       // fallback-routing is ON, \"CZ\" and \"PL\" countries has same priority \"1\", \"SK\" has prirority \"2\"\n\
```
dnwiebe, line 73: "countries have same priority"; also, "priority" instead of "prirority."

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+        masq> exit-location --country-codes \"CZ|SK\"                             // fallback-routing is OFF, \"CZ\" and \"SK\" countries has different prirority\n";
```
dnwiebe, line 74: Also here

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+
+// TODO update following help when GH-469 is done
+const COUNTRY_CODES_HELP: &str = "Establish a set of countries that your Node should try to use for exit Nodes. You should choose from the countries that host the \
+        Nodes in your Neighborhood. List the countries in order of preference, separated by vertical pipes (|). If your level of preference \
+        for a group of countries is the same, separate those countries by commas (,).\n\
+        To obtain codes you cant use 'country-codes-list' command. You can specify country codes followingly:\n\n\
```
dnwiebe, line 80: "To obtain codes, you can use the 'country-codes-list' command. You can specify country codes as follows:"

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+        masq> exit-location --country-codes \"CZ,PL|SK\"                          // \"CZ\" and \"PL\" countries has same priority \"1\", \"SK\" has prirority \"2\" \n\
```
dnwiebe, line 81: "countries have" here and on the next line; also, on the next line "priority" is misspelled.

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+    use crate::commands::set_exit_location_command::SetExitLocationCommand;
+    use crate::test_utils::mocks::CommandContextMock;
+    use masq_lib::messages::{
+        CountryCodes, ToMessageBody, UiSetExitLocationRequest, UiSetExitLocationResponse,
+    };
+    use std::sync::{Arc, Mutex};
+
+    #[test]
+    fn can_deserialize_ui_set_exit_location() {
+        let expected_request = UiSetExitLocationRequest {
```
dnwiebe, line 123: If this variable isn't used anywhere in Act or Arrange, could you move it down to Assert? It's part of the assertion logic, and having it up here in Act is confusing. It may sometimes be necessary, but if it's not, the confusion can be avoided.

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+        };
+        let transact_params_arc = Arc::new(Mutex::new(vec![]));
+        let mut context = CommandContextMock::new()
+            .transact_params(&transact_params_arc)
+            .transact_result(Ok(UiSetExitLocationResponse {}.tmb(0)));
+        let stderr_arc = context.stderr_arc();
+        let subject = SetExitLocationCommand::new(&[
+            "exit-location".to_string(),
+            "--country-codes".to_string(),
+            "CZ,SK|AT,DE|PL".to_string(),
```
dnwiebe, line 148: Q. What if there are spaces in this string?
A. Then it becomes several parameters to `--country-codes`.

Whoa!

If we could provide several parameters to `--country-codes`, we might be able to have more straightforward syntax, and not have to rely on commas and pipes. What about `--country-codes CZ,SK AT,DE PL`? Is that clearer?

masq/src/commands/set_exit_location_command.rs
```diff
@@ -0,0 +1,224 @@
...
+pub fn set_exit_location_subcommand() -> App<'static, 'static> {
+    SubCommand::with_name("exit-location")
+        .about(EXIT_LOCATION_ABOUT)
+        .arg(
+            Arg::with_name("country-codes")
+                .long("country-codes")
+                .value_name("COUNTRY-CODES")
+                .validator(common_validators::validate_exit_locations)
+                .help(COUNTRY_CODES_HELP)
+                .required(false),
```
dnwiebe, line 99: The way you're currently doing this, you should probably be specifying minimum and maximum value counts here. Since there has to be one value, and there can't be any more than one value, both minimum and maximum should be 1. Then `clap` will take care of more syntax validation, and you won't have to.

masq_lib/Cargo.toml
```diff
@@ -12,9 +12,11 @@ actix = "0.7.9"
 clap = "2.33.3"
 const_format = "0.2.22"
 crossbeam-channel = "0.5.1"
+csv = "1.3.0"
```
dnwiebe, line 15: I think csv shouldn't be added to the `masq_lib` production tree, at least if I understand things correctly.

It sounds like it's only used in a test. If it absolutely has to be in this `Cargo.toml`, then, perhaps it should be in the test tree only.

masq_lib/Cargo.toml
```diff
@@ -12,9 +12,11 @@ actix = "0.7.9"
...
 dirs = "4.0.0"
 ethereum-types = "0.9.0"
 itertools = "0.10.1"
+ip_country = { path = "../ip_country"}
```
dnwiebe, line 19: Bert suggests that perhaps `ip_country` shouldn't have to be part of `masq_lib`, but could reference `masq_lib` and be referenced  directly by other subprojects that need it.

masq_lib <- ip_country
masq_lib <- node
ip_country <- node

masq_lib/src/messages.rs
```diff
@@ -846,6 +846,43 @@ pub struct UiWalletAddressesResponse {
...
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct CountryCodes {
+    #[serde(rename = "countryCodes")]
+    pub country_codes: Vec<String>,
+    #[serde(rename = "priority")]
+    pub priority: usize,
+}
+
+impl From<(String, usize)> for CountryCodes {
```
dnwiebe, line 857: A couple things.
* Bert wants you to deconstruct the tuple and give names to the elements to make reading easier.
* I don't like this `From` implementation here in `messages.rs`, because it leaks syntax from `masq` into `masq_lib` where it doesn't belong. I think this `impl` should be in `masq` to confine that knowledge there.

dnwiebe (commented):
Resume at shared_schema.rs

masq_lib/src/shared_schema.rs
```diff
@@ -520,6 +522,43 @@ pub mod common_validators {
         }
     }
 
+    pub fn validate_country_code(country_code: String) -> Result<(), String> {
+        let country_range = StringRecord::from(vec!["1.2.3.4", "5.6.7.8", country_code.as_str()]);
```
dnwiebe, line 526: You can probably just use `INDEX_BY_ISO3166.keys().contains(&country_code)` (or something close) here.

masq_lib/src/shared_schema.rs
```diff
@@ -520,6 +522,43 @@ pub mod common_validators {
...
+            }
+        })
+    }
+
+    pub fn validate_separate_u64_values(values: String) -> Result<(), String> {
+        validate_pipe_separate_values(values, |segment: String| {
+            segment
+                .parse::<u64>()
+                .map_err(|_| {
+                    "Supply positive numeric values separated by vertical bars like 111|222|333|..."
```
dnwiebe, line 555: This is actually validating on nonnegative values, not positive values. It will accept zero without complaint.

masq_lib/src/shared_schema.rs
```diff
@@ -606,16 +645,24 @@ pub mod common_validators {
...
-        values_with_delimiters.split('|').try_for_each(|segment| {
-            segment
-                .parse::<u64>()
-                .map_err(|_| {
-                    "Supply positive numeric values separated by vertical bars like 111|222|333|..."
-                        .to_string()
-                })
-                .map(|_| ())
-        })
+    fn validate_pipe_separate_values(
```
dnwiebe, line 648: Probably should be `validate_pipe_separated_values`

node/src/bootstrapper.rs
```diff
@@ -400,7 +400,7 @@ impl BootstrapperConfig {
             neighborhood_config: NeighborhoodConfig {
                 mode: NeighborhoodMode::ZeroHop,
                 min_hops: DEFAULT_MIN_HOPS,
-                country: "ZZ".to_string(),
+                exit_locations_opt: None,
```
dnwiebe, line 403: If we're always going to establish exit conditions via the UI once the Node is running, then this information doesn't need to be in the `NeighborhoodConfig`, and doesn't do any good here. It's only useful if we're going to accept exit conditions via command-line parameters.

However, wherever `exit_locations_opt` is, `fallback_routing` should also be. (Perhaps they should both be in a single structure!)

masq/src/commands/set_exit_location_command.rs
dnwiebe, line 1: Synchronize things that say "set exit location" with things that say "exit location" so that they all say the same thing.

node/src/neighborhood/neighborhood_database.rs
```diff
@@ -561,6 +570,48 @@ mod tests {
...
+            (&root_node).into(),
+            Wallet::from_str("0x0000000000000000000000000000000000004444").unwrap(),
+            &CryptDENull::from(root_node.public_key(), TEST_DEFAULT_CHAIN),
+        );
+        subject.add_node(node_a.clone()).unwrap();
+        subject.add_node(node_b.clone()).unwrap();
+
+        let mut num: u16 = 7890;
+        let mut keys_nums: Vec<(PublicKey, u16)> = vec![];
+        let mutable_nodes = subject.nodes_mut();
```
dnwiebe, line 589: Looks like this is the Act; it should be set off by blank lines.

node/src/sub_lib/neighborhood.rs
```diff
@@ -380,6 +381,38 @@ pub enum Hops {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Country Codes: {:?}, Priority: {};",
+            self.country_code, self.priority
+        )
+    }
+}
+
+impl Display for ExitLocationSet {
```
dnwiebe, line 402: This could probably call `Display` for `ExitLocation`.

node/src/sub_lib/neighborhood.rs
```diff
@@ -406,7 +439,7 @@ impl Display for Hops {
 pub struct NeighborhoodConfig {
     pub mode: NeighborhoodMode,
     pub min_hops: Hops,
-    pub country: String,
+    pub exit_locations_opt: Option<HashMap<Vec<String>, usize>>,
```
dnwiebe, line 442: Should this be `Option<ExitLocationSet>`?

node/src/test_utils/neighborhood_test_utils.rs
```diff
@@ -66,13 +66,22 @@ impl From<(&NeighborhoodDatabase, &PublicKey, bool)> for AccessibleGossipRecord
...
 
-pub fn make_node_record(n: u16, has_ip: bool) -> NodeRecord {
+pub fn make_segments(n: u16) -> Vec<u8> {
     let seg1 = ((n / 1000) % 10) as u8;
     let seg2 = ((n / 100) % 10) as u8;
     let seg3 = ((n / 10) % 10) as u8;
     let seg4 = (n % 10) as u8;
-    let key = PublicKey::new(&[seg1, seg2, seg3, seg4]);
-    let ip_addr = IpAddr::V4(Ipv4Addr::new(seg1, seg2, seg3, seg4));
+    vec![seg1, seg2, seg3, seg4]
```
dnwiebe, line 74: This might be more elegant as a tuple.

node/src/neighborhood/mod.rs
```diff
@@ -106,6 +108,9 @@ pub struct Neighborhood {
     db_password_opt: Option<String>,
     logger: Logger,
     tools: NeighborhoodTools,
+    exit_locations_opt: Option<HashMap<usize, Vec<String>>>,
```
dnwiebe, line 111: I may come back and change this comment, but at the moment I think this could probably be eliminated, because it doesn't need to survive the processing of the set-exit-locations message from the UI.

node/src/neighborhood/mod.rs
```diff
@@ -106,6 +108,9 @@ pub struct Neighborhood {
...
+    exit_countries: Vec<String>, //if we cross number of countries used in one workflow, we want to change this member to HashSet<String>
+    fallback_routing: bool,
```
dnwiebe, line 113: I may also come back and change _this_ comment, but right now I think that these two fields could possibly be replaced with a three-valued enum: no exit-location preference, exit-location preference with fallback, and exit-location preference without fallback.

node/src/neighborhood/mod.rs
```diff
@@ -388,7 +393,9 @@ impl Handler<NodeFromUiMessage> for Neighborhood {
 
     fn handle(&mut self, msg: NodeFromUiMessage, _ctx: &mut Self::Context) -> Self::Result {
         let client_id = msg.client_id;
-        if let Ok((_, context_id)) = UiConnectionStatusRequest::fmb(msg.body.clone()) {
+        if let Ok((message, _)) = UiSetExitLocationRequest::fmb(msg.body.clone()) {
```
dnwiebe, line 396: I think you're going to need the `context_id` to be able to send back a response that gets to the right place in the UI.

node/src/neighborhood/mod.rs
```diff
@@ -1448,6 +1458,79 @@ impl Neighborhood {
         undesirability + node_undesirability
     }
 
+    fn handle_exit_location_message(&mut self, message: UiSetExitLocationRequest) {
```
dnwiebe, line 1461: It doesn't look as though this sends back a response. If that doesn't happen, the UI should wait forever for confirmation that the exit locations have been set.

This is probably why you didn't end up needing a `context_id`.

node/src/neighborhood/mod.rs
```diff
@@ -1448,6 +1458,79 @@ impl Neighborhood {
...
+        self.fallback_routing = message.fallback_routing;
+        let fallback_status = match self.fallback_routing {
+            true => "is",
+            false => "NOT",
+        };
+        let nodes = self.neighborhood_database.nodes_mut();
+        for node_record in nodes {
+            let country_code = match node_record.inner.country_code_opt.as_ref() {
+                Some(code) => code.clone(),
+                None => "ZZ".to_string(),
```
dnwiebe, line 1485: If this isn't already a manifest constant somewhere, it seems like it should be.

node/src/neighborhood/mod.rs
```diff
@@ -1448,6 +1458,79 @@ impl Neighborhood {
...
+                if exit_countries.contains(&country_code) {
+                    node_record.metadata.country_undesirability =
+                        1_000 * (*exit_priority - 1) as u32;
+                }
+                if !self.exit_countries.contains(
+                    node_record
+                        .inner
+                        .country_code_opt
+                        .as_ref()
+                        .expect("expected Country Code"),
```
dnwiebe, line 1497: This looks as though it will blow up the whole Node if one of these NodeRecords doesn't have a country code.

node/src/neighborhood/mod.rs
```diff
@@ -1448,6 +1458,79 @@ impl Neighborhood {
...
+                ) {
+                    node_record.metadata.country_undesirability = 1_000_000u32; //TODO change for constant COUNTRY_CODE_PENALTY
+                }
+                // TODO implement of reset of country_undesirability to 0 on all nodes when user disable exit-location
+                // --country-codes "CZ|DE" --falback-routing
+                // CZ = 0, DE = 1000, ALL OTHER COUNTRIES = 2000
+                // --country-codes "CZ|DE"
+                // CZ = 0, DE = 1000, ALL OTHER COUNTRIES = 100_000_000
```
dnwiebe, line 1505: Bunch of TODOs left here...

It's indeed an interesting problem to figure out exactly what `ALL_OTHER_COUNTRIES` should be. Maybe it should be dynamically determined: for example, if you have two priority levels, then `ALL_OTHER_COUNTRIES` should be 2000, which is what the third priority level would be; if you have ten priority levels, then `ALL_OTHER_COUNTRIES` should be 10,000, which is what the eleventh level would be. Or maybe it would be better to give it a constant value of some sort, like you have here.

However, if `--fallback-routing` is not set, then it _does not matter_ what `ALL_OTHER_COUNTRIES` is, because it should never be consulted. Under no `--fallback-routing`, a `NodeRecord` with a country not in the preferred list should be dropped from all consideration from exiting, just as if it were a Consume-Only Node.

dnwiebe (commented):
Resume at neighborhood/mod.rs:1505

node/src/neighborhood/mod.rs
```diff
@@ -1755,11 +1840,10 @@ mod tests {
             vec![make_node_descriptor(make_ip(2))],
             rate_pack(100),
         );
-        let country = "ZZ".to_string();
         let neighborhood_config = NeighborhoodConfig {
             mode,
             min_hops,
-            country,
+            exit_locations_opt: None,
```
dnwiebe, line 1846: This field could probably be that three-valued enum that we talked about above.

dnwiebe, line 1846: Since this will always be `None`, it doesn't need to be specified here.

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+        };
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
+
+        subject_addr.try_send(message).unwrap();
+        subject_addr.try_send(assertion_msg).unwrap();
+
+        System::current().stop();
+        system.run();
+
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
```
dnwiebe, line 3205: You can use a single `TestLogHandler` for all of these assertions.

`val tlh = TestLogHandler::new()`
`tlh.assert_logs_contain_in_order(...)`

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+            &format!("INFO: {}: Fallback Routing is Set.", test_name),
+            &format!("INFO: {}: Exit Location Set:", test_name),
+        ]);
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
+            &format!("INFO: {}: Exit Location Set:", test_name),
+            &format!("{}", "Country Codes: [\"CZ\", \"SK\"] - Priority: 1;"),
```
dnwiebe, line 3211: This could be just a `&str`; you don't need a `format!` macro here.

node/src/neighborhood/mod.rs
```diff
@@ -1448,6 +1458,79 @@ impl Neighborhood {
         undesirability + node_undesirability
     }
 
+    fn handle_exit_location_message(&mut self, message: UiSetExitLocationRequest) {
+        let exit_locations_by_priority: HashMap<usize, Vec<String>> = message
```
dnwiebe, line 1462: It doesn't look like you're ever looking up a list of countries by priority. That means you don't really need the hashiness of the `HashMap`; you could probably replace this with a `Vec<(usize, Vec<String>)>`. Then the contents would always be in the same order.

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+        let subject_addr = subject.start();
+        let peer_actors = peer_actors_builder().ui_gateway(ui_gateway).build();
+        let q_public_key = q.inner.public_key.clone();
+        let r_public_key = r.inner.public_key.clone();
+        let s_public_key = s.inner.public_key.clone();
+        let t_public_key = t.inner.public_key.clone();
+        let q_public_key_2 = q.inner.public_key.clone();
+        let r_public_key_2 = r.inner.public_key.clone();
+        let s_public_key_2 = s.inner.public_key.clone();
+        let t_public_key_2 = t.inner.public_key.clone();
```
dnwiebe, line 3279: Could you move these `xxx_2` variables down to before the closure where they're used? Thanks...

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+        };
+        let mut subject = make_standard_subject();
+        let system = System::new(test_name);
+        let (ui_gateway, _, _) = make_recorder();
+
+        subject.exit_locations_opt = None;
+        subject.logger = Logger::new(test_name);
+
+        let q = &mut make_node_record(3456, true);
+        q.inner.country_code_opt = Some("CZ".to_string());
```
dnwiebe, line 3253: I like the fact that you're making it obvious that `q` is from CZ. Could you make it similarly obvious that `s` is from FR? (Even though it automatically gets FR, that's not clear from this code.)

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+                    "We expecting zero, country is with Priority: 1"
+                );
+                assert_eq!(
+                    neighborhood
+                        .neighborhood_database
+                        .node_by_key(&r_public_key)
+                        .unwrap()
+                        .metadata
+                        .country_undesirability,
+                    1_000_000u32,
```
dnwiebe, line 3299: Should this be a manifest constant?

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+                        .neighborhood_database
+                        .node_by_key(&t_public_key_2)
+                        .unwrap()
+                        .metadata
+                        .country_undesirability,
+                    0u32,
+                    "We expecting zero, exit_location was unset"
+                );
+            }),
+        };
```
dnwiebe, line 3387: Do we need another assertion message here to make sure the field of the Neighborhood is properly reset? Or perhaps that could be part of the `AssertionMessage` above.

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+        subject_addr.try_send(BindMessage { peer_actors }).unwrap();
+        subject_addr.try_send(message).unwrap();
+        subject_addr.try_send(assertion_msg).unwrap();
+        subject_addr.try_send(assertion_msg_neighborhood).unwrap();
+        subject_addr.try_send(message_2).unwrap();
+        subject_addr.try_send(assertion_msg_2).unwrap();
```
dnwiebe, line 3393: Bert's right: these messages could have better names. Let's see...
`set_exit_location_message`
`assert_country_undesirability_populated`
`assert_neighborhood_exit_location`
`clear_exit_location_message`
`assert_country_undesirability_cleared`
`assert_neighborhood_no_exit_location`

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+            &format!("INFO: {}: Exit Location Set:", test_name),
+        ]);
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
+            &format!("INFO: {}: Exit Location Set:", test_name),
+            &format!("{}", "Country Codes: [\"CZ\"] - Priority: 1;"),
+        ]);
+        TestLogHandler::new().assert_logs_contain_in_order(vec![
+            &format!("INFO: {}: Fallback Routing is Set.", test_name),
+            &format!("INFO: {}: Exit Location Unset.", test_name),
+        ]);
```
dnwiebe, line 3409: It's possible that all these messages could go in one `assert_logs_contain_in_order()` call.

node/src/neighborhood/mod.rs
```diff
@@ -3064,6 +3148,267 @@ mod tests {
...
+            client_id: 0,
+            body: request.tmb(0),
+        };
+        let mut subject = make_standard_subject();
+        let system = System::new(test_name);
+        let (ui_gateway, _, _) = make_recorder();
+
+        subject.exit_locations_opt = None;
+        subject.logger = Logger::new(test_name);
+
```
dnwiebe, line 3251: These two blank lines make this look like the Act portion of the test. It's obvious upon further examination that it's not, but they're a little distracting.

masq/src/commands/set_exit_location_command.rs
czarte: Finally I used .value_delimiter("|") and there could be maximum c. 250 countries, but I do not thing there should be limit. Minimum values could be 0 to unset all exit-location preferences

masq_lib/Cargo.toml
czarte: shared_schema needs to validate country code, for that I am using const INDEX_BY_ISO3166 from ip_country

masq_lib/src/messages.rs
czarte: I can not move this impl to exit_location_command in masq, it complains about: Only traits defined in the current crate can be implemented for arbitrary types

PR #537, opened 2024-10-09
GH-525: Toggle for `entry_dns`

### LINKED ISSUE #437: Entry_DNS component toggle for Node
Currently when the MASQ starts, it always runs with `entry_dns` on and port 53 is listened to by Node.

Design a way to allow a startup parameter to be toggled that will turn off entry_dns and not require Node to listen on port 53, so other software can listen on the port unobstructed.

*Consider removing the entry_dns component, if system wide subversion is never to be a "thing"

Present a log to user to remind them that they will need to proxy network traffic/software to MASQ for privacy routing

ALSO check the daemon to ensure that entry_dns is not starting when daemon is running - there is a chance that this was turned on during some

SUGGESTIONS from Dan:

- Leave the entry_dns server in MASQNode for now, just in case it becomes useful in a system too small to have extensive proxying powers.
- In ServerInitializerReal, make the dns_socket_server field an Option. Start it out as None.
- Modify ServerInitializerReal::go() so that it needs a particular value from the multi_config before it will construct self.dns_socket_server and call .initialize_as_privileged() and .initialize_as_unprivileged() on it. That way, unless the user specifies something like --entry-dns on, there will be no contention over Port 53.
- Put help in shared_schema.rs that explains to the user the consequences of both --entry-dns on and --entry-dns off (which should be the default).

 
Command for inspecting processes with MASQNode
macOS
`ps -ef | grep MASQNode` (includes this command)

window
`netstat -ano`

List processes listening on port 53:
macOS
`lsof -Pn -i :53`

### LINKED ISSUE #525: Toggle for `entry_dns`
DNS subversion (using the `entry_dns` server) has been disabled essentially by commenting it out.

This works, unless the user finds himself in a situation where he needs DNS subversion, in which case he has to find and uncomment code and recompile and test.

`entry_dns` should be disabled by default, but we should have a command-line parameter (perhaps `entry-dns on`) available to bring it back if we need it. That's what this card is: establish such a command-line parameter.

utkarshg6:
Closes #525


masq_lib/src/shared_schema.rs
```diff
@@ -50,6 +50,7 @@ pub const EARNING_WALLET_HELP: &str =
      (case-insensitive). If you already have a derivation-path earning wallet, don't supply this. \
      If you have supplied an earning wallet address before, either don't supply it again or be \
      careful to supply exactly the same one you supplied before.";
+pub const ENTRY_DNS_HELP: &str = "Specify whether to use entry DNS.";
```
dnwiebe: I'd like to see something here about `dns_utility` and DNS subversion, so that users can connect that practice to this parameter. Maybe something like this?

If you can't use HTTP proxying to reroute your client's traffic through MASQ's Node, you can specify `entry-dns` to start up MASQ's local DNS server; but you will also need to subvert your network stack's DNS traffic using `dns_utility`. This practice is not recommended but may be necessary in certain situations.

node/src/server_initializer.rs
```diff
@@ -33,60 +35,90 @@ pub struct ServerInitializerReal {
...
+            result
+        }
+    }
+
+    fn initialize_unprivileged(
+        &mut self,
+        multi_config: &MultiConfig,
+        streams: &mut StdStreams<'_>,
+    ) -> RunModeResult {
+        let result = self
```
dnwiebe: Please switch these back so that `dns_socket_server` comes first.

dnwiebe (approved):

kauri-hero:
happy weekend @utkarshg6 

could you confirm what the default (if you configured one) boolean value is for entry_dns?

PR #540, opened 2024-10-10
GH-539: Don't Panic!

### LINKED ISSUE #539: Don't Panic!
Don't Panic

The thing is if an originating node receives a gossip from the other Node saying, "Hey! Meet this guy.", and it turns out that we were already trying to build a neighborship with that guy, we're going to panic. 😱 

We don't need to.

So, let's change the panic to an error log, here's your panic on node/src/neighborhood/overall_connection_status.rs:96 ==>

```rust
if self.connection_stage != ConnectionStage::TcpConnectionEstablished {
    panic!(
        "Can't update the stage from {:?} to {:?}",
        self.connection_stage,
        ConnectionStage::StageZero
    )
};
```

Now, we can in fact split this into following cases:

Case 1: `if self.connection_stage == StageZero` it means that we already know this guy 😉

Case 2: `if self.connection_stage == TcpConnectionEstablished` it means, we're on the ideal track 👍 

Case 3: Other cases, we should panic!

utkarshg6:
Closes #539

### DIFF
--- a/node/src/neighborhood/overall_connection_status.rs
+++ b/node/src/neighborhood/overall_connection_status.rs
@@ -394,7 +394,6 @@ mod tests {
     }
 
     #[test]
-    // #[should_panic(expected = "Can't update the stage from StageZero to StageZero")]
     fn connection_progress_logs_error_while_handling_pass_gossip_in_case_tcp_connection_is_not_established(
     ) {
         init_test_logging();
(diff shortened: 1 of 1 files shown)


node/src/neighborhood/overall_connection_status.rs
```diff
@@ -372,29 +379,58 @@ mod tests {
...
+                "DEBUG: {test_name}: Pass gossip received from Node with IP Address {:?} to a Node with \
+                IP Address {:?}. Updating the stage from TcpConnectionEstablished to StageZero.",
                 ip_addr, pass_target
             ),
         ]);
     }
 
     #[test]
-    #[should_panic(expected = "Can't update the stage from StageZero to StageZero")]
+    // #[should_panic(expected = "Can't update the stage from StageZero to StageZero")]
```
dnwiebe: Commented-out code

dnwiebe (approved):

PR #546, opened 2024-10-28
Gh 468 review4

czarte:


dnwiebe (commented):
Setting the country undesirability (if appropriate) for new Nodes that are added to the neighborhood database through Gossip needs to be part of this card to keep Node operation consistent.

PR #550, opened 2024-11-05
GH-800: Testing Issue Fix

utkarshg6:


node/src/proxy_client/stream_reader.rs
```diff
@@ -52,10 +51,9 @@ impl Future for StreamReader {
                     if self.logger.trace_enabled() {
                         trace!(
                             self.logger,
-                            "Read {}-byte chunk from {}: {}",
+                            "Read {}-byte chunk from {}",
                             len,
-                            self.peer_addr,
-                            utils::to_string(&Vec::from(&buf[0..len]))
```
dnwiebe, line 58: Can we now get rid of this function, or is it used elsewhere?

node/src/proxy_server/mod.rs
```diff
@@ -3757,6 +3746,25 @@ mod tests {
...
     fn proxy_server_schedules_stream_key_purge_once_shutdown_order_is_received_for_stream() {
+        let common_msg = StreamShutdownMsg {
+            peer_addr: SocketAddr::from_str("1.2.3.4:5678").unwrap(),
+            stream_type: RemovedStreamType::NonClandestine(NonClandestineAttributes {
+                reception_port: 0,
+                sequence_number: 0,
+            }),
+            report_to_counterpart: true,
+        };
+        stream_is_purged_with_a_delay(StreamShutdownMsg {
```
dnwiebe: This assertion function would be easier to read if it started with "assert" or "check" or something like that.

node/src/stream_handler_pool.rs
```diff
@@ -375,8 +375,30 @@ impl StreamHandlerPool {
...
+                );
+                false
+            }
+            Some(None) => {
+                trace!(
+                    self.logger,
+                    "While handling RemoveStreamMsg: Stream Writers contain an entry for key {}, but stream writer is missing",
+                    stream_writer_key
+                );
+                true
```
dnwiebe: Dan doesn't know about this; need to study the code more.

dnwiebe (commented):
Need to examine code for stream_handler_pool.rs and find out what line 392 should be.

node/src/proxy_client/stream_reader.rs
utkarshg6: It is still in use within the `http_packet_framer.rs` file, both in the production code and the tests.

node/src/proxy_server/mod.rs
utkarshg6: Done! 👍

PR #553, opened 2024-11-11
GH-744-Review-sub-2

Syther007:


multinode_integration_tests/tests/blockchain_interaction_test.rs
```diff
@@ -144,7 +144,7 @@ fn debtors_are_credited_once_but_not_twice() {
         let config_dao = config_dao(&node_name);
         assert_eq!(
             config_dao.get("start_block").unwrap().value_opt.unwrap(),
-            "2001"
+            "2000"
```
bertllll, line 147: This is serious. Apparently by the comment above, and from some memory I have of this test, its whole existence is predetermined by establishing an evidence that the next start block will be set to the current head block plus one (if the head block is known - and not represented as "latest") or by the youngest detected transaction (as the opposite of the previous text in brackets) but also plus one. The "plus one" part is integral. If you change your assertion like this, it breaks the intention of the test.

I noticed that the issue might be that somebody had probably set the test up wrong before you guys. 

If I'm right, and I'm not sure, then the following line 
`.response("0x5DC", 1) // eth_blockNumber 1500` isn't correct as it should be response to the query of the latest block number, but it is smaller than the block number which is supposed to hold in the only transaction detected which evaluates to 2000. I think the first value would never be smaller in the real world, because of its interstitial logic...

node/src/accountant/scanners/mod.rs
```diff
@@ -654,6 +654,8 @@ impl PendingPayableScanner {
         msg: ReportTransactionReceipts,
         logger: &Logger,
     ) -> PendingPayableScanReport {
+        // TODO: We want to ensure that failed transactions are not marked still pending,
+        // and also adjust log levels accordingly.
```
bertllll, line 658: What's this? Still relevant or can it be erased?

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -1,34 +1,24 @@
...
+use web3::{Error, Web3};
 
 #[derive(Debug, PartialEq, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub enum TransactionReceiptResult {
     NotPresent,
     Found(TransactionReceipt),
-    Error(String),
+    TransactionFailed(TransactionReceipt), // RemoteFailure
+    Error(String),                         // LocalFailure
```
bertllll, line 21: I might appreciate if you succeeded incorporating in this name that it is "ours" or "local" or using any other word of a similar meaning...
It would be more helpful than just a comment because it could co be immediately understood from anywhere it is used at.

node/src/actor_system_factory.rs
```diff
@@ -2063,14 +2067,24 @@ mod tests {
...
+            received_payments_message.payments_and_start_block,
+            PaymentsAndStartBlock {
+                payments: vec![BlockchainTransaction {
+                    block_number: 2000,
+                    from: Wallet::new("0x0000000000006561726e696e675f77616c6c6574"),
+                    wei_amount: 996000000
+                }],
+                new_start_block: 1000000000
+            }
+        );
```
bertllll, line 2087: I advised you to delete this test as I realized it had been a mistake in the first place to write it and add ex-post. This test was created as some sort of hardening. It didn't test-drive anything. I just became aware back then that we don't really guarantee in our tests that an actor, if it needs a connection to an external data source, ends up equipped with exactly such which would work as we think they should...
Later on, it started to be clearer than ever to me that I was pondering over integration tests which I was trying to squeeze into the suit of rather unit tests. This whole effort was rather sick. And for you guys, seeing you straggle with trying to keep a good maintenance of these, I came up with an argument that there are actually much much higher level integration tests (multinode tests) which prove the same crap. It was to me that you should be liberated from this task and I begun urging you about deleting at least this test in particular.

You seem to have got worried of this action, as you did multiple times when I asked you to delete some code. My opinion is you're overly careful sometimes, but I tend to get why you are too. 

Anyway, I do think the best solution is to get rid of this hybrid between unit and integration tests which has never worked well and has been indicating it to us constantly by the impossible elegance to be made over it. It's always a good sign something's just somebody's really wrong thinking, this time mine.   

PS: Why not keeping it? Mostly because even if we insisted the test only does half the objectives promised as it covers only the BlockcahinInterface connection (the blockchain service provider's one) and it has no assertions for the PersistentConfig (a database connection).

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -431,164 +356,11 @@ mod tests {
...
-            "Transport error: Error(Connect, Os { code: ",
-        );
-        assert_string_contains(
-            error_message.as_str(),
-            ", kind: ConnectionRefused, message: ",
-        );
-    }
-
-    #[test]
-    fn transaction_receipt_batch_works() {
```
bertllll, line 508: I don't get why you also deleted this test meant for the batched version. My take is that this is going to be in the production code. It should be tested. I find this place quite appropriate then. It should actually have maybe more than one test, with other than just happy results, unless you join all the possible situations together within a single test as it seems you did with this in the pick code.

Otherwise, deleting the suite related to the non-batch receipt processing is what I applaud for.

node/src/blockchain/blockchain_interface/lower_level_interface.rs
```diff
@@ -1,24 +1,18 @@
...
+use serde_json::Value;
+use web3::transports::{Batch, Http};
+use web3::types::{Address, Filter, Log, U256};
+use web3::{Error, Web3};
 
 pub trait LowBlockchainInt {
     // TODO: GH-495 The data structures in this trait are not generic, will need associated_type_defaults to implement it.
     // see issue #29661 <https://github.com/rust-lang/rust/issues/29661> for more information
 
+    // TODO: Address can be a wrapper type
```
bertllll, line 15: If you want to keep this TODO, maybe try thinking of some extension of the idea as it can easily become uninterpretable after some time when we lose the context we have now from our heads.

node/src/blockchain/blockchain_interface/mod.rs
```diff
@@ -3,24 +3,30 @@
...
+use crate::blockchain::blockchain_interface::blockchain_interface_web3::lower_level_interface_web3::TransactionReceiptResult;
 
 pub trait BlockchainInterface {
     fn contract_address(&self) -> Address;
 
     fn get_chain(&self) -> Chain;
 
+    // Initially this lower_interface wasn't wrapped with a box, but under the card GH-744 this design was used to solve lifetime issues
+    // with the futures.
+    // The downside to this method is we cant store persistent values, instead its being initialised where ever it being used.
```
bertllll, line 29: I don't understand this second sentence. Why should we actually want to store any values (results) in the interface? And if you mean values that are present the whole runtime epoch, then I don't think it would be undoable. 

At any rate, I get confused by this comment. Consider improving or deleting it.

node/src/blockchain/blockchain_interface/mod.rs
```diff
@@ -35,5 +41,19 @@ pub trait BlockchainInterface {
...
+    fn process_transaction_receipts(
+        &self,
+        transaction_hashes: Vec<H256>,
+    ) -> Box<dyn Future<Item = Vec<TransactionReceiptResult>, Error = BlockchainError>>;
+
+    fn submit_payables_in_batch(
+        &self,
+        logger: Logger,
+        chain: Chain,
+        agent: Box<dyn BlockchainAgent>,
```
bertllll, line 53: It would be quite cool, if the agent could simply answer what the chain is. This agent should know about the details around the blockchain as much as nobody else, therefore it looks to me a bit ugly if I see that there are two arguments, the agent and the chain, instead of passing in only one, the agent.

If you think you don't have time to do it, I'll understand.

node/src/blockchain/test_utils.rs
```diff
@@ -186,175 +161,13 @@ impl ReceiptResponseBuilder {
 
         let rpc_response = RpcResponse {
```
bertllll, line 162: I cannot comment in a higher place so I'm doing here:
I'm surprised that you've preserved the ReceiptResponseBuilder. Does it also mean we still fetch the entire TransactionReceipt during the retrieve_transactions process? We discussed and concluded that this structure is unbearably excessive compared our needs (just a little fraction).

node/src/proxy_server/mod.rs
```diff
@@ -2535,6 +2580,58 @@ mod tests {
...
+        subject.stream_key_factory = Box::new(stream_key_factory);
+        subject.logger = Logger::new(test_name);
+        let subject_addr: Addr<ProxyServer> = subject.start();
+        let mut peer_actors = peer_actors_builder()
+            .proxy_server(proxy_server_mock)
+            .neighborhood(neighborhood_mock)
+            .build();
+        // Get the dns_retry_result recipient so we can partially mock it...
+        let dns_retry_result_recipient = peer_actors.proxy_server.route_result_sub;
+        peer_actors.proxy_server.route_result_sub = dns_retry_result_recipient; //Partial mocking
```
bertllll, line 2624: Lines 2622 - 2624 are meaningless, if you think harder of them. Kick them off please.

node/src/proxy_server/mod.rs
```diff
@@ -1228,6 +1235,44 @@ impl Hostname {
         };
         Self { hostname }
     }
+
+    fn is_valid(&self) -> Result<(), String> {
```
bertllll, line 1239: The name `is_valid` isn't particularly well chosen. These names with "is" imply usually a boolean to be on the function's return.  

Maybe you want something like just "validate".

node/src/sub_lib/blockchain_bridge.rs
```diff
@@ -102,11 +104,11 @@ mod tests {
     #[test]
     fn blockchain_bridge_subs_factory_produces_proper_subs() {
         let subject = BlockchainBridgeSubsFactoryReal {};
-        let blockchain_interface = BlockchainInterfaceMock::default();
+        let blockchain_interface = make_blockchain_interface_web3(find_free_port());
         let persistent_config = PersistentConfigurationMock::new();
         let accountant = BlockchainBridge::new(
             Box::new(blockchain_interface),
-            Box::new(persistent_config),
+            Arc::new(Mutex::new(persistent_config)),
```
bertllll, line 111: :see_no_evil: :exploding_head: :scream:

node/src/sub_lib/wallet.rs
```diff
@@ -128,7 +128,7 @@ impl Wallet {
             WalletKind::Address(address) => H160(address.0),
             WalletKind::PublicKey(public) => H160(*public.address()),
             WalletKind::SecretKey(key_provider) => key_provider.address(),
-            WalletKind::Uninitialized => panic!("No address for an uninitialized wallet!"),
+            WalletKind::Uninitialized => panic!("No address for an uninitialized wallet!"), // TODO: If we can get rid of it, it'll be awesome!
```
bertllll, line 131: Have you tried out guys? Not saying that you necessarily should.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -720,17 +742,11 @@ mod tests {
...
-                .agent
-                .pending_transaction_id(),
-            35.into()
-        );
         assert_eq!(
             blockchain_agent_with_context_msg_actual
                 .agent
                 .agreed_fee_per_computation_unit(),
-            10
+            9395240960
```
bertllll, line 749: Again, if this number comes from what the mocked blockchain service client server sends towards us, then I'd like to see the format of the number in the server's setup area being the same as here, in other words, if you use hexadecimal notations first please use the same value in hex here. It is much easier to check then.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -742,7 +758,7 @@ mod tests {
             blockchain_agent_with_context_msg_actual
                 .agent
                 .estimated_transaction_fee_total(1),
-            733280
+            688_934_229_114_880
```
bertllll, line 761: And also here

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1582,27 +1508,33 @@ mod tests {
...
-                earning_wallet.address()
-            )]
-        );
+        let expected_transactions = RetrievedBlockchainTransactions {
+            new_start_block: 6040060u64,
+            transactions: vec![
+                BlockchainTransaction {
+                    block_number: 6040059,
+                    from: make_wallet("first_wallet"), // Points to topics of 1
+                    wei_amount: 42,                    // Its points to the field data
```
bertllll, line 1517: Sry, I don't understand the comments and so I don't find them useful in the state they are.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1652,20 +1584,20 @@ mod tests {
         let earning_wallet = make_wallet("earning_wallet");
         let amount = 996000000;
         let expected_transactions = RetrievedBlockchainTransactions {
-            new_start_block: 1000000001,
+            new_start_block: 1000000000,
```
bertllll, line 1587: Always careful with this...

Are you sure the new form is correct? I  think it is, but be it requires you to fully understand when you're doing such changes.

Again, I think everybody would appreciate if you could keep this number in hex as well as the number you fetch through the server's response. My assumption is this equals to 0x3B9ACA00.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1759,37 +1692,133 @@ mod tests {
...
+        let (accountant, _, accountant_recording_arc) = make_recorder();
+        let accountant = accountant.system_stop_conditions(match_every_type_id!(ScanError));
+        let earning_wallet = make_wallet("earning_wallet");
+        let blockchain_interface = make_blockchain_interface_web3(port);
+        let persistent_config = PersistentConfigurationMock::new()
+            .start_block_result(Ok(6))
+            .max_block_count_result(Err(PersistentConfigError::DatabaseError(
+                "my tummy hurts".to_string(),
+            )))
+            .set_max_block_count_result(Ok(()));
```
bertllll, line 1737: This is not enough. I don't like that your assertion at the bottom is a simple DEBUG log. It's an indirect, and therefore a weak assertion. Instead you should provide a container for params in this methods by using 
`set_max_block_count_params()`. You will assert on an actual call of this mocked method and that the value 1000 has been supplied. Such validation is much a stronger proof.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -1,12 +1,15 @@
-// Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+// Copyright (c) 2024, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+// TODO: GH-744: At the end of the review rename this file to: web3_blockchain_interface_utils.rs
+//       Or we should move this file into blockchain_interface_web3
```
bertllll, line 4: Yes please, don't forget about this.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -326,36 +308,23 @@ pub fn send_payables_within_batch(
...
+    )
 }
 
-pub fn convert_wei_to_gwei(wei: U256) -> u64 {
-    (wei / U256::from(GWEI_UNIT)).as_u64() + 1
-}
-
-pub fn create_blockchain_agent_web3(
-    gas_limit_const_part: u64,
+pub fn dynamically_create_blockchain_agent_web3(
```
bertllll, line 314: What do you mean by "dynamically"?? Does it really make sense or is that only distraction? As it is how it feels to me.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -875,18 +631,91 @@ mod tests {
             "INFO: {test_name}: {}",
             transmission_log(chain, &accounts, gas_price)
         ));
+        assert_eq!(result, expected_result);
+    }
+
+    #[test]
+    fn send_payables_within_batch_works() {
+        let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let expected_result = Ok(vec![
```
bertllll, line 640: Could you move the expected results to be right before the `execute_send_payables_test` function? Preferably with a blank line between these two?

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -875,18 +631,91 @@ mod tests {
...
+            port,
+        );
+    }
+
+    #[test]
+    fn send_payables_within_batch_fails_on_submit_batch_call() {
+        let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let os_code = transport_error_code();
+        let os_msg = transport_error_message();
+        let expected_result = Err(Sending {
```
bertllll, line 677: Same here, please.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -875,18 +631,91 @@ mod tests {
...
+            port,
+        );
     }
 
     #[test]
     fn send_payables_within_batch_all_payments_fail() {
-        init_test_logging();
-        let test_name = "send_payables_within_batch_all_payments_fail";
+        let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let expected_result = Ok(vec![
```
bertllll, line 697: And here too, please.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -875,18 +631,91 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn send_payables_within_batch_fails_on_submit_batch_call() {
+        let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let os_code = transport_error_code();
+        let os_msg = transport_error_message();
+        let expected_result = Err(Sending {
+            msg: format!("Transport error: Error(Connect, Os {{ code: {}, kind: ConnectionRefused, message: {:?} }})", os_code, os_msg).to_string(),
```
bertllll, line 678: Hmmm, this error should probably say something regarding the trigger point. So that we know, for example like here, that it happened during the call of `submit_batch`.

If there are more such places, then allow it to be a message compatible (given its variable part) with them all.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -899,107 +728,38 @@ mod tests {
...
-            transmission_log(chain, &accounts, gas_price)
-        ));
     }
 
     #[test]
     fn send_payables_within_batch_one_payment_works_the_other_fails() {
-        init_test_logging();
-        let test_name = "send_payables_within_batch_one_payment_works_the_other_fails";
+        let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let expected_result = Ok(vec![
```
bertllll, line 746: Here too please.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -1111,7 +826,7 @@ mod tests {
         let web3 = Web3::new(transport.clone());
         let chain = DEFAULT_CHAIN;
         let amount = 11_222_333_444;
-        let gas_price_in_gwei = 123000000000_u64;
+        let gas_price_in_wei = 123_000_000_000_000_000_000;
```
bertllll, line 829: You can use a pow() call. `123 * 10.pow(18)`. Also at other places :man_shrugging:

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -614,10 +705,11 @@ mod tests {
             .response("trash", 1)
             .raw_response(r#"{"jsonrpc":"2.0","id":2,"result":[{"address":"0xcd6c588e005032dd882cd43bf53a32129be81302","blockHash":"0x1a24b9169cbaec3f6effa1f600b70c7ab9e8e86db44062b49132a4415d26732a","data":"0x0000000000000000000000000000000000000000000000000010000000000000","logIndex":"0x0","removed":false,"topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef","0x0000000000000000000000003f69f9efd4f2592fd70be8c32ecd9dce71c472fc","0x000000000000000000000000adc1853c7859369639eb414b6342b36288fe6092"],"transactionHash":"0x955cec6ac4f832911ab894ce16aa22c3003f46deff3f7165b32700d2f5ff0681","transactionIndex":"0x0"}]}"#.to_string())
             .start();
-        let subject = make_blockchain_interface_web3(Some(port));
+        let subject = make_blockchain_interface_web3(port);
         let start_block_nbr = 42u64;
         let start_block = BlockNumber::Number(start_block_nbr.into());
-        let fallback_number = calculate_fallback_start_block_number(start_block_nbr, u64::MAX);
+        // let fallback_number = BlockchainBridge::calculate_fallback_start_block_number(start_block_nbr, u64::MAX);
```
bertllll, line 711: Shouldn't you delete this line? It might be needless now.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -768,29 +855,114 @@ mod tests {
...
+                contract_address: None,
+                logs: vec![],
+                status: Some(status_failed),
+                root: None,
+                logs_bloom: H2048::default()
+            })
+        );
+        assert_eq!(
+            result[5],
+            TransactionReceiptResult::Found(TransactionReceipt {
```
bertllll, line 953: Are you guys sure that we want to operate with these massive data structures from Web3. We held a discussion and I thought you might try to replace it with something closely ours with just as much in it as what we will utilize.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
Syther007: we deleted this test because the batched version is being tested at a higher level.


` blockchain_bridge_processes_requests_for_a_complete_and_null_transaction_receipt`
` handle_request_transaction_receipts_short_circuits_if_submit_batch_fails`
`handle_request_transaction_receipts_short_circuits_on_failure_from_remote_process_sends_back_all_good_results_and_logs_abort`

node/src/blockchain/test_utils.rs
Syther007: If we could we would really like to keep `ReceiptResponseBuilder`.
Being only test code I hope this isn't to much of a problem.
I predict in the future with us using MBCS more that `ReceiptResponseBuilder` will come in handy.

node/src/sub_lib/wallet.rs
Syther007: We have created a card for this task;
https://github.com/orgs/MASQ-Project/projects/8/views/1?pane=issue&itemId=90819582&issue=MASQ-Project%7CMASQ-Node-issues%7C816

Trying to keep things simple if we can :)

node/src/blockchain/blockchain_interface_utils.rs
Syther007: Hi Bert, Would you accept?

We used the word dynamically because we are returning a dyn.
and hope to stop anyone from refactoring this code by moving to the place its being called.

node/src/blockchain/blockchain_interface_utils.rs
Syther007: This error message will be delivered to Accountant, If you like we can concatenate it over there?

node/src/blockchain/blockchain_bridge.rs
Syther007: Is this better?

```rust
        let expected_transactions = RetrievedBlockchainTransactions {
            new_start_block: 6040060u64,
            transactions: vec![
                BlockchainTransaction {
                    block_number: 6040059,
                    from: make_wallet("first_wallet"), // Relates to RPC response topics of 1
                    wei_amount: 42,                            // Relates to RPC response field data
                },
                BlockchainTransaction {
                    block_number: 6040060,
                    from: make_wallet("second_wallet"), // Relates to RPC response topics of 1
                    wei_amount: 55,                             // Relates to RPC response field data
                },
            ],
        };
```

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
bertllll: Ok. Could you please add a one-liner to this place (which used to hold the deleted test)? But only if there have remained some similar tests, aimed at failures possibly, which you in turn didn't delete? It's because if anybody find some "sad path" tests they would expect to see them together with a "happy path" test nearby. You deleted a test saying "....something..._works". The word "works" is usually used if the test writer intended to create a full set of tests. It automatically raises eyebrows as it nearly indicates a missing test.  

Just ensure that this impression couldn't easily come on you. If you agree with my reason, use a comment to explain that there is another test somewhere else (the one you refer about in our conversation) which covers it properly.

node/src/blockchain/test_utils.rs
bertllll: Ehhh...that wasn't my point...but I think I didn't make myself clear. 

I just meant that I've been saying for a while that there will be time when we will have to get rid of our dependency on Web3. We don't want to work with such sizable structures as TransactionReceipt (from Web3) and therefore I hoped you would boil it down to our own and much smaller structure which could contain just that information we want. I'm certain it could be done while it would most likely also nicely work for any new chain added. However this isn't cheap in terms of time spent on refactoring it. So I will probably give you a green now. 

I myself wouldn't say it's completed if I were the author. I'd try to get rid of this boilerplate. If you don't feel up to this, let's throw it behind our back, but a TODO close to these problematic Web3 structures would be an advantage to have in the future. Please check the interface for any bloated structures fetched by an RPC code out of which we need just a little fragment of its whole. Write corresponding TODOS best at the trait declaration for an interface.

node/src/sub_lib/wallet.rs
bertllll: OK. Valid.

node/src/blockchain/blockchain_bridge.rs
bertllll: No, I'd need the comment to be more explicit about what is "topics of 1" and I'm not even sure about the rest. Which "field data"? Why is it important. 

It's so confusing. Above all, the comment probably has no or barely any value. I mean, maybe even if stated in better words. 

I can only judge what I can see on the screen and even though I give it an extra effort I must conclude the comment probably makes sense only to the person who wrote it and nobody else. Which is a bit irritating, because then no comment would be better than this.

node/src/blockchain/blockchain_interface_utils.rs
bertllll: I just noticed that it has a special error variant "Sending {...}". (See, it envelopes the data). If we work properly with these different categories in the Debug or even Display impl for this error type being an enum, than we probably don't have to care any further.

node/src/blockchain/test_utils.rs
Syther007: Ah I see, Sorry I must of misunderstood you.
We created a new Struct `TxReceipt` which is a striped down version.
I believe this is what you are asking for.

node/src/blockchain/blockchain_bridge.rs
Syther007: The `topics` are coming from the RPC response that was created in MBCS. I think these comments would make a lot more sense if we looked at the full test.

node/src/blockchain/blockchain_interface_utils.rs
Syther007: We’d love to keep this message string and prefer to keep it the way it is because we think it has some really important information in it.

node/src/blockchain/blockchain_bridge.rs
bertllll: I get that part still the comment can be confusing. Okay, I think, now that I'm sure we're talking about the same thing. What about this:

```
        let expected_transactions = RetrievedBlockchainTransactions {
            new_start_block: 6040060u64,
            transactions: vec![
                BlockchainTransaction {
                    block_number: 6040059,
                    // Wallet represented in the RPC response by the second 'topic' as: 0x000000000000000000000000000000000000000066697273745f77616c6c6574  
                    from: make_wallet("first_wallet"), 
                    // Paid amount read out from the field 'data' in the RPC  
                    wei_amount: 42,                         
                },
                BlockchainTransaction {
                    block_number: 6040060,
                    // Wallet represented in the RPC response by the second 'topic' as: 0x000000000000000000000000000000000000007365636f6e645f77616c6c6574  
                    from: make_wallet("second_wallet"),
                    // Paid amount read out from the field 'data' in the RPC  
                    wei_amount: 55,                             
                },
            ],
        };
```
This is a more explaining comment, I think. 
 
Just please check if I linked the right hex format to the wallet creating function it actually belongs to. :pray:

node/src/blockchain/blockchain_interface_utils.rs
bertllll: Yeah. I haven't definitely complained about a lengthiness of the presented message. The contrary, from the beginning, I was to suggest that we should add concatenate it with an identifier of the web3 function that threw this error out. 
However, as I commented last time, I've changed my mind since I noticed the `Sending { msg: ... }` case specifier. I've been satisfied since. It is misunderstanding therefore. Please have this conversation as closed.

PR #563, opened 2024-12-18
GH-744-Review-3

Syther007:


multinode_integration_tests/tests/communication_failure_test.rs
```diff
@@ -272,7 +272,9 @@ fn dns_resolution_failure_with_real_nodes() {
     );
 }
 
+// >>> TODO: GH-744: - Fix this test.
 #[test]
+#[ignore]
```
bertllll: What's this? Shouldn't you do it in this card? If you don't know what is causing it we can have a look in the team meeting.

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -233,7 +233,7 @@ fn verify_bill_payment() {
     assert_balances(
         &contract_owner_wallet,
         &blockchain_interface,
-        "99995074522000000000",
+        "99995231980000000000",
```
bertllll, line 236: Any idea why you changed the asserted value? It cannot be done just on random.

node/src/accountant/scanners/mid_scan_msg_handling/payable_scanner/agent_null.rs
```diff
@@ -37,6 +39,11 @@ impl BlockchainAgent for BlockchainAgentNull {
         &self.wallet
     }
 
+    fn get_chain(&self) -> Chain {
+        self.log_function_call("get_chain()");
+        TEST_DEFAULT_CHAIN
+    }
+
```
bertllll, line 46: This method call should've been added among the tests in the test section below.

node/src/accountant/scanners/mid_scan_msg_handling/payable_scanner/agent_web3.rs
```diff
@@ -86,6 +93,7 @@ mod tests {
             gas_limit_const_part,
             consuming_wallet.clone(),
             consuming_wallet_balances,
+            TEST_DEFAULT_CHAIN,
```
bertllll, line 96: This test should be broaden by another assertion of the get_chain() method.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -12,13 +12,31 @@ use web3::transports::{Batch, Http};
...
+    TransactionFailed(TxReceipt),
+    LocalError(String),
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct TxReceipt {
+    pub transaction_hash: H256,
+    pub block_hash: Option<H256>,
+    pub block_number: Option<U64>,
+    pub status: Option<bool>,
```
bertllll: This is mainly addressed to @utkarshg6. What do you think why couldn't we rely on this assumption?:  

If a transaction hasn't been added onto the blockchain, so it's not part of any block, it seems to me that all the optional fields in the struct will reflect it the same way, they will evaluate necessarily to `None`. Are you sure we cannot do something like: 

```
#[derive(Debug, PartialEq, Eq, Clone)]
pub struct TxReceipt {
    pub transaction_hash: H256,
    pub transaction_block_opt: Option<TransactionBlock>
}

#[derive(Debug, PartialEq, Eq, Clone)]
struct TransactionBlock {
    block_hash_opt: Option<H256>,
    block_number_opt: Option<U64>
}
  
```
I don't know if the responses can contain a divergence in these field but the common sense says that if a transaction doesn't have a block **then it is not confirmed**. That means that we may not need the `status` field at all.  

If you @utkarshg6 think that it's risky depending on that, you still, at minimum, can improve the name of the field as: 

`tx_confirmed_opt` 

where `false` means a failure :man_shrugging:...

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -372,7 +376,7 @@ impl BlockchainInterfaceWeb3 {
             );
 
             Ok(RetrievedBlockchainTransactions {
-                new_start_block: transaction_max_block_number,
+                new_start_block: transaction_max_block_number + 1,
```
bertllll, line 375: Yeah this is good. Just take caution for the start_block_fallback_value and its computation. I'm convinced that we've planted a mistake there when somebody decided to add another +1 as well as we do here. Because all together it makes +2. It should take a treatment at the other place by eliminating the plus one, not here.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -930,49 +933,59 @@ mod tests {
...
+        let subject = make_blockchain_interface_web3(port);
+        let tx_hash_1 =
+            H256::from_str("a128f9ca1e705cc20a936a24a7fa1df73bad6e0aaf58e8e6ffcc154a7cff6e0e")
+                .unwrap();
+        let tx_hash_2 =
+            H256::from_str("a128f9ca1e705cc20a936a24a7fa1df73bad6e0aaf58e8e6ffcc154a7cff6e0f")
+                .unwrap();
+        let tx_hash_vec = vec![tx_hash_1, tx_hash_2];
+
+        let result = subject
```
bertllll: You can rename "result" to "err" or "error". It's a formality but you maybe follow why I put the comment.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -654,13 +663,6 @@ mod tests {
...
         ]);
 
-        let port = find_free_port();
-        let _blockchain_client_server = MBCSBuilder::new(port)
-            .begin_batch()
-            .response("rpc_result".to_string(), 7)
-            .response("rpc_result_2".to_string(), 8)
-            .end_batch()
-            .start();
         execute_send_payables_test(
```
bertllll, line 664: Yeah, this test is so weak without the assertions of the RPC requests. :face_with_diagonal_mouth: 
But I know you created a card with the goal to get this right later, can you at least reference that card in a TODO in such tests? I'm really quite worried from seeing this going to be approved.

node/src/blockchain/blockchain_interface_utils.rs
```diff
@@ -637,6 +639,13 @@ mod tests {
     #[test]
     fn send_payables_within_batch_works() {
         let accounts = vec![make_payable_account(1), make_payable_account(2)];
+        let port = find_free_port();
+        let _blockchain_client_server = MBCSBuilder::new(port)
+            .begin_batch()
+            .response("rpc_result".to_string(), 7)
+            .response("rpc_result_2".to_string(), 8)
```
bertllll: Think about this both @Syther007  and @utkarshg6. Let me ask what you think. Imagine we just have sent our Json with the signed transactions in it, now we are waiting for the response and what we get back from the server is an error saying that we are rate-limited? How do you think our code would react upon its arrival? 
I'm a little worried, if we ignore the content of the response (Which we do because this test is a proof. We send in two garbage responses but the procedure ends up in two successes.), then we could also not to recognize that we're not receiving an OK, but something else. Can you please look into it carefully for me and disprove my hypothesis? Thanks.   
(Edit: Oh, I saw your telling apart `.response()` from `.err_response()`. It makes sense to me now. The problem is, it actually doesn't if you don't know about the specific ergonomics of your api for the MBCS. I can see how you could improve it easily. Just please modify it so that it says `.ok_response()` which will make it symmetrical to `.err_response`).

If all is fine, you also may want to use a bit more expressive strings:

Instead of 
```
      .response("rpc_result".to_string(), 7)
      .response("rpc_result_2".to_string(), 8)
```
rather
```
      .response("irrelevant_ok_rpc_response".to_string(), 7)
      .response("irrelevant_ok_rpc_response_2".to_string(), 8)
```

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -565,7 +459,7 @@ impl BlockchainBridge {
             if start_block_number == u64::MAX {
                 start_block_number
             } else {
-                start_block_number + 1u64 // TODO: GH-744 Way are we adding +1 can we just return the same value?
+                start_block_number + 1u64
```
bertllll, line 478: Yeah, I believe this is wrong. Careful here. But if you agree with me after I have explained you, do the cleanup.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -1730,14 +1628,12 @@ mod tests {
             received_payments,
             &ReceivedPayments {
                 timestamp: received_payments.timestamp,
+                new_start_block: 8675309u64 + 1,
```
bertllll, line 1662: I believe this is wrong. 
It should correspond **exactly** to the `new_start_block` in:
```
RetrievedBlockchainTransactions {
       new_start_block: 8675309u64,
       transactions: vec![BlockchainTransaction {
            block_number: 8675308u64,
            from: some_wallet.clone(),
            wei_amount: amount,
       }],
}
```
But you had to add one to make the assertion work and it only shows that the internal code is wrong. (We discussed the cause already. You will likely know by now how you should proceed)

multinode_integration_tests/tests/verify_bill_payment.rs
Syther007: The reason for this is because the multinode tests were not run before creating the second review.
It had mistakenly believed `./ci/all` ran multinode tests also. 😅

This is because of the changes we made to Gas fees (gwei to wei) from the first review.

bertllll: Ah, if that's how it is then it is probably all right.

PR #565, opened 2024-12-23
GH-744-Review-4

Syther007:


node/src/accountant/scanners/mod.rs
```diff
@@ -675,18 +675,24 @@ impl PendingPayableScanner {
         msg.fingerprints_with_receipts.into_iter().fold(
             scan_report,
             |scan_report_so_far, (receipt_result, fingerprint)| match receipt_result {
-                TransactionReceiptResult::Found(_receipt) => {
-                    handle_status_with_success(scan_report_so_far, fingerprint, logger)
+                TransactionReceiptResult::RpcResponse(tx_receipt) => {
+                    match tx_receipt.status {
+                        TxStatus::Pending => {
+                            handle_none_receipt(
```
bertllll: I just realised how you could improve the code arrangement easily so it shouldn't be too much bothering to do. 

Please move this `handle_none_receipt()` to the other file in which all these similar functions around are located, I think it's about "scanners::scanner_utils::pending_payable_scanner". The historically induced reason for writing this as an internal function (aka a formal function declared  inside the body of another function) has dissolved. 


Also, one more detail, I'd recommend to change its signature minority and that is to have `error_msg: &str` instead of `error_msg: String`. It's more idiomatic. Consider that you can supply a simple literal to such an argument, without having to always convert it into String first. On the other hand, if already have a String there is nothing easier than using it with &, which makes it a string slice instead.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -14,31 +14,51 @@ use web3::{Error, Web3};
...
+    RpcResponse(TxReceipt),
     LocalError(String),
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum TxStatus {
+    Failed,
+    Pending,
+    Succeeded(TransactionBlock),
+}
```
bertllll, line 26: I love it. 😊

bertllll (commented):
Continue node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs ln. 40

node/src/accountant/mod.rs
```diff
@@ -3800,6 +3802,8 @@ mod tests {
         let mut transaction_receipt_1 = TransactionReceipt::default();
         transaction_receipt_1.transaction_hash = transaction_hash_1;
         transaction_receipt_1.status = Some(U64::from(1)); //success
+        transaction_receipt_1.block_number = Some(U64::from(100));
+        transaction_receipt_1.block_hash = Some(Default::default());
```
bertllll: It seems to me like you're unnecessarily modifying an already instantiated structure. It might be cleaner if you simply constructed `TransactionReceipt` from scratch in place, by literal definition with all its fields and values. 
Why am I saying that? Well, the structure probably doesn't even compose of more than these four fields, making it slightly silly to go the more complicated way as you do.

node/src/accountant/mod.rs
```diff
@@ -3812,6 +3816,8 @@ mod tests {
         let mut transaction_receipt_2 = TransactionReceipt::default();
         transaction_receipt_2.transaction_hash = transaction_hash_2;
         transaction_receipt_2.status = Some(U64::from(1)); //success
+        transaction_receipt_2.block_number = Some(U64::from(200));
+        transaction_receipt_2.block_hash = Some(Default::default());
```
bertllll: Also here.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -403,8 +403,10 @@ impl BlockchainBridge {
                     let length = transaction_receipts_results.len();
                     let mut transactions_found = 0;
                     for transaction_receipt in &transaction_receipts_results {
-                        if let TransactionReceiptResult::Found(_) = transaction_receipt {
-                            transactions_found += 1;
+                        if let TransactionReceiptResult::RpcResponse(tx_receipt) = transaction_receipt {
+                            if let TxStatus::Succeeded(_) = tx_receipt.status {
+                                transactions_found += 1;
+                            }
```
bertllll: I have a couple of points about possibilities for minor optimisations.

- line 401: `move` is redundant 
- `if length != transactions_found {` coupled with "Aborting scanning; {} transactions succeed and {} transactions failed" is controversial. It's because the scan hasn't differed from a successful one in the length of the process (you do the same amount of work to figure out all the transactions were successful as you do if there are some pending or failed transactions), therefore speaking of "aborting" is misinterpreting the truth. We didn't abort. We completed everything. 

However there is more than that: you cannot count the succeeded transactions only and then assume that the rest of them have failed with no exception. That's not right. Be aware that the rest is a mix of legitimately **failed** transactions, yes, but also those still **pending**. I'm asking, do you think it's appropriate having the pending transactions counted in the failures? Because that's what you're doing in the debug!() log down here. 

If you still think we'd benefit from having that log, I'd recommend to keep track of all three categories: Confirmed, Pending, Failed. 
- inside the function `process_transaction_receipts()`, there is a passage in the chain of Future-devoted methods that says `.map_err(|e| e)` which is a redundant call with no effect, you can delete that line
-  in the impl of `get_transaction_receipt_in_batch()`, you can see this snipped: 
```
    let _ = hash_vec.into_iter().map(|hash| {
            self.web3_batch.eth().transaction_receipt(hash);
        });
```
Honestly, I'm surprised that this code even does anything because I've been taught that iterators in Rust are "lazy" and therefore their code isn't executed until a certain terminal method is reached and performed. Most of the time this method is the renowned `collect()`, then `fold()`, `partition()` and also `for_each()`. 
As you have used none of them in here, it makes me wonder how you got your code working, assuming this has proper test coverage. Anyway, without having to argue over this, we can easily put end to this question with the replacement of `.map()` by `.for_each()`. This swap will also rule out the assignment to an anonymous variable since `for_each()` is not allowed to return a value, naturally: It's meant to allow side effects only. I hope you'll be able to make out what I instruct you for.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -14,31 +14,51 @@ use web3::{Error, Web3};
...
     fn from(receipt: TransactionReceipt) -> Self {
+        let status = match (receipt.status, receipt.block_hash, receipt.block_number) {
+            (Some(status), Some(block_hash), Some(block_number)) if status == U64::from(1) => {
+                TxStatus::Succeeded(TransactionBlock {
+                    block_hash,
+                    block_number,
+                })
+            }
+            (Some(status), _, _) if status == U64::from(0) => TxStatus::Failed,
+            _ => TxStatus::Pending,
```
bertllll, line 50: I only hesitate at this wildcard situation. Shouldn't we test it thoroughly for every possibility that it embraces? The same with those two anonymous fields in the branch one higher.  We have thus probably 3 + 3 untested situations. Of course, I'm aware that all them or at least the vast majority of them are implicitly wrong and so, totally unanticipated, because they mismatch with the official Web3 protocol. Yeah, again, I know. Yet I think the standards we've had are that we should test drive them in.  

When I test drive code, despite I might know from the very start how many branches I want to keep, I sketch all the possibilities out, not shortened, and I use a single `todo!()` for each. I continue by knocking off one by one, writing tests that can hit them, until all `todo!()`s are gone. This is finally the moment that you have the test or tests ready, and after that you can do whatever you want, any possible rearrangement that comes to your mind. Therefore, you can reduce the number of branches in the match as you did guys in your code. Do you get the picture of what I'm trying to pass over to you? 

This is an example of those possible variants that resolve into TxStatus::Pending in your code. I believe you've got some higher level test that checks one of these, preferably the first one. However, because you probably didn't write a distinct test scoped just by testing this mechanism, you'd have to write a new one from scratch, instead of extending an old one.    

(None, None, None),
(None, Some(), None),
(None, None, Some()),
(None, Some(), Some())

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -202,29 +204,22 @@ impl BlockchainInterface for BlockchainInterfaceWeb3 {
     ) -> Box<dyn Future<Item = Vec<TransactionReceiptResult>, Error = BlockchainError>> {
         Box::new(
             self.lower_interface()
-                .get_transaction_receipt_in_batch(transaction_hashes)
+                .get_transaction_receipt_in_batch(transaction_hashes.clone())
                 .map_err(|e| e)
```
bertllll: Yeah. I commented on this one in on a different occasion. It's irrelevant if you either do or don't have this line in. Needless.

node/src/blockchain/blockchain_bridge.rs
utkarshg6: Thanks for this change! 😃 

I really appreciated your insights on tracking the different transaction statuses and the iterator usage. I went ahead and implemented those suggestions.

As for the `move` and the redundant error mapping, I decided to leave those as is to keep the original function signature intact.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
utkarshg6: The signature issue may cause a lot of trouble. I don't know if it's worth it.

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -400,33 +401,32 @@ impl BlockchainBridge {
...
+                                    TxStatus::Failed => (success, fail + 1, pending),
+                                    TxStatus::Pending => (success, fail, pending + 1),
+                                    TxStatus::Succeeded(_) => (success + 1, fail, pending),
+                                },
+                                TransactionReceiptResult::LocalError(_)=> (success, fail, pending + 1),
                             }
-                        }
-                    }
+                        });
+                    debug!(logger, "Scan results: Successful: {successful_count}, Pending: {pending_count}, Failed: {failed_count}");
```
bertllll: Do you think you could pack this whole fold for the debug into a separate function. 

I've got this in my mind: 

You can use the trick with 
```
logger.debug(||{
 <your function returns a String (the debug message)>
})
```

Therefore: 

```
fn transaction_receipts_results_debug(transaction_receipts_results: <The proper type>){
       logger.debug({
              let (successful_count, failed_count, pending_count) = transaction_receipts_results
                       .iter()
                       .fold((0, 0, 0), |(success, fail, pending), transaction_receipt| {
                               match transaction_receipt {
                                    TransactionReceiptResult::RpcResponse(tx_receipt) => match tx_receipt.status {
                                        TxStatus::Failed => (success, fail + 1, pending),
                                        TxStatus::Pending => (success, fail, pending + 1),
                                        TxStatus::Succeeded(_) => (success + 1, fail, pending),
                                    },
                                   TransactionReceiptResult::LocalError(_)=> (success, fail, pending + 1),
                              }
              format!("Scan results: Successful: {successful_count}, Pending: {pending_count}, Failed: {failed_count}");
      })
}
```

A praise for that nice use of `.fold()`. :wink:

PR #578, opened 2025-02-05
GH-744-QA-Review

Syther007:


multinode_integration_tests/tests/blockchain_interaction_test.rs
```diff
@@ -32,7 +32,7 @@ fn debtors_are_credited_once_but_not_twice() {
     // Create and initialize mock blockchain client: prepare a receivable at block 2000
     eprintln!("Setting up mock blockchain client");
     let blockchain_client_server = MBCSBuilder::new(mbcs_port)
-        .ok_response("0x5DC", 1) // eth_blockNumber 1500
+        .ok_response("0x9C4", 1) // eth_blockNumber 2500
```
bertllll: We discussed this and came to an idea that in order to preserve the full depth of the original test, which included picking the next start block based on the information coming back with the retrieved transaction, we could simulate an error to happen on the eth_block_number call. That would result in our ignoring of that event and we'd be forced to deduce the next start block from the only transaction that was found on the blockchain in the range of blocks we were scanning through.

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -307,6 +308,169 @@ fn verify_bill_payment() {
...
+
+    let ui_client = real_consuming_node.make_ui(ui_port);
+    ui_client.send_request(
+        UiScanRequest {
+            scan_type: ScanType::PendingPayables,
+        }
+        .tmb(0),
+    );
+
+    MASQNodeUtils::wrote_log_containing(
```
bertllll: Couldn't the name be more like "assert_node_wrote_log_containing"?

multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -307,6 +308,169 @@ fn verify_bill_payment() {
...
+    let real_consuming_node =
+        cluster.start_named_real_node(&consuming_node_name, consuming_node_index, consuming_config);
+    for _ in 0..6 {
+        cluster.start_real_node(
+            NodeStartupConfigBuilder::standard()
+                .chain(Chain::Dev)
+                .neighbor(real_consuming_node.node_reference())
+                .build(),
+        );
+    }
```
bertllll: This also can be done with the client only. There's no need to start those other Nodes. By saying that you want to run the scan for payables (similarly like you do with the pendingpayables scan), which has the power to override the Node startup conditions for launching the Accountant's services.

multinode_integration_tests/tests/verify_bill_payment.rs
bertllll: Instead of just reading the logs, there is probably an exacter assertion you could make, it involves reading from the payables table in the consuming node where there is a column that is usually empty but if we submit a transaction for the respective creditor (serving node), we first create a record in the pending_payables table and then we put the rowid of that record into the column meant for the rowid which are held in the payables table. By this we keep a relation between these two tables. Therefore, if a transaction is finally confirmed we need to blank this "linkage" out from that column. You should write an assertion which will check that the columns for all three wallets of the serving nodes are empty at the end of the test.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -103,22 +103,30 @@ impl BlockchainInterface for BlockchainInterfaceWeb3 {
...
         let contract_address = lower_level_interface.get_contract_address();
         let num_chain_id = self.chain.rec().num_chain_id;
+        let start_block_number = match start_block_marker {
+            BlockMarker::Uninitialized => self.chain.rec().contract_creation_block,
+            BlockMarker::Value(number) => number,
+        };
         Box::new(
             lower_level_interface.get_block_number().then(move |response_block_number_result| {
-                let end_block_marker = Self::calculate_end_block_marker(start_block_marker, scan_range, response_block_number_result, &logger);
+                let end_block_marker = Self::calculate_end_block_marker(start_block_number, scan_range, response_block_number_result, &logger);
```
bertllll: To be honest, I was grappling my thought, I had to, because the former version without the explicit transformation from start_block_marker into a number was quite neat. 
On the contrary, the facelift you did is a bit more chaotic. I don't mean really chaotic, but it doesn't make things particularly better. 

In the end, I assumed a position nearer your approach: however you should consider to modify sligthly the function `calculate_end_block_marker` to one which will return directly a `BlockNumber`. I think it's an unnecessary dressing outside, when you look at down here, it can be skipped if you change the return type as I say. For this reason, it'd better be renamed like `calculate_end_block`.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -350,16 +359,16 @@ impl BlockchainInterfaceWeb3 {
...
         response_block_number_result: Result<U64, BlockchainError>,
         logger: &Logger,
     ) -> BlockMarker {
-        let local_end_block_marker = match (start_block, scan_range) {
-            (BlockMarker::Value(start_block_number), BlockScanRange::Range(scan_range)) => {
-                BlockMarker::Value(start_block_number + scan_range)
+        let local_end_block_marker = match scan_range {
+            BlockScanRange::NoLimit => BlockMarker::Uninitialized,
+            BlockScanRange::Range(scan_range_number) => {
+                BlockMarker::Value(start_block + scan_range_number)
```
bertllll, line 370: The previous comment said it all, but for clarification, here I think there's no use of returning BlockMarker instead of BlockNumber already.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -577,7 +653,8 @@ mod tests {
                     },
                 ]
             }
-        )
+        );
+        TestLogHandler::new().exists_log_containing(&format!("DEBUG: {test_name}: Retrieving transactions {expected_log} for: 0x3f69…72fc chain_id: 137 contract: 0xee9a352f6aac4af1a5b9f467f6a93e0ffbe9dd35"));
```
bertllll: I'd prefer that you wouldn't insist on asserting on these logs. That's for in general. 

Here, it's even more clumsy, because your test performs this five times (with each call of assert_on_retrieves_transactions). It means that after the first call we only can't ever fail because if it worked once it will appease each attempt by which we check on the presence of this string in the list of printed logs. 

Again, I'd prefer to see that you deleted it wholly. There isn't much or almost any use at all in it.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -1026,84 +1103,48 @@ mod tests {
```
bertllll: If you want to keep the line
`type Subject = BlockchainInterfaceWeb3;`,
I'd move it probably to the absolute top. It's not an actual expression or statement, it's just a type alias. Strange use anyway.

node/src/proxy_server/mod.rs
```diff
@@ -1311,10 +1311,10 @@ impl Hostname {
     }
 
     fn validate_ipv4(addr: Ipv4Addr) -> Result<(), String> {
+        // TODO: 127.0.0.1 is being used in integration tests, that's why it's not here,
+        // but we should invalidate that IP too for safety
```
bertllll: Well, this has a solution and maybe we should apply it right away. 

This kind of thing is usually dealt by using the environment variables. 
Optimally, when the Node is starting up and the proxy_server in the actor_system_factory is being put together, we should make a query for a variable like "MASQ_INTEGRATION_TEST". If present among the current set of environment variables, let's set a flag in the body of the proxy_server, on a field being for example `is_running_in_integration_test` and with the value `true`. 

Elsewhere in the code, where this comment is placed by, you will add another condition into the combined if statement. something like:

```rust
   ...
   } else if !self.is_running_in_integration_test && addr.octets() == [127, 0, 0, 1] {
            Err("127.0.0.1".to_string())
   }
   ...
```
This code above can be even nicely unit-tested.  

For checking on the env var, use this code: `env::var("MASQ_INTEGRATION_TEST")`. It returns the value set behind this key. Therefore, also check that the value says "true".

Plus, of course, you have to go to the `ci` script for the integration tests and set this variable there. 
probably `EXPORT MASQ_INTEGRATION_TEST = true`

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
Syther007: It has another usage for `find_new_start_block`. 
If we match on `BlockNumber` instead of `BlockMarker`, we would have to cover more cases.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
Syther007: This `local_end_block_marker` is been return in the error case.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -575,9 +579,10 @@ mod tests {
         block_response: &str,
         expected_new_start_block: u64,
         expected_log: &str,
+        test_name: &str,
     ) {
         init_test_logging();
-        let test_name = "blockchain_interface_web3_retrieves_transactions";
+        // let test_name = "blockchain_interface_web3_retrieves_transactions";
```
bertllll: Please, obviously, this line is left without a purpose, but what I suggest even on top of it, you could change the argument's name "test_name" into "test_case", because technically, all those attempts fall under the same test. :wink:

node/src/proxy_server/mod.rs
```diff
@@ -1300,7 +1301,15 @@ impl Hostname {
...
     }
 
+    fn is_running_in_integration_test() -> bool {
+        env::var("MASQ_INTEGRATION_TEST").is_ok()
+    }
+
     fn validate_hostname(&self) -> Result<(), String> {
+        if Self::is_running_in_integration_test() {
+            return Ok(());
+        }
```
bertllll: I showed @utkarshg6 what meant originally. A flag to be set for the ProxyServer after the environment variable is read only once in the entire running laps of the Node. You would do it in `make_and_start_proxy_server`.  You could write a rare, special test which I have myself in the past but I have my own reservations to this technique. In this case, though, it is more meaningful given we need to come up with a way to address this requirement: Writing a test that will prove our Actor is set so that it does something if it's NOT running under the integration test environment. For that, we can only do it in a unit test, strictly, or alternatively, to use more than just one dev-only environment variable where I prefer the first variant if possible. 

The challenge is that in `make_and_start_proxy_server` and all those methods similar to it, it's extremely difficult and sometimes not doable at all to test that the our Actor is being set with proper initial inner state (the values in its fields). While normally we have some luxury of the use of AssertionsMsg we can do it only in the same file with the Actor's declaration. It's because the Actors fields tend to be private (Now I got a strange idea that we could possibly keep all fields public because it makes no big difference in terms of standard visibility. No other code that is outside the file of the declaration will ever be able to access them, because you can do it only through an Addr or Recipient. But that would be probably going too far). 
The more achievable solution stands upon asserting on side effects of handled Actor messages. In our case, we can probably call so some sort of warn or error logging (when we try to request a suspicious IP address, like the localhost).  

`make_and_start_proxy_server` returns a set of recipients, the so called "subs", and you need to identify the one which is meant for sending the Actor message whose handling results in hitting this `validate_hostname` function. Send this message set up so that it will request a route to the localhost. Then just find a way to test that you could find a log complaining about it.

node/src/proxy_server/mod.rs
```diff
@@ -1300,7 +1301,15 @@ impl Hostname {
         Self { hostname }
     }
 
+    fn is_running_in_integration_test() -> bool {
+        env::var("MASQ_INTEGRATION_TEST").is_ok()
+    }
+
     fn validate_hostname(&self) -> Result<(), String> {
```
bertllll: It just struck me that this could be renames as `validate_host_location_is_wan_rather_than_lan`. I'll leave it up to you guys. I thought something along these lines would be quite a nice name.

node/tests/connection_shutdown_test.rs
```diff
@@ -10,11 +10,12 @@ use std::net::{IpAddr, TcpListener, TcpStream};
 use std::net::{Shutdown, SocketAddr};
 use std::str::FromStr;
 use std::time::Duration;
-use std::{io, thread};
+use std::{env, io, thread};
 
 // 'node' below must not be named '_' alone or disappear, or the MASQNode will be immediately reclaimed.
 #[test]
 fn proxy_client_stream_reader_dies_when_client_stream_is_killed_integration() {
+    env::set_var("MASQ_INTEGRATION_TEST", "true");
```
bertllll, line 20: There is a possibility of making use of the guard we have in the test utils, it is placed like this:

```rust
```

to the first line of a test. Its purpose is to keep other tests away from being biased by something this test sets on. When this test ends, it will automatically invalidate all changes made to the environment.

If its visibility allows its use in integration tests (or you consider moving it over to a common place), you might want to give it a try also here.

PR #579, opened 2025-02-11
GH-744-QA-Review-2

Syther007:


node/src/accountant/scanners/mod.rs
```diff
@@ -856,40 +856,42 @@ impl ReceivableScanner {
...
+                    .persistent_configuration
+                    .set_start_block_from_txn(Some(start_block_number), &mut txn)
+                {
+                    Ok(()) => debug!(logger, "Start block updated to {}", start_block_number),
+                    Err(e) => panic!(
+                        "Attempt to set new start block to {} failed due to: {:?}",
+                        start_block_number, e
+                    ),
+                }
             }
```
bertllll, line 890: We discussed that it might be clearer or more illustrative if we had an else clause here resulting in a panic, because we don't expect something different than a value, the specific start block number. We cannot use `.expect()` on our enum, so now the choice should come with either `panic!()` or `unreachable!()`, dependable on your preference.

node/src/actor_system_factory.rs
```diff
@@ -411,6 +412,7 @@ impl ActorFactory for ActorFactoryReal {
         cryptdes: CryptDEPair,
         config: &BootstrapperConfig,
     ) -> ProxyServerSubs {
+        let is_running_in_integration_test = env::var("MASQ_INTEGRATION_TEST").is_ok();
```
bertllll: This is more a formality, but I think we want to be consistent with how we've done this up to now. It seems to me we should require this syntax: `MASQ_INTEGRATION_TEST = true`. In that case, we need to change it here to check:

```rust
if let Ok(value) = env::var("MASQ_INTEGRATION_TEST") {
     value.asci_lower_case() == "true"
}
```
This is more like a sudo code, with me unsure about the Result and the right function to convert a string into all lower case lettering. 

I know this equals to adding some new logic and thus more tests to write, I'm sorry. But those will be just very simple ones. (At least one for the value true to be written in different lettering, and another for checking different sequence of letters than t, r, u, e).

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -577,12 +587,11 @@ mod tests {
...
         scan_range: BlockScanRange,
         block_response: &str,
-        expected_new_start_block: u64,
+        expected_new_start_block: BlockMarker,
         expected_log: &str,
-        test_name: &str,
+        test_case: &str,
     ) {
         init_test_logging();
-        // let test_name = "blockchain_interface_web3_retrieves_transactions";
```
bertllll, line 585: A leftover

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -347,30 +348,29 @@ impl BlockchainInterfaceWeb3 {
...
-        start_block: u64,
+        start_block_marker: BlockMarker,
         scan_range: BlockScanRange,
         response_block_number_result: Result<U64, BlockchainError>,
         logger: &Logger,
     ) -> BlockMarker {
-        let local_end_block_marker = match scan_range {
-            BlockScanRange::NoLimit => BlockMarker::Uninitialized,
-            BlockScanRange::Range(scan_range_number) => {
+        let local_end_block_marker = match (start_block_marker, scan_range) {
```
bertllll: Maybe rather "locally_determined_end_block_marker". Up to you guys. But it's actually calculated or determined, we can't just look out for it and see the value right away.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -347,30 +348,29 @@ impl BlockchainInterfaceWeb3 {
...
+            (BlockMarker::Value(start_block), BlockScanRange::Range(scan_range_number)) => {
                 BlockMarker::Value(start_block + scan_range_number)
             }
+            (_, _) => BlockMarker::Uninitialized,
         };
-
         match response_block_number_result {
```
bertllll: We also can replace the word "response" with more specific "rpc".

node/src/blockchain/blockchain_interface/blockchain_interface_web3/mod.rs
```diff
@@ -1188,7 +1233,7 @@ mod tests {
                 BlockMarker::Uninitialized,
                 &logger
             ),
-            61
+            BlockMarker::Value(61)
         );
         // Case 3: end_block_marker is Uninitialized, highest block in transactions is Uninitialized, start_block_marker is Value
```
bertllll: We said that maybe "no transactions retrieved" or something of that meaning would be better than the middle segment of this and also those other comments where it says "uninitialized".

Issue #601, opened 2025-03-25
Create `sent_payable` table


bertllll:
This new table should eventually replace the `pending_payable` table. However, don't pay attention to this old table yet. It can stay in awhile which will make the integration more gradual, with the burden of complex changes as small as possible at the very beginning.

The schema of this table should probably consist of these columns:

`receiver_address`, `amount_high_b`, `amount_low_b`, `tx_hash`, `timestamp`, `gas_price_wei`, `nonce`, `status`, `retried`   

Possible `status` variants:

bn12345678  - block number of the mined block
p - pending
r - retry; 
f123 - failure with an opcode that can refer to a certain kind of failure (Maybe like f00 - general failure that we cannot interpret better, f01 - pending too long, f02 - transaction disappeared)

Important: Any status starting with f, that is every kind of failure, will be interpreted as a directive for a retry by the `payable_scanner`!   

Possible `retried` variants:
0 - so far no need to retry
1 - retried

Some thoughts on the implementation: 

Every table is always initialized by the `DbInitializer` class. That means, add a new function for this table that will be called once in the program's entire history. We must add a new method to `DbInitializerReal` which calls `rusqlite` code that will be able to generate the new table. It will be using SQL commands (`create table sent_payable (<definitions of the colums>) `). We do the same for each table so there is plenty of examples around. Get also some inspiration for writing tests looking at those already written for the other tables. 

This kind of modification has always required a DatabaseMigration. I think we can afford to do it even under the current circumstances since it shouldn't be so hard regarding this task. Reach the other files with individual migrations (each has its own file). Create a new one and try to guide yourself through by examination of those other migrations.
 
Don't forget to change the `CURRENT_DATABASE_SCHEMA` by one up (maybe it's recommended to start from there before 
you get started working on the migration). 

You don't need to work on an implementation a brand new DAO object for this table or even go modifying an old one. We'll reuse the old `PendingPayableDao` but preferably in a different card.

PR #609, opened 2025-03-28
GH-601: Introduce SentPayable Table

utkarshg6:


node/src/database/db_migrations/migrations/migration_10_to_11.rs
```diff
@@ -0,0 +1,95 @@
...
+        let _ = bring_db_0_back_to_life_and_return_connection(&db_path);
+        let subject = DbInitializerReal::default();
+
+        let result = subject.initialize_to_version(
+            &dir_path,
+            10,
+            DbInitializationConfig::create_or_migrate(make_external_data()),
+        );
+
+        assert!(result.is_ok());
```
bertllll: You can get rid of this line in the detriment of adding an unwrap() after the function. The validation from your concern will be preserved.

If you saw it in other migrations and got mislead by it, I understand, please, I'd be happy if you preferred to revisit those and fix it also there accordingly. :)

node/src/database/db_migrations/migrations/migration_10_to_11.rs
```diff
@@ -0,0 +1,95 @@
...
+            &["tx_hash", "text", "not", "null"],
+            &["receiver_address", "text", "not", "null"],
+            &["amount_high_b", "integer", "not", "null"],
+            &["amount_low_b", "integer", "not", "null"],
+            &["timestamp", "integer", "not", "null"],
+            &["gas_price_wei", "integer", "not", "null"],
+            &["nonce", "integer", "not", "null"],
+            &["status", "text", "not", "null"],
+            &["retried", "integer", "not", "null"],
+        ];
```
bertllll: We discussed that there are usually at least two tests at two different places, one here in the migrations, another for the DbInitializer, where we use exactly the same set of values for an assertion to prove that a certain table in our database is constructed with the correct SQL. I probably thought about it in the past already but when did it again now, I was like it wouldn't be such a bad idea if created a constant first. Maybe with a name like 

```Rust

pub const SQL_ATTRIBUTES_TO_CREATE_SENT_PAYABLE: &[&[&str]] = &[
            &["rowid", "integer", "primary", "key"],
            &["tx_hash", "text", "not", "null"],
            &["receiver_address", "text", "not", "null"],
            &["amount_high_b", "integer", "not", "null"],
            &["amount_low_b", "integer", "not", "null"],
            &["timestamp", "integer", "not", "null"],
            &["gas_price_wei", "integer", "not", "null"],
            &["nonce", "integer", "not", "null"],
            &["status", "text", "not", "null"],
            &["retried", "integer", "not", "null"],
        ];
```
It would work, I guess, if we stored it in a well accessible place, we pointed out, it could be in the `test_utils.rs` in node\src\database directory. However, I think, if we want to do it right, we'd need to write a test which would assert on this constant. And... testing a constant being a test util object is a new thing... I'm not 100% convinced that it's not clumsy a bit.

node/src/database/db_migrations/migrations/mod.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
 
 pub mod migration_0_to_1;
+pub mod migration_10_to_11;
```
bertllll: You could reorder these. Only if you wanted to. I understand why you didn't succeed the first time. You'd probably need to add `#[rustfmt::skip]` above the line.

bertllll (commented):
There are just marginal things left but if you're keen to do it, please fix them. :)

PR #617, opened 2025-04-10
GH-608: Redesign the PendingPayableScanner for `sent_payable table` (manily its DAO)

### LINKED ISSUE #608: Redesign the PendingPayableDAO for `sent_payable` table
The new table `sent_payable` is clearly different from the old `pending_payable` table. 
It will have these extra fields: `receiver_address`, `gas_price`, `nonce`, `status`, `retried`. 
Extra attention should be paid to the `status` column which may change its value in time.

You need to reimplement the methods of this DAO so that they refer to the new `sent_payables` table instead. 
There will be definitely some minor challenges, but in the big picture, it shouldn't be so much hard. Remember to review the names of the methods to fit their purpose now serving to a different table.

**NEW RECORDS**
When a new record for a pending transaction is created, put the status `p` (meant literally) in there, meaning a simple `pending`. 
This status isn't gonna be persistent, on the contrary, it's actually the only one being temporary. We'll always start with this status and will change it to either:

> a) `bnxxxxxxx` ("block number", followed by the respective number of the mined block on the blockchain) 
> b) `fxxx` ("failure", followed by a specific opcode that can be used to decode the given kind of failure). 

`p` should never stay in the record longer than what will be the time limit for a transaction to qualify as pending unbearably long, also resulting in another failure `fxxx`.
  
**PICKING RECORDS TO CHECK THEM**
There will be one important difference in the way we will do the selection of those past transactions that need to be queried about by an RPC. Up to this point, we've always concentrated on the transactions considered pending at the very moment (in the new design, those are represented by the status `p`). However, it's not enough. With some lower possibility, but still real, we could be exposed to a moment when we'd believe that transactions that we just claimed as taking too long could actually finish before we retry with transaction bearing the same nonce. It could cause that transactions left behind with a failure status led to creating its younger springs which, however, will never be able to complete, because they would always just repeat the same nonce which has been taken once and will not be approved again. Eventually, this would probably make the whole payment mechanics collapse. To avoid an endless cycle of terribly wrong resubmitting attempts, we need to take a short look back and recheck transactions that were marked with the failure identified as "tx pending too long" in the previous cycle, and so include them in the set of transactions of which we will ask the blockchain service provider. 
If we receive a response that the rechecked transactions (with the older hashes) have actually completed (been stuck in a block), we will correct our records which requires:

> a) Rewrite `fxxx` to `bnxxx` to express its completion.  
> b) Delete the `1` from the `retried` column, or writing something like `101` or some other code telling us later that this tx failure has been reverted 
> c) Rewrite status of the retrying transaction with a special failure type `fxxx` which will also help to explain the history of events if anybody is interested in tracking it down 

**UPDATING RECORD STATUSES AFTER SCAN**
Recently, the interface consists of methods like `increment_scan_attempts` and `mark_failures`. We will want to dump one and modify another. `mark_failures` should probably become something like `update_status` which will turn to be one of the most important methods. When we receive the data from the blockchain we will evaluate it which should end up with a determined status that will describe how each record stands after one scanning cycle. Having the most suitable status at hand, we will want to write it in the respective column in the table and that's the crucial moment which requires this `update_status` method. 
Most of the time, the writing will go only into the column labeled `status`, however, it's possible that the same method will also write into the `retried` column (or we will get an extra method for that purpose, alternatively). This dual writing should happen for transactions believed pending too long. Such will:

> a) receive a failure status code fxxx
> 
> b) value 1 will be writen into the `retried` column (this column is meant for conveying the urgency of a retry to the PayableScanner running immediately after).  

Probably redesign the struct `PendingPayableFingerprint` into `TransactionRecord`. It may be used anytime when we will read from the `sent_payables` table. For inserting a new record that belongs to a fresh pending transaction we may want to establish a different structure `SentTransaction`. 

```Rust

struct TransactionRecord {
   transaction: Transaction,
   status: TransactionStatus
}

struct Transaction {
   receiver_address: Address,
   amount: u128,
   tx_hash: String,
   timestamp: SystemTime,
   gas_price_wei: u64,
   nonce: u32,
}

pub trait SentPayablesDao {
//TODO maybe search by wallet addresses instead as they are shorter than hashes (remember, both are stored as `text` in the database)
    fn transaction_rowids(&self, hashes: &[H256]) -> TransactionHashes;
    fn return_all_pending_transactions(&self) -> Vec<TransactionRecord>;
    fn return_all_transactions_to_retry(&self) -> Vec<TransactionRecord> 
    fn insert_new_records(
        &self,
        transactions: Vec<Transactions>,
    ) -> Result<(), SentPayableDaoError>;
    fn delete_records(&self, ids: &[u64]) -> Result<(), SentPayableDaoError>;
    fn change_statuses(&self, ids: &[StatusChange]) -> Result<(), SentPayableDaoError>;
}

struct StatusChange {
    rowid: u64,
    new_status: TransactionStatus
}

```
Side note:
Change the struct `TransactionHashes` to maybe `TransactionRowids`.

utkarshg6:


node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+pub enum SentPayableDaoError {
+    EmptyInput,
+    NoChange,
+    InvalidInput(String),
+    PartialExecution(String),
+    SqlExecutionFailed(String),
+}
+
+type TxHash = H256;
+type RowID = u64;
```
bertllll: You thought this should be written as RowId.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+        if !self.get_tx_identifiers(&unique_hashes).is_empty() {
+            return Err(SentPayableDaoError::InvalidInput(
+                "Input hash is already present in the database".to_string(),
+            ));
+        }
+
+        let sql = format!(
+            "INSERT INTO sent_payable (\
+            tx_hash, receiver_address, amount_high_b, amount_low_b, \
+            timestamp, gas_price_wei, nonce, status
```
bertllll: If you want we could have an indentation in front of these lines with the column names.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+
+        let mut stmt = self
+            .conn
+            .prepare(&sql)
+            .expect("Failed to prepare SQL statement");
+
+        stmt.query_map([], |row| {
+            let tx_hash_str: String = row.get(0).expectv("tx_hash");
+            let hash = H256::from_str(&tx_hash_str[2..]).expect("Failed to parse H256");
+            let receiver_address_str: String = row.get(1).expectv("row_id");
```
bertllll: It should be "receivable_address".

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+        Ok(())
+    }
+
+    fn delete_records(&self, hashes: &HashSet<TxHash>) -> Result<(), SentPayableDaoError> {
+        if hashes.is_empty() {
+            return Err(SentPayableDaoError::EmptyInput);
+        }
+
+        let hash_strings: Vec<String> = hashes.iter().map(|h| format!("'{:?}'", h)).collect();
+        let hash_list = hash_strings.join(", ");
```
bertllll: You could reuse this utility comma_joined_stringifiable instead of the two lines but I understand if you don't.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+    #[test]
+    fn insert_new_records_returns_err_if_partially_executed() {
+        let setup_conn = Connection::open_in_memory().unwrap();
+        // Inject a deliberately failing statement into the mocked connection.
+        let failing_stmt = {
+            setup_conn
+                .execute("CREATE TABLE example (id integer)", [])
+                .unwrap();
+            setup_conn.prepare("SELECT id FROM example").unwrap()
+        };
```
bertllll: I think it would be less confusing if the first statement was made clear to be a correct one, simply returning zero results which is completely valid. 

This is the solution containing the whole setup.  
```Rust 
        setup_conn
                .execute("CREATE TABLE example (id integer)", [])
                .unwrap();
        let first_stmt_appropriate_for_a_check = setup_conn.prepare("SELECT id FROM example").unwrap();
        let second_stmt_invalid_for_insertion_itself = {
            setup_conn.prepare("SELECT id FROM example").unwrap()
        };
```

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+    fn get_tx_identifiers_works() {
+        let home_dir =
+            ensure_node_home_directory_exists("sent_payable_dao", "get_tx_identifiers_works");
+        let wrapped_conn = DbInitializerReal::default()
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let subject = SentPayableDaoReal::new(wrapped_conn);
+        let hash1 = H256::from_low_u64_le(1);
+        let hash2 = H256::from_low_u64_le(2);
+        let hash3 = H256::from_low_u64_le(3); // not present in the database
```
bertllll: I just talked about making the test stronger by choosing the hash number two as the one missing in this scenario. It would prove that when we bumped into one it did not stop the algorithm and we could continue in picking the remaining txs also being in the database.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+    pub receiver_address: Address,
+    pub amount: u128,
+    pub timestamp: i64,
+    pub gas_price_wei: u64,
+    pub nonce: u32,
+    pub status: TxStatus,
+}
+
+pub enum RetrieveCondition {
+    IsPending,
```
bertllll, line 41: The pending portion is the most important one, I agree, but we also need to include those "pending too long failures" from the previous run. 

It's a bit less trivial though...

We agreed that creating a new table to keep all the failured txs in would be a good idea. It should be put on a different card.

Once we have it we need to remember to check in the table for this very specific error, we'll decide to claim on our own consensus that each cycle is enough for a transaction to finish and otherwise we flip it into a failure this way. It needs to be properly backed up so that we know we don't miss a tx that has finished before we attempted with the new tx meant for a retry. 

It's a niche case but I consider it important one.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+            .timestamp(old_timestamp)
+            .status(TxStatus::Failed)
+            .build();
+        subject
+            .insert_new_records(&vec![tx1, tx2, tx3.clone()])
+            .unwrap();
+
+        let result = subject.retrieve_txs(Some(RetrieveCondition::ToRetry));
+
+        assert_eq!(result, vec![tx3]);
```
bertllll: I usually prefer to express the fact the returned values can be multiple rather than always just one. 
I know it's annoying but in terms of the inner functionality things really can be designed differently (either way), so I like to be more specific, especially for the case when somebody changed the code and the behavior from one to another and the test would not detect it or felt indifferent.  

It's makes my heart sink though because there are more than just one test I could comment on like this. I hate adding this extra work for you. So just think about it. You will how much energy you have left. :joy:

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+            .build();
+        subject
+            .insert_new_records(&vec![tx1.clone(), tx2.clone()])
+            .unwrap();
+
+        let result = subject.retrieve_txs(Some(ByHash(tx1.hash)));
+
+        assert_eq!(result, vec![tx1]);
+    }
+
```
bertllll, line 646: I think we also should have a separate test for the case when a None is supplied. It deserves its own test, not just to be included in other tests with an aim on a different function.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+        let wrapped_conn = DbInitializerReal::default()
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let subject = SentPayableDaoReal::new(wrapped_conn);
+        let existent_hash = H256::from_low_u64_le(1);
+        let tx = TxBuilder::default()
+            .hash(existent_hash)
+            .status(TxStatus::Pending)
+            .build();
+        subject.insert_new_records(&vec![tx.clone()]).unwrap();
```
bertllll: This clone is redundant. :rofl:

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+                .unwrap();
+        }
+        let read_only_conn = Connection::open_with_flags(
+            home_dir.join(DATABASE_FILE),
+            OpenFlags::SQLITE_OPEN_READ_ONLY,
+        )
+        .unwrap();
+        let wrapped_conn = ConnectionWrapperReal::new(read_only_conn);
+        let subject = SentPayableDaoReal::new(Box::new(wrapped_conn));
+
```
bertllll: If you don't mind..please...you know what I mean.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+        let result = subject.delete_records(&hashset);
+
+        let remaining_records = subject.retrieve_txs(None);
+        assert_eq!(result, Ok(()));
+        assert_eq!(remaining_records, vec![tx3]);
+    }
+
+    #[test]
+    fn delete_records_returns_error_when_input_records_are_invalid() {
```
bertllll: maybe rather "when_input_is_empty" or "when_no_input"

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -0,0 +1,844 @@
...
+        assert_eq!(
+            result,
+            Err(SentPayableDaoError::PartialExecution(
+                "Only 1 of the 2 hashes has been deleted.".to_string()
+            ))
+        );
+    }
+
+    #[test]
+    fn delete_records_returns_deletion_failed_error_when_an_error_occurs_in_sql() {
```
bertllll: maybe "delete_records_returns_a_general_error_from_sql"

Please make it match with the name used in ensure_node_home_directory_exists in this test.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -25,6 +27,50 @@ pub enum TxStatus {
     Succeeded(TransactionBlock),
 }
 
+impl FromStr for TxStatus {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Pending" => Ok(TxStatus::Pending),
+            "Failed" => Ok(TxStatus::Failed),
```
bertllll: Add a TODO mentioning the new card we talked about, for failed txs. 

Also remember that the new table should store also some information which will allow us to tell the failures apart by their ranking. We will definitely want to know as much as we can read out regarding the possible cause. (I'm writing it here because I was imagining it even in this table and as I see you did not go into this I just want to emphasize how important it is. Here I would've been expecting FailureMark<AndItsIdentification>, but we moved away from there and now the plans are different.)

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -612,6 +658,74 @@ mod tests {
...
+            TxStatus::from_str("Succeeded(123,0xinvalidhash)"),
+            Err("Invalid block hash".to_string())
+        );
+
+        // Test unknown status
+        assert_eq!(
+            TxStatus::from_str("InProgress"),
+            Err("Unknown status: InProgress".to_string())
+        );
+    }
```
bertllll, line 727: Since these are basically a set of alone-assertions I'd suggest to delete all the blank lines, I think it would work as well as this and be more concise.

Please consider that at the both of these tests above. 

You know what... do as you please to. I don't have a strong opinion on this. You can easily leave it this way. Cheers.

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: I've created this card - https://github.com/orgs/MASQ-Project/projects/8/views/1?filterQuery=59&pane=issue&itemId=3051162400&issue=MASQ-Project%7CNode%7C631

bertllll (approved):

PR #630, opened 2025-05-08
Gh v2 580b czarte and Dan

czarte:


.githooks/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+  exit 1
+fi
+
+# Get remote file size
+REMOTE_SIZE=$(git cat-file -s "$REMOTE_BLOB")
+
+# Get local file size (staged version, not working copy)
+LOCAL_SIZE=$(git ls-files --stage | grep "$FILENAME" | awk '{print $2}' | xargs git cat-file -s)
+
+if [ "$LOCAL_SIZE" != "$REMOTE_SIZE" ]; then
```
dnwiebe: Now that you have that `("NOEX", "Nonexistent")` country in the small file, you could find out whether the local `dbip_country.rs` file is the small one or the generated one by `grep`ping through it for that string.

Make sure the error message explains what's being searched for, though, so that the developer doesn't spend lots of time trying to figure out what's going on if he's making a change you don't expect.

.github/workflows/ci-matrix.yml
```diff
@@ -6,6 +6,9 @@ on:
     branches:
       - master
 
+env:
```
dnwiebe: Pretty sure this isn't needed. It doesn't hurt anything, but according to GitHub, `GITHUB_ACTIONS` is always set by the infrastructure in an Actions builder.

.github/workflows/dbip_download.yml
```diff
@@ -0,0 +1,63 @@
+name: Download DBIP data and generate dbip_country.rs
+
+on:
+  workflow_dispatch:
+#  schedule:
+#    - cron: "0 0 3 * *" # Runs at midnight on the 3rd of every month
+
+env:
+  SKIP_EXPORT: 'false'
```
dnwiebe: There's something you have to specifically set in a `.yml` file (I forget what it is) to make Actions keep going after an error. Normally, any error causes an abort. Therefore, I don't think this is necessary.

.github/workflows/dbip_download.yml
```diff
@@ -0,0 +1,63 @@
...
+
+permissions:
+  contents: write
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
dnwiebe: All the cool kids are using v4.

.github/workflows/dbip_download.yml
```diff
@@ -0,0 +1,63 @@
...
+          gunzip dbip-data/dbip-country-lite.mmdb.gz
+          if [ "$?" -ne "0" ]; then 
+            echo "SKIP_EXPORT=true" >> $GITHUB_ENV 
+          fi
+
+      - name: Generate Rust source file
+        if: ${{ env.SKIP_EXPORT == 'false' }}
+        run: |
+          cd ip_country
+          mkdir -p generated
```
dnwiebe: No longer used?

.github/workflows/dbip_download.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -B generated-source
+          git rm -rf .
+          mkdir ip_country/src
+          mv "${TEMP_DIR}"/dbip_country.rs ip_country/src/dbip_country.rs
+          git add ip_country/src/dbip_country.rs
+          git commit -m "Update generated dbip_country ${YEAR_MONTH} Rust source file"
+          git push origin generated-source --force
```
dnwiebe: `--force` seems a little violent, although I understand that there are no commits to conflict with yet. If we have to use `--force`, then let's go ahead and use it; but when I push the first commit to a new branch, I use `git push -u origin HEAD`. No `--force`. I wonder if that would work here.

ci/all.sh
```diff
@@ -1,12 +1,23 @@
 #!/bin/bash -xev
 # Copyright (c) 2019, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
 CI_DIR="$( cd "$( dirname "$0" )" && pwd )"
-PARENT_DIR="$1"
+export ARGV_1="$1"
 
-ci/format.sh
+#ci/format.sh
```
dnwiebe: Uncomment, please.

ci/all.sh
```diff
@@ -1,12 +1,23 @@
...
 
 export RUST_BACKTRACE=1
-
+#if github checkout fails, you can download csv database from db-ip.com and convert it locally by cargo run ip_country
+if [[ "$GITHUB_ACTIONS" == "true" || "$ARGV_1" == "dbip" ]]; then
```
dnwiebe: Modify the following code to be less obsolete.

ci/all.sh
```diff
@@ -1,12 +1,23 @@
...
+  ls
+  git remote add experiment https://github.com/MASQ-Project/DBIP-Experiment.git
+  git fetch experiment
+  git checkout experiment/generated-source -- ip_country/src/dbip_country.rs
+  git remote remove experiment
+  echo "chekout $ARGV_1"
+fi
+if [[ "$GITHUB_ACTIONS" != "true" && "$ARGV_1" == "dbip" ]]; then
+  bash ./install-hooks.sh
```
dnwiebe, line 15: Don't we want to install hooks every time, as long as we're not in Actions? I think the `&& "ARGV_1" == "dbip"` part should go away.

ci/all.sh
```diff
@@ -42,13 +53,15 @@ echo "**************************************************************************
...
 echo "*********************************************************************************************************"
 echo "*********************************************************************************************************"
 echo "***                                           IP COUNTRY HEAD                                         ***"
 cd "$CI_DIR/../ip_country"
-ci/all.sh "$PARENT_DIR"
+ci/all.sh
 echo "***                                           IP COUNTRY TAIL                                         ***"
 echo "*********************************************************************************************************"
-
+if [[ "$GITHUB_ACTIONS" == "" && "$ARGV_1" != "dbip" ]]; then
```
dnwiebe: I think the new commit hook will take care of guarding against this; I don't think you need this here anymore.

ip_country/ci/download_dbip.sh
dnwiebe, line 1: This can go away. Makes me a little sad, though.

ip_country/data/bak_digest_dbip_country.rs
dnwiebe, line 1: Probably this shouldn't have been checked in. Might want to `git rm`.

ip_country/src/countries.rs
```diff
@@ -1,329 +1,227 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Countries {
+    countries: Vec<Country>,
+    index_by_iso3166: HashMap<String, usize>,
 }
 
-impl TryFrom<&str> for Country {
-    type Error = String;
+impl Countries {
+    pub fn old_new(countries: Vec<Country>) -> Self {
```
dnwiebe: Maybe this can go away. The intention was for it to go away: that's why it's named old_new. If there's something still using it, that thing is probably deprecatable.

ip_country/src/countries.rs
```diff
@@ -1,329 +1,227 @@
...
+            .collect::<HashMap<String, usize>>();
+        Self {
+            countries,
+            index_by_iso3166,
+        }
+    }
 
-    fn try_from(iso3166: &str) -> Result<Self, Self::Error> {
-        let index = match INDEX_BY_ISO3166.get(&iso3166.to_ascii_uppercase()) {
+    pub fn new(mut country_pairs: Vec<(String, String)>) -> Self {
```
dnwiebe, line 27: Vojtek starts review here...

ip_country/src/countries.rs
```diff
@@ -1,329 +1,227 @@
...
+        }
+        country_pairs.sort_by(|a, b| a.0.cmp(&b.0));
+        let mut countries = country_pairs
+            .iter()
+            .enumerate()
+            .map(|(index, (iso3166, name))| Country::new(index + 1, iso3166, name))
+            .collect::<Vec<Country>>();
+        let sentinel = Country::new(0, "ZZ", "Sentinel");
+        countries.insert(0, sentinel);
+        Self::old_new(countries)
```
czarte: Maybe we can get rid of old_new and move its code in here

ip_country/src/countries.rs
```diff
@@ -1,329 +1,227 @@
...
             panic!(
                 "Data error: ISO3166 {} maps to index {}, but there is no such Country",
                 iso3166, index
             )
         });
         Ok(country)
     }
-}
 
-impl From<usize> for Country {
```
czarte, line 295: Maybe we can also implement TryFrom, because we can fail, and From, because we can want to handle Error message by ourselves. But consider if this is a reasonable effort.

multinode_integration_tests/ci/all.sh
```diff
@@ -14,11 +14,6 @@ export RUST_BACKTRACE=full
 # executable they need will not be in your filesystem but in your (and their) parent's filesystem.  If that's the
 # case, make sure you pass in as parameter 1 the path to the directory just above the 'node' module directory,
 # IN THE CONTEXT OF THE PARENT (host) FILESYSTEM.
```
dnwiebe: You could get rid of these comments too. They don't comment on anything anymore.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2104,10 +2104,7 @@ mod tests {
 
         let result_introducer: &NodeRecord =
             dest_db.node_by_key(&agrs[0].inner.public_key).unwrap();
-        let mut expected_introducer = NodeRecord::from(&agrs[0]);
-        expected_introducer.metadata.last_update = result_introducer.metadata.last_update;
-        expected_introducer.metadata.country_undesirability = COUNTRY_UNDESIRABILITY_FACTOR;
-        expected_introducer.resign();
+        let expected_introducer = NodeRecord::from(&agrs[0]);
```
dnwiebe: I'm curious about why the `last_update` and `country_undesirability` changes to `expected_introducer` are no longer required to make it line up with `result_introducer`.

node/src/neighborhood/gossip_acceptor.rs
```diff
@@ -2440,26 +2437,37 @@ mod tests {
         let src_root = make_node_record(1234, true);
         let dest_root = make_node_record(2345, true);
         let mut src_db = db_from_node(&src_root);
-        let node_a_fr = make_node_record_cc(5678, true, "FR");
-        let node_b_us = make_node_record_cc(4567, true, "US");
+        let node_a = make_node_record(5678, true);
+        let country_zero = node_a.inner.country_code_opt.clone().unwrap();
```
dnwiebe, line 2443: I think there's a way you can iterate on `COUNTRIES` to get codes instead of IP addresses; then you should be able to iterate until you find a country code that's not `country_zero` or `ZZ` or `NOEX`.

node/src/neighborhood/node_record.rs
```diff
@@ -402,6 +402,7 @@ mod tests {
         let mut expected_node_record = make_node_record(1234, true);
         expected_node_record.set_version(6);
         expected_node_record.resign();
+        let expexted_cc = expected_node_record.inner.country_code_opt.clone().unwrap();
```
dnwiebe: `expexted_cc` is misspelled

ip_country/src/countries.rs
dnwiebe: Done.

ip_country/src/countries.rs
dnwiebe: Done.

dnwiebe (approved):

PR #632, opened 2025-05-09
GH-608: Review 1

utkarshg6:


node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -395,22 +400,14 @@ mod tests {
...
-        };
+        setup_conn
+            .execute("CREATE TABLE example (id integer)", [])
+            .unwrap();
+        let pre_insert_stmt = setup_conn.prepare("SELECT id FROM example").unwrap();
+        let faulty_insert_stmt = { setup_conn.prepare("SELECT id FROM example").unwrap() };
         let wrapped_conn = ConnectionWrapperMock::default()
-            .prepare_result(Ok(failing_stmt))
-            .prepare_result(Ok(failing_stmt_2));
+            .prepare_result(Ok(pre_insert_stmt))
```
bertllll: Maybe it would be better if you just say what's the truth, that it is simply a SELECT statement. Or "meant_select_stmt", it's up to you. But the mention of "insert" in pre_insert_stm had me stumped. Then you could call the second one as "badly_working_insert_stm" or just leave in yours.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -462,19 +459,83 @@ mod tests {
...
+        subject
+            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone()])
+            .unwrap();
+        subject
+            .insert_new_records(&vec![tx4.clone(), tx5.clone()])
+            .unwrap();
+
+        let result = subject.retrieve_txs(None);
+
+        assert_eq!(result.len(), 5);
```
bertllll: When you do assertions where one asserts on the exact values and another only the count of the values in the set, the first one is included in the second one. So here you can simply delete this line. 

Also, there are tests where it actually makes sense to ask about the count of elements, but again, if you examine the values, especially one by one (not all of them in a single assertion against a vector), then you may want to remember it's always nicer if you put this very simple assertion, that answers the count only, to the very end. Because the other assertions may provide more insight in what's wrong with the data of the element or elements.
Alternatively, you can leave this way, but then I'd recommend to add a debug message:

```Rust 

assertion_eq!(result.len(), 5, "The result contains: {:?}", result);

```

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -540,23 +601,22 @@ mod tests {
             .timestamp(old_timestamp)
             .status(TxStatus::Failed)
             .build();
+        let tx4 = TxBuilder::default() // this should be picked for retry
```
bertllll, line 604: This comment will become misleading once we go off and implement the other table. The retry/not-retry decision will depend purely on what will be there, not this table. 

I just worry that you will forget to modify everything in the process of clean-up when you need to transfer some of the functions formerly planned for this table over to the other one. I trust your vow that you will go taking care of this in or after the next card is done, but there might be too many little things referring back to this state and you may miss some. It makes me a little nervous so please work with care.   

But I see now that this whole test will go away no matter what.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -730,32 +788,36 @@ mod tests {
...
                 block_hash: Default::default(),
                 block_number: Default::default(),
             }))
             .build();
         subject
-            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone()])
+            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone(), tx4.clone()])
             .unwrap();
-        let hashset = HashSet::from([tx1.hash, tx2.hash]);
+        let hashset = HashSet::from([tx1.hash, tx3.hash]);
```
bertllll, line 807: When I come to think of this HashSet, I may not like it. Look, what is the benefit? Even if there are duplicates which might be fought by the nature of the HashSet we might want to panic on such an event. Hmmm...
I'm just pondering here, I have the feeling that you actually implemented that for this very reason. Then ignore my youthful senility.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
```diff
@@ -33,7 +33,7 @@ impl FromStr for TxStatus {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "Pending" => Ok(TxStatus::Pending),
-            "Failed" => Ok(TxStatus::Failed),
+            "Failed" => Ok(TxStatus::Failed), // TODO: GH-631: This should be changed to "Retry"
```
bertllll: It may not actually...

Depends on your future decisions. 
It would be technically possible having just two variants with Pending and Suceeded. Because we could just take the failed transaction out and carry over to the failed_payables table. It's streamline, and I kinda like it that way. The only minus is that if the Node crashes at that time, we are short of this record, nothing, that's what we'll be left. If we carried that over and then deleted, we'd defend it, because then it would be stored in one of those two tables at any time. You know, this is actually an advance risk defense. I'd like to have something like that (and all wholes patched), but it's darn hard. I've dealt with it since the beginning when I started working on the payment-confirmation process a couple of years ago, maybe before you joined the team. I remember myself desperate from these unprotected (unguaranteed respectively) data transfers. There's even a card from back then in which I designed a data wrapper which would know how and where to write the data in at the persistent database if the Node is crashing, making use of the mighty Drop impl.

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: It's for the same reason, In case vector seems more convenient, I'll replace it and will add an input error check.

node/src/blockchain/blockchain_interface/blockchain_interface_web3/lower_level_interface_web3.rs
utkarshg6: This was a mistake from my side. I forgot to change this.

Issue #631, opened 2025-05-09
Create `FailedPayables` Table


utkarshg6:
We need to store all transactions generated by us that failed for various reasons, so we can track them for future use.

Keeping these transactions in the `SentPayables` table is not ideal. Instead, we should create a separate table for them.

We also need to update the `SentPayables` table to ensure there is no state called `Failed`; it should be renamed to `Retry`.

For the `FailedPayables` table, we will use the following structure:

```rust
create table if not exists failed_payable (
                rowid integer primary key,
                tx_hash text not null,
                receiver_address text not null,
                amount_high_b integer not null,
                amount_low_b integer not null,
                timestamp integer not null,
                gas_price_wei integer not null,
                nonce integer not null,
                failure_code text not null
            )
```

This table should support all `CRUD` operations.

Please include this in the same migration as the `SentPayables` table (at the time of writing this card, the migration was from version 10 to version 11).

Before starting on this card, be sure to read the comments.

dnwiebe:
From what I hear, a gas price lower than one gwei is conceivable, as is a gas price high enough that it can't be expressed as wei in 64 bits. Therefore, we probably need a two-column 127-bit value for `gas_price_wei`.

Unless! If we're never going to have to use the`gas_price_wei` (or, for that matter, the `amount_xxx` columns) in a `WHERE` clause, then they don't have to be represented as numeric data types at all. We can represent them somehow as strings (perhaps a string representation of a long decimal number, perhaps a sixteen-byte (128-bit) value encoded as a Base64 string, maybe in some other way) and store them as `TEXT` columns to avoid bringing the complicated two-column logic to bear.

dnwiebe:
Should there be some kind of status information in the row as well? For example, whether the failure is in the process of being addressed, or perhaps a retries-remaining count, or an enum with Unaddressed/InProcess/Fixed/Unfixed/Unfixable values, or whatever.

dnwiebe:
Think about what happens at end-of-life for a record here. Transaction information in the database is a security risk. At the very least, failures that have been resolved must be removed from this table. For more security, the sensitive parts of a transaction could be encrypted with the same logic used to encrypt some `CONFIG`-table entries.

If it's decided that some of the columns should be encrypted, anything encrypted will need to be of `TEXT` type; the encryption logic doesn't produce integers, even from integers.

utkarshg6:
Solutions for the above comments respectively:

[...]

PR #636, opened 2025-05-20
GH-631: Create `FailedPayables` Table (Old)

utkarshg6:


node/src/database/db_initializer.rs
```diff
@@ -283,6 +286,32 @@ impl DbInitializerReal {
...
+        conn.execute(
+            "create table if not exists failed_payable (
+                rowid integer primary key,
+                tx_hash text not null,
+                receiver_address text not null,
+                amount_high_b integer not null,
+                amount_low_b integer not null,
+                timestamp integer not null,
+                gas_price_wei_high_b integer not null,
+                gas_price_wei_low_b integer not null,
```
bertllll, line 299: This is problematic...

Last time when we talked, I think you expressed a lack of desire to include the gas price from the failed transaction into the computation of the gas price for their successor transactions (or replacement txs). You said we should always just take the recommended value fetched through the RPC and amplify it by some percents. 
Therefore, if you are certain that we cannot benefit from this information, it looks to me like we don't have to split it into two columns, because it could be simply represented by text. 

Maybe you really should change it to 
`gas_price_wei text not null` .

But only if you believe we won't need to process it except some debugging with a possible print.

node/src/database/db_initializer.rs
```diff
@@ -283,6 +286,32 @@ impl DbInitializerReal {
...
+                nonce integer not null,
+                failure_reason text not null,
```
bertllll: This may cause some clumsiness, if we have to pick words to describe it. I believe that using a numeric code with a legend somewhere else would be better. 
Especially because we will work with this data, applying a condition to identify what kind of error it is. If we only ask what number it is, it would become smoother.
Of course, the problem is, where would maintain the legend of these error codes. That I don't know an answer to at the moment. But I want to encourage you to think about it on your own despite the repulsion you may hold.

node/src/database/db_initializer.rs
```diff
@@ -283,6 +286,32 @@ impl DbInitializerReal {
...
+                failure_checked integer not null
```
bertllll: This name choice isn't ideal. Keep in mind that this column is not going to refer to any kind of error in general.
It speaks of a very specific case, only the single error. It is "pending too long".  

Don't fear to pick something heavier. It could even be like:
`last_pending_too_long_checked`.

Hopefully, you may even be able to come up with a better one.

node/src/database/db_initializer.rs
```diff
@@ -283,6 +286,32 @@ impl DbInitializerReal {
...
+            )",
```
bertllll, line 303: Also, and this is related rather to your other card, about the DAO of this table, I've figured out that even thought 
my idea about caching the <pending too long failures> in a hashmap for saving sparing us the hassle with the db, I do think we will need a special method to pull that information from the table. We aren't going to call it often, actually only once in the Node's life span, at the very beginning. Just to make sure that we haven't missed some which have succeeded over the time when the Node is down. 

I'd like you to introduce a method allowing this:

```
   SELECT 
   	    tx_hash 
   FROM
            failed_pyables
   WHERE         
	    failure_reason = <some representation of "pending too long"> AND
	    last_pending_too_long_checked = 0
``` 

It does not need to return only the hash, but even that alone would be fine for us to be able to check on that transaction :man_shrugging:  It's up to you what structure it be. I guess you've already established some standard data set from which and into which you write in one or another direction, so you will keep on even here.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -20,37 +20,33 @@ pub enum SentPayableDaoError {
...
 
-type TxIdentifiers = HashMap<TxHash, RowId>;
-type TxUpdates = HashMap<TxHash, TxStatus>;
+type TxIdentifiers = HashMap<Hash, RowId>;
+type TxUpdates = HashMap<Hash, TransactionBlock>;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Tx {
-    pub hash: TxHash,
+    pub hash: Hash,
```
bertllll: I've gone through this PR and I kinda grew fond of `TxHash`. Not sure why you gave that up. Was it me suggesting that?

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -574,166 +555,146 @@ mod tests {
...
-            .build();
-        let tx2 = TxBuilder::default()
-            .hash(H256::from_low_u64_le(2))
-            .status(TxStatus::Failed)
-            .build();
-        subject
-            .insert_new_records(&vec![tx1.clone(), tx2.clone()])
+        // Insert a record with block_hash but no block_number
+        {
+            let sql = "INSERT INTO sent_payable (tx_hash, receiver_address, amount_high_b, amount_low_b, timestamp, gas_price_wei_high_b, gas_price_wei_low_b, nonce, block_hash, block_number)
```
bertllll: If you could break this line into more lines it would be nice.

node/src/database/db_migrations/migrations/migration_10_to_11.rs
```diff
@@ -9,19 +9,38 @@ impl DatabaseMigration for Migrate_10_to_11 {
...
+                gas_price_wei_high_b integer not null,
+                gas_price_wei_low_b integer not null,
+                nonce integer not null,
+                failure_reason text not null,
+                failure_checked integer not null
+            )";
+
+        declaration_utils.execute_upon_transaction(&[
+            &sql_statement_for_sent_payable,
+            &sql_statement_for_failed_payable,
```
bertllll, line 42: Cool

node/src/database/test_utils/mod.rs
```diff
@@ -129,23 +145,46 @@ impl DbInitializerMock {
 
 #[cfg(test)]
 mod tests {
-    use crate::database::test_utils::SQL_ATTRIBUTES_FOR_CREATING_SENT_PAYABLE;
+    use crate::database::test_utils::{
+        SQL_ATTRIBUTES_FOR_CREATING_FAILED_PAYABLE, SQL_ATTRIBUTES_FOR_CREATING_SENT_PAYABLE,
+    };
 
     #[test]
     fn constants_have_correct_values() {
```
bertllll: Hmmmmm. Honestly, testing test code like this is really unusual. 

I have a feeling that I would've never done it myself. And I'd like to express this feeling and hope that it could change your mind.

It resemble a situation where somebody would write an assertion with some expected data and then another test that tests that the expected data are right. Now think about all those hundreds tests we have and try to tell yourself if we do such things.   

The existence of these constants is already a preserved state of the desired result.
If somebody changes it, the test will fail. And the cause will be traced down. If somebody changes both the table itself, plus this constant, and they matches then...it seems like legitimate modifications. 
So you've written a test that test the programmer writing the test. You say "I don't trust us programmers to know what we do".

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -502,39 +515,20 @@ mod tests {
...
-        let tx4 = TxBuilder::default()
-            .hash(H256::from_low_u64_le(4))
-            .status(TxStatus::Pending)
-            .build();
-        let tx5 = TxBuilder::default()
-            .hash(H256::from_low_u64_le(5))
-            .status(TxStatus::Failed)
+            .block(Default::default())
             .build();
+        let tx3 = TxBuilder::default().hash(H256::from_low_u64_le(3)).build();
```
bertllll: Please change the places where you construct the H256 so that it uses our utility function `make_tx_hash()`.
I believe that the big amount of lines where you do this can be substituted that way. 

Here, particularly, it would be 
```Rust
make_tx_hash(3).
```

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: It's because it can now be used to represent either Block's hash or transaction's hash, that's why I made this change to make it generic.

node/src/database/test_utils/mod.rs
utkarshg6: Ok!

node/src/database/db_initializer.rs
utkarshg6: We should use the old value only to check if the new value is greater; otherwise, the old value remains valid. Additionally, we need to fetch the latest gas price to determine the optimal value instead of relying solely on historical data. 

The sequence of events should be:
1. Fetch the latest gas price.
2. Compare it with the previous price for that nonce.
3. If the latest price is higher, use it; otherwise, increment the previous price to make the old transaction obsolete.
4. Calculate the new price by increasing it by a percentage if necessary due to batching.

node/src/database/db_initializer.rs
utkarshg6: After much thought and research, I realized that using text instead of error codes is better, as it offers improved readability and debugging. Plus, since it will be converted from an enum, it allows for flexibility in introducing new ones.

node/src/database/db_initializer.rs
utkarshg6: If that's the case, we should eliminate this column and instead introduce different failure reasons, updating them as needed.

For Example, `PendingTimeoutStage1` and `PendingTimeoutStage2`.

node/src/database/db_initializer.rs
utkarshg6: Sure, that can be easily done. I'll keep it in mind.

node/src/database/db_initializer.rs
bertllll: Good point. I like it. 

I just don't understand fully how the fourth step relates to the previous. :thinking:  Could you please elaborate?

node/src/database/db_initializer.rs
bertllll: This baffles me. 

Regarding your suggestion, it really isn't gonna be like you described. It's gonna always be just the same error, but only checked or unchecked. Saying that the failure reason shifts from one to "another" isn't accurate. 

By the way, is this what you mean by the text representation? A camel-case string like these? That looks better than what I expected. I worried that you wanted to write a whole cause via a sentence with spaces between words. HaHa.    

I'm gonna keep on thinking about a solution for this. If you have some before me, we'll go with yours.

node/src/database/db_initializer.rs
bertllll: Okay, I'm not gonna pull you away from that direction

node/src/accountant/db_access_objects/sent_payable_dao.rs
bertllll: I see. But then I'd prefer to simply introduce two different types. For clarity. 

```Rust
type TxHash = H256;
type BlockHash = H256;
```
Maybe not. What do you think? :man_shrugging:

node/src/database/db_initializer.rs
utkarshg6: Got it! Let me show you what I mean with some code:

```rust
    let latest_fetched_price: u128 = 100_000;
    let old_gas_price_for_nonce: u128 = 95_000;
    let batch_size: u8 = 10;

    let base_price = latest_fetched_price.max(old_gas_price_for_nonce + 1);

    // Calculate gas pricing by increasing the base price by a percentage based on batch size
    let gas_price = calculate_gas_price(base_price, batch_size);
```

The fourth step is not directly related to the previous steps; it has been included to provide a clearer understanding of the overall process.

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: That looks like a sound plan.

node/src/database/db_initializer.rs
utkarshg6: Okay, if we do need the column, I'll be happy to keep it.

PR #637, opened 2025-05-20
added dbip_download github workflow

czarte:


.github/workflows/dbip_download.yml
```diff
@@ -0,0 +1,52 @@
...
+          cd ip_country
+          cargo run < "dbip-data/dbip-country-lite.mmdb" > "$TEMP_DIR"/dbip_country.rs
+          ls "$TEMP_DIR"
+
+      - name: Commit and push generated file
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -B generated-source
+          git rm -rf .
```
dnwiebe, line 47: I'm a little nervous about using `-f` here in a script. If somebody messes with this script and we end up in a directory where we shouldn't be, `-f` could end up deleting something important. Does just `rm -r` work?

dnwiebe (approved):

Issue #602, opened 2025-03-25
Tie the `payable_scanner` with the `pending_payable` scanner and prefigure the retry semicycles


bertllll:
In the presence of the current state of things, it seems undesirable having those two scanners `pending_payable_scanner` and `payable_scanner` loosely configurable. It can cause an adverse effect if the person doesn't know what they are doing, and, it's been of a small advantage if otherwise, honestly. It comes to be more and more obvious that we can keep them distanced in a fixed manner. 

The keeping the configuration of those other scan intervals enabled can be also questioned but it doesn't belong into this card. 

We can actually run these two scanners consecutively, with basically no interval considered at all. We should definitely make sure the `pending_payable_scanner` goes always first! That's because it will be this scanner which will look after the status column. If it resolves a transaction attempt a failure (statuses beginning with f, followed by an opcode), it will be later picked up by the `payable_scanner` and included in the upcoming payment attempt, then executed immediately.

**Update: Cycles and semicycles** 

Pursuing the best design for the moment and trying to implement the retry mechanism more effectively, we started to question the original idea of waiting for the check simply until the upcoming try whose timing is set by the payable interval. Instead we've come up with a system of dynamic semicycles and from there derived cycles. 

In its utmost simplicity, it abides by this pattern: 

**P** 
Genuine scan for new payables producing payments 
**|**
**"retry interval"** 
Possibly also configurable
**|**
**PP/P** 
This is where most of the logic is centered. This scan will always run at least once within one "payable interval", which basically becomes the distance between two genuine payable scans, without any regards to pending payments. In fact, by the time the payable scanner seeks for new payments, there will never be any pending payments anymore. Here is how we will enforce it. When this bundled scanner is ending it will always evaluate if any tracked payments were found still pending. Until there are some, we will always command another retry. That implies that the "regular" payable scan can be theoretically postponed because we do not want it to collide with the verification scanner. If necessary, the repeated cycles of the verification scanner can roll through the regular milestone and will not stop until all known pending payments are taken care of. 
That is the ultimate condition for this verification scanner to conclude that the standard scanner should be scheduled. 
Now, there are two positions in which this may happen, and it determines the timing of the newly scheduled payable scanner. 
If the retries have so far taken less than a single payable interval, then we will compute how much time remains until the end of it, this will then be used as the time advancement of the scanner being scheduled. 
On the other hand, if the row of retries have already gone past the length of the large interval, we will simply schedule it just immediately, "wasting no more time". 
**|
P
[...]

PR #640, opened 2025-05-26
GH-602: Scanner system revamp

bertllll:


multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -432,7 +432,7 @@ fn verify_pending_payables() {
     );
     ui_client.send_request(
         UiScanRequest {
-            scan_type: ScanType::PendingPayables,
+            scan_type: ScanType::Payables,
```
utkarshg6: It was catchy to see that no line in the test was changed other than this assertion, which raised some concerns and after discussion, it turns out it could be a mistake, so you'd like to verify it and put it in the right manner. Good luck with your findings. 👍

node/src/accountant/scanners/scanners_utils.rs
```diff
@@ -321,6 +335,18 @@ pub mod pending_payable_scanner_utils {
         pub confirmed: Vec<PendingPayableFingerprint>,
     }
 
+    impl PendingPayableScanReport {
+        pub fn requires_payments_retry(&self) -> bool {
+            !self.still_pending.is_empty() || !self.failures.is_empty()
```
utkarshg6: You'd like to put a `todo!()` here.

node/src/accountant/scanners/scanners_utils.rs
```diff
@@ -843,4 +871,62 @@ mod tests {
...
+            PendingPayableScanReport {
+                still_pending: vec![],
+                failures: vec![PendingPayableId::new(456, make_tx_hash(1234))],
+                confirmed: vec![],
+            },
+            PendingPayableScanReport {
+                still_pending: vec![PendingPayableId::new(12, make_tx_hash(456))],
+                failures: vec![PendingPayableId::new(456, make_tx_hash(1234))],
+                confirmed: vec![],
+            },
```
utkarshg6: You'd like to comment this test and add a todo to fix it later, when you work on the production code related to this test.

node/src/accountant/scanners/test_utils.rs
```diff
@@ -2,9 +2,95 @@
...
+        self.compute_interval_results.borrow_mut().push(result);
+        self
+    }
+}
+
+pub enum ReplacementType<A, B> {
+    Real(A),
+    Mock(B),
+    Null,
+}
```
utkarshg6, line 354: You'd like to rename `A` to `ScannerReal` and `B` to `ScannerMock`.

node/src/accountant/scanners/test_utils.rs
```diff
@@ -2,9 +2,95 @@
...
+
+// The supplied scanner types are broken down to these detailed categories because they are
+// eventually represented by a private trait within the Scanners struct. Therefore, when
+// the values are constructed, they cannot be made into a trait object right away and needs to be
+// handled specifically.
```
utkarshg6: You'd like to reword this comment using this new text that you shared...

because they become eventually trait objects represented by a private trait. That one cannot be supplied, though, because it is unaccessable from outside, therefore, we provide the specific objects that will be then treated in an abstract manner.

node/src/accountant/test_utils.rs
```diff
@@ -133,7 +120,7 @@ pub enum DaoWithDestination<T> {
 enum DestinationMarker {
     AccountantBody,
     PendingPayableScanner,
-    PayableScanner,
+    SharedPayableScanner,
```
utkarshg6: You'd like to revert it back to the old name.

utkarshg6 (commented):
Continue at node/src/daemon/setup_reporter.rs

node/src/daemon/setup_reporter.rs
```diff
@@ -3440,10 +3440,10 @@ mod tests {
...
         let mut scan_intervals = *DEFAULT_SCAN_INTERVALS;
-        scan_intervals.pending_payable_scan_interval = scan_intervals
-            .pending_payable_scan_interval
+        scan_intervals.payable_scan_interval = scan_intervals
+            .payable_scan_interval
             .add(Duration::from_secs(15));
-        scan_intervals.pending_payable_scan_interval = scan_intervals
+        scan_intervals.receivable_scan_interval = scan_intervals
             .receivable_scan_interval
             .sub(Duration::from_secs(33));
```
utkarshg6, line 3451: Can you please add another case for `pending_payable_scan_interval`? Maybe you can test the boundary by adding or subtracting 1.

node/src/proxy_server/mod.rs
```diff
@@ -1380,7 +1380,7 @@ mod tests {
     use crate::test_utils::recorder::make_recorder;
     use crate::test_utils::recorder::peer_actors_builder;
     use crate::test_utils::recorder::Recorder;
-    use crate::test_utils::recorder_stop_conditions::{StopCondition, StopConditions};
+    use crate::test_utils::recorder_stop_conditions::{MsgIdentification, StopConditions};
```
utkarshg6: Inside the macro `match_lazily_every_type_id!()` you would like to put the full import for `MsgIdentification`, then there won't be any need for importing this wherever the macro is being used.

node/src/sub_lib/combined_parameters.rs
```diff
@@ -224,8 +224,7 @@ impl Display for ScanIntervals {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "{}|{}|{}",
-            self.pending_payable_scan_interval.as_secs(),
+            "{}|{}",
             self.payable_scan_interval.as_secs(),
```
utkarshg6, line 228: Printing pending payable scan interval is missing.

node/src/sub_lib/utils.rs
```diff
@@ -167,7 +167,7 @@ impl<T> NotifyLaterHandleReal<T> {
 
 impl<M, A> NotifyLaterHandle<M, A> for NotifyLaterHandleReal<M>
 where
-    M: Message + 'static,
+    M: Message + 'static + Send,
```
utkarshg6: You believe that you might not need the `Send` here anymore, and also on line `142`. So, you'd like to experiment removing them.

node/src/test_utils/mod.rs
```diff
@@ -950,19 +992,50 @@ pub mod unshared_test_utils {
                 self
             }
 
-            pub fn permit_to_send_out(mut self) -> Self {
+            pub fn capture_msg_and_let_it_fly_on(mut self) -> Self {
```
utkarshg6, line 995: You'd like to write an explanation for the messages that do not have `Default` trait via stringifying the type name.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl SingleCounterMsgSetup {
+    pub fn new(
+        trigger_msg_type_id: TypeId,
+        trigger_msg_id_method: MsgIdentification,
+        counter_messages: Vec<Box<dyn CounterMsgGear>>,
+    ) -> Self {
+        Self {
+            trigger_msg_type_id,
+            condition: trigger_msg_id_method,
```
utkarshg6: You'd like to keep the names consistent here, so the field name should be `trigger_msg_id_method` or `msg_identification` or something else that seems fir to you.

Also, this `SingleCounterMsgSetup` should be changed to `SingleTypeCounterMsgSetup`.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+            msg_gears: counter_messages,
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct CounterMessages {
+    msgs: HashMap<TypeId, Vec<SingleCounterMsgSetup>>,
+}
+
```
utkarshg6, line 99: You'd like a type alias to represent `TypeId` in the `HashMap`. You suggested to use: `TriggerMsgTypeId`.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+impl CounterMessages {
+    pub fn search_for_msg_setup<Msg>(&mut self, msg: &Msg) -> Option<Vec<Box<dyn CounterMsgGear>>>
```
utkarshg6: You'd also like to rename `Msg` to `TriggerMsg`. It was a new information to me that you use `setup` and `gear` synonymously, and that wasn't obvious to me, so after discussion you'd like to be consistently using either of them. I'd be grateful to you if you use `gear`, because it's easier to derive a deeper meaning out of this word.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+    where
+        Msg: ForcedMatchable<Msg> + 'static,
+    {
+        let type_id = msg.correct_msg_type_id();
```
utkarshg6: You'd like to rename the function name to `trigger_msg_type_id`. I appreciate the effort.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+        if let Some(msgs_vec) = self.msgs.get_mut(&type_id) {
+            msgs_vec
+                .iter_mut()
+                .position(|cm_setup| cm_setup.condition.resolve_condition(msg))
+                .map(|idx| {
+                    let matching_counter_msg = msgs_vec.remove(idx);
+                    matching_counter_msg.msg_gears
```
utkarshg6: You'd prefer to put in one line.

```rust
msgs_vec.remove(idx).msg_gears
```

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+            None
+        }
+    }
+
+    pub fn add_msg(&mut self, counter_msg_setup: SingleCounterMsgSetup) {
+        let type_id = counter_msg_setup.trigger_msg_type_id;
+        match self.msgs.entry(type_id) {
+            Entry::Occupied(mut existing) => existing.get_mut().push(counter_msg_setup),
+            Entry::Vacant(vacant) => {
```
utkarshg6: You like `vacancy` over `vacant`.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
...
+                })
+        } else {
+            None
+        }
+    }
+
+    pub fn add_msg(&mut self, counter_msg_setup: SingleCounterMsgSetup) {
+        let type_id = counter_msg_setup.trigger_msg_type_id;
+        match self.msgs.entry(type_id) {
+            Entry::Occupied(mut existing) => existing.get_mut().push(counter_msg_setup),
```
utkarshg6: You'd like to say `existing_vec` here.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) 2025, MASQ (https://masq.ai) and/or its affiliates. All rights reserved.
+
+#![cfg(test)]
+
+use crate::test_utils::recorder_stop_conditions::{ForcedMatchable, MsgIdentification};
+use actix::{Message, Recipient};
+use std::any::TypeId;
+use std::cell::RefCell;
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
```
utkarshg6, line 10: You'd like to leave an epic comment here mentioning how counter msgs can be used. How they can impersonate other actors. You'd also like to mention,"One message can trigger multiple messages."

node/src/test_utils/recorder.rs
```diff
@@ -70,6 +73,7 @@ pub struct Recorder {
     recording: Arc<Mutex<Recording>>,
     node_query_responses: Vec<Option<NodeQueryResponseMetadata>>,
     route_query_responses: Vec<Option<RouteQueryResponse>>,
+    expected_future_counter_msgs_opt: Option<CounterMessages>,
```
utkarshg6: `counter_msgs_opt`

node/src/test_utils/recorder.rs
```diff
@@ -162,8 +166,7 @@ recorder_message_handler_t_m_p!(ReportTransactionReceipts);
 recorder_message_handler_t_m_p!(RequestTransactionReceipts);
 recorder_message_handler_t_m_p!(RetrieveTransactions);
 recorder_message_handler_t_m_p!(ScanError);
-recorder_message_handler_t_m_p!(ScanForPayables);
-recorder_message_handler_t_m_p!(ScanForPendingPayables);
+recorder_message_handler_t_m_p!(ScanForNewPayables);
```
utkarshg6, line 169: remove all 4 of these types.

node/src/test_utils/recorder.rs
```diff
@@ -705,4 +815,213 @@ mod tests {
             TypeId::of::<PretendedMatchableWrapper<ExampleMsgB>>()
         )
     }
+
+    #[test]
+    fn diff_counter_msgs_with_diff_id_methods_can_be_used() {
```
utkarshg6: Let's look in the ways how to simplify these tests.

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -1,21 +1,26 @@
...
 
 use itertools::Itertools;
 use std::any::{Any, TypeId};
 
 pub enum StopConditions {
-    Any(Vec<StopCondition>),
-    All(Vec<StopCondition>),
+    Any(Vec<MsgIdentification>),
+    // Msg is tested against every ID method in the vector. In every case when they match, those
+    // methods are eliminated
```
utkarshg6: You'd like to rephrase the comment as this:`Single message can eliminate multiple ID Methods (previously stop conditions) by matching on them.`

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -1,21 +1,26 @@
...
+    AllGreedily(Vec<MsgIdentification>),
+    // Iterating through the vector of ID methods, only the first matching method is eliminated
+    // and any other one that would've matched needs to wait for the next received msg of this type
+    AllLazily(Vec<MsgIdentification>),
```
utkarshg6, line 15: You'd like to rephrase the comment as this: 

`Single message can eliminate _only one_ ID Method (previously stop conditions) by matching on them. To remove others, a new message must be received.`

Another suggestion:

`Iterating through ID methods, only the first matching method is eliminated and any foolowing one, that still can match, will need to wait for another message arival.`

utkarshg6 (commented):
Continue at node/src/test_utils/recorder_stop_conditions.rs:18

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -24,52 +29,70 @@ pub type BoxedMsgExpected = Box<dyn Any + Send>;
...
+        msg: &Msg,
+    ) -> bool {
+        if let Some(idx) = conditions
+            .iter()
+            .position(|condition| condition.resolve_condition::<Msg>(msg))
+        {
+            conditions.remove(idx);
+        }
+        conditions.is_empty()
+    }
```
utkarshg6, line 92: Great work! 🍻

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -84,43 +107,43 @@ impl StopConditions {
...
-    fn resolve_condition<T: ForcedMatchable<T> + Send + 'static>(&self, msg: &T) -> bool {
+impl MsgIdentification {
+    pub fn resolve_condition<Msg: ForcedMatchable<Msg> + Send + 'static>(&self, msg: &Msg) -> bool {
         match self {
-            StopCondition::StopOnType(type_id) => Self::matches_stop_on_type::<T>(msg, *type_id),
-            StopCondition::StopOnMatch { exemplar } => {
-                Self::matches_stop_on_match::<T>(exemplar, msg)
+            MsgIdentification::ByType(type_id) => Self::matches_by_type::<Msg>(msg, *type_id),
+            MsgIdentification::ByMatch { exemplar } => {
+                Self::matches_completely::<Msg>(exemplar, msg)
```
utkarshg6: Maybe you can rephrase the fn names as:

`matches_completely` -> `is_identical`, `matches_as_identical`
  
 `matches_by_type` -> `has_the_same_type` (optional)

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -336,10 +369,96 @@ mod tests {
...
+        let tested_msg_4 = NewPublicIp {
+            new_ip: IpAddr::V4(Ipv4Addr::new(45, 45, 45, 45)),
+        };
+
+        let kill_system = cond_set.resolve_stop_conditions::<NewPublicIp>(&tested_msg_4);
+
+        assert_eq!(kill_system, true);
+        assert_on_state_after_lazily_matched(4, &cond_set, |conds| {
+            assert!(conds.is_empty());
+        });
```
utkarshg6, line 454: In case you could introduce line breaks right after a particular set of assertion ends, that would really help in dividing the stages.

node/src/test_utils/recorder_stop_conditions.rs
```diff
@@ -336,10 +369,96 @@ mod tests {
...
+    }
+
+    fn assert_on_state_after_lazily_matched(
+        stage: usize,
+        cond_set: &StopConditions,
+        assertions: fn(&[MsgIdentification]),
+    ) {
+        match &cond_set {
+            StopConditions::AllLazily(conds) => assertions(conds),
```
utkarshg6: You might like to call this fn as `apply_assertions`.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+            if initial_pending_payable_scan {
+                SchedulingAfterHaltedScan::Schedule(ScanType::Payables)
+            } else {
+                unreachable!(
+                    "the automatic pending payable scan should always be requested only in need, \
+                    which contradicts the current StartScanError::NothingToProcess"
+                )
+            }
+        } else if err == &StartScanError::NoConsumingWalletFound {
+            SchedulingAfterHaltedScan::Schedule(ScanType::Payables)
```
utkarshg6: You'd like another check here which considers whether there is an awareness of unresolved pending payables.

Bert:
I first thought that it would require to bring the `aware_of_unresolved_pending_payables` which has had use only in the scanners/mod.rs so far, but then I realized that it could be simpler, understand as doing without it.
I only need to ask if the `initial_pending_payable_scan` flag says true and that it's not the externally triggered scan - but that's already clear that it isn't given the first if. But if it isn't the initial scan and it IS an automatic scan, then that is a problem, a state which technically shouldn't be possible (reason: In automatic scanners, it always all starts with scanning for the pending payables. So if the consuming wallet is not set, this is basically the first moment at scanners where we can run into it. And we should react by scheduling the same scan again (that is **ScanType::PendingPayables**), even though it will have no use, but it might alarm the user and he'll try to fix the void.). However, if it isn't the initial scan, then it's a bug, and we should simply panic through the **unreachable!()** macro. (reason: Only the NewPayableScanner can schedule this scan but it couldn't have been so because we wouldn't have ever gotten past the initial scan if the wallet is missing and so there is no way how we could reach the NewPayableScanner even a single time. Impossible. Period. )

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+            HintableScanner::PendingPayables { .. } => ScanType::PendingPayables,
+        }
+    }
+}
+
+pub struct PayableScanScheduler {
+    pub new_payable_notify_later: Box<dyn NotifyLaterHandle<ScanForNewPayables, Accountant>>,
+    pub dyn_interval_computer: Box<dyn NewPayableScanDynIntervalComputer>,
+    pub inner: Arc<Mutex<PayableScanSchedulerInner>>,
+    pub nominal_interval: Duration,
```
utkarshg6: You'd like to rename this field such that something else is used instead of nominal.

utkarshg6 (commented):
Continue at node/src/accountant/scanners/scan_schedulers.rs:78

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+    }
+
+    pub fn schedule_for_new_payable(&self, ctx: &mut Context<Accountant>, logger: &Logger) {
+        let inner = self.inner.lock().expect("couldn't acquire inner");
+        let last_new_payable_timestamp = inner.last_new_payable_scan_timestamp;
+        let nominal_interval = self.nominal_interval;
+        let now = SystemTime::now();
+        if let Some(interval) = self.dyn_interval_computer.compute_interval(
+            now,
+            last_new_payable_timestamp,
```
utkarshg6: You'd like to include `scan` in the name.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+                    response_skeleton_opt: None,
+                },
+                ctx,
+            );
+        }
+    }
+
+    // Can be triggered by a command, whereas the finished pending payable scanner is followed up
+    // by this scheduled message that can bear the response skeleton. This message is inserted into
+    // the Accountant's mailbox with no delay
```
utkarshg6: You'd like to revamp this comment.

utkarshg6 (commented):
Continue at node/src/accountant/scanners/scan_schedulers.rs:267

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+        err: &StartScanError,
+        is_externally_triggered: bool,
+    ) -> SchedulingAfterHaltedScan {
+        if is_externally_triggered {
+            SchedulingAfterHaltedScan::DoNotSchedule
+        } else {
+            unreachable!(
+                "{:?} should be impossible with RetryPayableScanner in automatic mode",
+                err
+            )
```
utkarshg6, line 314: I've noticed that there are several instances of unreachable!() in this code. It might be beneficial to explore ways to prevent the function calls that could lead to these situations. This could enhance the overall robustness of the code.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+        );
+        assert_eq!(
+            schedulers.pending_payable.interval,
+            scan_intervals.pending_payable_scan_interval
+        );
+        assert_eq!(
+            schedulers.receivable.interval,
+            scan_intervals.receivable_scan_interval
+        );
+        assert_eq!(schedulers.automatic_scans_enabled, true)
```
utkarshg6: You must assert against the variable created earlier, `automatic_scans_enabled`.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+                now.checked_sub(Duration::from_millis(1111)).unwrap(),
+                Duration::from_millis(3333),
+                Duration::from_millis(2222),
+            ),
+            (
+                now.checked_sub(Duration::from_secs(200)).unwrap(),
+                Duration::from_secs(204),
+                Duration::from_secs(4),
+            ),
+        ];
```
utkarshg6, line 392: I think you can refactor a helper a function out of it, which takes the following argument, `past_instant`, `standard_interval`, `expected_result`.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+                    acc.into_iter()
+                        .filter(|(num, _)| num != &current)
+                        .collect_vec()
+                })
+                .into_iter()
+                .map(|(_, err)| *err)
+                .collect_vec();
+            self.errors = adjusted;
+            self
+        }
```
utkarshg6, line 584: I think you can use the `retain()` method to prevent using too many iterators:

```rust
let error_variants_to_remove: Vec<_> = errors_to_eliminate.iter()
    .map(Self::number_variant)
    .collect();

// Retain only the errors that are not in the list of numbers to eliminate
self.errors.retain(|err| !error_variants_to_remove.contains(&Self::number_variant(*err)));
self
```

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+                    started_at: SystemTime::now(),
+                },
+                false,
+            );
+    }
+
+    #[test]
+    fn resolve_hint_for_new_payables_with_those_error_cases_that_result_in_future_rescheduling() {
+        let inputs =
+            AllStartScanErrorsAdjustable::default().eliminate_already_tested_variants(vec![
```
utkarshg6: Maybe you should rename `AllStartScanErrorsAdjustable` to `ConfigurableStartScanErrors`.

utkarshg6 (commented):
Continue at node/src/accountant/scanners/mod.rs

utkarshg6 (commented):
Continue at node/src/accountant/scanners/mod.rs:151

node/src/accountant/scanners/mod.rs
```diff
@@ -96,25 +108,280 @@ impl Scanners {
...
+            self.pending_payable.scan_started_at(),
+            self.payable.scan_started_at(),
+        ) {
+            (Some(pp_timestamp), Some(p_timestamp)) =>
+            // If you're wondering, then yes, this condition should be the sacre truth between
+            // PendingPayableScanner and NewPayableScanner.
+            {
+                unreachable!(
+                    "Both payable scanners should never be allowed to run in parallel. Scan for \
+                    pending payables started at: {}, scan for payables started at: {}",
```
utkarshg6: I think you'd like to say "Both payable and pending payable scanners should", and also in the comment, there's a typo, `sacre truth` -> `sacred truth`.

node/src/accountant/scanners/mod.rs
```diff
@@ -96,25 +108,280 @@ impl Scanners {
...
+        logger: &Logger,
+        automatic_scans_enabled: bool,
+    ) -> Result<RetrieveTransactions, StartScanError> {
+        let triggered_manually = response_skeleton_opt.is_some();
+
+        if triggered_manually && automatic_scans_enabled {
+            return Err(StartScanError::ManualTriggerError(
+                MTError::AutomaticScanConflict,
+            ));
+        }
```
utkarshg6, line 235: You might prefer to refactor the first few lines into a different function, as it is a duplicate in `start_new_payable_scan_guarded()`.

node/src/accountant/scanners/mod.rs
```diff
@@ -96,25 +108,280 @@ impl Scanners {
...
+        self.payable.perform_payment_adjustment(setup, logger)
+    }
+
+    pub fn initial_pending_payable_scan(&self) -> bool {
+        self.initial_pending_payable_scan
+    }
+
+    pub fn unset_initial_pending_payable_scan(&mut self) {
+        self.initial_pending_payable_scan = false
+    }
```
utkarshg6, line 308: Can you please change the name of fn `initial_pending_payable_scan()` to `get_initial_pending_payable_scan()` and `unset_initial_pending_payable_scan(&mut self)` to `set_initial_pending_payable_scan(&mut self, flag: bool)`. It's better to set it using a flag instead.

node/src/accountant/scanners/mod.rs
```diff
@@ -96,25 +108,280 @@ impl Scanners {
...
+    ) -> Result<QualifiedPayablesMessage, StartScanError>
+    where
+        TriggerMessage: Message,
+        (dyn MultistageDualPayableScanner + 'a):
+            StartableScanner<TriggerMessage, QualifiedPayablesMessage>,
+    {
+        <(dyn MultistageDualPayableScanner + 'a) as StartableScanner<
+            TriggerMessage,
+            QualifiedPayablesMessage,
+        >>::start_scan(scanner, wallet, timestamp, response_skeleton_opt, logger)
```
utkarshg6, line 329: You'd like to add a comment at the top of the function, explaining how it works, that'll be great.

node/src/accountant/scanners/mod.rs
```diff
@@ -96,25 +108,280 @@ impl Scanners {
...
+                },
+            ))
+        } else if !self.aware_of_unresolved_pending_payable {
+            unreachable!(
+                "Automatic pending payable scan should never start if there are no pending \
+                payables to process."
+            )
+        } else {
+            Ok(())
+        }
```
utkarshg6, line 357: I can't think of a serious risk that the if-else if-else, can pose here, but during our discussion it was a conclusion that using plain `if` statements here would be better, so go ahead with that change. 🙏

node/src/accountant/scanners/mod.rs
```diff
@@ -214,26 +481,40 @@ impl Scanner<QualifiedPayablesMessage, SentPayables> for PayableScanner {
...
-    fn finish_scan(&mut self, message: SentPayables, logger: &Logger) -> Option<NodeToUiMessage> {
+impl StartableScanner<ScanForRetryPayables, QualifiedPayablesMessage> for PayableScanner {
+    fn start_scan(
+        &mut self,
+        _consuming_wallet: &Wallet,
+        _timestamp: SystemTime,
+        _response_skeleton_opt: Option<ResponseSkeleton>,
+        _logger: &Logger,
+    ) -> Result<QualifiedPayablesMessage, StartScanError> {
+        todo!()
```
utkarshg6: Can you please write the card number inside the `todo!()`.

node/src/accountant/scanners/mod.rs
```diff
@@ -596,32 +885,52 @@ impl Scanner<RequestTransactionReceipts, ReportTransactionReceipts> for PendingP
...
-        let response_skeleton_opt = message.response_skeleton_opt;
+    ) -> PendingPayableScanResult {
+        let construct_msg_scan_ended_to_ui = move || {
+            message
+                .response_skeleton_opt
+                .map(|response_skeleton| NodeToUiMessage {
+                    target: MessageTarget::ClientId(response_skeleton.client_id),
+                    body: UiScanResponse {}.tmb(response_skeleton.context_id),
+                })
+        };
```
utkarshg6: Apparently, you can get rid of the closure and the `move` keyword.

node/src/accountant/scanners/mod.rs
```diff
@@ -596,32 +885,52 @@ impl Scanner<RequestTransactionReceipts, ReportTransactionReceipts> for PendingP
...
 
-        match message.fingerprints_with_receipts.is_empty() {
-            true => debug!(logger, "No transaction receipts found."),
+        let requires_payment_retry = match message.fingerprints_with_receipts.is_empty() {
+            true => {
+                debug!(logger, "No transaction receipts found.");
+                todo!(
+                    "requires payment retry...it must be processed across the new methods on \
+                the SentPaybleDAO"
```
utkarshg6: Can you please provide the card number here. For now, the card I'm currently working on is `GH-631`, it contains the implementation of `SentPayableDAO`.

node/src/accountant/scanners/mod.rs
```diff
@@ -946,146 +1269,413 @@ impl ReceivableScanner {
...
 
-#[derive(Debug, PartialEq, Eq)]
-pub enum BeginScanError {
+#[derive(Debug, PartialEq, Eq, Clone, VariantCount)]
+pub enum StartScanError {
     NothingToProcess,
     NoConsumingWalletFound,
-    ScanAlreadyRunning(SystemTime),
+    ScanAlreadyRunning {
+        pertinent_scanner: ScanType,
```
utkarshg6: You'd like to check on whether you could eliminate this line.

node/src/accountant/scanners/mod.rs
```diff
@@ -946,146 +1269,413 @@ impl ReceivableScanner {
...
-                false => debug!(logger, "{}", log_message),
-            }
+        match is_externally_triggered {
+            true => match log_message {
+                ErrorType::Temporary(msg) => info!(logger, "{}", msg),
+                ErrorType::Permanent(msg) => warning!(logger, "{}", msg),
+            },
+
+            false => match log_message {
+                ErrorType::Temporary(msg) | ErrorType::Permanent(msg) => debug!(logger, "{}", msg),
```
utkarshg6: You'd like to print all `Permanent` errors as `warning!()`, and `is_externally_triggered` should be moved to the case of `Temporary`, where you can use `info!()` or `debug!()` accordingly.

utkarshg6 (commented):
Continue at node/src/accountant/scanners/mod.rs:1474

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -0,0 +1,839 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+pub enum SchedulingAfterHaltedScan {
+    Schedule(ScanType),
+    DoNotSchedule,
+}
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum HintableScanner {
```
bertllll: Maybe `PayableSequenceScanner`

node/src/accountant/scanners/mod.rs
```diff
@@ -946,146 +1269,413 @@ impl ReceivableScanner {
...
+        }
+
+        fn perform_payment_adjustment(
+            &self,
+            _setup: PreparedAdjustment,
+            _logger: &Logger,
+        ) -> OutboundPaymentsInstructions {
+            intentionally_blank!()
+        }
+    }
```
utkarshg6, line 1386: Let's move the `ScannerMock` and `NullScanner` to a different file, meant for test utilities. You'd like to make the traits public in the scope of the scanners, and they won't be accessible in the accountant.

node/src/accountant/scanners/mod.rs
```diff
@@ -1220,48 +1889,150 @@ mod tests {
...
+            .payable_dao(payable_dao)
+            .build();
+
+        let result = Scanners::start_correct_payable_scanner::<ScanForNewPayables>(
+            &mut subject,
+            &consuming_wallet,
+            now,
+            None,
+            &Logger::new("test"),
+        );
```
utkarshg6, line 1724: You'd like to replace this:

```rust
        let result = subject.start_new_payable_scan_guarded(
            &consuming_wallet,
            SystemTime::now(),
            None,
            &Logger::new("test"),
            true,
        );
```

node/src/accountant/scanners/mod.rs
```diff
@@ -1270,49 +2041,106 @@ mod tests {
...
         let mut subject = PayableScannerBuilder::new()
             .payable_dao(payable_dao)
             .build();
 
-        let result = subject.begin_scan(consuming_wallet, now, None, &Logger::new("test"));
-
-        let is_scan_running = subject.scan_started_at().is_some();
-        assert_eq!(is_scan_running, false);
-        assert_eq!(result, Err(BeginScanError::NothingToProcess));
+        let _ = Scanners::start_correct_payable_scanner::<ScanForRetryPayables>(
```
utkarshg6, line 1864: you'd like to use the guarded fn here too

node/src/accountant/scanners/mod.rs
```diff
@@ -2876,21 +3890,26 @@ mod tests {
...
+        let is_scan_running = subject.scan_started_at(ScanType::PendingPayables).is_some();
+        assert_eq!(
+            result,
+            PendingPayableScanResult::NoPendingPayablesLeft(None)
+        );
         assert_eq!(is_scan_running, false);
         let tlh = TestLogHandler::new();
         tlh.exists_log_containing(&format!(
-            "DEBUG: {test_name}: No transaction receipts found."
+            "ERROR: {test_name}: No transaction receipts found."
```
utkarshg6: Please change it to `WARN: {test_name}: Expected {x} receipts, but none found.`

node/src/accountant/scanners/mod.rs
```diff
@@ -3367,38 +4413,270 @@ mod tests {
...
+    fn log_error_works_fine_for_automatic_scanning() {
+        init_test_logging();
+        let test_name = "log_error_works_fine_for_automatic_scanning";
+        let input = vec![
+            (
+                StartScanError::ScanAlreadyRunning {
+                    pertinent_scanner: ScanType::Payables,
+                    started_at: SystemTime::now()
+                },
+                format!("DEBUG: {test_name}: Payables scan was already initiated at" /*TODO suboptimal */)
```
utkarshg6: You'd also like to test the timestamp.

node/src/accountant/scanners/mod.rs
```diff
@@ -3367,38 +4413,270 @@ mod tests {
...
+    fn log_error_works_fine_for_externally_triggered_scanning() {
+        init_test_logging();
+        let test_name = "log_error_works_fine_for_externally_triggered_scanning";
+        let input = vec![
+            (
+                StartScanError::ScanAlreadyRunning {
+                    pertinent_scanner: ScanType::Payables,
+                    started_at: SystemTime::now()
+                },
+                format!("INFO: {test_name}: Payables scan was already initiated at" /*TODO suboptimal */)
```
utkarshg6: You'd like to merge the two test together, this and the one above, by introducing two more values in the same tuple, that represent different log levels.

node/src/accountant/scanners/mod.rs
```diff
@@ -3367,38 +4413,270 @@ mod tests {
...
+            initial_pending_payable_scan: false,
+            pending_payable: Box::new(NullScanner::new()),
+            receivable: Box::new(NullScanner::new()),
+        }
+    }
+
+    // Concatenated hours, minutes, seconds and milliseconds in a single integer
+    #[derive(PartialEq, Debug)]
+    struct PseudoTimestamp {
+        rep: u32,
```
utkarshg6: If you'd like to try you can find a way to convert `String` to `SystemTime`. Then it would be easier to compare. Furthermore, we can use UNIX timestamp for comparison.

utkarshg6 (commented):

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
-    fn handle(&mut self, msg: ReceivedPayments, _ctx: &mut Self::Context) -> Self::Result {
-        if let Some(node_to_ui_msg) = self.scanners.receivable.finish_scan(msg, &self.logger) {
-            self.ui_message_sub_opt
-                .as_ref()
-                .expect("UIGateway is not bound")
-                .try_send(node_to_ui_msg)
-                .expect("UIGateway is dead");
+    fn handle(&mut self, msg: ScanForPendingPayables, ctx: &mut Self::Context) -> Self::Result {
+        // By now we know this is an automatic scan process. The scan may be or may not be
+        // rescheduled. It depends on the findings. Any failed transactions will lead to the launch
```
utkarshg6: `Any failed transactions` -> `Any failed transaction` (singular)

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
+        // of the RetryPayableScanner, which finishes, and the PendingPayablesScanner is scheduled
+        // to run again. However, not from here.
+        let response_skeleton_opt = msg.response_skeleton_opt;
+        if let SchedulingAfterHaltedScan::Schedule(ScanType::Payables) =
```
utkarshg6: You'd like to rename `SchedulingAfterHaltedScan` to `StartScanErrorResponse`.

utkarshg6 (commented):
Continue at node/src/accountant/mod.rs:238

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
     type Result = ();
 
-    fn handle(
-        &mut self,
-        msg: BlockchainAgentWithContextMessage,
-        _ctx: &mut Self::Context,
-    ) -> Self::Result {
-        self.handle_payable_payment_setup(msg)
+    fn handle(&mut self, msg: ScanForNewPayables, ctx: &mut Self::Context) -> Self::Result {
+        // We know this must be a scheduled scan, but are yet clueless if where it's going to be
```
utkarshg6: You can drop off the `if` after `clueless`.

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
+        // rescheduled. If no payables qualify for a payment, we do it here right away. If some
```
utkarshg6: You'd like to change `no payables qualify` -> `no payable qualifies`.

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
-        if let Some(node_to_ui_msg) = self.scanners.payable.finish_scan(msg, &self.logger) {
-            self.ui_message_sub_opt
-                .as_ref()
-                .expect("UIGateway is not bound")
-                .try_send(node_to_ui_msg)
-                .expect("UIGateway is dead");
-        }
+    fn handle(&mut self, msg: ScanForRetryPayables, _ctx: &mut Self::Context) -> Self::Result {
+        // RetryPayableScanner is scheduled only when the PendingPayableScanner finishes discovering
+        // that there have been some failed pending payables. No place for it here.
```
utkarshg6: You may simply call `failed payables`.

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
-impl Handler<ScanForPayables> for Accountant {
+impl Handler<ScanForReceivables> for Accountant {
     type Result = ();
 
-    fn handle(&mut self, msg: ScanForPayables, ctx: &mut Self::Context) -> Self::Result {
-        self.handle_request_of_scan_for_payable(msg.response_skeleton_opt);
-        self.schedule_next_scan(ScanType::Payables, ctx);
+    fn handle(&mut self, msg: ScanForReceivables, ctx: &mut Self::Context) -> Self::Result {
+        // By now we know it is an automatic scan, which, for this particular scanner, is
+        // rescheduled pretty regularly, right here, no matter what its outcome is.
```
utkarshg6: You'd like to add that `Receivable scanner is independent of other scanners.`

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
+                } else {
+                    self.scan_schedulers
+                        .payable
+                        .schedule_for_new_payable(ctx, &self.logger)
+                }
+            }
+            PendingPayableScanResult::PaymentRetryRequired => self
+                .scan_schedulers
+                .payable
+                .schedule_for_retry_payable(ctx, response_skeleton_opt, &self.logger),
```
utkarshg6: It would be easier to identify if you'd call this function as `schedule_retry_payable_scan()`, similarly on the line 298, you can rename it to say `schedule_new_payable_scan()`.

node/src/accountant/mod.rs
```diff
@@ -183,94 +196,171 @@ impl Handler<StartMessage> for Accountant {
...
+        match self.scanners.finish_pending_payable_scan(msg, &self.logger) {
+            PendingPayableScanResult::NoPendingPayablesLeft(ui_msg_opt) => {
+                if let Some(node_to_ui_msg) = ui_msg_opt {
+                    self.ui_message_sub_opt
+                        .as_ref()
+                        .expect("UIGateway is not bound")
+                        .try_send(node_to_ui_msg)
+                        .expect("UIGateway is dead");
+                    // Externally triggered scan should never be allowed to spark a procedure that
+                    // would bring around payables with fresh nonces. The job's done.
```
utkarshg6: You'd like to change `around` to `over`. I'd suggest don't use any of them.

node/src/accountant/mod.rs
```diff
@@ -860,66 +945,129 @@ impl Accountant {
                     .expect("BlockchainBridge is unbound")
                     .try_send(scan_message)
                     .expect("BlockchainBridge is dead");
+                SchedulingAfterHaltedScan::DoNotSchedule
             }
-            Err(e) => e.handle_error(
-                &self.logger,
-                ScanType::Payables,
-                response_skeleton_opt.is_some(),
+            Err(e) => self.handle_start_scan_error_for_scanner_with_irregular_scheduling(
```
utkarshg6: You'd like to rename it `schedule_responsible_scanner`. A suggestion from my side: `handle_error_for_payable_sequence_scanner`.

node/src/accountant/mod.rs
```diff
@@ -1230,33 +1375,22 @@ mod tests {
...
+            result.scan_schedulers.pending_payable.interval,
             default_scan_intervals.pending_payable_scan_interval,
         );
-        assert_scan_scheduler(
-            ScanType::Receivables,
+        assert_eq!(
+            result.scan_schedulers.receivable.interval,
             default_scan_intervals.receivable_scan_interval,
         );
+        assert_eq!(result.scan_schedulers.automatic_scans_enabled, true);
```
utkarshg6, line 1418: If you'd like you can also add an assertion for `aware_of_initial_pending_payables`, you might prefer to keep only one to prevent an overhead.

node/src/accountant/mod.rs
```diff
@@ -1540,7 +1587,11 @@ mod tests {
         let payable_scanner = PayableScannerBuilder::new()
             .payment_adjuster(payment_adjuster)
             .build();
-        subject.scanners.payable = Box::new(payable_scanner);
+        subject
+            .scanners
+            .replace_scanner(ScannerReplacement::Payable(ReplacementType::Real(
+                payable_scanner,
```
utkarshg6, line 1624: It was weird that the `Real` variant can hold a mocked `payable_scanner`. You'd like to check it out.

node/src/accountant/mod.rs
```diff
@@ -1671,7 +1726,11 @@ mod tests {
         let payable_scanner = PayableScannerBuilder::new()
             .payment_adjuster(payment_adjuster)
             .build();
-        subject.scanners.payable = Box::new(payable_scanner);
+        subject
+            .scanners
+            .replace_scanner(ScannerReplacement::Payable(ReplacementType::Real(
```
utkarshg6, line 1762: here too

utkarshg6 (commented):
Continue at node/src/accountant/mod.rs:2032

node/src/accountant/mod.rs
```diff
@@ -1832,125 +1955,232 @@ mod tests {
...
+                            PendingPayable {
+                                recipient_wallet: make_wallet("abc"),
+                                hash: make_tx_hash(789)
+                            }
+                        )]),
+                        response_skeleton_opt
+                    },
+                    &subject_addr
+                ),
+            ]))
```
utkarshg6, line 2160: Egggscellent job!

A small request: Can you please migrate the elements of vector into different variables.

node/src/accountant/mod.rs
```diff
@@ -2078,208 +2576,303 @@ mod tests {
 
     #[test]
     fn accountant_scans_after_startup() {
```
utkarshg6: This test is quite long. I know it would be hard to refactor it because the preparatory lines are later used in assertions. So, please give it a try in case you see something that can be refactored out.

node/src/accountant/mod.rs
```diff
@@ -2078,208 +2576,303 @@ mod tests {
...
+            pending_payable_params.is_empty(),
+            "Should be empty but was {:?}",
+            pending_payable_params
+        );
+        let scan_for_pending_payables_notify_later_params =
+            scan_for_pending_payables_notify_later_params_arc
+                .lock()
+                .unwrap();
+        // The part with executing the scan for NewPayables is deliberately omitted here, but
+        // the scan for pending payables would've been scheduled just after that
```
utkarshg6: You'd like to rephrase the comment:

```
The part with running the `NewPayableScanner` is deliberately omitted here, we stop the test right before that, and the first occasion for scheduling the `PendingPayableScanner` would've lied no sooner than after the `NewPayableScan` finishes, having produced at least one blockchain transactions. 
```

node/src/accountant/mod.rs
```diff
@@ -2078,208 +2576,303 @@ mod tests {
...
+                },
+                receivable_scan_interval
+            )]
+        );
+        // Finally, an assertion to prove course of actions in time. We expect the pending payable
+        // scanner to take place before the payable scanner.
+        // I do believe it's impossible that these two instants would happen at the same time, until
+        // this is run on a real super-giga-computer.
+        assert!(pp_scan_started_at < p_scheduling_now);
     }
```
utkarshg6, line 3267: You'd like to write a new test which would take care of the pending transactions found in the initial scan unlike the test which has no such issue. Since, both of these tests are meant for the situation when the accountant starts up, it would be a good idea to combine most of it's functionality through a helper function (in case it seems a possibility).

node/src/accountant/mod.rs
```diff
@@ -2078,208 +2576,303 @@ mod tests {
...
-            }))
-            .stop_the_system_after_last_msg();
-        let earning_wallet = make_wallet("earning");
-        let mut config = bc_from_earning_wallet(earning_wallet.clone());
-        config.scan_intervals_opt = Some(ScanIntervals {
-            payable_scan_interval: Duration::from_secs(100),
-            receivable_scan_interval: Duration::from_millis(99),
-            pending_payable_scan_interval: Duration::from_secs(100),
-        });
+    fn initial_pending_payable_scan_if_some_payables_found() {
```
utkarshg6, line 3270: You'd like to assert on the message sent to the `BlockchainBridge`.

node/src/accountant/mod.rs
```diff
@@ -2078,208 +2576,303 @@ mod tests {
...
+        // PendingPayableScanner -> NewPayableScanner -> NewPayableScanner...
+        subject.scan_schedulers.payable.new_payable_notify = Box::new(NotifyHandleMock::default());
+        subject
+            .scanners
+            .replace_scanner(ScannerReplacement::PendingPayable(ReplacementType::Null));
+        subject
+            .scanners
+            .replace_scanner(ScannerReplacement::Receivable(ReplacementType::Mock(
+                receivable_scanner,
+            )));
```
utkarshg6, line 3371: A suggestion from your side: If you use `ScanForReceivables` instead of `StartMessage` you won't need to write the lines 2859...2869.

node/src/accountant/mod.rs
```diff
@@ -2289,155 +2882,264 @@ mod tests {
...
-        tlh.exists_log_containing(&format!("DEBUG: {test_name}: first attempt"));
-        tlh.exists_log_containing(&format!("DEBUG: {test_name}: second attempt"));
+        let mut start_scan_payable_params = start_scan_payable_params_arc.lock().unwrap();
+        let (wallet, timestamp, response_skeleton_opt, logger, _) =
+            start_scan_payable_params.remove(0);
+        assert_eq!(wallet, consuming_wallet);
+        assert!(time_before <= timestamp && timestamp <= time_after);
+        assert_eq!(response_skeleton_opt, None);
+        assert!(start_scan_payable_params.is_empty());
+        debug!(logger, "verifying payable scanner logger");
```
utkarshg6: You already have an auxiliary function to test this, you should try using that.

node/src/accountant/mod.rs
```diff
@@ -2499,14 +3207,14 @@ mod tests {
...
-        // no panics because of recalcitrant DAOs; therefore DAOs were not called; therefore test passes
+        // No panics because of recalcitrant DAOs; therefore DAOs were not called; therefore test passes
         TestLogHandler::new().exists_log_containing(
             &format!("{test_name}: Started with --scans off; declining to begin database and blockchain scans"),
         );
     }
 
     #[test]
-    fn scan_for_payables_message_does_not_trigger_payment_for_balances_below_the_curve() {
+    fn scan_for_new_payables_does_not_trigger_payment_for_balances_below_the_curve() {
```
utkarshg6, line 3749: This could be an overhead.

utkarshg6 (commented):
Continue at node/src/accountant/mod.rs:3296

node/src/accountant/mod.rs
```diff
@@ -3381,429 +4099,450 @@ mod tests {
...
+            *pending_payable_notify_later_params,
+            vec![(ScanForPendingPayables::default(), pending_payable_interval)]
+        );
+        // The accountant is unbound here. We don't use the bind message. It means we can prove
+        // none of those other scan requests could have been sent (especially ScanForNewPayables,
+        // ScanForRetryPayables)
+    }
+
+    #[test]
+    fn zero_sent_payables_so_the_payable_scan_is_rescheduled_without_scanning_for_pending_payables()
```
utkarshg6: I'd be happy if we have something other than zero, as it can be confused as empty vector.

```rust
no_payables_left_the_node_so_payable_scan_is_rescheduled_as_pending_payable_scan_was_omitted
```

node/src/accountant/mod.rs
```diff
@@ -3381,429 +4099,450 @@ mod tests {
...
+        assert_eq!(actual_sent_payable, sent_payable,);
+        test_use_of_the_same_logger(&logger, test_name, None);
+        let mut payable_notify_later_params = payable_notify_later_params_arc.lock().unwrap();
+        let (scheduled_msg, _interval) = payable_notify_later_params.remove(0);
+        assert_eq!(scheduled_msg, ScanForNewPayables::default());
+        assert!(
+            payable_notify_later_params.is_empty(),
+            "Should be empty but {:?}",
+            payable_notify_later_params
+        );
```
utkarshg6, line 4911: You'd like to configure the notify later of the pending payable and assert on it to verify that it stays empty.

node/src/accountant/mod.rs
```diff
@@ -3381,429 +4099,450 @@ mod tests {
...
+        subject
+            .scanners
+            .replace_scanner(ScannerReplacement::PendingPayable(ReplacementType::Mock(
+                pending_payable_scanner,
+            )));
+        subject.scan_schedulers.payable.retry_payable_notify =
+            Box::new(NotifyHandleMock::default().notify_params(&retry_payable_notify_params_arc));
+        let system = System::new(test_name);
+        let (mut msg, _) =
+            make_report_transaction_receipts_msg(TxStatus::Pending, TxStatus::Failed);
```
utkarshg6: If you could create a function that only outputs one `msg`, that would be ideal, then we can use it either once or multiple times as needed.

node/src/accountant/mod.rs
```diff
@@ -3381,429 +4099,450 @@ mod tests {
...
+            "expected actual {:?} to be between {:?} and {:?}",
+            actual_interval,
+            left_side_bound,
+            right_side_bound
+        );
+    }
+
+    fn make_report_transaction_receipts_msg(
+        status_of_tx_1: TxStatus,
+        status_of_tx_2: TxStatus,
```
utkarshg6: You may use a vector here as an argument. You'd like to use the `enumerate()` function for uniqueness.

utkarshg6 (commented):
Phase 1 Review Finished. Good Luck!

utkarshg6 (approved):

PR #646, opened 2025-06-04
GH-631: Create `FailedPayables` Table

utkarshg6:


node/src/database/db_initializer.rs
```diff
@@ -750,8 +781,7 @@ mod tests {
         let conn = subject
             .initialize(&home_dir, DbInitializationConfig::test_default())
             .unwrap();
-
-        let mut stmt = conn.prepare("select rowid, tx_hash, receiver_address, amount_high_b, amount_low_b, timestamp, gas_price_wei, nonce, status from sent_payable").unwrap();
+        let mut stmt = conn.prepare("select rowid, tx_hash, receiver_address, amount_high_b, amount_low_b, timestamp, gas_price_wei_high_b, gas_price_wei_low_b, nonce, block_hash, block_number from sent_payable").unwrap();
```
bertllll: Just for better readability, could you put in some line breaks?

node/src/database/db_initializer.rs
```diff
@@ -767,6 +797,33 @@ mod tests {
...
+        let home_dir = ensure_node_home_directory_does_not_exist(
+            "db_initializer",
+            "db_initialize_creates_failed_payable_table",
+        );
+        let subject = DbInitializerReal::default();
+
+        let conn = subject
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let mut stmt = conn.prepare("select rowid, tx_hash, receiver_address, amount_high_b, amount_low_b, timestamp, gas_price_wei_high_b, gas_price_wei_low_b, nonce, reason, checked from failed_payable").unwrap();
```
bertllll: Also here please...line breaks

node/src/database/db_initializer.rs
```diff
@@ -767,6 +797,33 @@ mod tests {
...
+        let mut sent_payable_contents = stmt.query_map([], |_| Ok(42)).unwrap();
```
bertllll: You can say `stmt.execute([]).unwrap()`. If you use that nobody will be wondering if the `Ok(42)` means anything. 

You said that you'd got inspired from other tests with this kind of check. If you can do it, please replace it with this solution also there.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -292,29 +319,19 @@ mod tests {
...
-                block_number: Default::default(),
-            }))
+            .hash(make_tx_hash(2))
+            .block(Default::default())
             .build();
         let subject = SentPayableDaoReal::new(wrapped_conn);
-        let txs = vec![tx1, tx2, tx3];
+        let txs = vec![tx1, tx2];
 
         let result = subject.insert_new_records(&txs);
```
bertllll, line 434: We were discussing that maybe the code should actually panic if an already confirmed transaction (having the block field populated) is supplied.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -545,27 +536,14 @@ mod tests {
...
-            .status(TxStatus::Succeeded(TransactionBlock {
-                block_hash: Default::default(),
-                block_number: Default::default(),
-            }))
+            .hash(make_tx_hash(3))
+            .block(Default::default())
             .build();
         subject
-            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3, tx4])
+            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone()])
```
bertllll: The clone of tx3 is unnecessary

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -574,166 +552,157 @@ mod tests {
...
         subject
-            .insert_new_records(&vec![tx1, tx2, tx3.clone(), tx4.clone(), tx5])
+            .insert_new_records(&vec![tx1.clone(), tx2.clone()])
             .unwrap();
 
-        let result = subject.retrieve_txs(Some(RetrieveCondition::ToRetry));
+        let result = subject.retrieve_txs(Some(ByHash(vec![tx1.hash])));
 
-        assert_eq!(result, vec![tx3, tx4]);
+        assert_eq!(result, vec![tx1]);
```
bertllll: This function potentially can return a collection of values and so it would be nice to see it in the assertion. Maybe two transactions retrieved and three in the database in total.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -574,166 +552,157 @@ mod tests {
...
+                0,
+                100,
+                1234567890,
+                0,
+                1000000000,
+                1,
+                "0x2345678901234567890123456789012345678901234567890123456789012345",
+                rusqlite::types::Null,
+            ])
             .unwrap();
```
bertllll, line 725: Wouldn't it be simpler if you used the normal `insert_new_records()` fn and then used the `update` sql statement just to null out the single cell? I'd expect much less code.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -574,166 +552,157 @@ mod tests {
...
-                    block_hash: H256::from_low_u64_le(3),
-                    block_number: U64::from(1),
-                }),
-            ),
+            (tx1.hash, tx_block_1.clone()),
+            (tx2.hash, tx_block_2.clone()),
         ]);
 
-        let result = subject.change_statuses(&hash_map);
+        let result = subject.update_tx_blocks(&hash_map);
```
bertllll, line 761: In this test, I'd maybe prefer to have another call like this before the act. So that you can assert that the records had the block column empty before the test itself starts.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+    NoChange,
+    InvalidInput(String),
+    PartialExecution(String),
+    SqlExecutionFailed(String),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FailureReason {
+    PendingTooLong,
+    NonceIssue,
```
bertllll, line 25: We could extend this in the future, right? If we begin to read the topics from the tx receipts...

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FailedTx {
+    pub hash: TxHash,
+    pub receiver_address: Address,
+    pub amount: u128,
+    pub timestamp: i64,
+    pub gas_price_wei: u128,
+    pub nonce: u64,
+    pub reason: FailureReason,
+    pub checked: u8,
```
bertllll: I wonder if `doublechecked` or `rechecked` would give a better idea of the purpose.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+}
+
+pub enum FailureRetrieveCondition {
+    ByReason(FailureReason),
```
bertllll: I'm staring at in awe how you go hardcore with enums. But I kinda like it hihi. :+1:

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+}
+
+impl Display for FailureRetrieveCondition {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            FailureRetrieveCondition::ByReason(reason) => {
+                write!(f, "WHERE reason IS '{:?}'", reason)
```
bertllll: I have mixed opinions about this, though. 

I usually don't like if some parts of the sql are composed hidden in a function. I prefer if all the parts can be seen over by a single glance. This Display impl looks cool, but its true convenience is dubious. Take a while to rethink it.

node/src/accountant/db_access_objects/utils.rs
```diff
@@ -9,18 +9,24 @@ use crate::database::db_initializer::{
...
 use std::fmt::{Debug, Display};
 use std::iter::FlatMap;
 use std::path::{Path, PathBuf};
 use std::string::ToString;
 use std::time::Duration;
 use std::time::SystemTime;
 
+pub type TxHash = H256;
+pub type RowId = u64;
+pub type TxIdentifiers = HashMap<TxHash, RowId>;
```
bertllll, line 28: Honestly, I'd prefer if these weren't so centralized and, instead, were placed near the use in the most relevant file.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+            }
+        }
+    }
+}
+
+pub trait FailedPayableDao {
+    fn get_tx_identifiers(&self, hashes: &HashSet<TxHash>) -> TxIdentifiers;
+    fn insert_new_records(&self, txs: &[FailedTx]) -> Result<(), FailedPayableDaoError>;
+    fn retrieve_txs(&self, condition: Option<FailureRetrieveCondition>) -> Vec<FailedTx>;
```
bertllll, line 69: Unfortunately, we cannot use this method as is for retrieving the LAST (or YET UNCHECKED) failures of pending too long. Your method would return even older and already checked entries, dozens, hundreds, who knows...
The condition for the SELECT needs to be hardened. You must make sure the checked column is empty or '0'

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError>;
+    fn delete_records(&self, hashes: &HashSet<TxHash>) -> Result<(), FailedPayableDaoError>;
+}
+
+#[derive(Debug)]
+pub struct FailedPayableDaoReal<'a> {
+    conn: Box<dyn ConnectionWrapper + 'a>,
```
bertllll, line 76: I would have to check out practically but it might be possible to have this object without the lifetime. Maybe not. :-)

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+
+        let mut stmt = self
+            .conn
+            .prepare(&sql)
+            .unwrap_or_else(|_| panic!("Failed to prepare SQL statement"));
+
+        stmt.query_map([], |row| {
+            let tx_hash_str: String = row.get(0).unwrap_or_default();
+            let tx_hash = TxHash::from_str(&tx_hash_str[2..]).unwrap_or_default();
+            let row_id: u64 = row.get(1).unwrap_or_default();
```
bertllll: Why do you allow all these defaults if unwrapping fails? It is not expected. What is the purpose of returning a key-value pair containing basically `0` and `0`?

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+
+            Ok((tx_hash, row_id))
+        })
+        .unwrap_or_else(|_| panic!("Failed to execute query"))
+        .filter_map(Result::ok)
```
bertllll: This is probably sub-optimal because we're sweeping failures away as if they had no meaning.  

When I used to write db queries, I had the vigilant_flatten() method. It made sure to panic if we stumble over an error this way... I considered that reasonable back then and I haven't changed my mind since, probably. 

My philosophy is: if something doesn't work we always want to know. That's why use unreachable!() and panics so often.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        if txs.is_empty() {
+            return Err(FailedPayableDaoError::EmptyInput);
+        }
+
+        let unique_hashes: HashSet<TxHash> = txs.iter().map(|tx| tx.hash).collect();
+        if unique_hashes.len() != txs.len() {
+            return Err(FailedPayableDaoError::InvalidInput(
+                "Duplicate hashes found in the input".to_string(),
+            ));
+        }
```
bertllll, line 129: Cool

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+
+        if !self.get_tx_identifiers(&unique_hashes).is_empty() {
+            return Err(FailedPayableDaoError::InvalidInput(
+                "Input hash is already present in the database".to_string(),
+            ));
+        }
```
bertllll, line 136: This might be quite expensive. Remember that this is communication between different processes (the db manager). Cannot you simply ignore this preliminary check (which would run every time!) and let the SQL be executed. I guess the insertion would show at its result if you count the number of changed rows (that's the output of execute). I guess we don't have to be so precise here, do we? I think this will run fairly a lot and I don't want to include this check every time so. :man_shrugging: 

On the other hand, if you really feel it like this you can leave it. It's just the overhead what I'm talking about.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+                    )))
+                }
+            }
+            Err(e) => Err(FailedPayableDaoError::SqlExecutionFailed(e.to_string())),
+        }
+    }
+
+    fn retrieve_txs(&self, condition: Option<FailureRetrieveCondition>) -> Vec<FailedTx> {
+        let raw_sql = "SELECT tx_hash, receiver_address, amount_high_b, amount_low_b, \
+        timestamp, gas_price_wei_high_b, gas_price_wei_low_b, nonce, reason, checked FROM failed_payable"
```
bertllll: This SQL is a bit hard to read. Could you try to rearrange it a bit, maybe even with indentation, keywords on a different level than the values?

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+                amount,
+                timestamp,
+                gas_price_wei,
+                nonce,
+                reason,
+                checked,
+            })
+        })
+        .expect("Failed to execute query")
+        .filter_map(Result::ok)
```
bertllll: I'm not a fan of this, you know it.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+    }
+
+    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError> {
+        if hash_map.is_empty() {
+            return Err(FailedPayableDaoError::EmptyInput);
+        }
+
+        for (hash, failure_reason) in hash_map {
+            let sql = format!(
+                "UPDATE failed_payable SET reason = '{:?}' WHERE tx_hash = '{:?}'",
```
bertllll: This is not what I expected.

First, why would we create a record with a wrong failure reason? More over, why would we get into a situation where we want to correct our previous opinion - the former reason?

Second, my understanding is that the only update that would ever happen on this table will be checking off the already back-checked transactions. So from 0 to 1, I suppose. I really cannot think of a different reason which brings me to saying that maybe we don't need such a generically looking method, but it could simply be `check_off_pending_too_long`.

However, go ahead and leave this generically considered interface. Only change it to a different modification in the table. Yours isn't useful in our use case. I hope to be right...

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        .filter_map(Result::ok)
+        .collect()
+    }
+
+    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError> {
+        if hash_map.is_empty() {
+            return Err(FailedPayableDaoError::EmptyInput);
+        }
+
+        for (hash, failure_reason) in hash_map {
```
bertllll: This could be demanding. Individual executions of the queries isn't optimal. 

Ideally, you'd do it in one command but with the use of CASE. 
You have your tests written, so maybe you could ask the AI for some help and experiment a bit.

```
 UPDATE table SET posX=CASE
      WHEN id=id[1] THEN posX[1]
      WHEN id=id[2] THEN posX[2]
      ...
      ELSE posX END [, posY = CASE ... END]
 WHERE id IN (id[1], id[2], id[3]...);
```

bertllll (commented):
continue in node/src/accountant/db_access_objects/failed_payable_dao.rs
at line 314

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        let home_dir =
+            ensure_node_home_directory_exists("failed_payable_dao", "insert_new_records_works");
+        let wrapped_conn = DbInitializerReal::default()
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let tx1 = FailedTxBuilder::default().hash(make_tx_hash(1)).build();
+        let tx2 = FailedTxBuilder::default()
+            .hash(make_tx_hash(2))
+            .reason(PendingTooLong)
+            .checked(1)
```
bertllll: I'm not a fan of this either. 

You're creating unrealistic conditions for this test. 
In the real world we'll never insert a record with `checked` already in. 
It's a question for you if we want to test it before going off to the attempt.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+            .build();
+        let subject = FailedPayableDaoReal::new(wrapped_conn);
+        let txs = vec![tx1, tx2];
+
+        let result = subject.insert_new_records(&txs);
+
+        let retrieved_txs = subject.retrieve_txs(None);
+        assert_eq!(result, Ok(()));
+        assert_eq!(retrieved_txs.len(), 2);
```
bertllll: Redundant, given the next assertion which is stronger and bears the same information.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        let tx1 = FailedTxBuilder::default().hash(hash).build();
+        let tx2 = FailedTxBuilder::default().hash(hash).checked(1).build();
+        let subject = FailedPayableDaoReal::new(wrapped_conn);
+
+        let result = subject.insert_new_records(&vec![tx1, tx2]);
+
+        assert_eq!(
+            result,
+            Err(FailedPayableDaoError::InvalidInput(
+                "Duplicate hashes found in the input".to_string()
```
bertllll: If you wanna keep this check, you should augment the message by the whole collection of hashes supplied (of which some are duplicated).

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        let subject = FailedPayableDaoReal::new(wrapped_conn);
+        let initial_insertion_result = subject.insert_new_records(&vec![tx1]);
+
+        let result = subject.insert_new_records(&vec![tx2]);
+
+        assert_eq!(initial_insertion_result, Ok(()));
+        assert_eq!(
+            result,
+            Err(FailedPayableDaoError::InvalidInput(
+                "Input hash is already present in the database".to_string()
```
bertllll: Also here, that you introduce this error, I'd expect show the already present hash, otherwise this message wouldn't be of big help.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+    fn insert_new_records_returns_err_if_partially_executed() {
+        let setup_conn = Connection::open_in_memory().unwrap();
+        setup_conn
+            .execute("CREATE TABLE example (id integer)", [])
+            .unwrap();
+        let get_tx_identifiers_stmt = setup_conn.prepare("SELECT id FROM example").unwrap();
+        let faulty_insert_stmt = { setup_conn.prepare("SELECT id FROM example").unwrap() };
+        let wrapped_conn = ConnectionWrapperMock::default()
+            .prepare_result(Ok(get_tx_identifiers_stmt))
+            .prepare_result(Ok(faulty_insert_stmt));
```
bertllll, line 482: Here you are preparing for two transactions but eventually the test performs over just one.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+            .hash(make_tx_hash(3))
+            .reason(FailureReason::NonceIssue)
+            .build();
+        subject
+            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone()])
+            .unwrap();
+
+        let result = subject.retrieve_txs(Some(ByReason(PendingTooLong)));
+
+        assert_eq!(result, vec![tx1, tx2]);
```
bertllll: This test should be adjusted and probably return only the first TX. Because that's the only one factually interesting for us. The older ones, already checked, would be just useless balast.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+    }
+
+    #[test]
+    fn update_tx_failures_works() {
```
bertllll: I still don't know what use case you had in your head when you put up this method.

Repeating myself, we will probably ever need just to update the checked column, nothing else should change once inserted, except deletions, those should also be possible.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        let home_dir =
+            ensure_node_home_directory_exists("failed_payable_dao", "update_tx_failures_works");
+        let wrapped_conn = DbInitializerReal::default()
+            .initialize(&home_dir, DbInitializationConfig::test_default())
+            .unwrap();
+        let subject = FailedPayableDaoReal::new(wrapped_conn);
+        let tx1 = FailedTxBuilder::default().hash(make_tx_hash(1)).build();
+        let tx2 = FailedTxBuilder::default().hash(make_tx_hash(2)).build();
```
bertllll: If you meant this test seriously you'd have to build these with the failure reason being different than those you try these update to by the tested method.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+            (existent_hash, PendingTooLong),
+            (non_existent_hash, NonceIssue),
+        ]);
+
+        let result = subject.update_tx_failures(&hash_map);
+
+        assert_eq!(
+            result,
+            Err(FailedPayableDaoError::PartialExecution(format!(
+                "Failed to update failure reason for hash {:?}",
```
bertllll: Maybe mentioning the desired failure reason...

But I think this method may change a lot. I don't want to comment it too much.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        let tx = FailedTxBuilder::default().hash(present_hash).build();
+        subject.insert_new_records(&vec![tx]).unwrap();
+        let hashset = HashSet::from([present_hash, absent_hash]);
+
+        let result = subject.delete_records(&hashset);
+
+        assert_eq!(
+            result,
+            Err(FailedPayableDaoError::PartialExecution(
+                "Only 1 of the 2 hashes has been deleted.".to_string()
```
bertllll: Maybe it would read better without this `the` there?

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -0,0 +1,778 @@
...
+        {
+            DbInitializerReal::default()
+                .initialize(&home_dir, DbInitializationConfig::test_default())
+                .unwrap();
+        }
+        let read_only_conn = Connection::open_with_flags(
+            home_dir.join(DATABASE_FILE),
+            OpenFlags::SQLITE_OPEN_READ_ONLY,
+        )
+        .unwrap();
```
bertllll: If you really wanted to do a bit of refactoring for us, you could create a function with all those lines above this comments up to the test name.  Because these are lines you repeat for each of those methods.

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: I'm doing it because of the `wrapped_conn`, it's complex to reuse it later.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: Definitely 👍 💯

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: I'd say, `rechecked` is better, let me know if I should rename it.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: Thanks! 🤩

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: I believe it's safe to keep it that way since it's a pattern and a read operation.

node/src/accountant/db_access_objects/utils.rs
utkarshg6: It is used by two files: `sent_payable_dao.rs` and `failed_payable_dao.rs`.

node/src/accountant/db_access_objects/failed_payable_dao.rs
bertllll: Yeah, it seems like a good idea, this change. Please use rechecked.

node/src/accountant/db_access_objects/failed_payable_dao.rs
bertllll: Different opinions. :man_shrugging: But I don't want to be oppressive.

node/src/accountant/db_access_objects/utils.rs
bertllll: hmm. ok

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -574,166 +552,157 @@ mod tests {
...
+                100,
+                1234567890,
+                0,
+                1000000000,
+                1,
+                "0x2345678901234567890123456789012345678901234567890123456789012345",
+                rusqlite::types::Null,
+            ])
             .unwrap();
+        }
```
bertllll: Why these blank lines above the subject?

node/src/accountant/db_access_objects/sent_payable_dao.rs
bertllll: There is no conflict though. 

If you can call `.prepare()` and then `.execute()` on it, then you can do whatever you want.
It would allow you to run the update statement. 
Really, try it. 
 
```
 subject
            .insert_new_records(&vec![tx.clone()])
 wrapped_conn
            .prepare("UPDATE failed_payable (block_number) values (null)")
            .unwrap()
            .execute([])
            .unwrap();
```

It should work... but my brain has its circuits in flames at the moment :man_shrugging:

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: It'll require both of them. Since, we need to check for unique hashes.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: I have removed it.

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: I think it would be too restrictive, maybe we want to insert confirmed transactions too.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: I want to explore the option of preventing data corruption by throwing the error earlier.

node/src/accountant/db_access_objects/sent_payable_dao.rs
bertllll: Maybe yes, but please don't write unrealistic tests. 
We know exactly, that our algorithm will always put in a transaction that is not confirmed at the time we are creating the record. 

If I were writing this, I wouldn't hesitate and include the panic constraint. What we are writing is a complex system where every little mistake can break something. I wish I could know of each behavior that we didn't consider as possible. That's why use unreachable!() so much these days. Those are border lines for me, which I don't think we step over but if we do then the code need a serious fix.
That's my take.

node/src/accountant/db_access_objects/failed_payable_dao.rs
bertllll: Did you obey this pls?

I have realized recently that our db operations need to be non-atomic, in other words, all the db changes needs to happen under on a single transaction (with only one commit, respectively).
If we don't we risk data inconsistent states cause by crash, which will be unrecoverable!!! It's just a question of fortune whether this inconsistency causes the Node to wreck, unusable.     

Oh pls, tell me you agree with these arguments. :)

PR #650, opened 2025-06-06
GH-631: Review 2

utkarshg6:


node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -68,7 +75,10 @@ pub trait FailedPayableDao {
     fn get_tx_identifiers(&self, hashes: &HashSet<TxHash>) -> TxIdentifiers;
     fn insert_new_records(&self, txs: &[FailedTx]) -> Result<(), FailedPayableDaoError>;
     fn retrieve_txs(&self, condition: Option<FailureRetrieveCondition>) -> Vec<FailedTx>;
-    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError>;
+    fn update_recheck_status(
+        &self,
+        hash_map: &HashSet<TxHash>,
```
utkarshg6: This should be renamed to `hash_set`.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -528,64 +566,106 @@ mod tests {
...
+            .hash(make_tx_hash(4))
             .reason(FailureReason::NonceIssue)
+            .rechecked(false)
+            .timestamp(now - 3600) // 1 hour ago
             .build();
+        let tx5 = FailedTxBuilder::default()
+            .hash(make_tx_hash(5))
+            .reason(FailureReason::PendingTooLong)
+            .rechecked(false) // This one is rechecked
+            .timestamp(now - 7200) // 2 hours ago
```
utkarshg6: This should be rechecked and within the time boundary, maybe pick 40 minutes.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -627,6 +638,8 @@ mod tests {
         let subject = SentPayableDaoReal::new(wrapped_conn);
         let tx1 = TxBuilder::default().hash(make_tx_hash(1)).build();
         let tx2 = TxBuilder::default().hash(make_tx_hash(2)).build();
+        let tx1_prior_block_opt = tx1.block_opt.clone();
+        let tx2_prior_block_opt = tx2.block_opt.clone();
```
utkarshg6: Instead of `clone()`, use `is_some()`.

node/src/accountant/db_access_objects/test_utils.rs
```diff
@@ -94,7 +108,22 @@ impl FailedTxBuilder {
             reason: self
                 .reason_opt
                 .unwrap_or_else(|| FailureReason::PendingTooLong),
-            checked: self.checked_opt.unwrap_or_default(),
+            rechecked: self.rechecked_opt.unwrap_or_else(|| false),
         }
     }
 }
+
+pub fn make_read_only_db_connection(home_dir: PathBuf) -> ConnectionWrapperReal {
```
bertllll, line 121: I just hope we hadn't had this function already. Could you just somehow quick-check? By a string search with some phrases that come on your mind as possible to be used in such context. Otherwise all good. Great job.

node/src/database/db_initializer.rs
```diff
@@ -727,7 +727,8 @@ mod tests {
         let mut stmt = conn
             .prepare("select name, value, encrypted from config")
             .unwrap();
-        let _ = stmt.query_map([], |_| Ok(42)).unwrap();
+        // TODO: The next line's execution returns Err(ExecuteReturnedResults). Don't know why.
```
bertllll: I think execute in the rusqlite library is meant to return only descriptive numbers not any data itself. 
If you're employing an SQL statement which is supposed to return a value (the one stored in the database column), you need to use `query`-like methods. For example `query_map`.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -47,18 +47,25 @@ pub struct FailedTx {
     pub gas_price_wei: u128,
     pub nonce: u64,
     pub reason: FailureReason,
-    pub checked: u8,
+    pub rechecked: bool,
 }
 
 pub enum FailureRetrieveCondition {
-    ByReason(FailureReason),
+    UncheckedPendingTooLong(u32), // u32 represents seconds ago
```
bertllll: Why do we need to work with seconds? :thinking: 

That's weird, I think we only consider a single "pending payable interval" the boundary of the lapse. It should technically the interval configured in the body of the Accountant (either given by the user or defaulted). However, given the strangeness of the real world conditions, this number of seconds will always differ. (It depends on the message queue). I just wonder if we do want to know...

Maybe yes, but only for some kind of talkative logging

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -47,18 +47,25 @@ pub struct FailedTx {
...
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            FailureRetrieveCondition::ByReason(reason) => {
-                write!(f, "WHERE reason IS '{:?}'", reason)
+            FailureRetrieveCondition::UncheckedPendingTooLong(seconds_ago) => {
+                let timestamp_threshold = current_unix_timestamp() - *seconds_ago as i64;
+                write!(
+                    f,
+                    "WHERE reason = 'PendingTooLong' AND rechecked = 0 \
+                     AND timestamp >= {} \
```
bertllll: Why do we have to care about the time?

You think we cannot rely on the `rechecked` column? We had some discussion and I think I give you enough trust in what I was saying that you now follow...
Let's keep time out of this whole business. The line 64 is all we need.
 
In every scan cycle we should always update the latest PendingTooLong errors from 0 to 1. Therefore we always can rely on the 0. If we ask about the last, we simply ask about these 0s.  

(Note: I mightn't have explained enough but this function isn't used regularly. I suppose we will call it only once in the life span of the Node. Otherwise, we will only keep writing these records in and also mirroring those in the runtime memory just to avoid the need of re-reading them like this so often...it'll be much easier to work with the hashmap. But if the Node panics, the hashmap will always be backed up in the persistent db).

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -68,7 +75,10 @@ pub trait FailedPayableDao {
     fn get_tx_identifiers(&self, hashes: &HashSet<TxHash>) -> TxIdentifiers;
     fn insert_new_records(&self, txs: &[FailedTx]) -> Result<(), FailedPayableDaoError>;
     fn retrieve_txs(&self, condition: Option<FailureRetrieveCondition>) -> Vec<FailedTx>;
-    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError>;
+    fn update_recheck_status(
```
bertllll: This should probably be either "update_status_rechecked" or simply "mark_as_rechecked".

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -209,7 +231,8 @@ impl FailedPayableDao for FailedPayableDaoReal<'_> {
             let reason_str: String = row.get(8).expectv("reason");
             let reason =
                 FailureReason::from_str(&reason_str).expect("Failed to parse FailureReason");
-            let checked: u8 = row.get(9).expectv("checked");
+            let checked_integer: u8 = row.get(9).expectv("rechecked");
```
bertllll: Maybe "rechecked_as_int"?

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -362,15 +386,29 @@ mod tests {
             .unwrap();
         let hash = make_tx_hash(123);
         let tx1 = FailedTxBuilder::default().hash(hash).build();
-        let tx2 = FailedTxBuilder::default().hash(hash).checked(1).build();
+        let tx2 = FailedTxBuilder::default()
+            .hash(hash)
+            .rechecked(true)
+            .build();
```
bertllll, line 386: Again... this recheck == true isn't technically possible for "insert_new_records". 
It irks me. :scream:

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -386,7 +424,10 @@ mod tests {
             .unwrap();
         let hash = make_tx_hash(123);
         let tx1 = FailedTxBuilder::default().hash(hash).build();
-        let tx2 = FailedTxBuilder::default().hash(hash).checked(1).build();
+        let tx2 = FailedTxBuilder::default()
+            .hash(hash)
+            .rechecked(true)
```
bertllll, line 423: Same here and in so many other places :cry:

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -528,64 +566,106 @@ mod tests {
...
+
         subject
-            .insert_new_records(&vec![tx1.clone(), tx2.clone(), tx3.clone()])
+            .insert_new_records(&vec![tx1.clone(), tx2, tx3.clone(), tx4, tx5])
             .unwrap();
 
-        let result = subject.retrieve_txs(Some(ByReason(PendingTooLong)));
-
-        assert_eq!(result, vec![tx1, tx2]);
+        // Retrieve unchecked PendingTooLong transactions from the last hour
```
bertllll: I don't really know what the time has to do with this. 

Could you please explain?

Edit: There is already a comment from me at the code this touches in particular. No need to reply to this question...

bertllll (commented):
continue in node/src/accountant/db_access_objects/failed_payable_dao.rs
`can_retrieve_unchecked_pending_too_long_txs`

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -68,7 +75,10 @@ pub trait FailedPayableDao {
     fn get_tx_identifiers(&self, hashes: &HashSet<TxHash>) -> TxIdentifiers;
     fn insert_new_records(&self, txs: &[FailedTx]) -> Result<(), FailedPayableDaoError>;
     fn retrieve_txs(&self, condition: Option<FailureRetrieveCondition>) -> Vec<FailedTx>;
-    fn update_tx_failures(&self, hash_map: &FailureUpdates) -> Result<(), FailedPayableDaoError>;
+    fn update_recheck_status(
+        &self,
+        hash_set: &HashSet<TxHash>,
+    ) -> Result<(), FailedPayableDaoError>;
     fn delete_records(&self, hashes: &HashSet<TxHash>) -> Result<(), FailedPayableDaoError>;
 }
```
bertllll, line 72: I can't comment on the `SentPayableDao` trait in this iteration, but I need to add an important request on you. 

We have discussed it in the chat roughly already. 

The thing is that we've got a void for the design of the procedure when we uncover that a transaction tracked as failed for pending too long was tagged so unrightfully. 

We have defined correctly that it should be taken out from the failed payable table. Now I think that actually is just the least important part. 

With regard to a possible panic-caused data loss and the resulted database inconsistent state dragging this system into the lands of no recovery, it is important to think about the steps so that the database must stay sane and serve its purpose even after a sudden death of the application. We need to consider the vulnerability at every moment separately. 
    
The steps should be:
a) rewrite the newer (successor) record dwelling recently in sent_payable of the nonce N with the data of the older transaction with the same nonce, now read from the failed_payable table. I say rewrite because I want to avoid the db state inconsistency precisely.
That means, I'd like to have a single method to call, which will internally perform: the old row deletion + insertion of the new row, but committed no earlier than both of them done. I say we should use an explicit rusqlite tx, that gives you the control over when it is committed. 
However, you may be able to work it around by writing an UPDATE sql which will update the whole row though (every valuei in the row.) I wonder if that is considered a hack in the community of the data-engineers? (According to the ChatGPT, it isn't a nonsense. I'd then highly recommend that as it is much simpler to implement)  
b) delete the record from failed_payble 

Maybe 
`fn replace_record()`
or
`fn rewrite_whole_record()`

node/src/accountant/db_access_objects/failed_payable_dao.rs
bertllll: Also, in relation to the confusions regarding the time test upon the retrieval of the pending payable, I wanna request you to change this method so that it actually always flip the flag of EVERY record conditioned as 
`reason = PendingTooLong AND rechecked = 0`.

Please raise an argument, if you can think of a good one, why we shouldn't do it. I myself reason about that 
the ultimate state which we want to achieve is supposed to be consistent with this idea I put forth. We actually EXPECT every 0 to change to 1 always just after a single cycle finishes, none of these 0 should stay longer than that to be resolved sometime later. That's not what we want. It definitely should be framed only by the stretch of one cycle. When we fully adopt this thinking we can perhaps agree on the stricter, but simpler at the same time, conditions I suggest. 

If you worry what if we are missing something, because I am partially hit by these questions too, you may free your spirit. I have given it some more time and assured myself that eventually, if this absoluteness (that means the use of no time parameter in the query) did not prove true, our whole mechanism will break down. And that will give us the answer. People with this issue would report the payments not working. In their logs, there would be an evidence of an endless loop of attempts, and we'd find out, checking the blockchain that their further transactions were always rejected due to the use of the wrong nonce.  
We can fix it then accordingly.     
     
Notice that this means that this method would need no additional arguments! :wink:

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -528,64 +566,106 @@ mod tests {
...
+        let tx3_pre_checked_state = tx3.rechecked;
         subject
-            .insert_new_records(&vec![tx1.clone(), tx2.clone()])
+            .insert_new_records(&vec![tx1, tx2.clone(), tx3.clone()])
             .unwrap();
-        let hash_map = HashMap::from([(tx1.hash, PendingTooLong), (tx2.hash, NonceIssue)]);
+        let hash_set = HashSet::from([tx2.hash, tx3.hash]);
 
-        let result = subject.update_tx_failures(&hash_map);
+        let result = subject.update_recheck_status(&hash_set);
```
bertllll: As stated by the trait definition, this method is suspected of a full loss of args, because it probably won't need them anymore.

node/src/accountant/db_access_objects/test_utils.rs
utkarshg6: I found `trick_rusqlite_with_read_only_conn`, which creates a read-only database connection but also creates a table.

node/src/database/db_initializer.rs
utkarshg6: Ok, then I'll just revert it back to the original statement.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: Ok, got it! I think I should eliminate it.

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: I was considering whether this function should be placed outside of these two DAOs, as it doesn't fully belong to either one. What are your thoughts?

bertllll: Nope, I believe it fits perfectly fine to the SentPayableDAO.

Please write it as a simple update statement which changes every value in the respective row. It's actually an idiomatic approach.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -139,6 +132,13 @@ impl FailedPayableDao for FailedPayableDaoReal<'_> {
             )));
         }
 
+        if let Some(_rechecked_tx) = txs.iter().find(|tx| tx.rechecked) {
+            return Err(FailedPayableDaoError::InvalidInput(format!(
+                "Already rechecked transaction(s) provided: {:?}",
+                txs
+            )));
+        }
+
```
bertllll, line 137: Thanks for this :hugs:

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -250,32 +250,18 @@ impl FailedPayableDao for FailedPayableDaoReal<'_> {
...
-                    Ok(())
-                } else {
-                    Err(FailedPayableDaoError::PartialExecution(format!(
-                        "Only {} out of {} records updated",
-                        updated_rows,
-                        hash_set.len()
-                    )))
-                }
-            }
+            Ok(_rows) => Ok(()),
```
bertllll: You could theoretically finish this branch the way you usually do. In terms of consistency, I'd prefer if you did that actually. 

How to proceed? Up there at the beginning of this method, you fetches all unchecked pending too long. That means you know how many of those there are. You can later check it against this returned number of updated 
rows ;) Those two numbers should match.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -51,20 +50,17 @@ pub struct FailedTx {
...
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            FailureRetrieveCondition::UncheckedPendingTooLong(seconds_ago) => {
-                let timestamp_threshold = current_unix_timestamp() - *seconds_ago as i64;
+            FailureRetrieveCondition::UncheckedPendingTooLong => {
                 write!(
                     f,
                     "WHERE reason = 'PendingTooLong' AND rechecked = 0 \
-                     AND timestamp >= {} \
                      ORDER BY timestamp DESC",
```
bertllll: I bet we don't even need to order the records now. It's of no use. We will treat them all the same. 

Feel free to delete that condition. It shouldn't kill your tests.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -578,50 +592,34 @@ mod tests {
...
         subject
-            .insert_new_records(&vec![tx1.clone(), tx2, tx3.clone(), tx4, tx5])
+            .insert_new_records(&vec![tx1.clone(), tx2, tx3.clone()])
             .unwrap();
 
-        // Retrieve unchecked PendingTooLong transactions from the last hour
-        let result = subject.retrieve_txs(Some(FailureRetrieveCondition::UncheckedPendingTooLong(
-            3600,
-        )));
+        let result = subject.retrieve_txs(Some(FailureRetrieveCondition::UncheckedPendingTooLong));
```
bertllll, line 621: Actually, this should be the act of this test. Don't isolate `insert_new_records`. (It's the fake prophet :laughing: )

PR #651, opened 2025-06-06
GH-602: Scanners system reconstruction. Review two.

bertllll:


utkarshg6 (commented):
Continue at node/src/test_utils/recorder_counter_msgs.rs.

PS: You need coffee to shake off the fatigue from today's travels.

node/src/test_utils/recorder_counter_msgs.rs
```diff
@@ -46,64 +60,72 @@ where
...
-                vacant.insert(vec![counter_msg_setup]);
+            Entry::Occupied(mut existing_vec) => existing_vec.get_mut().push(counter_msg_setup),
+            Entry::Vacant(vacancy) => {
+                vacancy.insert(vec![counter_msg_setup]);
             }
         }
     }
 }
 
+// Note that you're not limited to triggering an only message a time, but you can supply more
```
utkarshg6: You might prefer to change the following statement:

`triggering an only message a time` -> `triggering only one message at a time`

utkarshg6 (commented):
Continue at node/src/test_utils/recorder_stop_conditions.rs:355

node/src/test_utils/recorder.rs
```diff
@@ -824,104 +826,133 @@ mod tests {
...
-                respondent_addr.clone().recipient(),
-            ))],
-        );
-        let msg_4_matching = NewPublicIp {
-            new_ip: IpAddr::V4(Ipv4Addr::new(4, 5, 6, 7)),
+        // Case two
+        let cm_setup_2 = {
+            let counter_msg_strayed = StartMessage {};
+            let random_id = TypeId::of::<BlockchainBridge>();
+            let id_method = MsgIdentification::ByType(random_id);
```
utkarshg6: The `random_id` name is not good enough, it'll be better if we can point out that it's `screwed_id` for the trigger message. You'd like to give it a better name later on.

node/src/test_utils/recorder.rs
```diff
@@ -824,104 +826,133 @@ mod tests {
...
+            let counter_msg = NodeToUiMessage {
+                target: MessageTarget::ClientId(4),
+                body: UiUnmarshalError {
+                    message: "abc".to_string(),
+                    bad_data: "456".to_string(),
+                }
+                .tmb(0),
+            };
+            let id_method = MsgIdentification::ByMatch {
+                exemplar: Box::new(NewPublicIp { new_ip: make_ip(1) }),
```
utkarshg6: Maybe you want the ip to be different but constructed using a similar methodology.

node/src/test_utils/recorder.rs
```diff
@@ -824,104 +826,133 @@ mod tests {
             NodeToUiMessage
```
utkarshg6, line 825: Ultra Optional: If you'd like to figure out a way to improve this test further, give it a shot, but don't delve too deep into it, a TODO comment is fine too (along with your strategy), otherwise you can just ignore this comment.

utkarshg6 (commented):
Continue at node/src/accountant/scanners/scan_schedulers.rs

utkarshg6 (commented):
Continue at node/src/accountant/scanners/scan_schedulers.rs

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -302,20 +333,34 @@ impl RescheduleScanOnErrorResolverReal {
...
             }
         } else if err == &StartScanError::NoConsumingWalletFound {
-            SchedulingAfterHaltedScan::Schedule(ScanType::Payables)
+            if initial_pending_payable_scan {
+                // Cannot deduce there are strayed pending payables from the previous Node's run
+                // (StartScanError::NoConsumingWalletFound is thrown before
+                // StartScanError::NothingToProcess can be evaluated); but may be cautious and
+                // prevent starting the NewPayableScanner. Repeating this scan endlessly may alarm
+                // the user.
+                // TODO Correctly, a check-point during the bootstrap should be the solution,
```
utkarshg6: In case you can provide the card number too, that'll be awesome. ✌️

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -555,67 +588,87 @@ mod tests {
...
         assert_eq!(
             result,
-            SchedulingAfterHaltedScan::Schedule(ScanType::Payables),
+            ScanRescheduleAfterEarlyStop::Schedule(ScanType::Payables),
             "We expected Schedule(Payables) but got {:?}",
             result,
         );
+        TestLogHandler::new().exists_log_containing(&format!(
+            "INFO: {test_name}: Rescheduling strategy 'Schedule(Payables)' was chosen after \
+            the automatic PendingPayables scan encountered the 'NothingToProcess' error"
```
utkarshg6: I was questioning whether we should notify the user during the automatic scan process. I am thinking that for manually_triggered, it should be `info!()`, otherwise `debug!()`. What do you think?

utkarshg6 (commented):
Continue at node/src/accountant/scanners/mod.rs

utkarshg6 (commented):
node/src/accountant/scanners/mod.rs:1354

node/src/accountant/scanners/mod.rs
```diff
@@ -4510,105 +4251,102 @@ mod tests {
...
-        });
+                let expected_log_msg =
+                    form_expected_log_msg(severity_for_externally_triggered_scans);
+                test_log_handler.exists_log_containing(&expected_log_msg);
+
+                err.log_error(&logger, ScanType::Payables, false);
+
+                let expected_log_msg = form_expected_log_msg(severity_for_automatic_scans);
+                test_log_handler.exists_log_containing(&expected_log_msg);
+            },
```
utkarshg6, line 4348: Perhaps, you'd like to use an iterator:

```rust
        let cases = [
            (true, severity_for_externally_triggered_scans),
            (false, severity_for_automatic_scans),
        ];

        cases.iter().for_each(|(is_externally_triggered, severity)| {
            err.log_error(&logger, ScanType::Payables, *is_externally_triggered);

            let expected_log_msg = form_expected_log_msg(severity);
            test_log_handler.exists_log_containing(&expected_log_msg);
        });
```

utkarshg6 (commented):

node/src/accountant/mod.rs
```diff
@@ -260,18 +291,18 @@ impl Handler<ScanForRetryPayables> for Accountant {
 
     fn handle(&mut self, msg: ScanForRetryPayables, _ctx: &mut Self::Context) -> Self::Result {
         // RetryPayableScanner is scheduled only when the PendingPayableScanner finishes discovering
-        // that there have been some failed pending payables. No place for it here.
+        // that there have been some failed payables. No place for that here.
         let response_skeleton = msg.response_skeleton_opt;
-        let _ = self.handle_request_of_scan_for_retry_payable(response_skeleton);
+        self.handle_request_of_scan_for_retry_payable(response_skeleton);
```
utkarshg6, line 296: I wonder why don't you need to match on this one?

node/src/accountant/mod.rs
```diff
@@ -2606,6 +2640,222 @@ mod tests {
...
+        // I do believe it's impossible for these two events to happen within the same nanosecond,
+        // until somebody fiddles with MASQ tests on a true supercomputer.
+        assert!(
+            pp_scan_started_at < p_scheduling_happened_at,
+            "We failed to prove that the PendingPayableScanner runs before the PayableScanner."
+        );
+    }
+
+    #[test]
+    fn accountant_scans_after_startup_and_detects_straggled_pending_payable() {
```
utkarshg6: How about if we call it, `accountant_detects_straggling_pending_payable_during_startup`.

utkarshg6 (commented):
Continue at node/src/accountant/mod.rs:2696

node/src/accountant/mod.rs
```diff
@@ -2606,6 +2640,222 @@ mod tests {
...
+            test_name,
+            earning_wallet,
+            receivable_start_scan_params_arc,
+            receivable_scan_interval,
+            scan_for_receivables_notify_later_params_arc,
+        );
+        // Given the assertions proving that the pending payable scanner will run multiple times
+        // before the new payable scanner runs at least once (even not scheduled yet), its front
+        // position is clear
+    }
```
utkarshg6, line 2837: This test is a good addition. 👍 

In case it doesn't take much of your time, could you please group similar items and pass the whole objects to the helper functions? I mean something like this.


```rust
struct ScanParams {
   payable_start_scan: Arc<Mutex<Vec<...>>>,
   payable_finish_scan: Arc<Mutex<Vec<...>>>,
   pending_payable_start_scan: Arc<Mutex<Vec<...>>>,
   pending_payable_finish_scan: Arc<Mutex<Vec<...>>>,
   receivable_start_scan: Arc<Mutex<Vec<...>>>,
   receivable_finish_scan: Arc<Mutex<Vec<...>>>,
}

struct NotifyAndNotifyLaterParams {
    pending_payables_notify_later: Arc<Mutex<Vec<...>>>,
    new_payables_notify_later: Arc<Mutex<Vec<...>>>,
    new_payables_notify: Arc<Mutex<Vec<...>>>,
    retry_payables_notify: Arc<Mutex<Vec<...>>>,
    receivables_notify_later: Arc<Mutex<Vec<...>>>,
}

struct MockedScanners {
    pending_payable: ScannerMock<...>,
    receivable: ScannerMock<...>,
    payable: ScannerMock<...>,
}
```

Once you do, please use the `default()` call for these structs in the preparatory lines.

Also, the helper functions might look like this:

```rust
set_up_subject_for_some_pending_payable_found_startup_test(
   test_name,
   config,
   mocked_scanners,
   &notify_and_notify_later_params
)
```

For the assert functions, it would be better to pass the whole objects instead of the specific fields of these newly created structs.

```rust
assert_pending_payable_scanner_for_some_pending_payable_found(
   test_name,
   consuming_wallet,
   &scan_params,
   &notify_and_notify_later_params,
   expected_report_transaction_receipts
);

// Similarly the others
```

node/src/accountant/mod.rs
```diff
@@ -2638,32 +2888,183 @@ mod tests {
...
-                .stop_system_on_count_received(1),
+            NotifyHandleMock::default().notify_params(&scan_for_new_payables_notify_params_arc),
         );
-        // Important that this is short because the test relies on it to stop the system.
+        let receivable_notify_later_handle_mock = NotifyLaterHandleMock::default()
+            .notify_later_params(&scan_for_receivables_notify_later_params_arc)
+            .stop_system_on_count_received(1);
+        subject.scan_schedulers.receivable.handle = Box::new(receivable_notify_later_handle_mock);
+        // Important that this is made short because the test relies on it with the system stop.
         let receivable_scan_interval = Duration::from_millis(50);
```
utkarshg6: Could you please also migrate the instantiation of `new_payable_expected_computed_interval` and `receivable_scan_interval` to the top, along with the subject?

node/src/accountant/mod.rs
```diff
@@ -2638,32 +2888,183 @@ mod tests {
...
+            .replace_scanner(ScannerReplacement::Payable(ReplacementType::Mock(
+                payable_scanner,
+            )));
+        let pending_payable_notify_later_handle_mock = NotifyLaterHandleMock::default()
+            .notify_later_params(&scan_for_pending_payables_notify_later_params_arc)
+            // This should stop the system
+            .stop_system_on_count_received(1);
+        subject.scan_schedulers.pending_payable.handle =
+            Box::new(pending_payable_notify_later_handle_mock);
+        let pending_payable_scan_interval = Duration::from_secs(3600);
```
utkarshg6: Also here

node/src/accountant/mod.rs
```diff
@@ -2730,39 +3130,117 @@ mod tests {
...
+                    response_skeleton_opt: None
                 },
                 receivable_scan_interval
             )]
         );
-        // Finally, an assertion to prove course of actions in time. We expect the pending payable
-        // scanner to take place before the payable scanner.
-        // I do believe it's impossible that these two instants would happen at the same time, until
-        // this is run on a real super-giga-computer.
-        assert!(pp_scan_started_at < p_scheduling_now);
```
utkarshg6, line 2765: Would you mind sharing your thoughts on whether or not we still need this assertion?

node/src/accountant/mod.rs
```diff
@@ -2773,15 +3251,22 @@ mod tests {
...
 
+        System::current().stop();
+        system.run();
         let flag_after = subject.scanners.initial_pending_payable_scan();
-        assert_eq!(hint, SchedulingAfterHaltedScan::DoNotSchedule);
+        assert_eq!(hint, ScanRescheduleAfterEarlyStop::DoNotSchedule);
         assert_eq!(flag_before, true);
         assert_eq!(flag_after, false);
+        let blockchain_bridge_recording = blockchain_bridge_recording_arc.lock().unwrap();
+        let _recorded_msg = blockchain_bridge_recording.get_record::<RequestTransactionReceipts>(0);
```
utkarshg6: Instead of `get_record()`, you may use `get_record_opt`, and perhaps do something like this:

```rust
assert!(blockchain_bridge_recording.get_record_opt::<RequestTransactionReceipts>(0).is_some());
```

utkarshg6 (commented):
Continue at node/src/accountant/mod.rs:4068

node/src/accountant/mod.rs
```diff
@@ -3523,6 +4065,73 @@ mod tests {
...
+        subject.scan_schedulers.pending_payable.handle = Box::new(
+            NotifyLaterHandleMock::default()
+                .notify_later_params(&pending_payables_notify_later_params_arc)
+                .stop_system_on_count_received(1),
+        );
+        subject.scan_schedulers.pending_payable.interval = Duration::from_secs(60);
+        subject.scan_schedulers.payable.new_payable_notify =
+            Box::new(NotifyHandleMock::default().notify_params(&new_payables_notify_params_arc));
+        let system = System::new("test");
+        let subject_addr = subject.start();
```
utkarshg6, line 4107: I was wondering if it would be possible to explicitly represent the absence of a wallet.

utkarshg6 (commented):
Good Luck! I wish you the best. ✌️

node/src/accountant/scanners/scan_schedulers.rs
bertllll: It actually is more serious in the automation. 
With the manual scan, nothing really can get harm. 
On the other side, though, if the automatic scans run through these errors it's a bit eyebrow-raising. Actually, in most of the case, the Node even panics with no mercy. Those are situations considered not healthy or possible in the whole schema. 

The thing is the user will be hit by an error log before this info log. It logs the actual StartScanError. Before we even decide what to do next and whether we can manage to keep the scanning on or if we even give up entirely and kill the application (at only very serious accusations).

So, honestly, I'm thinking now that it's rather the other way around; the info log is more meaningful in the automatic scanning. (It explains that some kind of emergency scan scheduling had to be done to address the occurrence of the StartScanError)

bertllll: I chose a different approach. Realizing that there is already one log before this one that explains the StartScanError which is the cause of the rescheduling, I made message much briefer here and change it to be always DEBUG. It's of no particular use to the user in either way, to be honest. ;-)

node/src/accountant/scanners/mod.rs
bertllll: This isn't that easy. If I went your way, I'd have to supply the error and the form_expected_msg function for both cases to which I wanted to avoid.

node/src/accountant/mod.rs
bertllll: It almost always panic, except if externally triggered. That's why. I completely removed the return value. 

This is right. This scanner stands behind the PendingPayableScanner and it shield off all the problems at that stage, so there is really just a little what can happen with this next scan. And if it does, then something is really wrong.

node/src/accountant/mod.rs
bertllll: It is really the same thing. 
We wouldn't pass this line, if we didn't receive the RequestTransactionReceipts there

bertllll: It is alone a strong enough assertion. You cannot pass it falsely unless you panic.

bertllll: Haha. I accidentally replied twice on the same subject. Well, three times it is by now. :)

PR #653, opened 2025-06-15
GH-631: Review 3

utkarshg6:


node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -268,6 +270,61 @@ impl SentPayableDao for SentPayableDaoReal<'_> {
...
+               receiver_address = ?2,
+               amount_high_b = ?3,
+               amount_low_b = ?4,
+               timestamp = ?5,
+               gas_price_wei_high_b = ?6,
+               gas_price_wei_low_b = ?7,
+               nonce = ?8,
+               block_hash = ?9,
+               block_number = ?10
+               WHERE nonce = ?11";
```
bertllll: This is interesting. I've never used this syntax from rusqlite (it's a metalanguage). But I understand it, so no problem. 

I think it might work even without the enumeration of the question marks (as to what I remember from what their `params!` macro allows...)

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -60,6 +61,7 @@ pub trait SentPayableDao {
     fn insert_new_records(&self, txs: &[Tx]) -> Result<(), SentPayableDaoError>;
     fn retrieve_txs(&self, condition: Option<RetrieveCondition>) -> Vec<Tx>;
     fn update_tx_blocks(&self, hash_map: &TxUpdates) -> Result<(), SentPayableDaoError>;
+    fn replace_record(&self, hash_map: &Tx) -> Result<(), SentPayableDaoError>;
```
bertllll: You say it's a hash_map here but it actually is not. Either change it to address it by a right name or adjust it accordingly if you decide to allow supplying a collection.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -60,6 +61,7 @@ pub trait SentPayableDao {
     fn insert_new_records(&self, txs: &[Tx]) -> Result<(), SentPayableDaoError>;
     fn retrieve_txs(&self, condition: Option<RetrieveCondition>) -> Vec<Tx>;
     fn update_tx_blocks(&self, hash_map: &TxUpdates) -> Result<(), SentPayableDaoError>;
```
bertllll: Honestly, I don't like confusion this data alias can cause, because it hides the outer structure, the sort of the collection. I'd find this information handy, so I'm thinking it makes sense to type-alias values or even keys in a hash map but if you wrap up even the HashMap type with it, it's not really helping, rather the opposite.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -268,6 +270,61 @@ impl SentPayableDao for SentPayableDaoReal<'_> {
         Ok(())
     }
 
+    fn replace_record(&self, tx: &Tx) -> Result<(), SentPayableDaoError> {
```
bertllll: Okay, I've done some thinking regarding the concern of yours you told me about. My conclusion is, though, that it shouldn't cause any troubles even if we detect more than one falsely-failed transaction that we'd like to revive. I think we should allow an update possibly catering multiple replacements in a single operation. :pray: I'm taking the responsibility on myself :smiling_face_with_tear:

node/src/accountant/db_access_objects/sent_payable_dao.rs
utkarshg6: I attempted that, but it seems to require the question marks.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -941,42 +962,111 @@ mod tests {
...
+        let new_tx3 = TxBuilder::default()
+            .hash(make_tx_hash(33))
+            .block(TransactionBlock {
+                block_hash: make_block_hash(1),
+                block_number: U64::from(1),
+            })
+            .nonce(3)
+            .build();
+
+        let result = subject.replace_records(&[new_tx2.clone(), new_tx3.clone()]);
```
bertllll: Redundant clones

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -295,33 +298,49 @@ impl SentPayableDao for SentPayableDaoReal<'_> {
...
-        match result {
-            Ok(updated_rows) => {
-                if updated_rows == 1 {
-                    Ok(())
-                } else {
-                    Err(SentPayableDaoError::NoChange)
-                }
+        let mut updated_count = 0;
+
+        for new_tx in new_txs {
```
bertllll: Please :pray try rewriting this into the UPDATE + CASE stm. 

Maybe you will find it on the internet or the AI will generate it for you.

Or I'll try myself:

``` 
UPDATE failed_payable 
SET 
    tx_hash = CASE 
        WHEN nonce = 1 THEN 'txhash1' 
        WHEN nonce = 2 THEN 'txhash2'
        WHEN nonce = 3 THEN 'txhash3'
    END
    amount_high_b = CASE 
        WHEN nonce = 1 THEN 0
        WHEN nonce = 2 THEN 0 // This is not hard-coded; it could be theoretically other than 0
        WHEN nonce = 3 THEN 0
    END,
    amount_low_b = CASE 
        WHEN nonce = 1 THEN 4587897
        WHEN nonce = 2 THEN 8787979987
        WHEN nonce = 3 THEN 12874585
    END,
    timestamp = CASE 
        WHEN nonce = 1 THEN 134565658
        WHEN nonce = 2 THEN 134554688
        WHEN nonce = 3 THEN 13455599
    END
    gas_price_wei_high_b = CASE 
        WHEN nonce = 1 THEN 0
        WHEN nonce = 2 THEN 0 // Also here, the value should be provided, not assumed 0. 
        WHEN nonce = 3 THEN 0
    END
    gas_price_wei_low_b = CASE 
        WHEN nonce = 1 THEN 487879849
        WHEN nonce = 2 THEN 418798787
        WHEN nonce = 3 THEN 878787888
    END
    block_hash = CASE 
        WHEN nonce = 1 THEN 'blockHash1'
        WHEN nonce = 2 THEN 'blockHash2'
        WHEN nonce = 3 THEN 'blockHash3'
    END
    block_number = CASE 
        WHEN nonce = 1 THEN 4878787860
        WHEN nonce = 2 THEN 4878787877
        WHEN nonce = 3 THEN 4878787855
    END
WHERE nonce IN (1, 2, 3);

```

You see? It's kinda complex in terms of syntax, I agree, but should be efficient and especially, it's atomic. 

You'd have to find a way to construct this statement above as a string and then execute with no additional params supplied (I only bet this would be the best approach).

The positive thing is that you won't necessarily have to write any new tests. 
(However, I know how it could be done. With a fake connection. You could capture the params of the `prepare` method. When you assert on them you will see that it was called just once, not multiple times like if you approach each record separately just after you finish the previous. This kind of test would be really strong, but I don't know if you want to go that far. I'm willing to see you just introduce this different statement and be blind to any emerged needs for further testing.)

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -978,6 +1007,43 @@ mod tests {
...
+        let subject = SentPayableDaoReal::new(Box::new(wrapped_conn));
+        let tx1 = TxBuilder::default().hash(make_tx_hash(1)).nonce(1).build();
+        let tx2 = TxBuilder::default().hash(make_tx_hash(2)).nonce(2).build();
+        let tx3 = TxBuilder::default().hash(make_tx_hash(3)).nonce(3).build();
+
+        let _ = subject.replace_records(&[tx1, tx2, tx3]);
+
+        let captured_params = prepare_params.lock().unwrap();
+        let sql = &captured_params[0];
+        assert_eq!(captured_params.len(), 1);
```
bertllll: This assertion should be at the end. So that you allow to check the values from the first capture before this trivial assertion blows off.

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -280,67 +279,96 @@ impl SentPayableDao for SentPayableDaoReal<'_> {
...
+        let block_number_cases = build_case(|tx| match &tx.block_opt {
+            Some(block) => block.block_number.as_u64().to_string(),
+            None => "NULL".to_string(),
+        });
+
+        let nonces = new_txs
+            .iter()
+            .map(|tx| tx.nonce.to_string())
+            .collect::<Vec<_>>()
+            .join(", ");
```
bertllll: This could be simplifies by using “comma_separated_stringifyable”

node/src/accountant/db_access_objects/sent_payable_dao.rs
```diff
@@ -280,67 +279,96 @@ impl SentPayableDao for SentPayableDaoReal<'_> {
...
-               gas_price_wei_low_b = ?7,
-               nonce = ?8,
-               block_hash = ?9,
-               block_number = ?10
-               WHERE nonce = ?11";
+        let build_case = |value_fn: fn(&Tx) -> String| {
+            new_txs
+                .iter()
+                .map(|tx| format!("WHEN nonce = {} THEN {}", tx.nonce, value_fn(tx)))
+                .collect::<Vec<_>>()
```
bertllll: The collect() is redundant, the join() fn can replace its function entirely

utkarshg6: The `collect()` is necessary.

Issue #622, opened 2025-04-25
Establish And Maintain Public Key


dnwiebe:
There's a business requirement to allow a Node to keep the same public key over a shutdown/startup cycle.

We'll do this by having a newly-installed Node pick its own key pair and store it in the database; then assume it on startup each time. (Another card (GH-623) will allow the user to specify `--new-public-key on` or `--new-public-key off` on the command line; but that's not this card.)

* `CryptDEReal` has two important fields:
	`encryption_secret_key`
	`signing_secret_key`
  You should be able to reconstitute an entire `CryptDEReal` from just these two fields.

* These two fields should be stored in the `CONFIG` table in the database. Since they're really different parts of the same data object, there's a good argument for storing both fields in the same row. Suggestion: Get the binary data from each secret key, Base64-encode it into a string, and store both strings in the `VALUE` column of the `CONFIG` table, separated by some character that's not used in the Base64 encoding, like space or comma.

* Be sure to save these fields as encrypted: they're secret keys.

* This will require a database migration. Migrate from version 10 to version 11, adding a row to the `CONFIG` table. Look at `migration_0_to_1.rs` for an example.

* Add `From<String>` and `Into<String>` methods to `CryptDE` and implement them in both `CryptDEReal` and `CryptDENull`. Add a pair of methods to `PersistentConfiguration` for reading and writing `CryptDE`s to the database, using the new `From` and `Into` implementations.

* Definitely write the main `CryptDE` to the database. You can probably get away without writing the alias `CryptDE` to the database. The alias `CryptDE` shouldn't need to maintain its value between Node runs, only between the time a request is sent and the time the response arrives.

* The main `CryptDE` is established in `ActorSystemFactoryTools`, by the `cryptdes()` method. You'll want to modify this method to take `BootstrapperConfig` and `PersistentConfiguration` references and call the new `PersistentConfiguration::cryptde()` method with the database password (if any) from the `BootstrapperConfig`. If it returns `Some(cryptde)`, use that one. If it returns None, generate a new one with `CryptDEReal::default()` and call `PersistentConfiguration::set_cryptde()` with it.

* Edge cases:
	* Maybe `BootstrapperConfig::db_password_opt` is `None`. In that case, always generate a new public key, because you can't read or write an encrypted value without a password.

dnwiebe:
__Test Plan__

1. __Brand-New Database:__
    1. Install MASQ in a brand-new environment.
[...]

PR #664, opened 2025-07-02
GH-622 Retain public key in database

dnwiebe:


masq_lib/src/test_utils/environment_guard.rs
```diff
@@ -43,16 +42,45 @@ impl<'a> Drop for EnvironmentGuard<'a> {
         self.environment
             .iter()
             .for_each(|(name, value)| std::env::set_var(name, value));
+        let mut thread_id_guard = ENVIRONMENT_GUARD_THREAD_ID
```
dnwiebe, line 47: Enhanced `EnvironmentGuard` with a thread ID so that A) if we couldn't lock the guard we could produce a message saying who had it, B) we could tell if our thread already had the `EnvironmentGuard` locked and panic (instead of deadlock) if we tried to lock it again, and C) when we went to read or write the main and alias static `CryptDE`s, we could check to see if the `EnvironmentGuard` was owned by our thread and panic if it wasn't--that is, make tests that interact with static `CryptDE`s fail if they didn't lock the `EnvironmentGuard` first.

So...eventually we decided that static `CryptDE`s were a bad idea and got rid of them, but we're leaving the thread ID in the `EnvironmentGuard` anyway, for the first two reasons.

multinode_integration_tests/tests/bookkeeping_test.rs
```diff
@@ -112,6 +112,7 @@ pub fn start_real_node(cluster: &mut MASQNodeCluster, neighbor: NodeReference) -
     let index = cluster.next_index();
     cluster.start_real_node(
         NodeStartupConfigBuilder::standard()
+            .db_password(None)
```
dnwiebe, line 115: ...so that the recall-the-old-public-key machinery will not groan into action for this test, which doesn't require it.

node/src/actor_system_factory.rs
```diff
@@ -87,12 +85,10 @@ impl ActorSystemFactoryReal {
 pub trait ActorSystemFactoryTools {
     fn prepare_initial_messages(
         &self,
-        cryptdes: CryptDEPair,
```
dnwiebe, line 90: The `CryptDEPair` is now in `BootstrapperConfig`, so it doesn't have to be sent separately here.

node/src/actor_system_factory.rs
```diff
@@ -87,12 +85,10 @@ impl ActorSystemFactoryReal {
...
         config: BootstrapperConfig,
         persistent_config: Box<dyn PersistentConfiguration>,
         actor_factory: Box<dyn ActorFactory>,
     ) -> StreamHandlerPoolSubs;
-    fn cryptdes(&self) -> CryptDEPair;
```
dnwiebe, line 95: No longer necessary.

node/src/actor_system_factory.rs
```diff
@@ -380,12 +369,14 @@ pub trait ActorFactory {
 }
 
 pub struct ActorFactoryReal {
+    cryptde_pair: CryptDEPair,
```
dnwiebe, line 366: A lot of objects that formerly referred to the static `CryptDE`s now carry their own around with them.

node/src/actor_system_factory.rs
```diff
@@ -708,6 +699,11 @@ mod tests {
     use std::sync::{Arc, Mutex};
     use std::thread;
     use std::time::Duration;
+    use lazy_static::lazy_static;
+
+    lazy_static! {
+        static ref CRYPTDE_PAIR: CryptDEPair = CryptDEPair::null();
```
dnwiebe, line 693: Instead of having a single static pair of `CryptDE`s and trying to keep the tests both running in parallel and staying out of each other's way, this is how we do it now.  Each suite of tests has its own static read-only pair of `CryptDENull`s, and any tests in that suite that need `CryptDEReal`s, or to modify `CryptDE`s make their own and keep them local.

node/src/actor_system_factory.rs
```diff
@@ -813,7 +798,7 @@ mod tests {
         }
     }
 
-    struct ActorFactoryMock<'a> {
+    struct ActorFactoryMock {
```
dnwiebe, line 789: In a few places, I could get rid of lifetimes, because I'm no longer dealing with references to the static pair of `CryptDE`s.

node/src/bootstrapper.rs
```diff
@@ -386,8 +399,13 @@ impl BootstrapperConfig {
             port_configurations: HashMap::new(),
             data_directory: PathBuf::new(),
             node_descriptor: NodeDescriptor::default(),
-            main_cryptde_null_opt: None,
-            alias_cryptde_null_opt: None,
```
dnwiebe, line 390: These two fields were used by the `--fake-public-key` parameter; but since we now have `cryptde_pair` in here, we don't need them anymore.

node/src/bootstrapper.rs
```diff
@@ -544,51 +562,6 @@ impl Bootstrapper {
         }
     }
 
-    #[cfg(test)] // The real ones are private, but ActorSystemFactory needs to use them for testing
```
dnwiebe, line 547: No more static `CryptDE`s, so tests for the functions to manipulate them are obsolete.

node/src/daemon/setup_reporter.rs
```diff
@@ -999,13 +1014,19 @@ impl ValueRetriever for Neighbors {
 
     fn computed_default(
         &self,
-        _bootstrapper_config: &BootstrapperConfig,
+        bootstrapper_config: &BootstrapperConfig,
```
dnwiebe, line 1017: `node_descriptors_to_neighbors()` now needs a `CryptDE` because it can no longer access the static one.

node/src/dispatcher.rs
```diff
@@ -243,18 +261,10 @@ mod tests {
         );
     }
 
-    lazy_static! {
```
dnwiebe, line 246: This block was moved up and had `CRYPTDE_PAIR` added to it.

node/src/node_configurator/node_configurator_standard.rs
```diff
@@ -562,6 +724,7 @@ mod tests {
 
     #[test]
     fn can_read_dns_servers_and_consuming_private_key_from_config_file() {
+        let _guard = EnvironmentGuard::new();
```
dnwiebe, line 705: These tests don't involve the environment directly; but they use `MultiConfig` code that checks the environment, and if there are other tests that do use the environment running in parallel, these tests can get confused. So I added `EnvironmentGuard`s.

node/src/test_utils/persistent_configuration_mock.rs
```diff
@@ -79,6 +84,84 @@ pub struct PersistentConfigurationMock {
     arbitrary_id_stamp_opt: Option<ArbitraryIdStamp>,
 }
 
+impl Clone for PersistentConfigurationMock {
```
dnwiebe, line 87: I don't know why this `Clone` implementation looks new. I'm pretty sure I didn't write it, at least not for this card.

node/tests/initialization_test.rs
```diff
@@ -305,3 +307,40 @@ fn node_creates_log_file_with_heading_integration() {
...
+        .unwrap()
+        .get(1)
+        .unwrap()
+        .as_str();
+    node.kill().unwrap();
+    descriptor.to_string()
+}
+
+#[test]
+fn second_run_node_uses_same_public_key_as_first_run_node_integration() {
```
dnwiebe, line 339: This is the integration test that proves the change works.

node/src/crash_test_dummy.rs
```diff
@@ -5,40 +5,33 @@ use masq_lib::crash_point::CrashPoint;
 use masq_lib::logger::Logger;
 use tokio::prelude::future::Future;
 
-pub struct CrashTestDummy<C> {
-    pub configuration: C,
```
dnwiebe, line 9: The `configuration` field wasn't being used anywhere, so I took it out.

node/src/bootstrapper.rs
```diff
@@ -1320,15 +1280,15 @@ mod tests {
...
-                &NodeAddr::new(&IpAddr::from_str("1.2.3.4").unwrap(), &[5123]),
-                Chain::BaseSepolia,
-                main_cryptde_ref()
+            config.node_descriptor.node_addr_opt,
+            Some(NodeAddr::new(
+                &IpAddr::from_str("1.2.3.4").unwrap(),
+                &[5123]
             ))
         );
+        // Not checking the public key, because its value is not predictable.
```
czarte, line 1291: Maybe we can just `assert` on public key, that we know, there is one created?

node/src/db_config/persistent_configuration.rs
```diff
@@ -988,6 +1066,259 @@ mod tests {
         );
     }
 
+    #[test]
+    fn cryptde_null_success() {
+        let db_password = "bellybutton".to_string();
+        let cryptde = CryptDEReal::new(Chain::Dev);
```
czarte, line 1072: This should probably be CryptDENull here, otherwise this test is redundant

node/src/db_config/persistent_configuration.rs
```diff
@@ -988,6 +1066,259 @@ mod tests {
...
+                .into_iter()
+                .map(|s| s.to_string())
+                .collect_vec()
+        );
+    }
+
+    #[test]
+    fn cryptde_failure_null_value_in_database() {
+        let db_password = "bellybutton".to_string();
+        // TODO: See if we can get rid of this
```
czarte, line 1154: Did you already addressed this TODOs? there is couple of them, in case you have done these, you want to get rid of the TODOs

node/src/hopper/routing_service.rs
```diff
@@ -48,7 +48,7 @@ impl RoutingService {
         is_decentralized: bool,
     ) -> RoutingService {
         RoutingService {
-            cryptdes,
+            cryptde_pair: cryptdes,
```
czarte, line 51: probably want to change cryptdes to cryptde_pair

node/src/hopper/routing_service.rs
```diff
@@ -593,13 +606,13 @@ mod tests {
     #[test]
     fn logs_and_ignores_message_that_cannot_be_deserialized() {
         init_test_logging();
-        let cryptdes = make_cryptde_pair();
-        let route = route_from_proxy_client(&cryptdes.main.public_key(), cryptdes.main);
+        let cryptdes = CRYPTDE_PAIR.clone();
```
czarte, line 609: maybe you want to rename this as well

node/src/hopper/routing_service.rs
```diff
@@ -635,18 +648,15 @@ mod tests {
     #[test]
     fn logs_and_ignores_message_that_cannot_be_decrypted() {
         init_test_logging();
-        let (main_cryptde, alias_cryptde) = {
-            //initialization to real CryptDEs
-            let pair = Bootstrapper::pub_initialize_cryptdes_for_testing(&None, &None);
-            (pair.main, pair.alias)
-        };
+        let _ = EnvironmentGuard::new();
```
czarte, line 651: is EnvironmentGuard needed here?

node/src/neighborhood/node_record.rs
```diff
@@ -75,7 +75,7 @@ pub struct NodeRecordInputs {
 impl NodeRecord {
     pub fn new(
         public_key: &PublicKey,
-        cryptde: &dyn CryptDE, // Must be the new NodeRecord's CryptDE: used for signing
+        cryptde: &dyn CryptDE, // Do we need a separate CryptDE for each NodeRecord? Why?
```
czarte, line 78: Is this comment about to stay, or do we go to investigate?

node/src/test_utils/persistent_configuration_mock.rs
czarte: Looks like you did, at least I assume from previous state, where Clone was derived on PersistentConfigurationMock by annotation. And compared to Master too. I think dyn CryptDE could prevent from derive?

dnwiebe: Ah, yes--I remember now when I'm looking at it in the IDE instead of in GitHub. A simple `Clone` derivation would have worked except for the `cryptde_results` field, which is a collection of `dyn CryptDE`s which each have to be `.dup()`ed instead of `.clone()`d.

node/src/db_config/persistent_configuration.rs
dnwiebe: Done.

node/src/db_config/persistent_configuration.rs
dnwiebe: I believe so, yes.

node/src/hopper/routing_service.rs
dnwiebe: Done.

node/src/hopper/routing_service.rs
dnwiebe: Done.

node/src/hopper/routing_service.rs
dnwiebe: Doesn't seem like it should be, no.

node/src/neighborhood/node_record.rs
dnwiebe: Aha. We're not storing the `CryptDE`, we're just using it for signing. It's fine. I removed the comment.

node/src/bootstrapper.rs
dnwiebe: A placeholder `CryptDEPair` is created upon construction of the `BootstrapperConfig`, intended for replacement. We can't tell whether the public key belongs to the original or the replacement.

I thought about making the placeholder a sentinel--say, with an all-zeros public key--but I'm twitchy about that because of the way we were leaking `CryptDENull`s into production. If we start leaking predictable `CryptDEReal`s into production...but wait.

First, they should stick out like sore thumbs in the logs if the public keys are all zero, and second, the whole network should come crashing down because all the Nodes will have the same public key. Yeah, I think I'll do that. Not just for this test, but for future security.

PR #666, opened 2025-07-08
GH-665: Introduce new states in FailedPayableDao

### LINKED ISSUE #665: Introduce new states in `FailedPayableDao`
This card will introduce the following states:

```rust
pub enum FailureStatus {
    Retry,
    Recheck,
    Concluded,
}
```

Instead of `mark_as_rechecked()`, the new function looks like this:

```rust
    fn update_statuses(
        &self,
        status_updates: HashMap<TxHash, FailureStatus>,
    ) -> Result<(), FailedPayableDaoError>;
```

The query can be simplified to this:

```rust
"WHERE status = 'Retry'"
```

utkarshg6:


node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -238,35 +258,49 @@ impl FailedPayableDao for FailedPayableDaoReal<'_> {
...
+        status_updates: HashMap<TxHash, FailureStatus>,
+    ) -> Result<(), FailedPayableDaoError> {
+        if status_updates.is_empty() {
+            return Err(FailedPayableDaoError::EmptyInput);
+        }
+
+        let case_statements = status_updates
+            .iter()
+            .map(|(hash, status)| format!("WHEN tx_hash = '{:?}' THEN '{:?}'", hash, status))
+            .collect::<Vec<String>>()
```
bertllll: You don't need this to use the `join` afterwards. The trick is that there is another join() which resides in intertools.
Pls spare the single line by using that one.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -23,6 +25,26 @@ pub enum FailedPayableDaoError {
...
     PendingTooLong,
     NonceIssue,
+    General,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FailureStatus {
+    Retry,
+    Recheck,
+    Concluded,
```
bertllll, line 35: W hat about putting as (just throwing ideas):

PreRetry
PreRecheck
Concluded

PriorRetry
PriorRecheck
Concluded

RetryRequired
RecheckRequired
Concluded

AwaitingRetry
AwaitingRecheck
Concluded

NotRetriedYet
NotRecheckedYet
Concluded

...

The thing is it would be great if those three were more consistent in the word class.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -547,18 +563,29 @@ mod tests {
...
+    }
+
     #[test]
     fn retrieve_condition_display_works() {
-        let expected_condition = "WHERE reason = 'PendingTooLong' AND rechecked = 0";
         assert_eq!(
-            FailureRetrieveCondition::UncheckedPendingTooLong.to_string(),
-            expected_condition
+            FailureRetrieveCondition::ToRetry.to_string(),
+            "WHERE status = 'Retry'"
```
bertllll: I first complained (friendly) that we'd need also the possibility to retrieve the failed txs that require the second check, and you then said that you may make it so that it is flexible meaning any of those FailureStatuses can be provided and the corresponding records are retrieved.

node/src/accountant/test_utils.rs
```diff
@@ -1077,6 +1082,157 @@ impl PendingPayableDaoFactoryMock {
...
+    pub fn delete_records_params(mut self, params: &Arc<Mutex<Vec<HashSet<TxHash>>>>) -> Self {
+        self.delete_records_params = params.clone();
+        self
+    }
+
+    pub fn delete_records_result(self, result: Result<(), FailedPayableDaoError>) -> Self {
+        self.delete_records_results.borrow_mut().push(result);
+        self
+    }
+}
```
bertllll, line 1201: I don't think this is worth a comment but you insisted on adding a line between.

PR #673, opened 2025-07-20
GH-672: Improve FailureReason

### LINKED ISSUE #672: Improve `FailureReason`

utkarshg6:


node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -21,11 +23,83 @@ pub enum FailedPayableDaoError {
...
+    Unreachable,
+    Web3RpcError { code: i64, message: String },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum PoolError {
+    NonceIssue {
+        latest_used_nonce: u64,
+        tx_nonce: u64,
+    },
```
bertllll: How will we figure out that it was exactly this type of error? So far we've been evaluating only if the status is 0 or 1. 

We'd have to introduce some translator which could derive it from the tx topics, right? 
If so, do we want to define the error variant like this if we yet don't have such code?

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -21,11 +23,83 @@ pub enum FailedPayableDaoError {
...
+    OrphanedTransaction {
+        tx_hash: TxHash,
+        block_number: u64,
+    },
```
bertllll: What exactly is this "orphaned tx" supposed to represent? I think this term enables wider, multi-optional  interpretation.

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -21,11 +23,83 @@ pub enum FailedPayableDaoError {
...
+impl From<Web3Error> for FailureReason {
+    fn from(error: Web3Error) -> Self {
+        match error {
+            // Local Errors
+            Web3Error::Decoder(error) => FailureReason::Local(LocalError::Decoder(error)),
+            Web3Error::Internal => FailureReason::Local(LocalError::Internal),
+            Web3Error::Io(error) => FailureReason::Local(LocalError::Io(error.to_string())),
+            Web3Error::Signing(error) => {
+                // This variant cannot be tested due to import limitations.
+                FailureReason::Local(LocalError::Signing(error.to_string()))
```
bertllll: Do you remember what was the issue? It seems to me quite neat...

node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -21,11 +23,83 @@ pub enum FailedPayableDaoError {
     SqlExecutionFailed(String),
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum FailureReason {
+    Local(LocalError),
+    Api(ApiError),
+    Pool(PoolError),
```
bertllll: We've discussed this on Discord already.

Please introduce this distinguishing for the Api variant. 

```Rust

#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct ApiError {
     pub phase: TxPhase, 
     pub msg: String
}


#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub enum TxPhase {
    Sending,
    ReceiptCheck
}
```

or if you prefer, you can also add the two fields straight into the Api variant. With the struct-like syntax. 

```Rust
Api {
    phase: TxPhase,
    msg: String
}
```

node/src/accountant/db_access_objects/failed_payable_dao.rs
utkarshg6: It's with the import of `secp256k1`.

bertllll (approved):

PR #682, opened 2025-08-15
GH-818 Are Chunked Responses Being Handled Wrong?

dnwiebe:


node/src/proxy_server/mod.rs
```diff
@@ -58,7 +58,8 @@ use std::time::{Duration, SystemTime};
 use tokio::prelude::Future;
 
 pub const CRASH_KEY: &str = "PROXYSERVER";
-pub const RETURN_ROUTE_TTL: Duration = Duration::from_secs(120);
```
dnwiebe, line 62: Now there are two `TtlHashMap`s, so we need two TTLs.

node/src/proxy_server/mod.rs
```diff
@@ -163,12 +168,13 @@ impl Handler<AddRouteResultMessage> for ProxyServer {
     type Result = ();
 
     fn handle(&mut self, msg: AddRouteResultMessage, _ctx: &mut Self::Context) -> Self::Result {
-        let dns_failure = self
-            .dns_failure_retries
-            .get(&msg.stream_key)
-            .unwrap_or_else(|| {
-                panic!("AddRouteResultMessage Handler: stream key: {} not found within dns_failure_retries", msg.stream_key);
-            });
```
dnwiebe, line 171: Took out the panic, replaced it with an error log. This panic doesn't happen because of something that arrives from outside, but it does happen because of something that _doesn't_ arrive from outside; that should be against The Rule as well.

node/src/proxy_server/mod.rs
```diff
@@ -362,8 +369,12 @@ impl ProxyServer {
     }
 
     fn handle_dns_resolve_failure(&mut self, msg: &ExpiredCoresPackage<DnsResolveFailure_0v1>) {
+        let return_route_id = match self.rri_from_remaining_route(&msg.remaining_route) {
+            Some(rri) => rri,
+            None => return, // TODO: Eventually we'll have to do something better here, but we'll probably need some heuristics.
+        };
```
dnwiebe: `get_return_route_info()` has been modified to accept only the `return_route_id` (a `u32`) rather than the whole `Route`; so here we're pulling the `return_route_id` out of the `Route` we have.

node/src/proxy_server/mod.rs
```diff
@@ -923,31 +939,43 @@ impl ProxyServer {
         }
     }
 
-    fn get_return_route_info(
```
dnwiebe, line 927: This is hard to read, but I split the original `get_return_route_info()` into two:

`rri_from_remaining_route()` that pulls the ID out of the `Route`
`get_return_route_info()` (accepting an ID, not a whole `Route`) to traverse the `TtlHashMap`s

node/src/proxy_server/mod.rs
```diff
@@ -923,31 +939,43 @@ impl ProxyServer {
...
         }
     }
 
+    fn get_return_route_info(
+        &mut self,
+        return_route_id: u32,
+        source: &str,
+    ) -> Option<Rc<AddReturnRouteMessage>> {
+        match self.route_ids_to_return_routes_first_chance.remove(&return_route_id) {
+            Some(rri) => {
```
dnwiebe: There's no use case in which we leave a matching RRI in the `_first_chance` map. If we find it, we'll transfer it over to the `_stragglers` map, where it is subject to a much shorter TTL and will probably expire and be killed unless it's part of a stream.

node/src/proxy_server/mod.rs
```diff
@@ -923,31 +939,43 @@ impl ProxyServer {
...
+                self.route_ids_to_return_routes_stragglers.insert(return_route_id, (*rri).clone());
+                Some(rri)
+            },
+            None => match self.route_ids_to_return_routes_stragglers.get(&return_route_id) {
```
dnwiebe: If we don't find matching RRI in the `_first_chance` map, we immediately look in the `_stragglers` map.

node/src/proxy_server/mod.rs
```diff
@@ -1617,6 +1646,70 @@ mod tests {
         }
     }
```
dnwiebe, line 1768: The following tests address `get_return_route_info()` directly (white-box tests) and make sure it operates as intended.

node/src/proxy_server/mod.rs
```diff
@@ -3072,10 +3165,8 @@ mod tests {
...
 
     #[test]
-    #[should_panic(
-        expected = "AddRouteResultMessage Handler: stream key: AAAAAAAAAAAAAAAAAAAAAAAAAAA not found within dns_failure_retries"
-    )]
```
dnwiebe, line 3065: Converted panic to error log

node/src/proxy_server/mod.rs
```diff
@@ -3759,7 +3852,7 @@ mod tests {
         subject
             .keys_and_addrs
             .insert(stream_key.clone(), socket_addr.clone());
```
dnwiebe, line 4044: The rest of this file is just renames.

czarte (approved):

Issue #689, opened 2025-08-19
Amend scanner scheduling: Handling ScanError msg


bertllll:
_(First phase)_

If anything from the scanning goes wrong on the side of the BlockchainBridge, severe enough to halt the process, it results in a message called ScanError and addressed to the Accountant. So far what we've ever cared of was that we just canceled the scan inside the Scanners, which is important for being able to run a new scan in the future as it cannot be done until there is a timestamp stored in the respective scanner that signifies the time when an ongoing scanner begun. We do this still, but with the late big redesign in the scan timings and scheduling, we have now a responsibility to sustain the continuity of the constantly rescheduled scans and this is possibly a place where we could suffer discontinuity unless we make sure we will schedule a new scan when we receive this ScanError message.
This part, the responsible rescheduling is missing now and it would be fatal to the scanners, as there would be no other force to awake the scanner again. 

The bottom line: add scheduling.

PR #691, opened 2025-08-19
GH-689: Amend scanner scheduling: Handling ScanError msg

bertllll:


node/src/accountant/scanners/pending_payable_scanner/mod.rs
```diff
@@ -1376,7 +1376,7 @@ mod tests {
             FailureStatus::RecheckRequired(ValidationStatus::Waiting),
         );
         let failed_payable_dao = FailedPayableDaoMock::default()
-            .update_statuses_result(Err(FailedPayableDaoError::InvalidInput("bluh".to_string())));
+            .update_statuses_result(Err(FailedPayableDaoError::InvalidInput("blah".to_string())));
```
utkarshg6, line 1379: Wow, what a change! 👏

node/src/blockchain/blockchain_bridge.rs
```diff
@@ -601,6 +605,17 @@ mod tests {
         }
     }
 
+    impl SubsFactory<BlockchainBridge, BlockchainBridgeSubs>
+    for SubsFactoryTestAddrLeaker<BlockchainBridge>
```
utkarshg6, line 609: I would keep it in a `test_utils.rs` file.

node/src/sub_lib/accountant.rs
```diff
@@ -192,23 +192,44 @@ impl MessageIdGenerator for MessageIdGeneratorReal {
...
     use crate::sub_lib::wallet::Wallet;
     use crate::test_utils::recorder::{make_accountant_subs_from_recorder, Recorder};
     use actix::Actor;
+    use masq_lib::messages::ScanType;
     use std::str::FromStr;
     use std::sync::atomic::Ordering;
     use std::sync::Mutex;
     use std::time::Duration;
 
+    impl From<DetailedScanType> for ScanType {
```
utkarshg6, line 222: Shouldn't this be part of prod?

utkarshg6 (approved):

node/src/blockchain/blockchain_bridge.rs
bertllll: It implements closely on the BlockchainBridge. I think it belongs in that file only. This is not a piece of code of general use.

node/src/sub_lib/accountant.rs
bertllll: I guess I made use of it only in tests. The idea is that if it needs to be expanded to the prod code too, the compiler will complain about conflicting impls. You will have to resolve it by keeping just one, then in the prod code.

Issue #642, opened 2025-05-27
Teach the `PendingPayableScanner` on pending payble management


bertllll:
fill me in later

bertllll:
The PendingTooLong issue

Use caching of those transactions that end with this failure result so that you can have a structured list of the txs that you are obliged to doublecheck.  


The steps should be:
a) rewrite the newer (successor) record in sent_payable with the nonce N by the data of the older transaction, now read from the failed_payable table. I say rewrite because I want to avoid the db state inconsistency precisely,
That means, I'd like to have a single method to call, which will internally perform: the old row deletion + insertion of the new row, but committed no earlier than both of them done. I say we should use an explicit rusqlite tx, that gives you the control over when it is committed.
However, you may be able to work it around by writing an UPDATE sql which will update the whole row though (every valuei in the row.) I wonder if that is considered a hack in the community of the data-engineers? (According to the ChatGPT, it isn't a nonsense. I'd then highly recommend that as it is much simpler to implement)
b) delete the record from failed_payble

PR #702, opened 2025-09-11
GH-642: review 2

bertllll:


node/src/accountant/db_access_objects/failed_payable_dao.rs
```diff
@@ -28,6 +28,7 @@ pub enum FailedPayableDaoError {
 pub enum FailureReason {
     Submission(AppRpcErrorKind),
     Reverted,
+    Unrecognized,
```
utkarshg6: We don't want to keep things uncertain, when we know for sure that it has failed, just mark it as `Reverted` pls.

node/src/accountant/mod.rs
```diff
@@ -5024,10 +5037,84 @@ mod tests {
...
+                status: StatusReadFromReceiptCheck::Pending,
+            },
+            SeedsToMakeUpPayableWithStatus {
+                tx_hash: TxHashByTable::FailedPayable(make_tx_hash(456)),
+                status: StatusReadFromReceiptCheck::Reverted,
+            },
+        ]);
+        // Although it all begins with an external trigger, the response skeleton will not be
+        // propagated eventually to the further stages (see the NotifyLaterHandle assertion holding
+        // None for the response skeleton)
```
utkarshg6: but why's that?, that's more important to answer.

node/src/accountant/scanners/pending_payable_scanner/mod.rs
```diff
@@ -1584,7 +1554,8 @@ mod tests {
         subject
             .handle_confirmed_transactions(DetectedConfirmations::default(), &Logger::new("test"))
 
-        // Mocked payable DAO didn't panic, which means we skipped the actual process
+        // Mocked payable DAO without prepared results didn't panic, which none of its methods was
```
utkarshg6: which means none of it's

word "means" is misssing

utkarshg6 (commented):
Continue at node/src/accountant/scanners/pending_payable_scanner/utils.rs

utkarshg6 (commented):
Nicely done, so proud of you.

node/src/accountant/db_access_objects/failed_payable_dao.rs
bertllll: It was actually never used :rofl:

PR #710, opened 2025-09-25
GH-711: review three

bertllll:


multinode_integration_tests/tests/verify_bill_payment.rs
```diff
@@ -183,15 +186,15 @@ fn payments_were_adjusted_due_to_insufficient_balances() {
             .unwrap()
         };
         let transaction_fee_margin = PurePercentage::try_from(15).unwrap();
-        transaction_fee_margin.add_percent_to(gas_limit_dev_chain * gas_price_major)
+        transaction_fee_margin.increase_by_percent_for(gas_limit_dev_chain * gas_price_major)
```
utkarshg6, line 189: You don't need to keep the word `_for` in the functions's name. Please eliminate it.

node/src/accountant/mod.rs
```diff
@@ -1621,7 +1619,7 @@ mod tests {
             response_skeleton_opt: Some(response_skeleton),
         };
         let analyzed_accounts =
-            convert_qualified_into_analyzed_payables_in_test(unadjusted_qualified_accounts.clone());
+            convert_qualified_p_into_analyzed_p(unadjusted_qualified_accounts.clone());
```
utkarshg6, line 1622: This shorthand is not helping here, maybe even introducing jargon. I'd suggest if you use the name `convert_qualified_payable_into_analyzed_payable`, or if you want shorter than `qualified_to_analyzed_payable`.

node/src/accountant/mod.rs
```diff
@@ -1739,31 +1737,33 @@ mod tests {
...
+            of canceled payments: 1. Transaction fee per payment: 3,300,000,000,000,000 wei, while \
+            the wallet contains: 123,000,000,000 wei."
         ));
         log_handler
             .exists_log_containing(&format!("INFO: {test_name}: The Payables scan ended in"));
         log_handler.exists_log_containing(&format!(
-            "ERROR: {test_name}: Payable scanner is blocked from preparing instructions for payments. \
-            The cause appears to be in competence of the user."
+            "ERROR: {test_name}: Payable scanner is unable to generate payment instructions. \
+            It looks like only the user can resolve this issue."
```
utkarshg6, line 1766: Instead of saying, "It looks like...", we can just direct the user by saying, "The user will have to resolve the issue manually."

node/src/accountant/payment_adjuster/disqualification_arbiter.rs
```diff
@@ -280,7 +283,7 @@ mod tests {
     };
     use crate::accountant::payment_adjuster::miscellaneous::data_structures::UnconfirmedAdjustment;
     use crate::accountant::payment_adjuster::test_utils::local_utils::{
-        make_meaningless_weighed_account, make_non_guaranteed_unconfirmed_adjustment,
+        make_meaningless_unconfirmed_adjustment, make_meaningless_weighed_account,
```
utkarshg6, line 286: I think the word, `meaningless` is unnecessary, unless you have an alternate fn which says, `meaningful`, the prefix word `make` is enough to signify that it's meant for the test, and the developer should consider that it's fields were generated for testing purposes.

I mean this for both `make_meaningless_unconfirmed_adjustment` and `make_meaningless_weighed_account`.

node/src/accountant/payment_adjuster/disqualification_arbiter.rs
```diff
@@ -525,32 +528,14 @@ mod tests {
...
-                    .bare_account
-                    .wallet
-            })
-            .collect_vec();
-        assert_eq!(all_wallets.len(), 4);
-        let wallets_same_as_wallet_3 = all_wallets
-            .iter()
-            .filter(|wallet| wallet.address() == wallet_3)
-            .collect_vec();
-        assert_eq!(wallets_same_as_wallet_3.len(), 1);
```
utkarshg6, line 545: I've no idea why you decided to remove this... I thought you prefer stronger tests.

node/src/accountant/payment_adjuster/logging_and_diagnostics/diagnostics.rs
```diff
@@ -100,7 +100,7 @@ pub mod ordinary_diagnostic_functions {
     use thousands::Separable;
     use web3::types::Address;
 
-    pub fn thriving_competitor_found_diagnostics(
+    pub fn diagnostics_for_accounts_above_disqualification_limit(
```
utkarshg6, line 103: Does it mean that these accounts are disqualified or qualified? I'd suggest you either add it as a comment, or possibly rename the fn to make it clear.

node/src/accountant/payment_adjuster/miscellaneous/data_structures.rs
```diff
@@ -101,6 +101,11 @@ impl AffordableAndRequiredTxCounts {
     }
 }
 
+pub enum AccountsByFinalization {
+    Unexhausted(Vec<AdjustedAccountBeforeFinalization>),
+    Finalized(Vec<PayableAccount>),
```
utkarshg6, line 106: What does these enum variants? Their name doesn't appear to be mutually exclusive. For example, if you're using `Unexhausted`, then shouldn't the other be `Exhausted`? Similarly, if it's `Finalized`, then shouldn't the other variant be `Incomplete`?

Since they aren't mutually exclusive, I find them very confusing.

node/src/accountant/payment_adjuster/preparatory_analyser/mod.rs
```diff
@@ -104,16 +106,21 @@ impl PreparatoryAnalyzer {
...
+            TransactionFeeImmoderateInsufficiency,
+        >,
         service_fee_check_result: Result<(), ServiceFeeImmoderateInsufficiency>,
-    ) -> Result<Option<u16>, PaymentAdjusterError> {
+    ) -> Result<Option<TxCount>, PaymentAdjusterError> {
         let construct_error =
             |tx_fee_check_err_opt: Option<TransactionFeeImmoderateInsufficiency>,
              service_fee_check_err_opt: Option<ServiceFeeImmoderateInsufficiency>| {
-                PaymentAdjusterError::AbsolutelyInsufficientBalance {
+                PaymentAdjusterError::AbsoluteFeeInsufficiency {
```
utkarshg6, line 118: I'd prefer the older name. That's clearer.

node/src/accountant/payment_adjuster/preparatory_analyser/mod.rs
```diff
@@ -337,21 +344,21 @@ impl LateServiceFeeSingleTxErrorFactory {
...
+impl ServiceFeeErrorFactory<PaymentAdjusterError, WeighedPayable> for LaterServiceFeeErrorFactory {
     fn make(
         &self,
         current_set_of_accounts: &[WeighedPayable],
         cw_service_fee_balance_minor: u128,
     ) -> PaymentAdjusterError {
         let number_of_accounts = current_set_of_accounts.len();
-        PaymentAdjusterError::AbsolutelyInsufficientServiceFeeBalancePostTxFeeAdjustment {
-            original_number_of_accounts: self.original_number_of_accounts,
+        PaymentAdjusterError::AbsoluteFeeInsufficiency {
```
utkarshg6, line 354: Can't it be simply "InsufficientBalance"?

node/src/accountant/payment_adjuster/preparatory_analyser/mod.rs
```diff
@@ -630,33 +639,33 @@ mod tests {
...
+            let adjustment_is_needed_actual = subject
+                .recheck_if_service_fee_adjustment_is_needed(
+                    &accounts,
+                    *service_fee_balance,
+                    error_factory.clone(),
+                    &logger,
+                )
+                .unwrap();
+            assert_eq!(adjustment_is_needed_actual, *adjustment_is_needed_expected);
+        });
```
utkarshg6, line 662: Nicely done! 👏

node/src/accountant/payment_adjuster/preparatory_analyser/mod.rs
```diff
@@ -678,11 +687,11 @@ mod tests {
             .balance_wei = balance_2;
         let weighed_accounts = vec![account_1, account_2];
 
-        let result = LateServiceFeeSingleTxErrorFactory::new(&weighed_accounts);
+        let result = LaterServiceFeeErrorFactory::new(&weighed_accounts);
 
         assert_eq!(
             result,
-            LateServiceFeeSingleTxErrorFactory {
+            LaterServiceFeeErrorFactory {
```
utkarshg6, line 694: Shouldn't this be `LateServiceFeeErrorFactory` instead? I'm implying to use the word `Late` instead of `Later`.

node/src/accountant/payment_adjuster/service_fee_adjuster.rs
```diff
@@ -119,27 +122,31 @@ impl ServiceFeeAdjusterReal {
...
+                        disqualification_limit,
+                    );
+                    let mut adjusted = current;
+                    adjusted.proposed_adjusted_balance_minor = disqualification_limit;
+                    above_or_even_to_disq_limit.push(adjusted)
+                } else {
+                    below_disq_limit.push(current)
+                }
+                (above_or_even_to_disq_limit, below_disq_limit)
+            };
```
utkarshg6, line 141: I think this could be further refactored like this:

```rust
let fold_guts = |(mut above_limit, mut below_limit): (Vec<_>, Vec<_>), current: UnconfirmedAdjustment| {
    let disqualification_limit = current.disqualification_limit_minor();
    
    if current.proposed_adjusted_balance_minor >= disqualification_limit {
        diagnostics_for_accounts_above_disqualification_limit(&current, disqualification_limit);
        
        let adjusted = UnconfirmedAdjustment {
            proposed_adjusted_balance_minor: disqualification_limit,
            ..current
        };
        above_limit.push(adjusted);
    } else {
        below_limit.push(current);
    }
    
    (above_limit, below_limit)
};
```

node/src/accountant/scanners/scanners_utils.rs
```diff
@@ -314,6 +314,11 @@ pub mod payable_scanner_utils {
                 .payable_threshold_gauge
                 .calculate_payout_threshold_in_gwei(payment_thresholds, debt_age);
 
+            eprintln!(
+                "Balance wei: {}\n\
+                 Threshold:   {}",
+                payable.balance_wei, threshold
+            );
```
utkarshg6, line 321: Remove this...

utkarshg6 (commented):
Continue with node/src/accountant/payment_adjuster/non_unit_tests/mod.rs

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -81,16 +81,16 @@ fn loading_test_with_randomized_params() {
...
     // unbanned. In bulky accounts, this until-some-time forgiven portion stands only as a fraction
     // of a whole. Small accounts, however, if it can be applied (as opposed to the account having
     // to be excluded) might get shrunk a lot, and therefore many percents are to be reported as
-    // missing. This is what the numbers like 99% and 90% illustrates. That said, the letter account
-    // comes across as it should take precedence for its expectedly larger weight, and gain at the
-    // expanse of the other, but the percents speak otherwise. Yet, it's correct. The interpretation
+    // missing. This is what the numbers like 99% and 90% illustrate. That said, the letter account
+    // comes across as it should take precedence for its expectedly larger weight and gain at the
+    // expanse of the other, but the percents speak otherwise. Yet it's correct. The interpretation
     // is the key. (Caution: this test displays its output with those accounts sorted).
```
utkarshg6, line 94: Bert, this explanation is not clear and concise.

Here's an output from ChatGPT that I found easy to understand. It took me few iterations to get the right output. I request you to pick an inspiration from this text and add it accordingly.

```
// This is a fuzz test for the PaymentAdjuster. It generates a large number of different 
// scenarios to see how well the PaymentAdjuster can handle various situations that might 
// occur in real life. The goal is to ensure it works reliably without causing problems in Node.

// The test has an exo-parameter, which is the number of scenarios to generate. This number 
// is flexible because many scenarios will be discarded during setup if they don’t meet 
// certain requirements, like whether the accounts involved are valid for payment adjustments.

// Some scenarios will be thrown out early if the PaymentAdjuster finds mistakes, ensuring 
// that only valid scenarios are tested. This helps identify corner cases that may not have 
// been considered during the design phase.

// When the test is finished, the results are saved in a file located at:
// node/generated/test/payment_adjuster/tests/home/loading_test_output.txt. This file 
// contains important numbers and statistics that show how well the PaymentAdjuster performed.

// The output may include confusing results, especially regarding percentages that indicate 
// how much of a debt is covered. These percentages might not match the account balances, 
// which can be tricky to understand. The adjustment depends on a concept called the 
// "disqualification limit," which affects how much of the total amount is required for 
// payment. This can lead to discrepancies in reported percentages.

// It’s important to interpret the results carefully. The way payments are adjusted can 
// depend on these limits, which may cause differences in the percentages reported for 
// different accounts.
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -171,27 +171,35 @@ fn loading_test_with_randomized_params() {
...
-            administrate_single_scenario_result(
-                payment_adjuster_result,
-                account_infos,
-                scenario.applied_thresholds,
-                required_adjustment,
-                cw_service_fee_balance_minor,
-            )
-        })
-        .collect();
+    let (test_overall_output_collector, scenario_adjustment_results) =
```
utkarshg6, line 174: Let's just call these arguments as `output_collector` and `results`.

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -171,27 +171,35 @@ fn loading_test_with_randomized_params() {
...
+                    scenario.applied_thresholds,
+                    required_adjustment,
+                    cw_service_fee_balance_minor,
+                );
+
+                scenario_results.push(scenario_result);
+
+                (t_o_c_i_f, scenario_results)
+            },
+        );
```
utkarshg6, line 202: Please use this refactored code:

```rust
    let output_collector_init = first_stage_output.output_collector;
    let (output_collector, results) = second_stage_scenarios.into_iter().fold(
        (output_collector_init, vec![]),
        |(collector, mut results), scenario| {
            let adjustment = scenario.prepared_adjustment;

            let account_info =
                preserve_account_infos(&adjustment.adjustment_analysis.accounts, now);
            let adjustment_type = adjustment.adjustment_analysis.adjustment.clone();
            let wallet_balance = adjustment.agent.service_fee_balance_minor();

            let result = subject.adjust_payments(adjustment);

            let (updated_collector, single_result) = administrate_single_scenario_result(
                collector,
                result,
                account_info,
                scenario.applied_thresholds,
                adjustment_type,
                wallet_balance,
            );

            results.push(single_result);
            (updated_collector, results)
        },
    );

    render_results_to_file_and_attempt_basic_assertions(
        results,
        number_of_requested_scenarios,
        output_collector,
    )
```
    
    
To reduce the number of arguments passed in the fn `administrate_single_scenario_result` you may use this new structure.
    
```rust
    
// Create a struct to group scenario data
struct ScenarioData {
    applied_thresholds: AppliedThresholds,
    adjustment_type: Adjustment,
    wallet_balance: u128,
    account_info: Vec<AccountInfo>,
}

// Then your function call becomes:
let scenario_data = ScenarioData {
    applied_thresholds: scenario.applied_thresholds,
    adjustment_type,
    wallet_balance,
    account_info,
};

let (updated_collector, single_result) = administrate_single_scenario_result(
    collector,
    result,
    scenario_data,
);
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -256,21 +264,28 @@ fn generate_debt_age(gn: &mut ThreadRng, thresholds: &PaymentThresholds) -> u64
...
-    // occurrence of accounts with balances having different magnitudes in the frame of a single
-    // scenario. This was crucial to me so much that I was ready to write even this piece of code
-    // a bit crazy by look.
-    // This setup worked well to stress the randomness I needed, a lot more significant compared to
-    // what the naked number generator can put for you. Using some nesting, it broke the rigid
-    // pattern and gave an existence to accounts with diverse balances.
-    let mut generate_u128 = || generate_non_zero_usize(gn, 100) as u128;
+    // This seems overcomplicated, damn. Yet it's a result of good simple intentions. I wanted
+    // to ensure the occurrence of accounts with balances of different magnitudes to be generated
+    // for a single scenario. This was crucial to me so much that I didn't stop myself from writing
```
utkarshg6, line 276: The second sentence will be clearer if you write it like this:

```
I wanted to make sure that accounts with balances of different sizes are generated for a single scenario.
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -256,21 +264,28 @@ fn generate_debt_age(gn: &mut ThreadRng, thresholds: &PaymentThresholds) -> u64
...
+    // this fishy-looking piece of code.
+    // This setup worked well for the randomness I needed, a lot significantly more compared to
+    // what the default number generator from this library seemed to be able to provide only.
+    // Using some nesting, it scattered the distribution better and allowed me to have accounts
+    // with diverse balances.
```
utkarshg6, line 281: Also, for the second paragraph please use this:

```
// This setup worked better for the randomness I needed compared to the default number 
// generator from the library, which didn't provide enough variety. By using some nesting, 
// I was able to improve the distribution of balances and create accounts with more diverse 
// amounts.
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -311,40 +314,47 @@ fn try_make_qualified_payables_by_applied_thresholds(
...
 }
 
+fn make_qualified_payables_for_individualized_thresholds(
+    payable_accounts: Vec<PayableAccount>,
+    now: SystemTime,
+    payment_inspector: &PayableInspector,
+    individual_thresholds: &HashMap<Address, PaymentThresholds>,
+) -> Vec<QualifiedPayableAccount> {
+    let vec_of_thresholds = individual_thresholds.values().collect_vec();
+    let zipped = payable_accounts.into_iter().zip(vec_of_thresholds.iter());
```
utkarshg6, line 344: Please don't rely on `zip()` here, it can cause ordering mismatches. Please refer to the code below:

```rust
fn make_qualified_payables_for_individualized_thresholds(
    payable_accounts: Vec<PayableAccount>,
    now: SystemTime,
    payment_inspector: &PayableInspector,
    individual_thresholds: &HashMap<Address, PaymentThresholds>,
) -> Vec<QualifiedPayableAccount> {
    payable_accounts
        .into_iter()
        .flat_map(|payable_account| {
            let wallet_address = payable_account.wallet.address();
            let thresholds = individual_thresholds.get(&wallet_address).unwrap();
            
            make_single_qualified_payable_opt(
                payable_account,
                payment_inspector,
                thresholds,
                false,
                now,
            )
        })
        .collect()
}
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -378,14 +388,19 @@ fn pick_appropriate_cw_service_fee_balance(
     qualified_payables: &[QualifiedPayableAccount],
     accounts_count: usize,
 ) -> u128 {
-    // Value picked empirically
-    const COEFFICIENT: usize = 1000;
+    // Values picked empirically
+    const COEFFICIENT_A: usize = 1000;
+    const COEFFICIENT_B: usize = 2;
```
utkarshg6, line 393: I don't mind this code here, but these magic numbers causes problems in the future, please avoid them whenever possible.

Apart from that, the names only say that they are two coefficients but they tell nothing what these coefficients represent.

Here are alternative names:

```rust
    const BALANCE_MULTIPLIER: usize = 1000;
    const MIN_FRACTION_PIECES: usize = 2;
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -378,14 +388,19 @@ fn pick_appropriate_cw_service_fee_balance(
...
+
     let balance_average = sum_as(qualified_payables, |account| {
         account.initial_balance_minor()
     }) / accounts_count as u128;
```
utkarshg6, line 397: Also, it's better to name it in a format like: `average_debt_per_account`, and it should be calculated in two steps:


```rust
let total_debt = sum_as(qualified_payables, |account| {
        account.initial_balance_minor()
});
let average_debt_per_account = total_debt / accounts_count as u128;
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -378,14 +388,19 @@ fn pick_appropriate_cw_service_fee_balance(
...
-    let max_pieces = accounts_count * COEFFICIENT;
-    let number_of_pieces = generate_usize(gn, max_pieces - 2) as u128 + 2;
-    balance_average / COEFFICIENT as u128 * number_of_pieces
+
+    let max_pieces = accounts_count * COEFFICIENT_A;
+    let number_of_pieces =
+        generate_usize(gn, max_pieces - COEFFICIENT_B) as u128 + COEFFICIENT_B as u128;
+
+    balance_average / COEFFICIENT_A as u128 * number_of_pieces
```
utkarshg6, line 403: Another suggestion, use this please, explanation is necessary here:

```rust
    // This ensures the wallet has insufficient funds (less than total debt needed) 
    // by generating a random fraction of the total debt
    let max_fraction_pieces = accounts_count * BALANCE_MULTIPLIER;
    let random_fraction_pieces = generate_usize(gn, max_fraction_pieces - MIN_FRACTION_PIECES) as u128 
        + MIN_FRACTION_PIECES as u128;

    // This creates a wallet balance that's typically much less than total debt
    let base_unit = average_debt_per_account / BALANCE_MULTIPLIER as u128;
    base_unit * random_fraction_pieces
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -576,6 +598,35 @@ fn merge_information_about_particular_account(
         .collect()
 }
 
+fn prepare_interpretable_adjustment_results(
+    merged: Vec<(AccountInfo, Option<PayableAccount>)>,
+) -> (Vec<InterpretableAccountAdjustmentResult>, usize) {
+    merged.into_iter().fold(
```
utkarshg6, line 604: Instead of `fold()` a better choice would be to simply iterate over it.

```rust
fn prepare_interpretable_adjustment_results(
    merged: Vec<(AccountInfo, Option<PayableAccount>)>,
) -> (Vec<InterpretableAccountAdjustmentResult>, usize) {
    let mut eliminated_accounts_count = 0;
    
    let interpretable_accounts = merged
        .into_iter()
        .map(|(account_info, maybe_adjusted_account)| {
            let (interpretable_account, was_eliminated) =
                InterpretableAccountAdjustmentResult::new(account_info, maybe_adjusted_account);
            
            if was_eliminated {
                eliminated_accounts_count += 1;
            }
            
            interpretable_account
        })
        .collect();

    (interpretable_accounts, eliminated_accounts_count)
}
```

node/src/accountant/payment_adjuster/non_unit_tests/mod.rs
```diff
@@ -576,6 +598,35 @@ fn merge_information_about_particular_account(
...
+fn look_after_adjustment_statistics(
+    test_overall_output_collector: &mut TestOverallOutputCollector,
+    adjusted_accounts_within_this_scenario: usize,
+) {
+    if adjusted_accounts_within_this_scenario > 1 {
+        test_overall_output_collector.scenarios_with_more_than_one_adjustment += 1
+    }
+    test_overall_output_collector.adjusted_account_count_total +=
+        adjusted_accounts_within_this_scenario;
+}
```
utkarshg6, line 628: I don't know why this specific statistic is required, if you could answer `Why it exist?` as a comment inside this code, that'll be great.

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -54,15 +54,43 @@ use masq_lib::utils::convert_collection;
...
+// by service fee follows up (or it may take the first place if the need for the previous step was
+// missing).
+
+// Here comes the recursive part. The algorithm iterates through the weighted accounts and assigns
+// a proportional portion of the available means to them. Since the initial stage of the Payment-
+// Adjuster, where accounts were tested on causing insolvency, they've remained equipped with
+// a computed parameter of the so-called disqualification limit. This limit determines
+// if the weight-derived assignment of the money is too low or enough. If it is below the limit,
+// the account is removed from consideration. However, only a single account with the smallest
+// weight is eliminated per each recursion.
```
utkarshg6, line 82: You want to say, `per each recursive call` here.

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -54,15 +54,43 @@ use masq_lib::utils::convert_collection;
...
 //
-// Once the new calculator exists, its place belongs in the vector of calculators which is the heart
-// of this module.
+// If it is necessary to adjust the set by the transaction fee, the accounts are sorted
+// by the weights and only accounts that fit together under the limit are kept in. The adjustment
+// by service fee follows up (or it may take the first place if the need for the previous step was
+// missing).
+
+// Here comes the recursive part. The algorithm iterates through the weighted accounts and assigns
+// a proportional portion of the available means to them. Since the initial stage of the Payment-
```
utkarshg6, line 77: Change `means` to `funds` here.

Also the next sentence, needs clarity, change it from:

```
// Since the initial stage of the Payment-Adjuster, where 
// accounts were tested on causing insolvency, they've 
// remained equipped with a computed parameter of the
// so-called disqualification limit.
```

to

```
// During the initial stage of the Payment Adjuster, 
// accounts were evaluated for their potential to 
// cause insolvency. Each account now has a computed 
// parameter known as the disqualification limit.
```

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -54,15 +54,43 @@ use masq_lib::utils::convert_collection;
...
+// Adjuster, where accounts were tested on causing insolvency, they've remained equipped with
+// a computed parameter of the so-called disqualification limit. This limit determines
+// if the weight-derived assignment of the money is too low or enough. If it is below the limit,
+// the account is removed from consideration. However, only a single account with the smallest
+// weight is eliminated per each recursion.
+
+// The pool of money remains the same, but the set of accounts is reduced. The recursion repeats.
+// If none of the accounts disqualifies, it means all accounts were proposed with enough money.
+// Although the proposals may exceed the disqualification limits, only the value of the limit is
+// dedicated to the selected accounts.
```
utkarshg6, line 87: Also for this:

```
// The total pool of funds remains unchanged, but the number of accounts is reduced. 
// The recursion process repeats. If none of the accounts are disqualified, 
// it indicates that all accounts have received adequate funding. 
// Although the proposed amounts may exceed the disqualification limits, 
// only the amount equal to the limit is allocated to the selected accounts.
```

You know this algorithm is confusing because first you are allocating funds but later there's another check based on disqualification limit, which confuses what will be the behaviour of the algorithm will be during a dry-run.

I'm not suggesting changing the algorithm here, just wanted to share my pov.

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -54,15 +54,43 @@ use masq_lib::utils::convert_collection;
...
+
+// The remaining assets are later allocated to the accounts based on the order by their weights,
+// exhausting them fully one account after another up their 100% allocation until there is any
+// money that can be distributed.
```
utkarshg6, line 91: Few grammatical changes:

`based on the order by their weights` -> `based on the order of their weights`
`up their 100% allocation` -> `up to their 100% allocation`
`until there is any money that can be distributed` -> `until there is no money left to be distributed`

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -184,35 +212,29 @@ impl PaymentAdjusterReal {
...
     }
 
     fn resolve_initial_adjustment_dispatch(
         &self,
         weighed_payables: Vec<WeighedPayable>,
-    ) -> Result<
-        Either<Vec<AdjustedAccountBeforeFinalization>, Vec<PayableAccount>>,
-        PaymentAdjusterError,
-    > {
+    ) -> Result<AccountsByFinalization, PaymentAdjusterError> {
```
utkarshg6, line 232: That's good!

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -248,12 +267,16 @@ impl PaymentAdjusterReal {
...
+            ))
         } else {
             let accounts_not_needing_adjustment =
                 convert_collection(weighed_accounts_affordable_by_transaction_fee);
 
-            Ok(Either::Right(accounts_not_needing_adjustment))
+            Ok(AccountsByFinalization::Finalized(
+                accounts_not_needing_adjustment,
+            ))
         }
```
utkarshg6, line 280: Optional (If you're too excited): I think this fn is too long, maybe refactor the if-else block to an outboard method.

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -437,18 +456,35 @@ impl AdjustmentAnalysisReport {
...
+    //     number_of_accounts: usize,
+    //     transaction_fee_opt: Option<TransactionFeeImmoderateInsufficiency>,
+    //     service_fee_opt: Option<ServiceFeeImmoderateInsufficiency>,
+    // },
+    // AbsolutelyInsufficientServiceFeeBalancePostTxFeeAdjustment {
+    //     original_number_of_accounts: usize,
+    //     number_of_accounts: usize,
+    //     original_total_service_fee_required_minor: u128,
+    //     cw_service_fee_balance_minor: u128,
+    // },
```
utkarshg6, line 473: Remove these commented out lines.

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -466,12 +502,14 @@ pub struct ServiceFeeImmoderateInsufficiency {
...
+                detection_phase, ..
+            } => match detection_phase {
+                DetectionPhase::InitialCheck { .. } => true,
+                DetectionPhase::PostTxFeeAdjustment { .. } => true,
+            },
+            PaymentAdjusterError::RecursionEliminatedAllAccounts => true,
+            // We haven't needed to worry about this matter, yet this is rather a future alarm that
             // will draw attention after somebody adds a possibility for an error not necessarily
             // implying that an insolvency was detected before. At the moment, each error occurs
             // only alongside an actual insolvency. (Hint: There might be consequences for
```
utkarshg6, line 515: Some rephrasing for clarity:

"We haven't needed to worry about this matter" -> "We have not had to worry about this issue." 

"yet this is rather a future alarm" should be "but this is rather a future alarm"

"whose forming takes place back out" should be "whose formation takes place outside."

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -957,7 +1005,7 @@ mod tests {
         inputs
             .into_iter()
             .for_each(|(error, expected_msg)| assert_eq!(error.to_string(), expected_msg));
-        assert_eq!(inputs_count, PaymentAdjusterError::VARIANT_COUNT + 2)
+        assert_eq!(inputs_count, PaymentAdjusterError::VARIANT_COUNT + 3)
```
utkarshg6, line 1008: Why did you increase it by one?

node/src/accountant/payment_adjuster/mod.rs
```diff
@@ -1018,7 +1076,7 @@ mod tests {
             .map(|err| err.insolvency_detected())
             .collect::<Vec<_>>();
         assert_eq!(results, vec![true, true, true, true, true]);
-        assert_eq!(inputs_count, PaymentAdjusterError::VARIANT_COUNT + 2)
+        assert_eq!(inputs_count, PaymentAdjusterError::VARIANT_COUNT + 3)
```
utkarshg6, line 1079: here too

utkarshg6 (commented):
Continue with masq_lib/src/percentage.rs

masq_lib/src/percentage.rs
```diff
@@ -51,174 +29,110 @@ macro_rules! impl_percentage_integer {
...
-        i16: TryFrom<N>,
-        <i16 as TryFrom<N>>::Error: Debug,
-    {
-        let to_add = self.of(num);
+    // This function helps to correct the last digit of the resulting integer to be as close
+    // as possible to the hypothetical fractional number, if we could go beyond the decimal point.
+    fn __derive_rounding_increment(remainder: N) -> N {
+        let is_negative = remainder < N::try_from(0).expect("Each integer has 0");
+        let is_minor =
+            Self::__abs(remainder, is_negative) < N::try_from(50).expect("Each integer has 50");
```
utkarshg6, line 121: This part is hard to understand, why's that? Specifically the number 50, it's origin.

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+            .expect("Mathematically impossible")
+    }
+
+    fn __handle_upper_overflow(&self, num: N) -> N {
+        let (base, remainder) = base_and_rem_from_div_100(num);
+        let percents = N::try_from(self.degree as i8).expect("Each integer has 100");
+        let percents_of_base = base * percents;
+        let (percents_of_remainder, nearly_lost_tail) =
+            base_and_rem_for_ensured_i16(remainder, percents);
+        let final_rounding_element = Self::__derive_rounding_increment(nearly_lost_tail);
```
utkarshg6, line 156: Let's call it `rounding_adjustment` instead.

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+}
+
+fn base_and_rem_from_div_100<N>(num: N) -> (N, N)
+where
+    N: PercentageInteger,
+    <N as TryFrom<i8>>::Error: Debug,
+{
+    let hundred = N::try_from(100i8).expect("Each integer has 100");
+    let modulo = num % hundred;
+    (num / hundred, modulo)
```
utkarshg6, line 223: I'm noticing a pattern here, maybe it's time to create a new structure for this, what about `QuotientAndRemainder`?

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+}
+
+// This is a wider type that allows to specify cumulative percents of more than only 100.
+// The expected use of this would look like requesting percents meaning possibly multiples of 100%,
+// roughly, of a certain base number. Similarly to the PurePercentage type, also signed numbers
+// would be accepted.
```
utkarshg6, line 229: Some grammatical corrections:

"cumulative percents of more than only 100" -> "cumulative percents greater than 100"

"requesting percents meaning possibly multiples of 100%" -> "requesting percents that may represent multiples of 100%"

"Similarly to the PurePercentage type, also signed numbers would be accepted." -> "Similar to the PurePercentage type, signed numbers would also be accepted."

utkarshg6 (commented):
Continue at masq_lib/src/percentage.rs:232

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LoosePercentage {
+    multiplier_of_100_percent: u32,
+    degrees_from_remainder: PurePercentage,
```
utkarshg6, line 234: Instead of calling `multiplier_of_100_percent` you can simply call it `quotient` and `degrees_from_remainder` to `remainder`.

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+}
+
+impl LoosePercentage {
+    pub fn new(percents: u32) -> Self {
+        let multiples_of_100_percent = percents / 100;
+        let remainder = (percents % 100) as u8;
+        let degrees_from_remainder =
+            PurePercentage::try_from(remainder).expect("Should never happen.");
```
utkarshg6, line 242: You may just call these values for clearer expression:

`multiples_of_100_percent` -> `quotient`
`remainder` -> `remainder_value`
`degrees_from_remainder` -> `remainder`

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+
+    // If this returns an overflow error, you may want to precede this by converting the base
+    // number to a larger integer
+    pub fn of<N>(&self, num: N) -> Result<N, BaseTypeOverflow>
+    where
+        N: PercentageInteger + TryFrom<u32>,
+        <N as TryFrom<i8>>::Error: Debug,
+        <N as TryFrom<u32>>::Error: Debug,
+        i16: TryFrom<N>,
+        <i16 as TryFrom<N>>::Error: Debug,
```
utkarshg6, line 257: Over here, please reorder them such that the error types are together, and the core and type conversions are separate, even separating with the blank lines are sufficient or you can add comments and do something like this:

```rust
        // Core
        N: PercentageInteger + TryFrom<u32>,
        // Error
        <N as TryFrom<i8>>::Error: Debug,
        <N as TryFrom<u32>>::Error: Debug,
        <i16 as TryFrom<N>>::Error: Debug,
        // Reverse Conversion
        i16: TryFrom<N>,
```

Apparently, you could create a trait:

```rust
pub trait ExtendedPercentageInteger: 
    PercentageInteger + TryFrom<u32>
where
        // Error
        <Self as TryFrom<i8>>::Error: Debug,
        <Self as TryFrom<u32>>::Error: Debug,
        <i16 as TryFrom<Self>>::Error: Debug,
        // Reverse Conversion
        i16: TryFrom<Self>,
{
}
```

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+                        multiplier,
+                        type_name::<N>()
+                    ),
+                })
+            }
+        };
+
+        let remainder = self.degrees_from_remainder.of(num);
+
+        match wholes.checked_add(&remainder) {
```
utkarshg6, line 289: Optional: You may save the result in the variable and then return that variable from this fn.

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+                    type_name::<N>()
+                ),
+            }),
+        }
     }
+
+    // Note that functions like 'add_percents_to' or 'subtract_percents_from' don't need to be
+    // implemented here, even though they are at the 'PurePercentage'. You can substitute them
+    // simply by querying 100 + <your desired addition in percents> or 100 - <your desired
+    // subtraction in percents in the interval (1..=99) >
```
utkarshg6, line 305: Instead of the second sentence, you can directly provide usage examples:

```rust
    // - To increase by X%: use `LoosePercentage::new(100 + X)`
    // - To decrease by X%: use `LoosePercentage::new(100 - X)` where X is in range 1..=99
```

utkarshg6 (commented):
Continue at masq_lib/src/percentage.rs:345

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+        N: PercentageInteger,
+        <N as TryFrom<i8>>::Error: Debug,
+        i16: TryFrom<N>,
+        <i16 as TryFrom<N>>::Error: Debug,
+    {
+        assert_against_literal_value(subject, num, expected_literal_num);
+
+        let trivially_calculated_half = num / N::try_from(2).unwrap();
+        // Widening the bounds to compensate the extra rounding
+        let one = N::try_from(1).unwrap();
```
utkarshg6, line 359: Let's call this variable `rounding_tolerance`.

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
+        expected_literal_num: N,
+    ) where
+        N: PercentageInteger,
+        <N as TryFrom<i8>>::Error: Debug,
+        i16: TryFrom<N>,
+        <i16 as TryFrom<N>>::Error: Debug,
+    {
+        assert_against_literal_value(subject, num, expected_literal_num);
+
+        let trivially_calculated_half = num / N::try_from(2).unwrap();
```
utkarshg6, line 357: Maybe just call it: `half_of_input` 🤷

masq_lib/src/percentage.rs
```diff
@@ -227,91 +141,271 @@ impl PurePercentage {
...
         N: PercentageInteger,
         <N as TryFrom<i8>>::Error: Debug,
         i16: TryFrom<N>,
         <i16 as TryFrom<N>>::Error: Debug,
     {
-        assert_eq!(subject.of(num), expected);
-        let half = num / N::try_from(2).unwrap();
+        assert_against_literal_value(subject, num, expected_literal_num);
+
+        let trivially_calculated_half = num / N::try_from(2).unwrap();
```
utkarshg6, line 382: Similarly, you can do the same here.

masq_lib/src/percentage.rs
```diff
@@ -421,78 +512,66 @@ mod tests {
...
-                898784545,
-                RoundingTo::SmallerPositive,
-                RoundingTo::SmallerNegative,
-            ),
+            (49, 0),
+            (50, 1),
+            (51, 1),
+            (5, 0),
+            (99,1),
+            (0,0)
```
utkarshg6, line 520: This is simple, yet so effective in explaining what's going on... nicely done @bertllll.

multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs
```diff
@@ -41,46 +40,39 @@ use std::thread;
...
+use web3::types::{
+    Address, Bytes, SignedTransaction, TransactionParameters, TransactionReceipt,
+    TransactionRequest,
+};
 use web3::Web3;
 
 pub type StimulateConsumingNodePayments = fn(&mut MASQNodeCluster, &MASQRealNode, &WholesomeConfig);
 
-pub type StartServingNodesAndLetThemPerformReceivablesCheck =
+pub type StartServingNodesAndLetThemActivateTheirAccountancy =
```
utkarshg6, line 51: This name is too long, prefer this `ActivateServingNodesWithAccountancy` or maybe `NodeActivationHandler`. Please change this name, for god's sake, that's toooooooooooo long.

multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs
```diff
@@ -41,46 +40,39 @@ use std::thread;
...
     fn(&mut MASQNodeCluster, &WholesomeConfig) -> [MASQRealNode; 3];
 
 pub fn test_body(
-    test_inputs: TestInputs,
+    test_inputs: TestInput,
     assertions_values: AssertionsValues,
-    stimulate_consuming_node_to_pay: StimulateConsumingNodePayments,
-    start_serving_nodes_and_activate_their_accountancy: StartServingNodesAndLetThemPerformReceivablesCheck,
+    stimulate_consuming_node_payments: StimulateConsumingNodePayments,
+    start_serving_nodes_and_let_them_activate_their_accountancy: StartServingNodesAndLetThemActivateTheirAccountancy,
```
utkarshg6, line 58: You can call this argument name as `serving_nodes_activator`.

multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs
```diff
@@ -94,7 +86,7 @@ pub fn test_body(
     }
     wholesome_config.assert_payments_via_direct_blockchain_scanning(&assertions_values);
 
-    let _ = start_serving_nodes_and_activate_their_accountancy(
+    let _ = start_serving_nodes_and_let_them_activate_their_accountancy(
```
utkarshg6, line 89: Please do the same here

utkarshg6 (commented):
Continue at multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs:282

multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs
```diff
@@ -259,21 +245,277 @@ impl FinalServiceFeeBalancesByServingNodes {
...
+
+    fn await_transaction_receipt(
+        &self,
+        tx_hash: H256,
+    ) -> Result<Option<TransactionReceipt>, web3::error::Error> {
+        self.web3_transport
+            .eth()
+            .transaction_receipt(tx_hash)
+            .wait()
+    }
```
utkarshg6, line 511: Code in this file is well written. 👏

multinode_integration_tests/tests/verify_bill_payment_utils/utils.rs
```diff
@@ -376,56 +617,38 @@ impl NodeProfile for ServingNodeProfile {
 }
 
 pub fn establish_test_frame(
-    test_inputs: TestInputs,
+    test_inputs: TestInput,
```
utkarshg6, line 620: You can make the arg name singular too for consistency.

utkarshg6 (commented):
Finished! 👏

Issue #662, opened 2025-07-01
Remove pending_payable_rowid column and the related processing


bertllll:
_(Planned in the second phase)_

More description may be provided later. 

This card was created to unload a good portion of work from GH-642 and pass it on separated. 
It bears on the pending_payable_rowid column in the payable table. This column becomes redundant as there are now different mechanisms which control if a new scan for payable should be run or rather prevented. It further stretches across the PayableDao trait, the method mark_pending_payable and all functions of similar name which also were left in the code yet.

The reason why this part was split off is that the payments will work even if this card isn't played yet and therefore not having to deal with this card could bring the whole-feature testing phase in less time. However, this card shouldn't be forgotten to be played because it implies a lot messy dead code staying in the codebase.

utkarshg6:
I propose we link the payable table using `tx_hash` instead of `rowid`. 

1. **Linking Transactions**: Use `tx_hash` to connect transactions in the payable table.
  
2. **Handling Retries**:
   - If a transaction needs to be retried, update the `tx_hash` of the retry transaction to the existing record.
   - Once the transaction is successful, delete the record.

3. **Failed Retries**: 
   - If a retry fails and a new transaction is created, link the new `tx_hash` to the same record.

This approach will help us efficiently track transactions and their retries.

bertllll:
> I propose we link the payable table using `tx_hash` instead of `rowid`.
> 
>     1. **Linking Transactions**: Use `tx_hash` to connect transactions in the payable table.
> 
>     2. **Handling Retries**:
>        
>        * If a transaction needs to be retried, update the `tx_hash` of the retry transaction to the existing record.
>        * Once the transaction is successful, delete the record.
> 
>     3. **Failed Retries**:
>        
>        * If a retry fails and a new transaction is created, link the new `tx_hash` to the same record.
> 
> 
> This approach will help us efficiently track transactions and their retries.

I'm sorry but I disagree. This efforts would cost time and energy despite it would produce just an overhead not helping anything. I have multiple arguments to explain why:

1) 
[...]

PR #715, opened 2025-09-30
GH-662: Clean-up of the obsolete pending payable param in PayableAccount and financials command fix with some SQL work 

bertllll:


USER-INTERFACE-INTERFACE.md
```diff
@@ -826,9 +829,25 @@ payment was later also confirmed on the blockchain.
...
+
+The second, `failures`, is a positive number of times we've tried to send this same tx to the blockchain in this latest 
+payment, but it has failed repeatedly. The tx always bears the same nonce and is bound for the same wallet. 
+There are three possible observable states:
+
+1. `pendingTxHashOpt` is not null and `failures` is 0. This means that we've yet made only a single attempt to send 
+a tx to the blockchain, and the tx is pending.
+2. `pendingTxHashOpt` is not null and `failures` is greater than 0. This means that we've made more than one attempt. 
+There is an alive pending tx, plus we've experienced X failures before that.
+3. `pendingTxHashOpt` is null and `failures` is greater than 0. This means that the payment mechanism is in a mid-state
```
utkarshg6, line 848: This case is bit confusing, we should always have a transaction hash, as it is generated locally.

masq/src/commands/financials_command/mod.rs
```diff
@@ -1028,15 +1028,27 @@ mod tests {
...
                         },
                         UiPayableAccount {
                             wallet: "0x6DbcCaC5596b7ac986ff8F7ca06F212aEB444440".to_string(),
                             age_s: 150000,
-                            balance_gwei: 8,
-                            pending_payable_hash_opt: Some(
-                                "0x0290db1d56121112f4d45c1c3f36348644f6afd20b759b762f1dba9c4949066e"
-                                    .to_string(),
+                            balance_gwei: 999888777,
+                            tx_processing_info_opt: Some(TxProcessingInfo{pending_tx_hash_opt: Some("0x0290db1d56121112f4d45c1c3f36348644f6afd20b759b762f1dba9c4949066e".to_string()),failures: 0}
```
utkarshg6, line 1037: According to me, the inner Some shouldn't exist, tx hash should be a compulsory value.

node/src/accountant/db_access_objects/utils.rs
```diff
@@ -303,27 +304,34 @@ impl<T: Copy> From<&RangeQuery<T>> for CustomQuery<T> {
...
+                        gwei
+                    } else {
+                        panic!(
+                            "Broken code: PayableAccount with less than 1 gwei passed through \
+                            db query constraints; wallet: {}, balance: {}",
+                            account.wallet, account.balance_wei
+                        )
+                    }
+                },
+                tx_processing_info_opt: account_with_tx_info.tx_opt,
```
utkarshg6, line 333: Names could be consistent here.

node/src/accountant/db_access_objects/utils.rs
```diff
@@ -303,27 +304,34 @@ impl<T: Copy> From<&RangeQuery<T>> for CustomQuery<T> {
...
+            let account = account_with_tx_info.account;
+            UiPayableAccount {
+                wallet: account.wallet.to_string(),
+                age_s: to_age(account.last_paid_timestamp),
+                balance_gwei: {
+                    let gwei = (account.balance_wei / (WEIS_IN_GWEI as u128)) as u64;
+                    if gwei > 0 {
+                        gwei
+                    } else {
+                        panic!(
```
utkarshg6, line 326: This panic could be abstracted if you use a `TryFrom` conversion here, it'll look cleaner if you do that way.

node/src/accountant/db_access_objects/utils.rs
```diff
@@ -334,14 +342,23 @@ pub fn remap_receivable_accounts(accounts: Vec<ReceivableAccount>) -> Vec<UiRece
...
-                                                 account.wallet,
-                                                 account.balance_wei
-                )}
+            balance_gwei: {
+                let gwei = (account.balance_wei / (WEIS_IN_GWEI as i128)) as i64;
+                if gwei != 0 {
+                    gwei
+                } else {
+                    panic!(
+                        "Broken code: ReceivableAccount with balance between {} and 0 gwei \
```
utkarshg6, line 351: You may prefer to use the same code for conversion. If you decide to do so then prefer this piece of code over the other.

node/src/accountant/db_access_objects/utils.rs
```diff
@@ -564,17 +581,21 @@ mod tests {
...
+                tx_opt: None,
             },
-            PayableAccount {
-                wallet: make_wallet("ac3665c"),
-                balance_wei: 565_122_333,
-                last_paid_timestamp: SystemTime::now(),
-                pending_payable_opt: None,
+            PayableAccountWithTxInfo {
+                account: PayableAccount {
+                    wallet: make_wallet("ac3665c"),
```
utkarshg6, line 594: Optional: Prefer Using Address over Wallet. (I know it's a time consuming change, that's why I kept it as optional)

node/src/accountant/mod.rs
```diff
@@ -6253,11 +6247,13 @@ mod tests {
...
-        let payable_accounts_retrieved = vec![PayableAccount {
-            wallet: make_wallet("abcd123"),
-            balance_wei: 58_568_686_005,
-            last_paid_timestamp: SystemTime::now().sub(Duration::from_secs(5000)),
-            pending_payable_opt: None,
+        let payable_accounts_retrieved = vec![PayableAccountWithTxInfo {
+            account: PayableAccount {
+                wallet: make_wallet("abcd123"),
+                balance_wei: 58_568_686_005,
+                last_paid_timestamp: SystemTime::now().sub(Duration::from_secs(5000)),
```
utkarshg6, line 6254: Timestamp should be an `u64`, instead of `SystemTime` because precision here is useless.

utkarshg6 (commented):
Continue at node/src/accountant/db_access_objects/payable_dao.rs

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -347,48 +307,104 @@ impl PayableDaoReal {
...
                     wallet,
                     balance_wei: checked_conversion::<i128, u128>(BigIntDivider::reconstitute(
                         high_bytes, low_bytes,
                     )),
                     last_paid_timestamp: utils::from_unix_timestamp(last_paid_timestamp),
-                    pending_payable_opt: None,
-                })
-            }
+                },
+                tx_opt: Self::maybe_construct_tx_info(tx_hash_opt, previous_failures),
```
utkarshg6, line 342: This function is not a good idea, there is a lot of uncertainty. In this whole PR, these Optional values are my biggest concern, prefer using absolute values so that guaranteed values are returned.

utkarshg6 (commented):
Continue at node/src/accountant/db_access_objects/payable_dao.rs:367

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -347,48 +307,104 @@ impl PayableDaoReal {
...
-               balance_low_b,
-               last_paid_timestamp
-           from
-               payable
-           {} {}
-           order by
-               {},
-               {}
-           {}",
+            "SELECT
```
utkarshg6, line 367: Please add a comment at the top of this fn saying:

```rust
// A dynamic SQL query that joins payable, sent_payable, and failed_payable
// to provide comprehensive information about accounts that need to be paid, 
// including their current transaction processing status.
```

This comment only explains what this SQL does, but it's still a mystery _Why_ it exists?

The _Why_ is a more important question to answer here. So, please write an explanation why you wrote this code.

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -347,48 +307,104 @@ impl PayableDaoReal {
...
+                    WITH nonces_of_failures_by_wallet AS
+                    (
+                        SELECT nonce
+                        FROM failed_payable
+                        WHERE receiver_address = p.wallet_address
+                    )
+                    SELECT COUNT(*)
+                    FROM nonces_of_failures_by_wallet
+                    WHERE nonce = (SELECT MAX(nonce) FROM nonces_of_failures_by_wallet)
+                )
```
utkarshg6: The `THEN` clause can be refactored like this:

```rust
    THEN (
        SELECT COUNT(*) FROM failed_payable
        WHERE receiver_address = p.wallet_address
            AND nonce = (
                SELECT MAX(nonce) FROM failed_payable
                WHERE receiver_address = p.wallet_address
            )
    )
```

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -347,48 +307,104 @@ impl PayableDaoReal {
...
+                ELSE 0
+            END AS recent_failures_count
```
utkarshg6, line 398: You can replace the entire `EXISTS` clause with this code:

```rust
(
    SELECT COUNT(*)
    FROM failed_payable
    WHERE receiver_address = p.wallet_address
        AND status NOT LIKE '%Concluded%'
        AND nonce = (
            SELECT MAX(nonce)
            FROM failed_payable
            WHERE receiver_address = p.wallet_address
                AND status NOT LIKE '%Concluded%'
        )
) AS recent_failures_count
```

It'll automatically return 0 if no cases match.

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -347,48 +307,104 @@ impl PayableDaoReal {
...
-           {}",
+            "SELECT
+            p.wallet_address,
+            p.balance_high_b,
+            p.balance_low_b,
+            p.last_paid_timestamp,
+            CASE WHEN s.status LIKE '%Pending%'
+                THEN s.tx_hash
+                ELSE NULL
+            END AS pending_tx_hash,
```
utkarshg6, line 375: Always fetch the `tx_hash` and store the transaction status in a boolean called `is_pending`. Using `NULL` can cause confusion about whether the hash was ever present or removed conditionally. Since `NULL` values can lead to complications in SQL, it's best to avoid them.

You may use this code below:

```rust
s.tx_hash AS pending_tx_hash,
CASE WHEN s.status LIKE '%Pending%'
    THEN 1
    ELSE 0
END AS is_pending
```

utkarshg6 (commented):
Continue at node/src/accountant/db_access_objects/payable_dao.rs:871

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -1450,31 +1112,313 @@ mod tests {
...
+        let now = current_unix_timestamp();
+        let wallet_addr = "0x1111111111111111111111111111111111111111";
+        let mut sent_tx_1 = make_sent_tx(789);
+        sent_tx_1.receiver_address = Wallet::new(wallet_addr).address();
+        let sent_tx_hash_1 = sent_tx_1.hash;
+        let wallet_addr = "0x3333333333333333333333333333333333333333";
+        let mut sent_tx_2 = make_sent_tx(345);
+        sent_tx_2.receiver_address = Wallet::new(wallet_addr).address();
+        let sent_tx_hash_2 = sent_tx_2.hash;
+        let mut failed_tx_1 = make_failed_tx(123);
```
utkarshg6, line 1154: Optional: Call for a builder pattern here (everywhere in this test tbh, but it's up to you).

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -1450,31 +1112,313 @@ mod tests {
...
+            .custom_query(CustomQuery::TopRecords {
+                count: 3,
+                ordered_by: Balance,
+            })
+            .unwrap();
+
+        assert_eq!(
+            result,
+            vec![
+                PayableAccountWithTxInfo {
```
utkarshg6, line 1235: This is general comment, you don't need to fix anything in specific here. 

The thing is making assertions on everything can weaken test clarity, causing readers to overlook the specific fields or parameters we intended to highlight. This reduces code readability. I prefer to assert only on the items that we truly want to test.

node/src/accountant/db_access_objects/payable_dao.rs
```diff
@@ -1496,56 +1440,66 @@ mod tests {
...
-                "0x7777777777777777777777777777777777777777",
-                gwei_to_wei::<_, u64>(2_500_647_000u64),
+            insert_payable_record(
                 now - 80_333,
-                None,
+                "0x7777777777777777777777777777777777777777",
+                2_500_647_000,
             );
+            sent_payable_dao
+                .insert_new_records(&btreeset![SentTx {
```
utkarshg6, line 1491: Prefer creating `SentTx` using the builder method.

utkarshg6 (commented):
Continue at node/src/accountant/db_access_objects/payable_dao.rs:1716

utkarshg6 (commented):
Done!

Issue #598, opened 2025-03-21
Improve Stability in the Payment Process


utkarshg6:
I and Bert identified a couple of problems which are likely responsible for frequent delinquency bans in the network. 

The current mechanism doesn't seem reliable enough because of rather a chaotic rate of successes in pending payments being evaluated incorrectly compared to inspections done through the common online blockchain explorers. We shouldn't probably strive to pinpoint every little eventuality of this process, convinced for that, besides other things, by the differences in the local manifestation at particular Node runners participating in QA tests. It seems like it has quite a variable nature. What we have been able to conclude, though, is that our old design can be greatly improved by implementing these following ideas:

- Thorough the whole research it became more and more obvious that our code lacks the ability of resubmitting transactions which for some reason have never completed. It's a crucial feature which is probably a gold standard of the web3 ecosystem. If we implement a reasonable time limit for transactions to complete, it makes sense to retry with a slightly higher gas price than of that one at those previously given. That may imply the need of remembering more information about pending transactions than we do with at the moment. Particularly, we should newly also keep their nonce and gas price.
It's unclear regarding the importance of having the gas price because with every retry we will probably query the blockchain fro the recommended average gas price, which should be more meaningful than what we remember about the previously submitted transactions. Of course, if we found out that the recommended price goes lower than the one used previously, something would be quite wrong and we'd have to care of it. Therefore, maybe let's check it with an if statement if this always holds just to be sure the code is prepared for the reality well. 
Besides we do try to get the most relevant gas price by the RPC call, it's recommended that we add on some extra and ensure the transaction will be attractive for the miners.   

- From what we found out from a research, our choice of using the method from the Web3 JSON-RPC standard `eth_getTransactionReceipt` has been suboptimal or even wrong, with the big difference to `eth_getTransactionByHash` which can also answer if the so-believed pending transaction is still kept waiting in the transaction pool or otherwise it has actually disappeared and so it becomes a fool's errand to go on checking on its completion. It's commonly stated that transactions can be dropped by the transaction pool under certain circumstances. We think we witnessed some cases of this kind in the collected logs that a transaction not being findable by the third-party search tools was still regarded as pending which resulted into a high counts of attempts of querying its receipt. 
[...]

PR #722, opened 2025-10-03
GH-598: First QA fix

bertllll:


node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -79,8 +79,8 @@ impl From<PayableSequenceScanner> for ScanType {
 pub struct PayableScanScheduler {
     pub new_payable_notify_later: Box<dyn NotifyLaterHandle<ScanForNewPayables, Accountant>>,
     pub dyn_interval_computer: Box<dyn NewPayableScanDynIntervalComputer>,
-    pub inner: Arc<Mutex<PayableScanSchedulerInner>>,
-    pub new_payable_interval: Duration,
+    // pub inner: Arc<Mutex<PayableScanSchedulerInner>>,
+    // pub new_payable_interval: Duration,
```
utkarshg6, line 83: You may remove this code.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -89,24 +89,18 @@ impl PayableScanScheduler {
...
         Self {
             new_payable_notify_later: Box::new(NotifyLaterHandleReal::default()),
-            dyn_interval_computer: Box::new(NewPayableScanDynIntervalComputerReal::default()),
-            inner: Arc::new(Mutex::new(PayableScanSchedulerInner::default())),
-            new_payable_interval,
+            dyn_interval_computer: Box::new(NewPayableScanDynIntervalComputerReal::new(
+                new_payable_interval,
+            )),
+            // inner: Arc::new(Mutex::new(PayableScanSchedulerInner::default())),
+            // new_payable_interval,
```
utkarshg6, line 96: this too

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -132,6 +126,10 @@ impl PayableScanScheduler {
         }
     }
 
+    pub fn update_last_new_payable_scan_timestamp(&mut self) {
+        self.dyn_interval_computer.zero_out();
```
utkarshg6, line 130: What does this mean?

node/src/blockchain/errors/validation_status.rs
```diff
@@ -261,7 +246,7 @@ mod tests {
     // #[test]
     // fn previous_attempts_hash_works_correctly() {
     //     let now = SystemTime::now();
-    //     let clock = ValidationFailureClockMock::default()
+    //     let clock = SimpleClockMock::default()
```
utkarshg6, line 249: Why are we keeping this test like this, as a comment, shouldn't we decide whether we want to keep it or not?

utkarshg6 (commented):
Continue with node/src/accountant/test_utils.rs

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -152,49 +150,51 @@ impl PayableScanScheduler {
     }
 }
 
-pub struct PayableScanSchedulerInner {
-    pub last_new_payable_scan_timestamp: SystemTime,
-}
+pub trait NewPayableScanDynIntervalComputer {
```
utkarshg6, line 153: Please just call it as `ScanIntervalComputer`.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -152,49 +150,51 @@ impl PayableScanScheduler {
...
+    fn compute_interval(&self) -> Option<Duration>;
 
-impl Default for PayableScanSchedulerInner {
-    fn default() -> Self {
-        Self {
-            last_new_payable_scan_timestamp: UNIX_EPOCH,
-        }
-    }
-}
+    fn zero_out(&mut self);
```
utkarshg6, line 156: Let's call it `reset_timestamp` instead.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -152,49 +150,51 @@ impl PayableScanScheduler {
...
+}
 
 impl NewPayableScanDynIntervalComputer for NewPayableScanDynIntervalComputerReal {
-    fn compute_interval(
-        &self,
-        now: SystemTime,
-        last_new_payable_scan_timestamp: SystemTime,
-        interval: Duration,
-    ) -> Option<Duration> {
+    fn compute_interval(&self) -> Option<Duration> {
```
utkarshg6, line 168: These names are too vague, they don't represent what you're trying to achieve here.

Please use this instead:

```rust
#[derive(Debug, PartialEq, Eq, Clone)]
pub enum ScanTiming {
    ReadyNow,
    WaitFor(Duration),
}

fn time_until_next_scan(&self) -> ScanTiming {
    let current_time = self.clock.now();
    let time_since_last_scan = current_time
        .duration_since(self.last_scan_timestamp)
        .unwrap_or_else(|_| {
            panic!(
                "Current time ({:?}) is earlier than last scan timestamp ({:?})",
                current_time, self.last_scan_timestamp
            )
        });
    
    if time_since_last_scan >= self.scan_interval {
        ScanTiming::ReadyNow
    } else {
        ScanTiming::WaitFor(self.scan_interval - time_since_last_scan)
    }
}
```

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -427,7 +433,7 @@ mod tests {
 
     #[test]
     fn scan_dyn_interval_computer_computes_remaining_time_to_standard_interval_correctly() {
-        let now = SystemTime::now();
+        let (clock, now) = fill_simple_clock_mock_and_return_now();
```
utkarshg6, line 436: Let's use this name instead:

```
make_mock_clock_and_time
```

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -445,12 +451,17 @@ mod tests {
                 Duration::from_secs(4),
             ),
         ];
-        let subject = NewPayableScanDynIntervalComputerReal::default();
+        let mut subject = make_subject();
```
utkarshg6, line 454: This is too vague, please keep the prefix `make` but use the struct name for suffix. For example, `make_scan_interval_computer`.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -538,15 +553,76 @@ mod tests {
...
 
-        let _ = subject.compute_interval(
-            now.checked_sub(Duration::from_secs(1)).unwrap(),
-            now,
-            Duration::from_secs(32),
+        let _ = subject.compute_interval();
+    }
+
+    #[test]
+    fn zero_out_works_for_default_subject() {
```
utkarshg6, line 564: After you rename this function from `zero_out`, please rename this test and the test below.

node/src/accountant/scanners/scan_schedulers.rs
```diff
@@ -538,15 +553,76 @@ mod tests {
...
+            NewPayableScanDynIntervalComputerMock::default().zero_out_params(&zero_out_params_arc),
+        );
+
+        subject.payable.update_last_new_payable_scan_timestamp();
+
+        let zero_out_params = zero_out_params_arc.lock().unwrap();
+        assert_eq!(*zero_out_params, vec![()])
+    }
+
+    fn make_subject() -> NewPayableScanDynIntervalComputerReal {
```
utkarshg6, line 613: Migrate these functions and other code to the test_utils file.

node/src/accountant/mod.rs
```diff
@@ -973,6 +973,10 @@ impl Accountant {
                 None => Err(StartScanError::NoConsumingWalletFound),
             };
 
+        self.scan_schedulers
+            .payable
+            .update_last_new_payable_scan_timestamp();
```
utkarshg6, line 978: Let's call it `reset_scan_timer`.

node/src/accountant/mod.rs
```diff
@@ -2338,54 +2347,87 @@ mod tests {
...
+        )
+    }
+
+    fn accountant_sends_qualified_payable_msg_when_qualified_payable_found<ActorMessage>(
+        act_msg: ActorMessage,
+        initial_templates: Either<NewTxTemplates, RetryTxTemplates>,
+        zero_out_params_expected: Vec<()>,
+    ) where
+        ActorMessage: Message + Send + 'static,
+        ActorMessage::Result: Send,
```
utkarshg6, line 2380: Very Very Good, I'll have to give you 100/100 for this.

node/src/accountant/mod.rs
```diff
@@ -2338,54 +2347,87 @@ mod tests {
...
+        Accountant: Handler<ActorMessage>,
+    {
+        let zero_out_params_arc = Arc::new(Mutex::new(vec![]));
```
utkarshg6, line 2383: Also, you would like to rename this variable too.

node/src/accountant/mod.rs
```diff
@@ -2431,14 +2472,29 @@ mod tests {
...
+            .payable_scan_interval
+            .checked_sub(Duration::from_secs(5))
+            .unwrap();
+        let right_bound = default_scan_intervals
+            .payable_scan_interval
+            .checked_add(Duration::from_secs(5))
+            .unwrap();
+        // The divergence should be only a few milliseconds, definitely not seconds; the tested
+        // interval should be safe for slower machines too.
+        assert!(left_bound <= actual_interval && actual_interval <= right_bound);
```
utkarshg6, line 2497: Please change the last few lines to make it clearer.

```rust
    let default_interval = ScanIntervals::compute_default(TEST_DEFAULT_CHAIN).payable_scan_interval;
    let tolerance = Duration::from_secs(5);
    let min_interval = default_interval.checked_sub(tolerance).unwrap();
    let max_interval = default_interval.checked_add(tolerance).unwrap();
    
    assert!(
        min_interval <= scheduled_interval && scheduled_interval <= max_interval,
        "Expected interval between {:?} and {:?}, got {:?}",
        min_interval, max_interval, scheduled_interval
    );
```

node/src/accountant/mod.rs
```diff
@@ -2431,14 +2472,29 @@ mod tests {
...
             msg,
             ScanForNewPayables {
                 response_skeleton_opt: None
             }
         );
-        assert_eq!(interval, Duration::from_secs(500));
+        let default_scan_intervals = ScanIntervals::compute_default(TEST_DEFAULT_CHAIN);
+        // The previous last_new_payable_scan_timestamp is UNIX_EPOCH, if the interval was derived
+        // from that timestamp, it would result in an immediate-scan command. This implies that
+        // the last_new_payable_scan_timestamp was reset to zero, which is how it is meant to be.
```
utkarshg6, line 2486: Please rewrite this comment in the way:

```
        // Since no immediate scan was triggered, the timestamp was properly reset to zero
        // (rather than using UNIX_EPOCH which would cause immediate scheduling).
```

node/src/accountant/mod.rs
```diff
@@ -2794,7 +2850,7 @@ mod tests {
             .start_scan_params(&scan_params.receivable_start_scan)
             .start_scan_result(Err(StartScanError::NothingToProcess));
         let (subject, new_payable_expected_computed_interval, receivable_scan_interval) =
-            set_up_subject_for_no_pending_payables_found_startup_test(
+            set_up_subject_for_no_p_p_found_startup_test(
```
utkarshg6, line 2853: How about - `configure_accountant_for_empty_pending_payables_at_startup`

node/src/accountant/mod.rs
```diff
@@ -2814,20 +2870,19 @@ mod tests {
         let before = SystemTime::now();
         system.run();
         let after = SystemTime::now();
-        assert_pending_payable_scanner_for_no_pending_payable_found(
+        assert_pending_payable_scanner_for_no_p_p_found(
```
utkarshg6, line 2873: Previous name was far better, using `_p_p` is more like introducing jargon.

node/src/accountant/mod.rs
```diff
@@ -2893,8 +2948,8 @@ mod tests {
             .scan_started_at_result(None)
             .start_scan_params(&scan_params.receivable_start_scan)
             .start_scan_result(Err(StartScanError::NothingToProcess));
-        let (subject, pending_payable_expected_notify_later_interval, receivable_scan_interval) =
-            set_up_subject_for_some_pending_payable_found_startup_test(
+        let (subject, expected_pending_payable_notify_later_interval, receivable_scan_interval) =
+            set_up_subject_for_some_p_p_found_startup_test(
```
utkarshg6, line 2952: Similarly, this should be called `configure_accountant_for_some_pending_payables_at_startup`

node/src/accountant/mod.rs
```diff
@@ -2948,17 +3003,17 @@ mod tests {
         let before = SystemTime::now();
         system.run();
         let after = SystemTime::now();
-        assert_pending_payable_scanner_for_some_pending_payable_found(
+        assert_pending_payable_scanner_for_some_p_p_found(
```
utkarshg6, line 3006: Same issue here, previous name was better.

node/src/accountant/mod.rs
```diff
@@ -3308,22 +3373,29 @@ mod tests {
...
         notify_and_notify_later_params: &NotifyAndNotifyLaterParams,
         expected_sent_payables: SentPayables,
+    ) {
+        assert_payable_scanner_ran_for_some_p_p_found(
+            test_name,
+            consuming_wallet,
+            scan_params,
+            expected_sent_payables,
+        );
+        assert_scan_scheduling_for_some_p_p_found(notify_and_notify_later_params);
```
utkarshg6, line 3391: Please don't use `_p_p` anywhere.

node/src/accountant/mod.rs
```diff
@@ -5802,12 +5841,6 @@ mod tests {
             |_scanners: &mut Scanners, scan_schedulers: &mut ScanSchedulers| {
                 // Setup
                 let notify_later_params_arc = Arc::new(Mutex::new(vec![]));
-                scan_schedulers
-                    .payable
-                    .inner
-                    .lock()
-                    .unwrap()
-                    .last_new_payable_scan_timestamp = SystemTime::now();
```
utkarshg6, line 5810: I'm confused why you removed these lines now, instead of replacing with something else.

utkarshg6 (commented):
Done!

node/src/blockchain/errors/validation_status.rs
bertllll: I was on the track to prove a hypothesis that this code structure, PreviousAttempts, doesn't need the Hash impl. I disabled it together with this test that used to cover it. It turned out right.I just forgot to delete this test.

node/src/accountant/scanners/scan_schedulers.rs
bertllll: A compromise: NewPayableScanIntervalComputer.
I dropped "Dyn" from within. 

I do think the name is important to give the idea what it relates to.

node/src/accountant/scanners/scan_schedulers.rs
bertllll: Huh. Fairly annoyed, I did do it. 

Sometimes you're overdoing as a reviewer. This was on the edge. Note for the next time that Option may be much a better choice, as it has the whole pallet of methods and so you can easily work with it. Not so with a newly constructed enum. Still, I understand your intent. And as said, I did agree to go through the hassle. Just please don't push too much in refactoring.

node/src/accountant/scanners/scan_schedulers.rs
bertllll: Only code that is general enough that it finds usages in more than a single file belongs there. 

I'm not going to flip this code into something public while it can be safely private and just local.

node/src/accountant/mod.rs
bertllll: This parameter has moved inside the guts of the NewPayableScanIntervalComputer. It's been an object which is mocked in these tests as you can see on the next lines here in this closure. 

That's basically it. No need to do anything about the mock as it makes an impact on the environment only via the results it produces.

PR #732, opened 2025-10-21
GH-818 Rearch away Return Route Information IDs

dnwiebe:


node/src/proxy_server/mod.rs
```diff
@@ -1121,77 +1210,85 @@ impl IBCDHelper for IBCDHelperReal {
             let msg = TransmitDataMsg {
                 endpoint: Endpoint::Socket(client_addr),
                 last_data: true,
```
czarte, line 1212: why do we sending in `handle_normal_cloent_data` message `last_data: true` and `sequence_number_opt: Some(0)` ?

czarte (approved):

node/src/proxy_server/mod.rs
dnwiebe: We're turning off the stream from the browser because the user has no consuming wallet balance.

PR #735, opened 2025-10-27
Refine regex for max block count extraction and add test

### LINKED ISSUE #734: Update Error parsing for Nodies v2 max-block range on get_ethLogs
Update the error parsing to catch newly updated max block range error.

From live testing, the error is produced below as an example:

`WARN: BlockchainBridge: Error while retrieving transactions: QueryFailed("RPC error: Error { code: ServerError(-32001), message: \"Block range too large: maximum allowed is 20000 blocks\", data: None }")`

kauri-hero:
Updated regex to capture maximum block count from error messages more accurately - namely update to Nodies provider. Added a new test case for extracting this specific error response.

Closes #734

### DIFF
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -1541,7 +1541,7 @@ mod tests {
         system.run();
         let after = SystemTime::now();
         let expected_transactions = RetrievedBlockchainTransactions {
-            new_start_block: BlockMarker::Value(42 + 9_000_000 + 1),
+            new_start_block: BlockMarker::Value(42 + 9_000_000),
             transactions: vec![
                 BlockchainTransaction {
                     block_number: 6040059,
@@ -1742,7 +1742,7 @@ mod tests {
         let received_payments_message = accountant_recording.get_record::<ReceivedPayments>(0);
         check_timestamp(before, received_payments_message.timestamp, after);
         let expected_transactions = RetrievedBlockchainTransactions {
-            new_start_block: BlockMarker::Value(6 + 5000 + 1),
+            new_start_block: BlockMarker::Value(6 + 5000),
             transactions: vec![BlockchainTransaction {
                 block_number: 2000,
                 from: earning_wallet.clone(),
--- a/node/src/blockchain/blockchain_bridge.rs
+++ b/node/src/blockchain/blockchain_bridge.rs
@@ -507,8 +507,8 @@ impl BlockchainBridge {
 
     pub fn extract_max_block_count(error: BlockchainError) -> Option<u64> {
         let regex_result =
-            Regex::new(r".* (max: |allowed for your plan: |is limited to |block range limit \(|exceeds max block range )(?P<max_block_count>\d+).*")
-                .expect("Invalid regex");
+            Regex::new(r".* (max: |allowed for your plan: |is limited to |block range limit \(|exceeds max block range |maximum allowed is )(?P<max_block_count>\d+).*")
+        .expect("Invalid regex");
         let max_block_count = match error {
             BlockchainError::QueryFailed(msg) => match regex_result.captures(msg.as_str()) {
                 Some(captures) => match captures.name("max_block_count") {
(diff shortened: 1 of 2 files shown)


node/src/blockchain/blockchain_bridge.rs
```diff
@@ -507,8 +507,8 @@ impl BlockchainBridge {
 
     pub fn extract_max_block_count(error: BlockchainError) -> Option<u64> {
         let regex_result =
-            Regex::new(r".* (max: |allowed for your plan: |is limited to |block range limit \(|exceeds max block range )(?P<max_block_count>\d+).*")
-                .expect("Invalid regex");
+            Regex::new(r".* (max: |allowed for your plan: |is limited to |block range limit \(|exceeds max block range |maximum allowed is )(?P<max_block_count>\d+).*")
```
dnwiebe, line 510: So now this is a six-fold regular expression. I think that's probably confusing enough: if there turns out to be a seventh option here, we should redesign this to be a list of single-option regexes. We might even want to put the list in the database so that we can distribute little scripts that customers can run to update it.

dnwiebe (approved):

node/src/blockchain/blockchain_bridge.rs
kauri-hero: Thats a very good idea.
I added the sixth, so our original approach was getting stacked many-fold.

Ill write a card and we can see if the adding to db is a lot of extra work

kauri-hero:
I have QA'd this myself (CC: @Syther007 ) running on this build.

Node is handling the parsing and max-block-range change successfully in a test run (using base-mainnet)

However, it seems the Nodies endpoint still returns the error when max-block-range is set to 20,000 as indicated in their error:
```
(\"RPC error: Error { code: ServerError(-32001), message: \\\"Block range too large: maximum allowed is 20000 blocks\\\", data: None }\")
```

I have passed to Nodies team to diagnose to see if there is any further aspects we have to consider.

Also, I managed to nail a test scan of 19,999 blocks and that worked successfully with Nodies without error.

So Node side testing is a pass, but before we push to production I want to check with Nodies provider with our results

kauri-hero:
I have discovered something interesting which maybe has been interfering with the max-block-range logic and handling with the block scans.

After testing and meeting with our RPC blockchain service provider, we realized that when a max block range is parsed out and added to our config, the next scan is performed across the limited range but the starting block number counts as "1 block" and the end range block number is simply start block number + the max range.

that actually makes the scan range max-block-range +1 and the error will repeat because its over the max range by 1 block, since it is including the starting block number of the scan.

In the test example:

```
BlockchainInterface: Retrieving transactions from start block: Number(37564645) to end block: Number(37584645) for: 0x27d9…788d chain_id: 8453 contract: 0x45d9c101a3870ca5024582fd788f4e1e8f7971c3
2025-11-02 21:25:24.436 Thd23: DEBUG: web3::transports::http: [5] Sending: {"jsonrpc":"2.0","method":"eth_getLogs","params":[{"address":"0x45d9c101a3870ca5024582fd788f4e1e8f7971c3","fromBlock":"0x23d30e5","toBlock":"0x23d7f05","topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",null,"0x00000000000000000000000027d9a2ac83b493f88ce9b4532edcf74e95b9788d"]}],"id":5} to https://lb.nodies.app/v1/c903375d5db44fe89de5736d47840f34
2025-11-02 21:25:24.768 Thd23: TRACE: web3::transports::shared: [5] Extracting result.
2025-11-02 21:25:24.769 Thd23: TRACE: BlockchainInterface: Transaction logs retrieval completed: Err(QueryFailed("RPC error: Error { code: ServerError(-32001), message: \"Block range too large: maximum allowed is 20000 blocks\", data: None }"))
```

The block range of `(37564645) to end block: Number(37584645)` is actually 20,001 blocks as it includes the first block number in the range!

kauri-hero:
QA Pass on Windows build:
✅ block range scan produces error when max range error received from RPC (in case of Nodies provider used for testing range is 20000)
 manually changed start-block to 20005 blocks in past using set-configuration and triggered another scan to ensure it proceeds
✅ Next scan proceeds with correct block range, capped by the correct range and start block inclusive bounds
✅ Changed start block again the 50000 blocks in past
✅ Next scans proceed as expected, in scans of 20000. Each scan starting and ended with correct block numbers

PR #743, opened 2025-11-11
GH-598: fixes driven from debugging - review 2

bertllll:


node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -153,304 +148,263 @@ pub enum AccountingMsgType {
...
 
-pub struct NewPostingsDebugContainer {
+pub struct NewChargessDebugContainer {
     debug_enabled: bool,
-    vec: Vec<NewPosting>,
+    vec: Vec<NewCharge>,
 }
 
-impl NewPostingsDebugContainer {
+impl NewChargessDebugContainer {
```
utkarshg6, line 156: We want single 's' in the `Charges`.

node/src/accountant/logging_utils/msg_id_generator.rs
```diff
@@ -4,20 +4,29 @@ use crate::sub_lib::accountant::MSG_ID_INCREMENTER;
...
 #[derive(Default)]
 pub struct MessageIdGeneratorReal {}
 
 impl MessageIdGenerator for MessageIdGeneratorReal {
-    fn id(&self) -> u32 {
+    fn new_id(&self) -> u32 {
         MSG_ID_INCREMENTER.fetch_add(1, Ordering::Relaxed)
     }
+    fn last_used_id(&self) -> u32 {
+        MSG_ID_INCREMENTER.load(Ordering::Relaxed) - 1
```
utkarshg6, line 20: Why is there -1 at the end? Correct me if I'm wrong but the last I'd should be the current value, not current value -1.

node/src/accountant/logging_utils/msg_id_generator.rs
```diff
@@ -47,10 +56,26 @@ mod tests {
...
+    fn msg_id_generator_last_used_id() {
+        let _guard = MSG_ID_GENERATOR_TEST_GUARD.lock().unwrap();
+        let subject = MessageIdGeneratorReal::default();
+        let new_id = subject.new_id();
+
+        let same_id_1 = subject.last_used_id();
+        let same_id_2 = subject.last_used_id();
+        let new_id_2 = subject.new_id();
+
+        assert_eq!(new_id, same_id_1);
```
utkarshg6, line 77: I would suggest if you increment the value first and then return it in the new_id method. That way you won't need to use -1 in the load_id method.

node/src/accountant/test_utils.rs
```diff
@@ -1475,9 +1475,13 @@ pub struct MessageIdGeneratorMock {
 }
 
 impl MessageIdGenerator for MessageIdGeneratorMock {
-    fn id(&self) -> u32 {
+    fn new_id(&self) -> u32 {
         self.ids.borrow_mut().remove(0)
     }
+
+    fn last_used_id(&self) -> u32 {
+        todo!()
```
utkarshg6, line 1483: You can write something here, I don't like it empty.

node/src/hopper/routing_service.rs
```diff
@@ -420,11 +420,13 @@ impl RoutingService {
                 }
                 match self.routing_service_subs.to_accountant_routing.try_send(
                     ReportRoutingServiceProvidedMessage {
-                        timestamp: SystemTime::now(),
-                        paying_wallet: payer.wallet,
-                        payload_size,
-                        service_rate: self.per_routing_service,
-                        byte_rate: self.per_routing_byte,
+                        service: ServiceProvided {
```
utkarshg6, line 423: I dunno why you nested it like that, looks unnecessary, unless there's a sibling field.

node/src/sub_lib/accountant.rs
```diff
@@ -124,18 +126,82 @@ impl SubsFactory<Accountant, AccountantSubs> for AccountantSubsFactoryReal {
...
 
-// TODO: These four structures all consist of exactly the same five fields. They could be factored out.
-#[derive(Clone, PartialEq, Eq, Debug, Message)]
-pub struct ReportRoutingServiceProvidedMessage {
-    pub timestamp: SystemTime,
-    pub paying_wallet: Wallet,
-    pub payload_size: usize,
-    pub service_rate: u64,
-    pub byte_rate: u64,
+pub trait AccountableServiceWithTraceLog {
```
utkarshg6, line 129: This would be additional complexity while searching for logs in the production code. I personally prefer a direct print of log using macros. Introducing a layer of abstraction here apparently can make it harder to search for logs.

utkarshg6 (commented):
Continue at node/src/sub_lib/accountant.rs:136

node/src/sub_lib/accountant.rs
```diff
@@ -124,18 +126,82 @@ impl SubsFactory<Accountant, AccountantSubs> for AccountantSubsFactoryReal {
...
+    fn log_trace(&self, (logger, msg_id): (&Logger, u32)) {
+        logger.trace(|| self.trace_log_msg(msg_id))
+    }
+    fn trace_log_msg(&self, msg_id: u32) -> String;
+}
+
+pub trait MessageWithServicesProvided {
```
utkarshg6, line 136: Prefer enum based approach here

```rust

pub struct ReportServiceProvidedMessage {
    pub service: ServiceProvided,
    pub service_type: ServiceType,
}

#[derive(Clone, PartialEq, Eq, Debug)]
pub enum ServiceType {
    Exit,
    Routing,
}

impl ReportServiceProvidedMessage {
    pub fn new_exit_service(service: ServiceProvided) -> Self {
        Self {
            service,
            service_type: ServiceType::Exit,
        }
    }
    
    pub fn new_routing_service(service: ServiceProvided) -> Self {
        Self {
            service,
            service_type: ServiceType::Routing,
        }
    }
}

impl ReportServiceProvidedMessage {
    fn services_provided(&self) -> &ServiceProvided {
        &self.service
    }

    fn service_type(&self) -> ServiceType {
        self.service_type
    }
}
```

node/src/sub_lib/accountant.rs
```diff
@@ -147,25 +213,58 @@ pub struct ReportExitServiceProvidedMessage {
 pub struct ReportServicesConsumedMessage {
     pub timestamp: SystemTime,
     pub exit: ExitServiceConsumed,
-    pub routing_payload_size: usize,
-    pub routing: Vec<RoutingServiceConsumed>,
+    pub routing: RoutingServicesConsumed,
+}
+
+impl AccountableServiceWithTraceLog for ExitServiceConsumed {
```
utkarshg6, line 219: Lines 219...234 could be replaced by a single log message which just prints the service type similar to other values in the trace log message.

node/src/sub_lib/accountant.rs
```diff
@@ -147,25 +213,58 @@ pub struct ReportExitServiceProvidedMessage {
...
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct RoutingServiceConsumed {
     pub earning_wallet: Wallet,
-    pub payload_size: usize,
     pub service_rate: u64,
     pub byte_rate: u64,
 }
 
+pub enum ServiceProvidedTraceLogWrapper<'a> {
```
utkarshg6, line 258: Again it should be done other way around. You should use the `enum ServiceType` here. Refer to this comment - https://github.com/MASQ-Project/Node/pull/743/files#r2533460766

node/src/sub_lib/accountant.rs
```diff
@@ -292,4 +395,39 @@ mod tests {
...
+            timestamp: SystemTime::now(),
+            paying_wallet: make_wallet("abc"),
+            payload_size: 0,
+            service_rate: 0,
+            byte_rate: 0,
+        };
+        let exit_service_msg = ReportExitServiceProvidedMessage {
+            service: service.clone(),
+        };
+        let routing_service_msg = ReportRoutingServiceProvidedMessage { service };
```
utkarshg6, line 411: This test will change slightly. You should consider a test with 2 cases along with a helper function.

utkarshg6 (commented):
Continue at node/src/accountant/logging_utils/accounting_msgs_debug.rs

node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -12,80 +12,68 @@ use web3::types::Address;
...
-    ) {
-        if logger.debug_enabled() {
-            if let Some(loggable_stats) = self.manage_log(msg_type, new_postings, log_window_size) {
-                debug!(logger, "{}", loggable_stats);
-            }
-        }
+    ) -> Option<LoggableStats> {
+        self.record_new_charges_by_msg_type(new_charges, msg_type);
+
+        self.maybe_dump_stats_by_msg_type(log_window_size, msg_type)
```
utkarshg6, line 30: I'd safely remove the suffix `_by_msg_type` from the fn names.

node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -153,304 +148,263 @@ pub enum AccountingMsgType {
...
-                    .msg_count_since_last_logged
-            },
+        test_process_debug_stats(AccountingMsgType::ServicesConsumed, 8);
+    }
+
+    fn test_process_debug_stats(msg_type: AccountingMsgType, log_window_size: u16) {
+        let mut new_charge_feeds_per_msg =
+            generate_new_charge_feeds_representing_msgs(log_window_size);
+        let expected_sorted_stats =
+            construct_expected_sorted_stats_from_generated_new_charges(&new_charge_feeds_per_msg);
```
utkarshg6, line 212: The variable names are less clear and verbose. Kindly change it to something familiar like:

`new_charge_feeds_per_msg` -> `test_msgs`
`expected_sorted_stats` -> `expected_stats`

node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -153,304 +148,263 @@ pub enum AccountingMsgType {
...
+        let mut subject = AccountingMessageTracker::default();
+        assert_empty_stats(&subject);
+        let charge_msg_matching_the_window_size =
+            new_charge_feeds_per_msg.remove(log_window_size as usize - 1 - 1);
```
utkarshg6, line 216: These two `-1` are less clear. It's hard to decipher what individual subtractions are trying to achieve. I'd suggest to break this into two steps and use different variable names. It will shed some light on why these subtractions happened.

node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -153,304 +148,263 @@ pub enum AccountingMsgType {
...
+        let initial_charge_msgs = new_charge_feeds_per_msg;
+
+        test_msgs_of_count_window_size_minus_one(
+            &mut subject,
+            msg_type,
+            log_window_size,
```
utkarshg6, line 222: Also, the use of window size is unclear in this testing fn as whole. I'd like if we can make it clearer.

node/src/accountant/logging_utils/accounting_msgs_debug.rs
```diff
@@ -153,304 +148,263 @@ pub enum AccountingMsgType {
...
+                msg_type,
+                charge_msg_matching_the_window_size,
+                log_window_size,
+            )
+            .expect("first try: expected stats dump");
+
+        assert_provided_loggable_stats(result, msg_type, log_window_size, expected_sorted_stats);
+        assert_empty_stats(&subject);
+
+        retest_after_emptied(&mut subject, msg_type);
```
utkarshg6, line 237: Optional: There are so many fns being used, I'd converted it into a builder pattern and would implement these fns on the structure. If you're ready for more adventure then only embark on it, otherwise it's totally fine to leave it this way.

utkarshg6 (commented):

node/src/accountant/mod.rs
```diff
@@ -622,39 +625,33 @@ impl Accountant {
...
-        timestamp: SystemTime,
-        payload_size: usize,
-        wallet: &Wallet,
-    ) -> u128 {
-        let byte_charge = byte_rate as u128 * (payload_size as u128);
-        let total_charge = service_rate as u128 + byte_charge;
+    fn record_service_provided(&self, service: &ServiceProvided) -> Option<NewCharge> {
+        let byte_charge = service.byte_rate as u128 * (service.payload_size as u128);
+        let total_charge = service.service_rate as u128 + byte_charge;
+        let wallet = &service.paying_wallet;
```
utkarshg6, line 631: Nicely Done!

node/src/accountant/mod.rs
```diff
@@ -745,133 +742,103 @@ impl Accountant {
...
-        self.logging_utils.accounting_msgs_stats.manage_debug_log(
-            &self.logger,
-            AccountingMsgType::RoutingServiceProvided,
-            DEFAULT_ACCOUNTING_MSG_LOG_WINDOW,
-            vec![NewPosting::new(msg.paying_wallet.address(), sum)],
-        )
-    }
+    fn handle_report_service_provided_message<AccountingMessage>(&mut self, msg: AccountingMessage)
+    where
+        AccountingMessage: MessageWithServicesProvided,
```
utkarshg6, line 747: This should go away once you decide to use the structure instead of the trait implementation. The enum based approach will make things easier for you here. You can thank me later.

node/src/accountant/mod.rs
```diff
@@ -745,133 +742,103 @@ impl Accountant {
...
-            )
-        });
-        let sum = self.record_service_provided(
-            msg.service_rate,
-            msg.byte_rate,
-            msg.timestamp,
-            msg.payload_size,
-            &msg.paying_wallet,
-        );
+        trace_log_wrapper.log_trace(self.logging_kit(MsgIdRequested::New));
```
utkarshg6, line 753: This code is worrisome. I think if we had to find the log once we have the log file with us, then at that time, it'll be difficult for us to find this.

This layer of abstraction is the issue, if it wasn't for a use case like logging, I'd appreciate it and here, I'm concerned.

utkarshg6 (commented):
Done!

PR #748, opened 2025-11-20
GH-826 zero-hop as default neighborhood-mode

czarte:
> [!NOTE]
> Switch default `neighborhood-mode` from `standard` to `zero-hop` and update tests/expectations accordingly.
> 
> - **Config/Setup**:
>   - Change `NeighborhoodMode` computed default to `zero-hop` in `node/src/daemon/setup_reporter.rs`.
> - **Tests**:
>   - Update integration and unit tests to expect `neighborhood-mode` = `zero-hop`, including setup output assertions and default setup clusters.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1874585d2821fd99621b4dd41abfef6562d2e86. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dnwiebe:
Note to @kauri-hero and others:

Changing a default is a serious matter. People build things on top of infrastructure based on assumptions about that infrastructure's defaults. When those defaults change, the things break. Before this is merged into master, it should be proclaimed from the rooftops that a default is changing, and anyone who doesn't want their Nodes coming up in Zero-Hop Mode needs to specify Standard Mode ASAP.

Also, we might want to think about changing the name of Standard Mode. A default that isn't standard--or a standard that isn't default--is kind of a contradiction in terms; it's confusing.

kauri-hero:
Thanks Dan - I totally understand and appreciate the significance of changing these defaults.<br>I'll ensure that we have an immediate announcement and also include several notification areas in our guides and setup instructions.<br>I am aware that most of the users who are using MASQ will not be starting up Node with default values in 99% of cases - especially since the previous default values for chain and rate-packs etc were far from a usable set of default values in the "real-world" live mainnet environment. So every user was having to change the defaults, either using the daemon, config file, or most commonly within the apps we provide where it's taken care of by the user easily without manual config.<br><br>But overall, point taken and we will ensure its over-communicated on this change

PR #749, opened 2025-11-20
GH-826 zero-hop as default neighborhood-mode review addressed

czarte:


node/src/bootstrapper.rs
```diff
@@ -1655,6 +1661,8 @@ mod tests {
                     "1.2.3.4",
                     "--data-directory",
                     data_dir.to_str().unwrap(),
+                    "--neighborhood-mode",
+                    "standard",
```
dnwiebe, line 1669: See if this will pass with neither `--neighborhood-mode` nor `--ip`.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -304,10 +304,7 @@ fn make_neighborhood_mode(
             "--neighborhood-mode {} has not been properly provided for in the code",
             s
         ),
-        None => {
-            let rate_pack = configure_rate_pack(multi_config, persistent_config)?;
-            neighborhood_mode_standard(multi_config, neighbor_configs, rate_pack)
-        }
+        None => Ok(NeighborhoodMode::ZeroHop),
```
dnwiebe: I think this `None` clause should have the same functionality as the `Some(ref s) if s == "zero-hop"` clause. Can you put that functionality in an outboard method and call it from both places?

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -1537,33 +1534,7 @@ mod tests {
                     .unwrap()
             )),
         );
-        assert_eq!(
```
dnwiebe, line 1540: This assert should probably not be removed, but should appear somewhere else where the neighborhood mode is Standard. It protects against Standard-mode sloppiness, and we still need that protection in Standard mode.

node/src/node_configurator/unprivileged_parse_args_configuration.rs
```diff
@@ -1597,15 +1568,7 @@ mod tests {
             .mode
             .neighbor_configs()
             .is_empty());
-        assert_eq!(
```
dnwiebe, line 1600: Check and see if we have a "When neighborhood mode is Standard and IP is not specified, IP becomes 0.0.0.0" assert somewhere else. If not, make one.

dnwiebe (approved):

node/src/bootstrapper.rs
czarte: In case of no setting of `--neighborhood-mode` and `--ip` clandestine port is not defined

PR #766, opened 2026-05-27
GH-598: cosmetics from debugging review3

bertllll:


bertllll:
@kauri-hero ignore this PR; it's been a dead-end road. Valuable as an experiment but the code is going away. It's an ending of a sequence of commits even longer than what's included here. There have been good moves among the bad ones, I might try to save some in a small PR before this piece of work is definitely gone. Cheers

